PR #32, opened 2021-06-28
Add type transformation

### LINKED ISSUE #24: Support "type" transformation
Would it be possible to add a "type" transformation? In TypeScript, for example:

```typescript
const logs: { user: string; event: string }[] = [];
```

I think being able to "take type" and select `{ user: string; event: string }[]` it would be very useful.

Less important imo, but there are also the cases of generics and type parameters to consider?

brxck:
This should allow referencing type declarations and annotations in Typescript, as I brought up in #24 

The transformation can be used on variable declarations, parameters, functions, and properties within type declarations.

This doesn't cover generics, i.e. type parameters or arguments. I started on them, but they complicate things a bit more. Personally I don't think I'm going to miss supporting them at least for now. 

Any tweaks or feedback welcome 👍🏼 

Closes #24


pokey (commented):
Awesome!!  Tested locally and it works well.  I had an [idea](https://github.com/brxck/cursorless-vscode/pull/2) for some fancy unification; lmk if you can get that approach working / if it seems like a good direction

src/languages/typescript.ts
```diff
@@ -131,6 +155,18 @@ const nodeMatchers: Record<ScopeType, NodeMatcher> = {
...
+    // Typed parameters, properties, and functions
+    typedParentNode,
+    // Type alias/interface declarations
+    possiblyWrappedNode(
+      (node) => node.type === "export_statement",
+      (node) =>
+        node.type === "type_alias_declaration" ||
+        node.type === "interface_declaration",
+      (node) => [getDeclarationNode(node)]
+    )
```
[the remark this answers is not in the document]
brxck: It does! Hadn't noticed that function 😅

brxck:
Got this working with your changes. Pulling out the type node isn't quite as clean as you might hope, but all the pieces seem to fit together nicely.

src/nodeMatchers.ts
```diff
@@ -7,6 +7,39 @@ export function hasType(...typeNames: string[]): NodeMatcher {
     typeNames.includes(node.type) ? simpleSelectionExtractor(node) : null;
 }
 
+export function childNodeMatcher(
+  getNode: (node: SyntaxNode) => SyntaxNode | null,
```
pokey, line 11: I think I'd rename this to
```suggestion
  getMatchingChildNode: (node: SyntaxNode) => SyntaxNode | null,
```

src/nodeMatchers.ts
```diff
@@ -7,6 +7,39 @@ export function hasType(...typeNames: string[]): NodeMatcher {
...
+  extractor: (node: SyntaxNode) => SelectionWithContext
+): NodeMatcher {
+  return (editor: TextEditor, node: SyntaxNode) => {
+    const returnNode = getNode(node);
```
pokey, line 15:
```suggestion
    const childNode = getMatchingChildNode(node);
```

pokey (approved):
This looks awesome @brxck.  I left one more nit to make things slightly clearer but then it's good to go.  I can pull it down to mess around with tomorrow my time then I'll ship it

pokey:
Ok I pulled this one locally and it works pretty well.  One that I noticed tho:

```typescript
export type DecorationMap = {
  [k in SymbolColor]?: vscode.TextEditorDecorationType;
};
```
If cursor is inside `vscode`, I think I'd expect it to select `vscode.TextEditorDecorationType`, but it selects the whole type instead.  Is that expected?

brxck:
Apparently the type of this node is different: `opting_type_annotation` I hadn't seen that one before. Adding it to `getTypeNode` seems to handle that as expected.

PR #33, opened 2021-06-28
Updated regexp to support fixed tokens and sequential symbols

AndreasArvidsson:


src/constants.ts
```diff
@@ -1,6 +1,35 @@
...
+  "/",
+  "=",
+  "<",
+  ">",
+  "_",
+  "#",
+  "\\.",
+  "\\|",
+  "&"
+].map(s => `${s}+`).join("|");
```
pokey, line 14: Let's leave it for now, but def starting to feel a bit complex for `constants.py` 😅

pokey (approved):
LGTM! I'll mess around with it locally then merge and release

src/constants.ts
AndreasArvidsson: Yea that is true.
I'm running it locally now as well.

src/tokenizer.ts
```diff
@@ -0,0 +1,45 @@
...
+];
+
+const FIXED_TOKENS = ["!==", "!=", "+=", "-=", "*=", "/=", "%=", "<=", ">="];
+const REPEATABLE_SYMBOLS_REGEX = REPEATABLE_SYMBOLS.map(escapeRegExp)
+  .map((s) => `${s}+`)
+  .join("|");
+const FIXED_TOKENS_REGEX = FIXED_TOKENS.map(escapeRegExp).join("|");
+const IDENTIFIERS_REGEX = "[a-zA-Z_]+[a-zA-Z_0-9]*";
+const SINGLE_SYMBOLS_REGEX = "[^\\s\\w]";
+const NUMBERS_REGEX = "(?<![.])\\d+.\\d+(?![.])|\\d+"; // (not-dot digits dot digits not-dot) OR digits
```
pokey, line 23: Any performance issues with the look ahead? I know these can be a bit slower but I'd imagine not enough to be noticeable here

pokey (commented):
I find this confusing:
image
I'd prefer that it either totally bailed on a semver and split it into dots and number strings (like today) or highlighted the whole thing.  Tho I think I'd prefer the former?  Usually I"ll want to select one of the sections to bump, eg major, minor or patch, rather than the whole thing

AndreasArvidsson:
Your wish is my command. Semantic versioning should now be separate tokens.

src/tokenizer.ts
AndreasArvidsson: I haven't noticed anything but I haven't measured either.

src/test/suite/tokenizer.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    test(`tokenizer ${input}`, () => {
+      const output = tokenize(input, (match) => match[0]);
+      assert.deepStrictEqual(output, expectedOutput);
+    });
+  });
+});
+
+function getSymbols() {
+  const results = <string[]>[];
+  const range = [33, 48, 58, 65, 91, 97, 123, 126];
```
pokey, line 42: I'd split this into pairs of start-range, end-range rather than flattening into one list. And I'd probably rename for clarity. Then you could just do something like `flatten(rangesToTest.map(([start, end])) => range(start, end).map(charCode => String.fromCharCode(charCode)))`

I'd also add a comment after each range to indicate what strings they correspond to. Or use single-char strings in the first place and convert during processing

src/test/suite/tokenizer.test.ts
```diff
@@ -0,0 +1,49 @@
...
+suite("tokenizer", () => {
+  tests.forEach(([input, expectedOutput]) => {
+    test(`tokenizer ${input}`, () => {
+      const output = tokenize(input, (match) => match[0]);
+      assert.deepStrictEqual(output, expectedOutput);
+    });
+  });
+});
+
+function getSymbols() {
```
pokey, line 40: I'd find a better name. `singleSymbolTests`? And a quick docstring

pokey (approved):
This looks great! Left a few more nits then I think we can merge this one in

src/test/suite/tokenizer.test.ts
```diff
@@ -0,0 +1,58 @@
+import * as assert from "assert";
+import { tokenize } from "../../tokenizer";
+import { flatten, range } from "lodash";
+
+type TestType = [string, string[]][];
```
pokey, line 5: I think I'd do
```suggestion
type TestCase = [string, string[]];
```

And then put the `[]` on the end in the two places you use it

src/test/suite/tokenizer.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  ["my-variable", ["my", "-", "variable"]],
+  ["my.variable", ["my", ".", "variable"]],
+  ["my/variable", ["my", "/", "variable"]],
+  ["my::variable", ["my", ":", ":", "variable"]],
+  // Strings
+  ['"my variable"', ['"', "my", "variable", '"']],
+  ["'my variable'", ["'", "my", "variable", "'"]],
+  // Symbols incl repeatable
+  ...singleSymbolTests,
+  ["!_|||>{.-----()", ["!", "_", "|||", ">", "{", ".", "-----", "(", ")"]],
```
pokey, line 30: Nice. Would be good to have multiple parens in a row I mean, eg `(((` and `)))` because I prefer these to be split so want to make sure they don't end up in repeated symbols regex by accident

src/test/suite/tokenizer.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  tests.forEach(([input, expectedOutput]) => {
+    test(`tokenizer ${input}`, () => {
+      const output = tokenize(input, (match) => match[0]);
+      assert.deepStrictEqual(output, expectedOutput);
+    });
+  });
+});
+
+/**
+    Return an array of all non-alnum symbols in the ascii range
```
pokey, line 47:
```suggestion
    Returns an array of single non-alphanumeric symbols in various ascii ranges to make sure they don't get dropped by the tokeniser
```

pokey (approved):
Much better! I left 3 more nits sorry 🙈

Issue #27, opened 2021-06-27
Support "name" transformation


pokey:
Eg name of var in a declaration, class name, function name, etc.  Could be extended to support specifying "funk name", "class name", etc.  So eg

```python
class Foo:
    def bar():
        baz = 5
```

If your cursor is on the `5`, then:
- "take name" selects `baz`,
- "take state name" also selects `baz`,
- "take funk name" selects `bar`, and
- "take class name" selects `Foo`

pokey:
Fwiw I'd imagine this would commonly be used with the "rename" action, in which case the syntax becomes a bit awkward: "rename name" / "rename class name".  I wonder if we should consider a way for "rename" to automatically add the "name" transformation if it gets a different scope.  Eg support "rename state" / "rename class" / "rename funk"

But that's def extra credit 😄

First cut would just be to support "name", then support "funk name", then slick syntax for "rename"

pokey:
closed by #56

PR #56, opened 2021-06-30
Refactor transformations and add name transformation

brxck:
The scope of this PR has changed a bit!

This refactors the transformations code to increase reusability and lower the cost to support new transformations (modifiers). This also adds the name, function name, and class name transformations.

NodeMatchers are split into component functions NodeFinders and SelectionExtractors. Functions have been added to combine and compose these into new NodeMatchers.

---

> Here's the first step of #27, relatively simple.
> 
> I'd be happy to look into extending this for "funk name," "funk class," etc, but I am not entirely sure what approach to take or if there are already provisions for this sort of thing.
>
> This includes/depends on work in #32. I wish stacked PRs were a thing in GitHub, lmk if you want to handle to this kind of PR differently.


brxck:
Oh this also includes a commit which adds some logging of the types of the selected node and its ancestors. This has been pretty handy developing these transformations. If you prefer, I can easily drop that commit.

brxck:
@pokey I'm still a little hung up on how to go about implementing a function name transformation.

Selecting a function in both our languages is non trivial. For example, the Typescript implementation is three cascading matchers:

```ts
  namedFunction: cascadingMatcher(
    possiblyExportedDeclaration("function_declaration", "method_definition"),
    (editor: TextEditor, node: SyntaxNode) =>
      node.type === "public_field_definition" &&
      getValueNode(node)!.type === "arrow_function"
        ? simpleSelectionExtractor(node)
        : null,
    possiblyWrappedNode(
      (node) => node.type === "export_statement",
      isNamedArrowFunction,
      (node) => [getDeclarationNode(node)]
    )
  ),
```

This ends up being a substantial amount of logic to extract and duplicate for this one transformation. It feels like we should have the ability to compose/extend matchers if we want to implement more things like this.

If we separated finding a node and extracting a selection, then we could compose the node finder functions. Which could look very roughly like:

```ts
import { flowRight } from "lodash";

const nodeMatchers: Record<ScopeType, NodeMatcher> = {
  functionName: flowRight([simpleSelectionExtractor, findName, findNamedFunction]),
  namedFunction: flowRight([simpleSelectionExtractor, findNamedFunction])
}
```

Curious what you think about this!

pokey:
Sounds good to me if you can make it work! My worry is that extracting the name will be different depending which of the cascading marchers hits. For example if the matcher hits an export, it returns the whole export statement, so I believe it would then have to drill back down to get the name. I think it's simpler to avoid that because the name extractor doesn't need to try to include the export, whereas the function matcher goes to great lengths to do so. So I think the function matching process will actually be a bit different if you just want the name

But if you think you can make it work give it a shot! Otherwise I'd just copy the matchers in the cascade and tear off the stuff you don't need and get the name from there. If after that it still looks really similar to the original then yeah maybe makes sense to abstract it

Not sure if any of this made sense it's pretty late here 😅

src/Types.ts
```diff
@@ -253,4 +255,11 @@ export interface DecorationColorSetting {
   dark: string;
   light: string;
   highContrast: string;
-}
\ No newline at end of file
+}
+
+export type NodeFinder = (node: SyntaxNode) => SyntaxNode | null;
```
pokey, line 260: I might move the `NodeMatcher` above down to right above this one; it got separated in the rebase

I also might add a docstring for this type and the next one as they're important abstractions

src/extension.ts
```diff
@@ -197,6 +198,25 @@ export async function activate(context: vscode.ExtensionContext) {
     addDecorations();
   };
 
+  function logBranchTypes(event: vscode.TextEditorSelectionChangeEvent) {
```
pokey, line 201: Mind moving this one to a separate file?  Like `debug.ts` or something?

src/extension.ts
```diff
@@ -197,6 +198,25 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    const getBranch = (branch: SyntaxNode[]): SyntaxNode[] => {
+      if (branch[0].parent) {
+        return getBranch([branch[0].parent, ...branch]);
+      }
+      return branch;
+    };
```
pokey, line 207: whoah recursion.  this one is a bit tough to follow 😅.  would `getAncestors` be a better name?  Presuming I've understood the function correctly...

I think this one might be a bit clearer without recursion too.  Just iterate until the parent is null and add the nodes as you go up.  I also prefer `== null` comparisons to `boolean` transformations

Also I generally try to avoid nesting functions unless it needs to capture scope

src/extension.ts
```diff
@@ -197,6 +198,25 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    const location = new vscode.Location(
+      vscode.window.activeTextEditor!.document.uri,
+      event.selections[0]
+    );
+    const leaf: SyntaxNode = getNodeAtLocation(location);
+    const branch = getBranch([leaf]);
+    branch.forEach((node, i) => console.log(">".repeat(i + 1), node.type));
+    const leafText = leaf.text.replace(/\s+/g, " ").substring(0, 100);
+    console.log(">".repeat(branch.length), `"${leafText}"`);
+  }
```
pokey, line 218: nice.  yeah this one could be useful.  I think I might use `console.debug` for these tho

src/nodeMatchers.ts
```diff
@@ -1,200 +1,35 @@
...
-      trailingDelimiterRange = makeRange(
-        node.endPosition,
-        nextNonDelimiterNode.startPosition
-      );
-
-      containingListDelimiter = editor.document.getText(trailingDelimiterRange);
+  return function (editor: TextEditor, initialNode: SyntaxNode) {
+    let returnNode: SyntaxNode | null = initialNode;
+    for (const finder of finders) {
+      returnNode = returnNode ? finder(returnNode) : null;
```
pokey, line 24: should we not short-circuit if one of the finders returns null?

src/nodeMatchers.ts
```diff
@@ -216,3 +51,10 @@ export function cascadingMatcher(...matchers: NodeMatcher[]): NodeMatcher {
     return null;
```
pokey, line 51: While we're here, maybe `cascadingMatcher` isn't the best name?  Maybe something like `unionMatcher`?  `matchAny`?  Idk.  But I had to scratch my head to remember what it did and I'm the one that named it in the first place 😅

src/nodeSelectors.ts
```diff
@@ -0,0 +1,129 @@
...
+      return node;
+    }
+
+    node = node.previousSibling;
+  }
+
+  return node;
+}
+
+export function selectDelimited(
```
pokey, line 78: Should this be `delimitedSelector`?

src/nodeFinders.ts
```diff
@@ -0,0 +1,53 @@
+import { SyntaxNode } from "web-tree-sitter";
+import { NodeFinder } from "./Types";
+
+export const findNode =
+  (isTargetNode: (node: SyntaxNode) => boolean): NodeFinder =>
+  (node: SyntaxNode) => {
+    return isTargetNode(node) ? node : null;
+  };
+
+export const findNodeOfType =
```
pokey, line 10: `typedNodeFinder`?  Idk maybe verb convention is better than noun.  But this does return a finder, it doesn't find.  I don't feel strongly as long as we're consistent

src/nodeFinders.ts
```diff
@@ -0,0 +1,53 @@
+import { SyntaxNode } from "web-tree-sitter";
+import { NodeFinder } from "./Types";
+
+export const findNode =
+  (isTargetNode: (node: SyntaxNode) => boolean): NodeFinder =>
+  (node: SyntaxNode) => {
+    return isTargetNode(node) ? node : null;
+  };
```
pokey, line 8: has this been prettified?  I would think it would indent this one a bit differently.  I run prettier and eslint

src/nodeFinders.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const findNodeOfType =
+  (...typeNames: string[]): NodeFinder =>
+  (node: SyntaxNode) => {
+    return typeNames.includes(node.type) ? node : null;
+  };
```
pokey, line 14: Could you just implement this function in terms of `findNode`?

src/languages/getPojoMatchers.ts
```diff
@@ -1,42 +1,38 @@
...
-      return simpleSelectionExtractor(getKeyNode(node)!);
-    },
-    value: childNodeMatcher(getValueNode, getNodeWithLeadingDelimiter),
-    list: hasType(...listTypes),
-    listElement: delimitedMatcher(
-      (node) =>
-        listTypes.includes(node.parent?.type ?? "") && listElementMatcher(node),
-      (node) => node.type === "," || node.type === "[" || node.type === "]",
-      ", "
+    pairKey: composedMatcher([findNodeOfType("pair"), getKeyNode]),
```
pokey, line 21: oooh shiny 😊

src/nodeMatchers.ts
```diff
@@ -1,200 +1,35 @@
...
-  }
-
-  return node;
-}
-
-export function delimitedMatcher(
-  nodeMatches: (node: SyntaxNode) => boolean,
-  isDelimiterNode: (node: SyntaxNode) => boolean,
-  defaultDelimiter: string
+export function composedMatcher(
```
pokey, line 17: hmm i tend to like variadic signatures for this type of thing, so you can say eg `composedMatcher(findNodeOfType("pair"), getKeyNode)` (notice no square brackets).  But obv that doesn't work with this selector param.  Not sure what the right answer is here.  Prob fine as you have it

src/languages/python.ts
```diff
@@ -130,23 +139,32 @@ const nodeMatchers: Record<ScopeType, NodeMatcher> = {
...
-    (node) => node.type === "," || node.type === "(" || node.type === ")",
-    ", "
+  ifStatement: typeMatcher("if_statement"),
+  class: matcher(possiblyDecoratedDefinition("class_definition")),
+  statement: typeMatcher(...STATEMENT_TYPES),
+  name: cascadingMatcher(
+    matcher(getNameNode),
+    matcher((node) => (node.type === "assignment" ? getLeftNode(node) : null))
+  ),
+  functionName: notSupported,
```
pokey, line 149: Would it be hard to implement this one?

src/languages/typescript.ts
```diff
@@ -135,55 +147,68 @@ const nodeMatchers: Record<ScopeType, NodeMatcher> = {
...
+    matcher(findClassPropertyArrowFunction),
 
     // eg:
     // const foo = () => "hello"
-    possiblyWrappedNode(
-      (node) => node.type === "export_statement",
-      isNamedArrowFunction,
-      (node) => [getDeclarationNode(node)]
+    matcher(
+      findPossiblyWrappedNode(
```
pokey, line 204: I wonder why this one can't use `possiblyExportedDeclaration` 🤔.  Looks like it was already this way but is a bit curious

src/languages/typescript.ts
```diff
@@ -135,55 +147,68 @@ const nodeMatchers: Record<ScopeType, NodeMatcher> = {
...
-  class: possiblyExportedDeclaration("class_declaration", "class"),
-  statement: possiblyExportedDeclaration(...STATEMENT_TYPES),
-  arrowFunction: hasType("arrow_function"),
-  functionCall: hasType("call_expression", "new_expression"),
+  ifStatement: typeMatcher("if_statement"),
+  class: matcher(possiblyExportedDeclaration("class_declaration", "class")),
+  statement: matcher(possiblyExportedDeclaration(...STATEMENT_TYPES)),
+  arrowFunction: typeMatcher("arrow_function"),
+  functionCall: typeMatcher("call_expression", "new_expression"),
+  functionName: cascadingMatcher(
```
pokey, line 155: Do you not also support `name` for typescript?

pokey (approved):
@brxck this is awesome!!  great improvement.  I left a bunch of nits.  I don't feel strongly on `findX` vs `xFinder` but I do think we should try to be super consistent about our policy.  Also, I'd update the PR description: def no longer a simple change 😄

src/extension.ts
brxck: Yeah I hear you, this was mostly meant for me at the time of writing. This is probably an artifact of the fact that my first intro to programming around syntax trees was in SML & Racket 😆 I can clean this one up and move it to a separate file.

src/nodeMatchers.ts
brxck: I think it makes more sense to short circuit. Otherwise you could be effectively skipping a finder, which could yield the incorrect node. 

For instance, take this completely made up example. We want to get the return type of a function:

```ts
returnType: composedMatcher([
    findNodeOfType("function_declaration"),
    findNodeOfType("return_type"),
    getTypeNode
])
```

Imagine a situation where we find a declaration but we don't find anything for `findNodeOfType("return_type")`. If we continue to the next finder, we could match any other encountered type (like a parameter type) in the function declaration, returning an incorrect node.

src/nodeMatchers.ts
brxck: I didn't have trouble with the name, but `matchAny` would probably do just as well. I liked cascading because it implied that the matchers would be tried in order.

src/nodeFinders.ts
brxck: This is after prettier! This is how it formats an arrow function implicitly returning another arrow function. If we don't like the look of this I could always throw some curly braces and a return in there.

src/nodeMatchers.ts
brxck: Yeah I would have preferred that too.

I think maybe it can actually be done ([see this PR for examples](https://github.com/microsoft/TypeScript/pull/39094)), but when I tried it seemed to break type inference within the function so I backed off. I did not spend much time looking into this though.

It would look something like this (?):

```ts
export function composedMatcher(
    ...args: [...NodeFinder[], SelectionExtractor]
)
```

src/languages/python.ts
brxck: I doubt it would! I think I just got distracted by what this PR turned into 😅

src/nodeMatchers.ts
pokey: ah oh right you *are* short-circuiting.  Missed that.  I think maybe would be a bit easier to understand if you just had a `return` statement when you get a `null`?  Then it's really obvious you're short-cirtuiting and can remove ternary at bottom

src/nodeMatchers.ts
pokey: Yeah that's fair.  I guess cascading has precedent with "cascading style sheets".  Happy to leave it if you find it clear.  problem with `matchAny` is that it wouldn't play nicely if we align on having `fooMatcher` as convention

src/nodeFinders.ts
pokey: ah right ok.  hmm.  could go either way.  i'm not super used to this style of function but people coming from FP languages prob dream in code like this.  fine with either

src/nodeMatchers.ts
pokey: Hmm I think that's potentially more confusing.  Maybe let's just leave it

src/languages/python.ts
pokey: ha fair enough; there was a lot going on in this PR 😄

pokey:
Do we also want to implement class names as part of this PR or leave that for follow on work?

src/languages/typescript.ts
brxck: It's using the one defined in `getPojoMatchers`. Unsure if it should stay there? Currently typescript is the only language using it, but it's the most generic implementation possible.

pokey: Yeah I don't think it should be in `pojoMatchers`.  "Pojo" stands for "plain old javascript object", so more or less should only contain json stuff.  But if you want to do a rename / refactor I'm fine with it

pokey (commented):
Ok looking really good!  Left a couple more minor comments

src/debug.ts
```diff
@@ -0,0 +1,24 @@
...
+      vscode.window.activeTextEditor!.document.uri,
+      event.selections[0]
+    );
+
+    const ancestors: SyntaxNode[] = [];
+    let node: SyntaxNode = getNodeAtLocation(location);
+    while (node.parent) {
+      ancestors.unshift(node.parent);
+      node = node.parent;
+    }
```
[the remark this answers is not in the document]
brxck: Dang, I forgot about that! I will probably continue to forget about it unless an eslint rule is instituted...

brxck:
Thank you for the solid review @pokey! Think I've addressed everything now.

src/debug.ts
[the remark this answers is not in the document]
pokey: haha yeah good point.  I think [this](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md) would do the trick?

PR #57, opened 2021-07-01
Use web view to compute font size

AndreasArvidsson:


src/Decorations.ts
```diff
@@ -18,16 +19,16 @@ export default class Decorations {
   decorations: NamedDecoration[];
   decorationMap: DecorationMap;
 
-  constructor() {
-    const fontWidthPx = 8.04;
+  constructor(fontSize: FontSize) {
+    console.log(fontSize)
```
pokey, line 23: We'll prob want to either remove or migrate to `console.debug` before merging into master

src/Decorations.ts
```diff
@@ -18,16 +19,16 @@ export default class Decorations {
...
     const iconWidthPx = 4;
-    const iconWidth = `${iconWidthPx}px`;
-    const iconHeight = "1.7em";
+    const iconPadding = 0;
+    const iconWidth = `${iconWidthPx}`;
```
pokey, line 26: Is this idiomatic?  I usually use `toString` but if this is common practice I'm fine with it

src/Decorations.ts
```diff
@@ -18,16 +19,16 @@ export default class Decorations {
...
+    const iconHeight = `${fontSize.fontHeight + iconWidthPx + iconPadding}px`;
```
pokey, line 27: Why add `iconWidthPx` for the height?

src/Decorations.ts
```diff
@@ -18,16 +19,16 @@ export default class Decorations {
   decorations: NamedDecoration[];
   decorationMap: DecorationMap;
 
-  constructor() {
-    const fontWidthPx = 8.04;
+  constructor(fontSize: FontSize) {
+    console.log(fontSize)
     const iconWidthPx = 4;
```
pokey, line 24: do we want this to be fixed or vary with the font size?

src/Decorations.ts
```diff
@@ -18,16 +19,16 @@ export default class Decorations {
...
-    const iconWidth = `${iconWidthPx}px`;
-    const iconHeight = "1.7em";
+    const iconPadding = 0;
+    const iconWidth = `${iconWidthPx}`;
+    const iconHeight = `${fontSize.fontHeight + iconWidthPx + iconPadding}px`;
+    const spanWidthPx = iconWidthPx + (fontSize.fontWidth - iconWidthPx) / 2;
```
pokey, line 28: yessss this is great to have that font size now haha

src/computeFontSize.ts
```diff
@@ -0,0 +1,63 @@
+import * as vscode from "vscode";
+
+export interface FontSize {
+  fontSize: number;
+  fontWidth: number;
+  fontHeight: number;
+}
+
+export default async function (): Promise<FontSize> {
```
pokey, line 9: wow i didn't know you could define an anonymous function for export this way

src/computeFontSize.ts
```diff
@@ -0,0 +1,63 @@
...
+  const fontRatios = await getFontRatios();
+  const fontSize = getFontSize();
+  return {
+    fontSize,
+    fontWidth: fontSize * fontRatios[0],
+    fontHeight: fontSize * fontRatios[1],
+  };
+}
+
+function getFontRatios(): Promise<[number, number]> {
```
pokey, line 19: I'd make a type rather than returning a tuple.  That way you can name them to make it clearer what's being returned, eg `fontWidthToFontSizeRatio`

src/computeFontSize.ts
```diff
@@ -0,0 +1,63 @@
...
+  <body>
+      <h1>Loading Cursorless</h1>
+      <span id="wrapper" style="visibility:hidden; font-size: 1000px; font-family: var(--vscode-editor-font-family);  font-weight: var(--vscode-editor-font-weight);">A</span>
+      <script>
+        const wrapper = document.getElementById("wrapper");
+        const vscode = acquireVsCodeApi();
+        vscode.postMessage([
+          wrapper.offsetWidth / 1000,
+          wrapper.offsetHeight / 1000
+        ]);
```
pokey, line 59: and then you can just return that same dictionary all the way from here rather than a tuple

src/computeFontSize.ts
```diff
@@ -0,0 +1,63 @@
...
+  fontHeight: number;
+}
+
+export default async function (): Promise<FontSize> {
+  const fontRatios = await getFontRatios();
+  const fontSize = getFontSize();
+  return {
+    fontSize,
+    fontWidth: fontSize * fontRatios[0],
+    fontHeight: fontSize * fontRatios[1],
```
pokey, line 15: Cool you get the height now; that's nice

src/extension.ts
```diff
@@ -15,7 +16,8 @@ import {
 import makeGraph from "./makeGraph";
 
 export async function activate(context: vscode.ExtensionContext) {
-  const decorations = new Decorations();
+  const fontSize = await computeFontSize();
```
pokey, line 19: yeah I guess prob better to hoist all that stuff out of the constructor 😂

pokey (approved):
Left a bunch of minor nits but overall this looks awesome!  Other than these nits, I think the things we should consider before merging are:
- Making icon padding a setting
- Caching the font width using `context.globalState`, along with the font family and size so that we can just reuse it if it hasn't changed
- Listening for settings changes to recompute if necessary
- Getting relative path of the icon using `context.asAbsolutePath` instead of the node.js join nonsense
- Making dot size configurable?  If we did that we'd prob want to move to svg wrapper approach so that it can be tweaked more precisely.  Prob not worth doing before merge tho?

src/Decorations.ts
AndreasArvidsson: Well you did it first so I just continued :D

In js you can concatenate numbers and strings so just `iconWidthPx + "px"` is probably easiest.

src/Decorations.ts
AndreasArvidsson: Gonna remove most logs

src/Decorations.ts
AndreasArvidsson: For the dot its the height as well. We have to look more into that tomorrow.

src/Decorations.ts
AndreasArvidsson: Hmm unsure.

src/Decorations.ts
AndreasArvidsson: Oh yes

src/computeFontSize.ts
AndreasArvidsson: Neither did I. I just trying what I think TS should do :D

src/computeFontSize.ts
AndreasArvidsson: Yeah I know. I was just lazy.

src/computeFontSize.ts
AndreasArvidsson: Agreed.

src/computeFontSize.ts
AndreasArvidsson: Yeah did that at the same time.

src/extension.ts
AndreasArvidsson: Hah yeah.  async constructor is problematic also ;)

src/Decorations.ts
pokey: Ah ok so this was a mistake then. You dropped a `px` in that format string

AndreasArvidsson: Ah my bad. Will fix that tomorrow :)

Issue #11, opened 2021-06-23
Add an action for scrolling a line to the top middle or bottom of the screen


pokey:
Can use the built-in vscode command `revealLine` as defined on [this page](https://code.visualstudio.com/api/references/commands)

pokey:
closed by #58

PR #58, opened 2021-07-01
Added actions to scroll target top/center

AndreasArvidsson:
Added two actions to scroll mark to either top or middle of screen/editor.

* What action words should we use? I'm tempted to just use top and center.
*  Right now it only uses the first target. How should we handle additional targets?


AndreasArvidsson:
#11

src/actions/Scroll.ts
```diff
@@ -0,0 +1,117 @@
...
+  async run([targets]: [TypedSelection[]]): Promise<ActionReturnValue> {
+    const selectionGroups = groupBy(
+      targets,
+      (t: TypedSelection) => t.selection.editor
+    );
+
+    const selections = <SelectionWithEditor[]>[];
+    const lines = <LineWithEditor[]>[];
+
+    [...selectionGroups].forEach(([editor, targets]) => {
```
pokey, line 30: I guess you decided `runOnTargetsForEachEditor` was overkill because of the promise stuff?

Also,
```suggestion
    selectionGroups.forEach(([editor, targets]) => {
```

src/actions/Scroll.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  async run([targets]: [TypedSelection[]]): Promise<ActionReturnValue> {
+    const selectionGroups = groupBy(
+      targets,
+      (t: TypedSelection) => t.selection.editor
+    );
+
+    const selections = <SelectionWithEditor[]>[];
+    const lines = <LineWithEditor[]>[];
```
pokey, line 28:
```suggestion
    const selections: SelectionWithEditor[] = [];
    const lines: LineWithEditor[] = [];
```

src/actions/Scroll.ts
```diff
@@ -0,0 +1,117 @@
...
+        editor,
+      });
+    });
+
+    await displaySelectionDecorations(
+      selections,
+      this.graph.editStyles.referencedLine
+    );
+
+    const originalEditor = <TextEditor>window.activeTextEditor;
```
pokey, line 47: why the cast?

src/actions/Scroll.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    await Promise.all(
+      lines.map(async (lineWithEditor) => {
```
pokey, line 50:
```suggestion
      lines.map(async ({lineNumber, editor}) => {
```

src/actions/Scroll.ts
```diff
@@ -0,0 +1,117 @@
...
+    lineNumber = lineNumberDecorate = startLine;
+  } else if (at === "bottom") {
+    // Necessary change to actually get vscode to put line at absolute bottom.function
+    lineNumber = Math.max(0, endLine - 1);
+    lineNumberDecorate = endLine;
+  } else {
+    lineNumber = lineNumberDecorate = Math.floor((startLine + endLine) / 2);
+  }
+
+  return { lineNumber, lineNumberDecorate };
```
pokey, line 111:
```suggestion
  return { lineNumber, lineNumberToDecorate };
```

src/actions/Scroll.ts
```diff
@@ -0,0 +1,117 @@
...
+export class ScrollToBottom extends Scroll {
+  constructor(graph: Graph) {
+    super(graph, "bottom");
+  }
+}
+
+function getLineNumber(targets: TypedSelection[], at: string) {
+  let startLine = Number.MAX_SAFE_INTEGER;
+  let endLine = 0;
+  targets.forEach((t: TypedSelection) => {
```
pokey, line 95: Do you actually need the type annotation here?

src/actions/Scroll.ts
```diff
@@ -0,0 +1,117 @@
...
+    const lines = <LineWithEditor[]>[];
+
+    [...selectionGroups].forEach(([editor, targets]) => {
+      const { lineNumber, lineNumberDecorate } = getLineNumber(
+        targets,
+        this.at
+      );
+      lines.push({ lineNumber, editor });
+      selections.push({
+        selection: new Selection(lineNumberDecorate, 0, lineNumberDecorate, 0),
```
pokey, line 37: If you used [`TextLine`](https://code.visualstudio.com/api/references/vscode-api#TextLine)s instead of raw line numbers, you could maybe use `line.range.start` and `line.range.end` to construct the selection.  Up to you tho; not sure it's much of an improvement but just figured I'd point out the API in case you'd not seen it before

src/actions/Scroll.ts
```diff
@@ -0,0 +1,117 @@
...
+    await displaySelectionDecorations(
+      selections,
+      this.graph.editStyles.referencedLine
+    );
+
+    const originalEditor = <TextEditor>window.activeTextEditor;
+
+    await Promise.all(
+      lines.map(async (lineWithEditor) => {
+        // For reveal line to the work we have to have the correct editor focused
```
pokey, line 51: I appreciate the comment here

src/actions/setSelectionsAndFocusEditor.ts
```diff
@@ -25,3 +25,9 @@ export async function setSelectionsAndFocusEditor(
   editor.selections = selections;
   editor.revealRange(editor.selections[0]);
 }
+
+export async function focusEditor(editor: TextEditor) {
```
pokey, line 29: Would be good to use this function on lines 21-23 above to avoid the duplication, eg `await focusEditor(editor)`

src/editDisplayUtils.ts
```diff
@@ -44,8 +44,34 @@ export default async function displayPendingEditDecorations(
 
   await decorationSleep();
 
-  await runOnTargetsForEachEditor(targets, async (editor, selections) => {
+  await runOnTargetsForEachEditor(targets, async (editor) => {
```
pokey, line 47: I guess typescript doesn't complain about the missing parameter?

src/editDisplayUtils.ts
```diff
@@ -44,8 +44,34 @@ export default async function displayPendingEditDecorations(
...
     editor.setDecorations(tokenStyle, []);
     editor.setDecorations(lineStyle, []);
   });
 }
+
+export async function displaySelectionDecorations(
+  selections: SelectionWithEditor[],
+  lineStyle: TextEditorDecorationType
```
pokey, line 55:
```suggestion
  decorationType: TextEditorDecorationType
```

pokey (approved):
LGTM with a bunch of minor nits.  This feature is great; loving it

src/actions/Scroll.ts
```diff
@@ -0,0 +1,117 @@
...
+        this.at
+      );
+      lines.push({ lineNumber, editor });
+      selections.push({
+        selection: new Selection(lineNumberDecorate, 0, lineNumberDecorate, 0),
+        editor,
+      });
+    });
+
+    await displaySelectionDecorations(
```
pokey, line 42: Ok so after using this for a little bit, I think the UX would be better if we showed the edit decorations first, then scroll, then remove them. We might not even need the decoration sleep timeout. Should be a fairly easy change right?

src/actions/Scroll.ts
AndreasArvidsson: Is just because I don't have a typed selection. I just have the line and the editor and I didn't want to actually recreate entire context

src/actions/Scroll.ts
AndreasArvidsson: Just to make sure it doesn't complain about possible undefined.

src/actions/Scroll.ts
AndreasArvidsson: Thanks but I seen it before. For this use the case since we on really care about one line number not the range I think this worked out well.

src/actions/setSelectionsAndFocusEditor.ts
AndreasArvidsson: Yes it would but I wanted to set all selections before I actually did any scrolling

src/editDisplayUtils.ts
AndreasArvidsson: No unused arguments can be left out just as in regular javascript seems. I'm glad for that otherwise you had make your lambdas much larger for each foreach and filter function.

src/actions/Scroll.ts
pokey: But it might actually be undefined right?  Where does it complain about possibly undefined?

src/actions/setSelectionsAndFocusEditor.ts
pokey: I'm confused.  This function is identical to lines 21-23 above.  Why not just remove those lines and use this function now that we have it?

src/actions/Scroll.ts
AndreasArvidsson: Unlikely but I added a check for it instead of the cast. Is used below when focusing back the original editor.

src/actions/setSelectionsAndFocusEditor.ts
AndreasArvidsson: Then we are two

src/actions/Scroll.ts
AndreasArvidsson: I agree. We still need a sleep timeout because the scrolling itself only takes a few milliseconds and that's not enough to actually see the decoration at all. I implemented it so that time it takes to actually do the proper scrolling is subtracted from the sleep timeout.

src/actions/setSelectionsAndFocusEditor.ts
AndreasArvidsson: Now I get what you mean. Fixed now.

src/actions/Scroll.ts
```diff
@@ -28,42 +28,43 @@ class Scroll implements Action {
...
+      await Promise.all(
+        lines.map(async (lineWithEditor) => {
+          // For reveal line to the work we have to have the correct editor focused
+          if (lineWithEditor.editor !== window.activeTextEditor) {
+            await focusEditor(lineWithEditor.editor);
+          }
+          await commands.executeCommand("revealLine", {
+            lineNumber: lineWithEditor.lineNumber,
+            at: this.at,
+          });
```
pokey, line 52: Is there not a race condition here?  Ie the first `await` yields control, and then the other promise may get to run, so then when we do the `revealLine`, we may be in the wrong editor.  Might want to just run the promises sequentially rather than in parallel

src/actions/Scroll.ts
```diff
@@ -0,0 +1,118 @@
...
+        })
+      );
+
+      // If necessary focus back original editor
+      if (originalEditor && originalEditor !== window.activeTextEditor) {
```
pokey, line 57: Idk you guys both seem to prefer the more compact for this one but I do tend to prefer

```suggestion
      if (originalEditor != null && originalEditor !== window.activeTextEditor) {
```

pokey (commented):
Ok almost there haha just a few more comments

src/actions/Scroll.ts
AndreasArvidsson: Why wouldn't focus editor and reveal line be run in sequence?

pokey: That's how `await` works; the subsequent lines don't run immediately afterwards.  They are essentially put into a `.then()` block, so the other promise will get kicked off in the meantime.  There's no guarantee about which one will resolve first.  So it's possible the other function will pick up first.  Does that make sense?

pokey: Can do something like 

```typescript
for (const lineWithEditor of lines) {
             // For reveal line to the work we have to have the correct editor focused
          if (lineWithEditor.editor !== window.activeTextEditor) {
            await focusEditor(lineWithEditor.editor);
          }
          await commands.executeCommand("revealLine", {
            lineNumber: lineWithEditor.lineNumber,
            at: this.at,
          });
}
```

That will ensure they run in sequence

AndreasArvidsson: Damn you are right. I'll get on that.

AndreasArvidsson: Fixed now. Great that you caught that!

pokey (approved):
few more tiny changes; most can just accept but may need to follow the renames around

PR #64, opened 2021-07-03
Test case recorder

### LINKED ISSUE #59: Make test case recorder
**Goal**: to make it super lightweight to make a new test case so that you don’t think twice about it, while keeping them robust enough not to rot if we rework internals

- New command `cursorless.recordTestCase`, which asks for a test case name / maybe directory
- Waits for next cursorless command
- On next cursorless command, takes a snapshot of current file, including cursor locations, visible lines, file type, and relevant decorations
- Takes snapshot after cursorless command, including cursor and visible lines
- Saves before, command, context, and after as a test case
- Then add code that can parse this type of fixture and run it as a test
- I'd argue that cursorless command should be captured pre-inference, but don't feel strongly
- This technique can be used for testing transformations, inside / outside, actions, etc
- For example, to check whether the delimiter / inside outside stuff is working for eg "type" transformation, can run the command with "take" and "Chuck". This makes it robust to changes in internal representations, at the expense of being slightly less granular
- Also save highlights to make sure those don't break?
- could then also use these test cases to automatically record GIFs to add to the docs. Could do this either by having talon send the full command text and save it with the test case, or build something to generate talon command from json
- For the GIF generation, we could easily automatically find all test cases that use a particular action to put into the doc for that action.  Same with transformations, etc
- Ideas for making a good GIF:
  - making text larger,
  - making command text show a good delay before action,
  - having one command per GIF so you can see the same one over and over,
  - slowing down highlight
  - put something in the gif where you can clearly see when the loop point is (thanks @timo!).  Can possibly use https://github.com/nwtgck/gif-progress
  - have 1 sec delay at start of gif (thanks @timo)
  - use longer `decorationSleepTimeout` than usual
  - somehow call out the decoration that is referenced in the command, possibly by making it bigger?  Could either do this from the start of GIF, or right after they say the command before it executes

- [ ] let's use this one to capture https://github.com/pokey/cursorless-vscode/pull/32#issuecomment-872564310 as a test case

Extra credit is:

- [ ] capture clipboard before and after
- [ ] capture visible lines before and after
- [ ] test ability to work across splits
- [ ] capture highlights

pokey:
Closes #59

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,10 @@
 			{
 				"command": "cursorless.toggleDecorations",
 				"title": "Cursorless: Toggle decorations"
+			},
+			{
+				"command": "cursorless.recordTestCase",
+				"title": "Cursorless: Record test case"
 			}
 		],
 		"colors": [
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -72,7 +72,12 @@ export async function activate(context: vscode.ExtensionContext) {
 
   const graph = makeGraph(graphConstructors);
   var thatMark: SelectionWithEditor[] = [];
-
+  const cursorlessRecordTestCaseDisposable = vscode.commands.registerCommand(
+    "cursorless.recordTestCase",
+    () => {
+      console.log("Hello world");
+    }
+  );
   const cursorlessCommandDisposable = vscode.commands.registerCommand(
     "cursorless.command",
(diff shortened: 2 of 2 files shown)


pokey:
Closed in favour of #87

PR #67, opened 2021-07-04
Fix subword for all caps tokens

### LINKED ISSUE #65: Bug with sub token matcher on all caps snake token
For example with the token `DEFAULT_HAT_WIDTH_TO_CHARACTER_WITH_RATIO`, can't say "take first word harp"

brxck:
This allows matching subwords in all caps snake and kebab cases.

[Here's a demo of the regex.](https://regex101.com/r/lWanj2/2)

Closes #65

### DIFF
--- a/src/test/suite/fixtures/constants.fixture.ts
+++ b/src/test/suite/fixtures/constants.fixture.ts
@@ -52,10 +52,12 @@ export const subwordFixture: Fixture[] = [
     input: "QUICK::BROWN::FOX",
     expectedOutput: ["QUICK", "BROWN", "FOX"],
   },
+  // Not currently handled properly: https://github.com/pokey/cursorless-vscode/issues/79
   {
     input: "APIClientFactory",
     expectedOutput: ["APIClientFactory"],
   },
+  // Not currently handled properly: https://github.com/pokey/cursorless-vscode/issues/79
   {
     input: "MockAPIClientFactory",
     expectedOutput: ["Mock", "APIClientFactory"],
--- a/src/test/suite/fixtures/constants.fixture.ts
+++ b/src/test/suite/fixtures/constants.fixture.ts
@@ -1,0 +1,63 @@
+interface Fixture {
+  input: string;
+  expectedOutput: string[];
+}
+
+export const subwordFixture: Fixture[] = [
+  {
+    input: "QuickBrownFox",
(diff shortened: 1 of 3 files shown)


pokey (commented):
This looks great!  Mind adding some simple unit tests?  Something like these should be pretty easy to use here:
- https://github.com/pokey/cursorless-vscode/blob/svg-hats/src/test/suite/inferFullTargets.test.ts
- https://github.com/pokey/cursorless-vscode/blob/svg-hats/src/test/suite/fixtures/inferFullTargets.fixture.ts

pokey (approved):
Looks great! Might be worth adding a test with something that has a sequence of uppercase eg "APIMatcher". I'm not sure this case is handled correctly by the regex, or even what the desired behaviour is in every case. But maybe worth having a test just to check that it stays the same?

brxck:
> Looks great! Might be worth adding a test with something that has a sequence of uppercase eg "APIMatcher". I'm not sure this case is handled correctly by the regex, or even what the desired behaviour is in every case. But maybe worth having a test just to check that it stays the same?

Yeah it's not properly handled right now.   `APIMatcher` gets treated as a single word. I can imagine that handling that would get pretty hairy. Worth adding a test though, no problem.

src/test/suite/fixtures/constants.fixture.ts
```diff
@@ -0,0 +1,63 @@
...
+    input: "quick::brown::fox",
+    expectedOutput: ["quick", "brown", "fox"],
+  },
+  {
+    input: "QUICK::BROWN::FOX",
+    expectedOutput: ["QUICK", "BROWN", "FOX"],
+  },
+  {
+    input: "APIClientFactory",
+    expectedOutput: ["APIClientFactory"],
```
pokey, line 57: Oh wow it chokes really hard on this.  Didn't think it would be that bad 😅.  Worth at least opening an issue and leaving a comment on this and the next one mentioning the issue

pokey (approved):
Ok looks great left one last nit

src/test/suite/fixtures/constants.fixture.ts
brxck: Yeah and handling it properly will probably be a pain. The current matcher just skips "API" entirely. I've opened #79 and added a comment to these cases.

PR #75, opened 2021-07-06
Fix purple color setting name

brxck:
This got left behind in the rename from mauve to purple. It is currently breaking extension activation.


pokey (approved):

pokey:
Nice catch thanks @brxck

PR #76, opened 2021-07-06
Add a couple more token fixed symbols

pokey:
To support `::` and `=>`


AndreasArvidsson:
Arrow i like, both thin and fat.
Should we just make : repeatable?

PR #87, opened 2021-07-08
Test Case Recorder

brxck:
This is the test case recorder for #59 

Adds a command which records info about the next cursorless command and saves it to a yaml fixture:

```yaml
talonCommand: swap harp with crunch
command:
  actionName: swap
  partialTargets:
    - type: primitive
      mark:
        type: decoratedSymbol
        symbolColor: default
        character: h
    - type: primitive
      mark:
        type: decoratedSymbol
        symbolColor: default
        character: k
  extraArgs: []
languageId: python
targets:
  - type: primitive
    mark:
      type: decoratedSymbol
      symbolColor: default
      character: h
    selectionType: token
    position: contents
    modifier:
      type: identity
    insideOutsideType: inside
  - type: primitive
    mark:
      type: decoratedSymbol
      symbolColor: default
      character: k
    selectionType: token
    position: contents
    modifier:
      type: identity
    insideOutsideType: inside
marks:
  default.h:
    start:
      line: 3
      character: 12
    end:
      line: 3
      character: 17
  default.k:
    start:
      line: 2
      character: 22
    end:
      line: 2
      character: 27
initialState:
  document: |
    import sys

    def hello_world(name="pokey"):
        print(f"hello, {name}")
      
    if __name__ == "__main__":
        hello_world(sys.argv[0])
  selections:
    - active:
        line: 7
        character: 0
      anchor:
        line: 7
        character: 0
  visibleRanges: []
  clipboard: ""
  thatMark: []
finalState:
  document: |
    import sys

    def hello_world(name="hello"):
        print(f"pokey, {name}")
      
    if __name__ == "__main__":
        hello_world(sys.argv[0])
  selections:
    - active:
        line: 7
        character: 0
      anchor:
        line: 7
        character: 0
  visibleRanges: []
  clipboard: ""
  thatMark:
    - active:
        line: 3
        character: 17
      anchor:
        line: 3
        character: 12
    - active:
        line: 2
        character: 27
      anchor:
        line: 2
        character: 22
returnValue: null
```

On recording the user is prompted for the talon command and test fixture file nam

During testing the editor initial state is restored, the recorded command is run, and the result is compared to the recorded one.


pokey:
Great! Would be super helpful to paste an example snapshot yaml into a comment to give a sense of what it looks like

brxck:
Sure thing!

```yaml
command:
  actionName: swap
  partialTargets:
    - type: primitive
      mark:
        type: decoratedSymbol
        symbolColor: default
        character: h
    - type: primitive
      mark:
        type: decoratedSymbol
        symbolColor: default
        character: p
  extraArgs: []
languageId: typescript
decorations:
  default.h:
    start:
      line: 1
      character: 15
    end:
      line: 1
      character: 20
  default.p:
    start:
      line: 4
      character: 12
    end:
      line: 4
      character: 17
initialState:
  document: |
    function helloWorld(name: string) {
      console.log(`hello, ${name}`)
    }

    helloWorld("pokey")
  selections:
    - active: &ref_0
        _line: 5
        _character: 0
      anchor: &ref_1
        _line: 5
        _character: 0
  visibleRanges:
    - start:
        line: 0
        character: 0
      end:
        line: 5
        character: 0
finalState:
  document: |
    function helloWorld(name: string) {
      console.log(`pokey, ${name}`)
    }

    helloWorld("hello")
  selections:
    - active: *ref_0
      anchor: *ref_1
  visibleRanges:
    - start:
        line: 0
        character: 0
      end:
        line: 5
        character: 0
```

And I already see that I should probably look into serializing values instead of the references for snapshot selections 🧐

pokey:
Dude. This. Is. Epic 😂

pokey:
Looks great in yaml. Love how the it uses that block for the doc text. Super readable

pokey:
Gonna save us so much time writing test cases

brxck:
Now able to run some test cases. Plenty to improve here, including the fact that cases with inferred targets seem to throw an error...

For example,

[this line](https://github.com/brxck/cursorless-vscode/blob/a0e99bbf951b2f5b33009198ff7664ff91d53515/src/processTargets.ts#L160) 

```ts
function transformSelection(
  context: ProcessedTargetsContext,
  target: PrimitiveTarget,
  selection: SelectionWithEditor
): { selection: SelectionWithEditor; context: SelectionContext }[] {
  const { transformation } = target;

  switch (transformation.type) {
    case "identity":
      return [{ selection, context: {} }];
    case "containingScope":
      var node: SyntaxNode | null = context.getNodeAtLocation( // !!!
        new vscode.Location(selection.editor.document.uri, selection.selection)
      );
```

throws `Trace: Cannot read property 'rootNode' of undefined` 

on this fixture 🤔  I'm a little stumped at the moment.

```yaml
command:
  actionName: swap
  partialTargets:
    - type: primitive
      transformation:
        type: containingScope
        scopeType: namedFunction
    - type: primitive
      transformation:
        type: containingScope
        scopeType: ifStatement
      mark:
        type: decoratedSymbol
        symbolColor: blue
        character: i
  extraArgs: []
languageId: python
decorations:
  blue.i:
    start:
      line: 5
      character: 0
    end:
      line: 5
      character: 2
initialState:
  document: |
    import sys

    def hello_world(name="pokey"):
        print(f"hello, {name}")
      
    if __name__ == "__main__":
        hello_world(sys.argv[0])
  selections:
    - active:
        line: 2
        character: 20
      anchor:
        line: 2
        character: 20
  visibleRanges:
    - start:
        line: 0
        character: 0
      end:
        line: 7
        character: 0
  clipboard: ""
finalState:
  document: |
    import sys

    if __name__ == "__main__":
        hello_world(sys.argv[0])
      
    def hello_world(name="pokey"):
        print(f"hello, {name}")
  selections:
    - active:
        line: 2
        character: 20
      anchor:
        line: 2
        character: 20
  visibleRanges:
    - start:
        line: 0
        character: 0
      end:
        line: 7
        character: 0
  clipboard: ""
```

src/TestCase.ts
```diff
@@ -0,0 +1,173 @@
...
+    return targetedDecorations;
+  }
+
+  static async getSnapshot(): Promise<Snapshot> {
+    const activeEditor = vscode.window.activeTextEditor!;
+    return {
+      document: activeEditor.document.getText(),
+      selections: activeEditor.selections.map(serializeSelection),
+      visibleRanges: activeEditor.visibleRanges.map(serializeRange),
+      clipboard: await vscode.env.clipboard.readText(),
```
pokey, line 128: I'd be tempted to only bother with the clipboard if it's a copy or paste command

src/TestCase.ts
```diff
@@ -0,0 +1,173 @@
...
+    this.command = command;
+    this.languageId = activeEditor.document.languageId;
+    this.decorations = this.extractTargetedDecorations(targets, navigationMap);
+  }
+
+  extractPrimitiveTargetKeys(...targets: PrimitiveTarget[]) {
+    const keys: string[] = [];
+    targets.forEach((target) => {
+      if (target.mark.type === "decoratedSymbol") {
+        keys.push(`${target.mark.symbolColor}.${target.mark.character}`);
```
pokey, line 86: I tried to use the function in navigation map for this so that we don't duplicate the key generation here

src/TestCase.ts
```diff
@@ -0,0 +1,173 @@
...
+  };
+}
+
+export function serializePosition(position: Position): SerializedPosition {
+  return { line: position.line, character: position.character };
+}
+
+type Command = {
+  actionName: ActionType;
+  partialTargets: PartialTarget[];
```
pokey, line 42: For completeness I'd be tempted to also capture the full target, even if we don't end up using it for anything

src/TestCase.ts
```diff
@@ -0,0 +1,173 @@
...
+    active: serializePosition(selection.active),
+    anchor: serializePosition(selection.anchor),
+  };
+}
+
+export function serializePosition(position: Position): SerializedPosition {
+  return { line: position.line, character: position.character };
+}
+
+type Command = {
```
pokey, line 40: It would be cool to capture the talon command. I wonder if we could use something like keeper to have talon automatically send it over

src/TestCase.ts
```diff
@@ -0,0 +1,173 @@
...
+      languageId: this.languageId,
+      decorations: this.decorations,
+      initialState: this.initialState,
+      finalState: this.finalState,
+    };
+    return yaml.dump(fixture, { noRefs: true });
+  }
+
+  async presentFixture() {
+    // TODO: naming convention for fixture files?
```
pokey, line 161: For now I would just be tempted to prompt the user for a name. In the future we could automatically convert the talon command to camel case or generate it based on the partial target

src/extension.ts
```diff
@@ -82,7 +83,22 @@ export async function activate(context: vscode.ExtensionContext) {
...
   var thatMark: SelectionWithEditor[] = [];
-
+  var recordTestCase = false;
+  const cursorlessRecordTestCaseDisposable = vscode.commands.registerCommand(
+    "cursorless.recordTestCase",
+    () => {
+      console.log("Recording test case for next command");
+      recordTestCase = true;
+      // Empty clipboard to avoid accidentally cluttering fixtures
+      vscode.env.clipboard.writeText("");
```
pokey, line 93: See above comment; I'd argue we should just only grab the clipboard content in the case of a paste or copy

src/extension.ts
```diff
@@ -82,7 +83,22 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    }
+  );
+  const cursorlessSetNavigationMapDisposable = vscode.commands.registerCommand(
+    "cursorless.setNavigationMap",
+    (newNavigationMap: NavigationMap) => {
+      navigationMap = newNavigationMap;
+    }
+  );
```
pokey, line 101: Not sure this one should be a command.  Maybe we could use a mock?  I'd be tempted for now to just check whether the given decoration is the same on the navigation map as it was during recording.  If it is, we don't need to change anything.  If it isn't, we can throw an error.  It's a bit fragile but I believe that color allocation is deterministic, and if it changes / isn't, we will see the error and can do something fancier

The advantage of not touching the token map is that it will make recording gifs easier I believe because then everything will look normal

src/test/suite/inferFullTargets.test.ts
```diff
@@ -3,7 +3,7 @@ import * as assert from "assert";
 import { inferFullTargets } from "../../inferFullTargets";
 import fixture from "./fixtures/inferFullTargets.fixture";
 
-suite("inferFullTargets", () => {
+suite.skip("inferFullTargets", () => {
```
pokey, line 6: We should probably put a big comment above this so that people noticed that it's skipped

src/test/suite/processTargets.test.ts
```diff
@@ -8,7 +8,7 @@ import processTargets from "../../processTargets";
 import { Target, Token } from "../../Types";
 // import * as myExtension from '../../extension';
 
-suite("processTargets", () => {
+suite.skip("processTargets", () => {
```
pokey, line 11: Here too

src/test/suite/recorded.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  const anchor = deserializePosition(selection.anchor);
+  return new vscode.Selection(anchor, active);
+}
+
+suite("recorded test cases", async function () {
+  const directory = path.join(
+    __dirname,
+    // TODO What's the best way to handle this?
+    "../../../src/test/suite/recordings"
+  );
```
pokey, line 39: not saying this is better or worse than what you have here, but [here's](https://github.com/microsoft/vscode-extension-samples/blob/63d86e3a0e459b2adf26487204891f7af62898ce/lsp-sample/client/src/test/helper.ts) one approach that I had bookmarked in case helpful

src/test/suite/recorded.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      const editor = await vscode.window.showTextDocument(document);
+
+      await vscode.env.clipboard.writeText(fixture.initialState.clipboard);
+      editor.selections =
+        fixture.initialState.selections.map(deserializeSelection);
+
+      // TODO restore visible ranges?
+      // Not sure of a straightforward way to do this. Maybe just use to test folding?
+
+      const navigationMap = new NavigationMap();
```
pokey, line 62: See above comment; if we can get away with it it may be worth just asserting that the map matches for the tokens we care about and then hopefully the assert doesn't fail

src/test/suite/recorded.test.ts
```diff
@@ -0,0 +1,95 @@
...
+        const token: Token = {
+          text,
+          range,
+          startOffset: editor.document.offsetAt(range.start) + index,
+          endOffset: editor.document.offsetAt(range.end) + index + text.length,
+          displayLine: value.start.line, // TODO depends on visible ranges? See above
+          editor: vscode.window.activeTextEditor!,
+        };
+        const [color, character] = key.split(".");
+        // @ts-ignore TODO should probably add a decoration color type?
```
pokey, line 76: not sure I follow; there is a [token decoration color type](https://github.com/pokey/cursorless-vscode/blob/ff120118c5ee083559744afccdd53c3a86ff5608/src/constants.ts#L14)

src/test/suite/recorded.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      const fixture = yaml.load(buffer.toString()) as TestCaseFixture;
+
+      await vscode.commands.executeCommand("workbench.action.closeAllEditors");
+      const document = await vscode.workspace.openTextDocument({
+        language: fixture.languageId,
+        content: fixture.initialState.document,
+      });
+      const editor = await vscode.window.showTextDocument(document);
+
+      await vscode.env.clipboard.writeText(fixture.initialState.clipboard);
```
pokey, line 55: I'd prefer to use a mock for this, so that we don't overwrite the real clipboard

pokey (commented):
Nice!!

For that test that's not working, it's possible that the `rootNode` can't be found because the syntax tree extension is not getting a chance to create a syntax tree?  Easiest thing would be to try an async [sleep](https://github.com/pokey/cursorless-vscode/blob/master/src/editDisplayUtils.ts#L7) (eg `await sleep(500)`) right after you open the document. You might also check what that URI is that gets passed into the `Location`. Problem could be that it's an "anonymous" doc. Might see if the way of loading up the fixture from my link in [comment](https://github.com/pokey/cursorless-vscode/pull/87#discussion_r667851032) works any better

brxck:
Thanks @pokey, exactly what I needed to keep this going!

pokey:
> Thanks @pokey, exactly what I needed to keep this going!

Awesome!  btw feel free to ping me on slack if you'd like when you post something in case I miss the email 😊.  Always excited to see progress

src/extension.ts
brxck: We're on the same page that this shouldn't be a command. Was more straightforward than mocking it, so I went with this so I could keep going.

I really like the idea of not having to touch the token map if it's not necessary, so let's go that route.

brxck: What do you think is the best way to you get at the extensions internals for this, in this case `navigationMap`? I could think of a bunch of different ways but none seems much better than another.

pokey: Ah right.  What if we added `navigationMap` to the graph (something like https://github.com/pokey/cursorless-vscode/commit/2963d73c7c71ca4e2a9cf9372a846c8835b271c0)?  Then we could use `makeGraph` to construct a graph in the test, and mock `makeGraph` to return our graph?  Something like that?

brxck: Mocking `makeGraph` has been giving me some trouble, since the mock needs to be in place before the extension is activated -- which happens before any tests are run 🤔

src/TestCase.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  static async getSnapshot(
+    thatMark: SelectionWithEditor[]
+  ): Promise<TestCaseSnapshot> {
+    const activeEditor = vscode.window.activeTextEditor!;
+    return {
+      document: activeEditor.document.getText(),
+      selections: activeEditor.selections.map(serializeSelection),
+      visibleRanges: activeEditor.visibleRanges.map(serializeRange),
+      clipboard: await Clipboard.readText(),
```
pokey, line 162: I think probably better to only capture the clipboard in case of a copy or paste

src/TestCase.ts
```diff
@@ -0,0 +1,241 @@
...
+    const decorationRanges = navigationMap.serializeRanges();
+    const targetedDecorations: DecorationRanges = {};
+    const targetKeys = targets.map(this.extractTargetKeys, this).flat();
+    targetKeys.forEach((key) => {
+      targetedDecorations[key] = decorationRanges[key];
+    });
+    return targetedDecorations;
+  }
+
+  isThatMarkTargeted() {
```
pokey, line 148: This won't work if a "that" mark is in a compound target, right?  I think you need to descend like you do for finding decorated marks

src/NavigationMap.ts
```diff
@@ -43,15 +44,27 @@ export default class NavigationMap {
...
+    return this.map[NavigationMap.getKey(color, character)];
+  }
+
+  public serializeRanges() {
+    const rangeMap: { [coloredSymbol: string]: SerializedRange } = {};
+    Object.entries(this.map).forEach(([key, value]) => {
+      rangeMap[key] = serializeRange(value.range);
+    });
+    return rangeMap;
+  }
```
pokey, line 65: What is this for?  Don't you just serialize the targets that actually get referred to?

src/TestCase.ts
```diff
@@ -0,0 +1,241 @@
...
+    return {
+      document: activeEditor.document.getText(),
+      selections: activeEditor.selections.map(serializeSelection),
+      visibleRanges: activeEditor.visibleRanges.map(serializeRange),
+      clipboard: await Clipboard.readText(),
+      thatMark: thatMark.map((mark) => serializeSelection(mark.selection)),
+    };
+  }
+
+  async getSnapshot(): Promise<TestCaseSnapshot> {
```
pokey, line 167: I think it's a bit confusing to have both this function and the static one

src/TestCase.ts
```diff
@@ -0,0 +1,241 @@
...
+  command: TestCaseCommand;
+  targets: Target[];
+  languageId: string;
+  marks: DecorationRanges;
+  initialState: TestCaseSnapshot;
+  finalState: TestCaseSnapshot;
+  returnValue: any;
+};
+
+export default class TestCase {
```
pokey, line 84: Something doesn't quite feel right here but it's tough to see everything with all the functions in this file.  Maybe let's split things up a bit to get a better feel?

src/TestCase.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    this.talonCommand = talonCommand;
+    this.command = command;
+    this.languageId = activeEditor.document.languageId;
+    this.marks = this.extractTargetedDecorations(targets, navigationMap);
+    this.targets = targets;
+    this.context = context;
+  }
+
+  extractPrimitiveTargetKeys(...targets: PrimitiveTarget[]) {
```
pokey, line 107: I'd get all these primitive target extraction functions into another file

src/TestCase.ts
```diff
@@ -0,0 +1,241 @@
...
+        return [];
+    }
+  }
+
+  extractTargetedDecorations(targets: Target[], navigationMap: NavigationMap) {
+    if (!navigationMap) {
+      return {};
+    }
+
+    const decorationRanges = navigationMap.serializeRanges();
```
pokey, line 139: Seems wasteful / not quite the right separation of concerns to have navigation map dump everything.  Might be better to do `getToken` directly for what you need

src/TestCase.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  async getSnapshot(): Promise<TestCaseSnapshot> {
+    return await TestCase.getSnapshot(this.context.thatMark.get());
+  }
+
+  async saveSnapshot() {
+    const snapshot = await this.getSnapshot();
+
+    if (!["copy", "paste"].includes(this.command.actionName)) {
+      snapshot.clipboard = "";
```
pokey, line 175: i'm not in love with setting things then deleting them later, tho not a showstopper

src/TestCase.ts
```diff
@@ -0,0 +1,241 @@
...
+  navigationMap: NavigationMap;
+};
+
+type TestCaseSnapshot = {
+  document: string;
+  clipboard: string;
+  visibleRanges: SerializedRange[];
+  selections: SerializedSelection[];
+  thatMark: SerializedSelection[] | null;
+};
```
pokey, line 69: I would make `visibleRanges`, `clipboard`, and `thatMark` optional, and then not include them rather than having them be empty / `null`

src/TestCase.ts
```diff
@@ -0,0 +1,241 @@
...
+type TestCaseCommand = {
+  actionName: ActionType;
+  partialTargets: PartialTarget[];
+  extraArgs: any[];
+};
+
+type TestCaseContext = {
+  talonCommand: string;
+  thatMark: ThatMark;
+  targets: Target[];
```
pokey, line 59: It's a funny smell to have `targets` here but the actual cursorless command somewhere else

src/TestCase.ts
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  async saveSnapshot() {
+    const snapshot = await this.getSnapshot();
+
+    if (!["copy", "paste"].includes(this.command.actionName)) {
+      snapshot.clipboard = "";
+    }
+
+    if (!["fold", "unfold"].includes(this.command.actionName)) {
```
pokey, line 178:
```suggestion
if (
      ![
        "fold",
        "unfold",
        "scrollToBottom",
        "scrollToCenter",
        "scrollToTop",
      ].includes(this.command.actionName)
    ) {
```

src/TestCase.ts
```diff
@@ -0,0 +1,241 @@
...
+      clipboard: await Clipboard.readText(),
+      thatMark: thatMark.map((mark) => serializeSelection(mark.selection)),
+    };
+  }
+
+  async getSnapshot(): Promise<TestCaseSnapshot> {
+    return await TestCase.getSnapshot(this.context.thatMark.get());
+  }
+
+  async saveSnapshot() {
```
pokey, line 171: The difference between `saveSnapshot`, the two different `getSnapshot` functions, and `writeSnapshot` is a bit confusing.  Might rework / rename

src/TestCase.ts
```diff
@@ -0,0 +1,241 @@
...
+  async writeFixture(filename: string) {
+    if (filename === "") {
+      throw new Error("Filename required");
+    }
+
+    const fixture = this.toYaml();
+    const workspacePath = vscode.workspace.workspaceFolders?.[0].uri.path;
+    let document;
+
+    if (workspacePath) {
```
pokey, line 223: when will this one be null?

src/extension.ts
```diff
@@ -81,8 +81,33 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    "cursorless.recordTestCase",
+    async () => {
+      console.log("Recording test case for next command");
+
+      const talonCommand = await vscode.window.showInputBox({
+        prompt: "Talon Command",
+        validateInput: (input) =>
+          input.trim().length ? "" : "Missing command",
+      });
+      const filename = await vscode.window.showInputBox({
```
pokey, line 96: if filename missing, maybe just show it in an untitled file like you used to?

pokey, line 96: if filename missing, maybe just show it in an untitled file like you used to?

src/test/suite/recorded.test.ts
```diff
@@ -0,0 +1,98 @@
...
+): vscode.Selection {
+  const active = deserializePosition(selection.active);
+  const anchor = deserializePosition(selection.anchor);
+  return new vscode.Selection(anchor, active);
+}
+
+suite("recorded test cases", async function () {
+  const directory = path.join(__dirname, "../../../testFixtures");
+  const files = await fsp.readdir(directory);
+
```
pokey, line 32: Better to drop this here, so that it still tears down mock if test fails

```ts
  teardown(() => {
    sinon.restore();
  });
```

src/test/suite/recorded.test.ts
```diff
@@ -0,0 +1,98 @@
...
+        ...fixture.command.extraArgs
+      );
+
+      const resultState = await TestCase.getSnapshot(
+        cursorlessApi.thatMark.get()
+      );
+      assert.deepStrictEqual(fixture.finalState, resultState);
+      assert.deepStrictEqual(fixture.returnValue, returnValue);
+
+      sinon.restore();
```
pokey, line 95: Per above, let's put this in a `teardown` func

src/test/suite/recorded.test.ts
```diff
@@ -0,0 +1,98 @@
...
+        fixture.command.actionName,
+        fixture.command.partialTargets,
+        ...fixture.command.extraArgs
+      );
+
+      const resultState = await TestCase.getSnapshot(
+        cursorlessApi.thatMark.get()
+      );
+      assert.deepStrictEqual(fixture.finalState, resultState);
+      assert.deepStrictEqual(fixture.returnValue, returnValue);
```
pokey, line 93: As mentioned above, we'll not include properties in the fixture that aren't relevant, eg clipboard, so we can avoid checking them if not necessary

src/test/suite/recorded.test.ts
```diff
@@ -0,0 +1,98 @@
...
+      if (fixture.initialState.thatMark) {
+        const initialThatMark = fixture.initialState.thatMark.map((mark) => ({
+          selection: deserializeSelection(mark),
+          editor,
+        }));
+        cursorlessApi.thatMark.set(initialThatMark);
+      }
+
+      let clip = fixture.initialState.clipboard;
+      sinon.replace(Clipboard, "readText", async () => clip);
```
pokey, line 67: maybe only mock if clipboard is in fixture

testFixtures/chuckThat.yml
```diff
@@ -0,0 +1,61 @@
...
+        character: 0
+  visibleRanges: []
+  clipboard: ""
+  thatMark:
+    - active:
+        line: 2
+        character: 0
+      anchor:
+        line: 2
+        character: 0
```
pokey, line 60: Is this the correct `thatMark`?  This test didn't work for me

testFixtures/swap.yml
```diff
@@ -0,0 +1,103 @@
...
+    if __name__ == "__main__":
+        hello_world(sys.argv[0])
+  selections:
+    - active:
+        line: 7
+        character: 0
+      anchor:
+        line: 7
+        character: 0
+  visibleRanges: []
```
pokey, line 88: visible ranges broke for me here, but as mentioned we can just leave this property out and not check for it if it's not there

also, we should prob raise an exception in setup if they include visible ranges, because I don't think we properly support it yet

pokey (commented):
Alright cool stuff!  Left a bunch of comments.  Something feels slightly off about the `TestCase` recording setup, but tough to tell with the file so big so let's extract some pieces to separate files and have another look.  I think it's getting close; the meat of the code looks really good it's just some top-level restructuring that might need to happen which should be quick using vscode's refactoring tools

src/TestCase.ts
brxck: You're right, that would feel better. I did it this way because internal map is private, so that the only way to get all the marks out is to try every color/character combo with `getToken`, which also didn't feel right.

src/TestCase.ts
brxck: Not sure I follow. Do you mean you would rather construct test case with a single argument/type ex. `{ command, talonCommand, thatMark, targets }` ?

src/TestCase.ts
brxck: If we are in a fresh, untitled window, ex. `File > New Window`, or if only a file is opened, not a folder.

src/test/suite/recorded.test.ts
brxck: I didn't know the name of the hook! Related, do we want to keep mocha in [BDD mode](https://mochajs.org/#interfaces)?  Sort of threw me cause I'm used to the TDD-style hooks and that's what their documentation uses, but also not a big deal.

testFixtures/chuckThat.yml
brxck: 🤔  it is, I just checked.

src/test/suite/recorded.test.ts
pokey: I'm happy to switch to TDD.  Idk why it is in BDD in the template.  The TDD hooks that the documentation suggests silently disable the test if you use them in BDD 😳.  Took me forever to figure that one out haha

testFixtures/chuckThat.yml
```diff
@@ -0,0 +1,65 @@
+talonCommand: chuck that
+command:
+  actionName: delete
+  partialTargets:
+    - type: primitive
+      mark:
+        type: that
+  extraArgs: []
+languageId: python
+targets:
```
pokey, line 10: I think I might call this `fullTargets` or `inferredFullTargets`? We might add a docstring to that field as well to indicate it's just for why-nots right now

pokey (commented):
This is beautiful. Gotten super crisp. I left a dozen minor nits. I'm happy to knock them out but if you get to them first feel free to take them down

pokey:
And one more  minor one: for location of the yaml files, I'd be inclined to use `src/test/suite/fixtures/recorded`.  Does that make sense?

src/TestCase.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  async recordInitialState() {
+    const excludeFields = this.getExcludedFields();
+    this.initialState = await takeSnapshot(
+      this.context.thatMark,
+      excludeFields
+    );
+  }
+
+  async recordFinalState(returnValue: unknown) {
```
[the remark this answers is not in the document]
brxck: It's basically like a type safe `any`. You can use `unknown` in the same places you would use `any`, but it requires you to assert a type if you want to do anything with it.

src/TestCase.ts
```diff
@@ -0,0 +1,145 @@
...
+      return [target.start, target.end].some(this.includesThatMark, this);
+    }
+    return false;
+  }
+
+  private getExcludedFields() {
+    const excludableFields = {
+      clipboard: !["copy", "paste"].includes(this.command.actionName),
+      thatMark:
+        this.initialState == null &&
```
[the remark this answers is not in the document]
brxck: Yeah looking at this I have to think about it too, maybe I can clarify the code.

If we are taking the first snapshot, then we only want to record `thatMark` when it is targeted. If we're taking the final snapshot, we always want to record `thatMark`.

src/TestCase.ts
```diff
@@ -0,0 +1,145 @@
...
+    this.returnValue = returnValue;
+    this.finalState = await takeSnapshot(this.context.thatMark, excludeFields);
+  }
+
+  async writeFile(filename: string) {
+    const fixture = this.toYaml();
+    const workspacePath = vscode.workspace.workspaceFolders?.[0].uri.path;
+    let document;
+
+    if (workspacePath && filename) {
```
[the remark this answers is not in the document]
brxck: Yeah, wasn't really sure how to handle this. I wasn't assuming we were in the cursorless workspace, and that's why I moved fixtures from `src/test/suite/fixtures/recorded` to a folder in the root.

Now that I'm thinking about it, I could actually check if we're in the cursorless workspace or not. If we're not in the cursorless workspace, just show the yaml in a pane?

src/extractTargetedMarks.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+}
+
+export function extractTargetedMarks(
+  targets: Target[],
+  navigationMap: NavigationMap
+) {
+  if (!navigationMap) {
+    return {};
+  }
```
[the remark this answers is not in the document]
brxck: I don't think it actually does, but that's how it's typed due to the way the extension sets up on activation.

src/makeGraph.ts
```diff
@@ -25,7 +25,7 @@ function makeGetter<GraphType, K extends keyof GraphType>(
   };
 }
 
-export default function makeGraph<GraphType extends object>(
+export function makeGraph<GraphType extends object>(
```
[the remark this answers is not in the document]
brxck: No reason, this is left over from when I was trying to mock `makeGraph`. I can revert it.

src/serializers.ts
```diff
@@ -0,0 +1,46 @@
...
+    active: serializePosition(selection.active),
+    anchor: serializePosition(selection.anchor),
+  };
+}
+
+export function serializePosition(position: Position): SerializedPosition {
+  return { line: position.line, character: position.character };
+}
+
+export function serializeMarks(marks: { [coloredSymbol: string]: Token }) {
```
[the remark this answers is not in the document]
brxck: You're totally right, but I couldn't think of a better thing to call them 😂 `toPojo` is more accurate if less pretty

src/test/suite/recorded.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      if (fixture.initialState.clipboard) {
+        let mockClipboard = fixture.initialState.clipboard;
+        sinon.replace(Clipboard, "readText", async () => mockClipboard);
+        sinon.replace(Clipboard, "writeText", async (value: string) => {
+          mockClipboard = value;
+        });
+      }
+
+      // Wait for cursorless to set up decorations
+      await new Promise((resolve) => setTimeout(resolve, 300));
```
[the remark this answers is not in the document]
brxck: Yeah, I haven't played with the timeout, it could probably be lower. Maybe we add a function/property to see if decorations are finished that we can check periodically?

src/TestCase.ts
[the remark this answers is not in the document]
pokey: Yeah I figured that out later; I'd just pass something into the function?

src/TestCase.ts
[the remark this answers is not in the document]
pokey: TIL thanks

src/TestCase.ts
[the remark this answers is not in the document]
pokey: Yeah that sounds good to me

Btw I'm happy to call this out of scope but from a usability standpoint I think my ideal sequence would be:

1. Ask user for talon command
2. Wait for them to run command
3. Have subfolders in the fixture directory and present them with a list of possible subdirectories and allow them to select "add new" if they type something that doesn't exist
4. Ask them for a file name
5. Display a message that the fixture was made, with a link to open it, but don't open it by default

Note that steps 3–5 only happen if in cursorless workspace

So then what we'd do is to have a command in talon "record", which will grab the parsed command, call record on cursorless, type in literal command, then issue command. So workflow is then:

1. Open a file to test on. We might consider storing a few common ones somewhere
2. Say "record take air"
3. Select subdirectory
4. Say test name
5. Reset file and repeat

Should be able to rapidly record dozens of tests the way. Alternately, we could keep things as they are extension side but ask for dir and name before asking for talon command. Then we'd have two record commands. The first one is eg "setup recording" and the second is "record". So you'd say

1. "setup recording in <dir name> named <name>", which would call extension record and fill out dir and name
2. "record take air", which would type in "take air" literally, then issue the command

Maybe the second approach is better?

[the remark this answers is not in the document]
pokey: Yeah that sounds good to me

Btw I'm happy to call this out of scope but from a usability standpoint I think my ideal sequence would be:

1. Ask user for talon command
2. Wait for them to run command
3. Have subfolders in the fixture directory and present them with a list of possible subdirectories and allow them to select "add new" if they type something that doesn't exist
4. Ask them for a file name
5. Display a message that the fixture was made, with a link to open it, but don't open it by default

Note that steps 3–5 only happen if in cursorless workspace

So then what we'd do is to have a command in talon "record", which will grab the parsed command, call record on cursorless, type in literal command, then issue command. So workflow is then:

1. Open a file to test on. We might consider storing a few common ones somewhere
2. Say "record take air"
3. Select subdirectory
4. Say test name
5. Reset file and repeat

Should be able to rapidly record dozens of tests the way. Alternately, we could keep things as they are extension side but ask for dir and name before asking for talon command. Then we'd have two record commands. The first one is eg "setup recording" and the second is "record". So you'd say

1. "setup recording in <dir name> named <name>", which would call extension record and fill out dir and name
2. "record take air", which would type in "take air" literally, then issue the command

Maybe the second approach is better?

src/extractTargetedMarks.ts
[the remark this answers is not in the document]
pokey: Hmm but the type signature of function doesn't allow undefined or null right?

src/test/suite/recorded.test.ts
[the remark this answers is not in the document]
pokey: Let's leave it for now. They run pretty quick. If it becomes a problem we can do something fancier

testFixtures/chuckThat.yml
brxck: Added a comment and moved it to the end. Should be easy to remove if we decide it's not worth keeping.

brxck:
I hit most of these. I'll see about the recording flow and fixture location later today!

brxck:
@pokey Moved the fixtures and took a shot at the recording flow you described:


>  1. Ask user for talon command
> 
>  2. Wait for them to run command
> 
>  3. Have subfolders in the fixture directory and present them with a list of possible subdirectories and allow them to select "add new" if they type something that doesn't exist
> 
>  4. Ask them for a file name
> 
>  5. Display a message that the fixture was made, with a link to open it, but don't open it by default
> 
> 
> Note that steps 3–5 only happen if in cursorless workspace

pokey:
Nice! Will have one last look tomorrow and get it merged in

Fwiw I am planning to make the following quick changes tomorrow before merging:

- Add an env var that indicates cursorless workspace dir. This env var will be set when you run in debug mode. Then if the var is set it will behave the same as if you're in cursorless workspace. This behaviour will make it trivial to add tests while you're developing a new feature.
- Remove the full target output to make things more compact
- Implement flow style for simple objects to make more compact
- have another look at order of keys in yaml

My goal is to have:

- super compact and readable output for ease of reviewing test cases
- rapid effortless test case recording for new and old features

I'm planning to sit down for a bit tomorrow and will try to knock these out myself. But if I don't, I think I may just file some of them and merge in this PR without.

pokey (approved):

AndreasArvidsson:
May I suggest one improvement? Using `"pre:phrase"` event you can just send the spoken form to cursorless instead of having to type it. If you also use the spoken form as the file name we can skip that step and actually do bulk recordings.

1. `"cursorless record"`
2. popup to chose directory
3. `take air` => file created
4. `clear bat` => file created
5. `"cursorless stop"`

Issue #79, opened 2021-07-07
Subword matching fails on upper case sequences


brxck:
This is a bug knowingly introduced by #67 

Subwords in camel/parcal-cased tokens with a sequence of upper case letters are not distinguished correctly. Anything after a sequence of upper case letters will register as one word.

Visualization of current behavior:

```
APIMatcherFactory -> APIMatcherFactory
```

vs desired behavior:

```
APIMatcherFactory -> API Matcher Factory
```

AndreasArvidsson:
Have a look at the regexp I made for the reformatter:
https://github.com/knausj85/knausj_talon/blob/12229e932d9d3de85fa2f9d9a7c4f31ed6b6445b/code/formatters.py#L270

Specificly:
`
(?<=[A-Z])(?=[A-Z][a-z])
`

pokey:
@AndreasArvidsson sorry my lookahead regex syntax is a bit rusty 😅.  Would that be the whole regex?  I don't see any repeats in there

AndreasArvidsson:
That is not the whole regex. Check the link.

That part just matches where to break when there are sequiential upper case characters.

pokey:
So do we wanna just use that whole regex then?

AndreasArvidsson:
I have to look up how it's used first. This regular expression matches where to break not the tokens itself. So I'm going to have to do some modifying probably.

AndreasArvidsson:
PR #88

pokey:
closed by #88

PR #88, opened 2021-07-09
Updated regular expression for sub word matcher and added additional …

AndreasArvidsson:
…tests

Subword matcher can now correctly match tokens like:
"mockAPIClient123Factory" => ["mock", "API", "Client", "123", "Factory"]

Fixes issue #79

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,4 +1,4 @@
-export const SUBWORD_MATCHER = /([A-Z]?[a-z]+|[A-Za-z]+)/g;
+export const SUBWORD_MATCHER = /[A-Z]?[a-z]+|[A-Z]+(?![a-z])|[0-9]+/g;
 
 export const DEBOUNCE_DELAY = 175;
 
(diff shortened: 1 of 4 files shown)


src/constants.ts
```diff
@@ -1,4 +1,4 @@
-export const SUBWORD_MATCHER = /([A-Z]?[a-z]+|[A-Za-z]+)/g;
+export const SUBWORD_MATCHER = /[A-Z]?[a-z]+|[A-Z]+(?![a-z])|[0-9]+/g;
```
pokey, line 1: @knausj85 shall we steal this regex for `create_spoken_forms`?

pokey (approved):
Legend.

AndreasArvidsson:
@pokey @brxck 
FYI `deepStrictEqual` has arguments `(actual, expected)` which is a bit unusual ordering.

pokey:
> @pokey @brxck
> FYI `deepStrictEqual` has arguments `(actual, expected)` which is a bit unusual ordering.

yeah that's def unexpected but good to know

src/constants.ts
knausj85: Nice, I'll steal this.

PR #108, opened 2021-07-12
Move action

AndreasArvidsson:


src/actions/BringMoveSwap.ts
```diff
@@ -0,0 +1,266 @@
...
+  editor: TextEditor;
+  range: Selection;
+  newText: string;
+  targetsIndex: number;
+  originalSelection: TypedSelection;
+}
+
+interface ThatMarkEntry {
+  editor: TextEditor;
+  targetsIndex: number;
```
pokey, line 32: ohhh this is how you preserve the highlight color in "swap".  Nice

src/actions/BringMoveSwap.ts
```diff
@@ -0,0 +1,266 @@
...
+  private getDecorationStyles(): DecorationTypes {
+    if (this.type === "use") {
+      return {
+        sourceStyle: this.graph.editStyles.referenced,
+        sourceLineStyle: this.graph.editStyles.referencedLine,
+        destinationStyle: this.graph.editStyles.pendingModification0,
+        destinationLineStyle: this.graph.editStyles.pendingLineModification0,
+      };
+    }
+    return {
```
pokey, line 68: For "move", I wonder if we want to use the `delete` style after?  Maybe that's what happens already

src/actions/BringMoveSwap.ts
```diff
@@ -0,0 +1,266 @@
...
+
+        // Add source edit for move and swap
+        // Prevent multiple instances of the same expanded source.
+        if (this.type !== "use" && !usedSources.includes(source)) {
+          let newText = ""; // Defaults to delete source/move
+          if (this.type === "swap") {
+            newText = destination.selection.editor.document.getText(
+              destination.selection.selection
+            );
+          }
```
pokey, line 127: hmm this is one place where the branches I think make things slightly more confusing, but I can definitely follow the code and the comments help a ton so it's probably fine

src/actions/BringMoveSwap.ts
```diff
@@ -0,0 +1,266 @@
...
+            originalSelection: source,
+          });
+        }
+
+        return result;
+      })
+    );
+  }
+
+  private async performDocumentEdits(edits: Edit[]) {
```
pokey, line 144: btw this one is prob worth extracting into a different file in this PR as it's fairly general purpose.  I think there are a couple files that would already be a good fit for it

pokey (approved):
Sooo good!!  Great improvement to code clarity while adding a feature we will use all the time.  Left only a few minor comments; tried to keep the nits to a minimum, tho tbh I think we're converging on a style that we're both happy with so there really weren't many nits to be had 😊

src/Types.ts
```diff
@@ -223,6 +223,7 @@ export interface Action {
 }
 
 export type ActionType =
+  | "bring"
```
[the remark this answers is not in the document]
AndreasArvidsson: Wouldnt be the first time ;)
But yeah I can add both

src/actions/BringMoveSwap.ts
AndreasArvidsson: :)

src/actions/BringMoveSwap.ts
AndreasArvidsson: Ahh no its modify now. I disnt even realize they had different colors ^^

src/actions/BringMoveSwap.ts
AndreasArvidsson: Yeah agree but I think its fine.

src/actions/BringMoveSwap.ts
```diff
@@ -0,0 +1,266 @@
...
+              editBuilder.replace(edit.range, edit.newText);
+            }
+          });
+        });
+      }
+    );
+  }
+
+  private async getThatMark(edits: Edit[]): Promise<ThatMarkEntry[]> {
+    // Only swap has source as a that mark
```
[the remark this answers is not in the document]
AndreasArvidsson: Yes but just a quick one.

src/actions/BringMoveSwap.ts
AndreasArvidsson: Btw after? There are no sources decoration after since the sources are gone. Nothing to decorate.

src/actions/BringMoveSwap.ts
```diff
@@ -0,0 +1,266 @@
...
+    return runForEachEditor(
+      edits,
+      (edit) => edit.editor,
+      async (editor, edits) => {
+        await editor.edit((editBuilder) => {
+          edits.forEach((edit) => {
+            if (edit.range.isEmpty) {
+              editBuilder.insert(edit.range.start, edit.newText);
+            } else {
+              editBuilder.replace(edit.range, edit.newText);
```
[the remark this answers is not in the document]
AndreasArvidsson: Could do that but dont know if it changes anything?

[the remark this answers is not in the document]
pokey: Not sure bout with `delete`, but I know that eg with `insert`, it behaves slightly from a `replace` with empty range:

> Although the equivalent text edit can be made with replace, insert will produce a different resulting selection (it will get moved).
(from [docs](https://code.visualstudio.com/api/references/vscode-api#TextEditorEdit))

src/actions/BringMoveSwap.ts
pokey: yeah was thinking bout that.  I wonder if we should highlight it as a delete before then?

AndreasArvidsson: Agreed

src/actions/BringMoveSwap.ts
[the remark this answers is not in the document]
AndreasArvidsson: Yes I know selected or not. But delete is working fine with replace as is. Gonna think about it.

src/performDocumentEdits.ts
```diff
@@ -0,0 +1,24 @@
+import { Edit } from "./types";
```
brxck, line 1:
```suggestion
import { Edit } from "./Types";
```

This breaks compile on case-sensitive file systems.

Issue #36, opened 2021-06-29
Support "paste" action


pokey:
- Trims clipboard then splits lines
- If number of lines divisible by number of targets, groups clipboard lines per target
- If not divisible, broadcasts clipboard text for each target

PR #115, opened 2021-07-15
Implemented paste action

AndreasArvidsson:
Issue #36


src/actions/Paste.ts
```diff
@@ -0,0 +1,107 @@
...
+    const text = await env.clipboard.readText();
+
+    if (text.length === 0) {
+      throw new Error("Can't paste empty clipboard");
+    }
+
+    const lines = text.trim().split("\n");
+
+    const getText =
+      targets.length === lines.length
```
pokey, line 38: I think we decided to match VSCode behaviour and check that number of lines is divisible by number of targets and if so split clipboard into blocks

src/actions/Paste.ts
```diff
@@ -0,0 +1,107 @@
...
+            return {
+              editor,
+              typedSelection: update(edit.originalSelection, {
+                selection: {
+                  selection: { $set: selection },
+                },
+              }),
+              selection,
+            };
+          });
```
pokey, line 94: This block appears elsewhere right? Might stand nicely on its own as a utility func

src/actions/Paste.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  async run([targets]: [TypedSelection[]]): Promise<ActionReturnValue> {
+    await displayPendingEditDecorations(
+      targets,
+      this.graph.editStyles.pendingModification0,
+      this.graph.editStyles.pendingLineModification0
+    );
+
+    const text = await env.clipboard.readText();
```
pokey, line 29: The clipboard text is also acquired in extension.ts I believe. The reason for this is that cursorless looks at clipboard contents during inference to try to automatically infer that it is a line and then automatically apply that transformation to target

The main goal was to try to mimic vim's behaviour, where if you copy a line, and then paste it somewhere, rather than trying to drop the line into the middle or the line you're on, it will put the pasted line after the current line. 

That was also the reason behind "paste" having an action position preference of "after". This way you could say "copy line air", then "paste bat" and it would paste after line bat. 

But maybe that's too much smarts; simple might be less surprising here. To approximate the behaviour, you'd say "paste after line bat" I guess.

pokey (commented):
Nice! Left a few minor comments. Also, you'll prob want to add the new delimiter insertion behaviour that I tacked onto the "move" PR so that you can say eg "paste after arg air" or "paste after line air" and it will do the right thing

src/actions/Paste.ts
AndreasArvidsson: That's what line 37 does

src/actions/Paste.ts
AndreasArvidsson: Yes I think it would be unexpected if the paste would be after when I wanted it to replace a token. As you say you could always add an after line if you want

src/actions/Paste.ts
AndreasArvidsson: One everything is merged I'm going to try to recombine it into utility/base class. right now that code is not in master yet. This is more less a stop gap to actually get the paste function out there

src/actions/Paste.ts
pokey: That checks equality, not divisibility, no? What am I missing?

pokey:
Also, I don't think this code matches vscode's paste behaviour wrt trimming. But maybe let's file that as a follow-up issue. When we add tests for this action we can test that it matches VSCode on various edge cases

src/actions/Paste.ts
AndreasArvidsson: That is how vscode paste work.

AndreasArvidsson:
What about the trimming is not correct?

src/actions/Paste.ts
pokey: Nope!  Try it out.  Copy 4 lines, then make 2 cursors and paste.  It will split up the 4 lines into 2 blocks of 2

AndreasArvidsson: Not on windows at least. I tried it multiple times before I made implementation and I tried it once again now.

pokey: Haha wait really? That's really odd that that would differ on platforms. I'll try it one more time today; maybe I messed something up in my testing

AndreasArvidsson: Yes. Really strange if it differs.

pokey (approved):

pokey (approved):

Issue #100, opened 2021-07-09
Add "find all" action


pokey:
To open the global search sidebar with the given target pasted in

PR #120, opened 2021-07-16
added find all action

AndreasArvidsson:
issue #100


src/actions/Find.ts
```diff
@@ -0,0 +1,26 @@
+import { ActionReturnValue, Graph, TypedSelection } from "../Types";
+import { commands } from "vscode";
+import GetText from "./GetText";
+
+class FindBase extends GetText {
```
pokey, line 5: Why does this extend `GetText`? If you need to get text, I'd just do it directly, or if you'd like to use the action just call it from the graph. See eg `carve`

src/actions/GetText.ts
```diff
@@ -0,0 +1,35 @@
+import {
+  Action,
+  ActionPreferences,
+  ActionReturnValue,
+  Graph,
+  TypedSelection,
+} from "../Types";
+import displayPendingEditDecorations from "../editDisplayUtils";
+
+export default class GetText implements Action {
```
pokey, line 10: This one will be great to have, eg for "phones"

pokey (commented):
One minor tweak but otherwise looks good!

src/actions/Find.ts
AndreasArvidsson: I was thinking that we could have optional constructor arguments. So a subclass is better in that regard then calling via graph, but for now it doesn't really matter so calling via graph is fine if you prefer that.

src/actions/GetText.ts
AndreasArvidsson: Yes phones and formatters. I find myself missing both.

src/actions/Find.ts
pokey: Sorry I'm not sure I follow. You're not passing any constructor arguments to the base class, and it doesn't have any optional arguments

AndreasArvidsson: Not yet. For phones and such. Was a future provision. But for find it does t matter.

pokey (approved):

Issue #68, opened 2021-07-05
Properly support "puff", "drop", and "float"


pokey:
To add empty lines surrounding, above, and below, respectively, without moving cursor.  For example "puff if" would add newlines above and below the containing if statement without moving the cursor

PR #122, opened 2021-07-17
implemented actions puff, float, drop

AndreasArvidsson:
Issue #68


AndreasArvidsson:
@pokey Could you try to push this one through? the latest changes in talon develop requires this one with the new action names for drink and pour.

pokey (approved):
LGTM but see inline comment re follow-on work

src/actions/InsertEmptyLines.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    for (const edit of edits) {
+      await edit.editor.edit((editBuilder) => {
+        edit.lines.forEach((line) => {
+          editBuilder.insert(new Position(line, 0), "\n");
+        });
+      });
+    }
+
+    const thatMark = targets.map((target) => {
```
pokey, line 56: So I guess it probably makes sense to have "that" refer to the targets as you've done here, tho it might be useful in the future to support something like a "changed" mark, because I could see wanting to refer to the newly insterted lines as well.

Also, this code can be removed once we have `listenForDocumentChanges`, right?

pokey (approved):

pokey (approved):

src/actions/InsertEmptyLines.ts
AndreasArvidsson: My assumption was that if you wanted the new lines you would use drink and pour instead but I see your point.

Yes I believe so.

PR #125, opened 2021-07-19
Added selection type paragraph

AndreasArvidsson:


src/processTargets.ts
```diff
@@ -443,8 +479,12 @@ function getLineSelectionContext(
   selection: SelectionWithEditor,
   selectionContext: SelectionContext
 ): SelectionContext {
-  const document = selection.editor.document;
-  const { start, end } = selection.selection;
+  if (selectionContext.isInDelimitedList) {
+    return selectionContext;
+  }
```
pokey, line 484: We actually do want to clobber the delimiter for "line", because they've explicitly asked for it

src/processTargets.ts
```diff
@@ -458,13 +498,62 @@ function getLineSelectionContext(
 
   const trailingDelimiterRange =
     end.line + 1 < document.lineCount
-      ? new Range(end.line, 0, end.line + 1, 0)
+      ? new Range(end.line, end.character, end.line + 1, 0)
```
pokey, line 501: ha nice catch

src/processTargets.ts
```diff
@@ -458,13 +498,62 @@ function getLineSelectionContext(
...
+  const end = selection.selection.end;
+
+  const leadingLine =
+    start.line > 1 ? start.line - 2 : start.line > 0 ? start.line - 1 : null;
+  const trailingLine =
+    end.line + 2 < document.lineCount
+      ? end.line + 2
+      : end.line + 1 < document.lineCount
+      ? end.line + 1
+      : null;
```
pokey, line 542: I think these should eat all empty lines until next non-empty.  This would be consistent with our space behaviour

src/processTargets.ts
```diff
@@ -474,9 +563,12 @@ function getLineSelectionContext(
...
   );
 
+  const isInDelimitedList =
+    leadingDelimiterRange != null || trailingDelimiterRange != null;
+
   return {
-    isInDelimitedList: true,
-    containingListDelimiter: "\n",
+    isInDelimitedList,
+    containingListDelimiter: isInDelimitedList ? "\n" : undefined,
```
pokey, line 571: should be 
```suggestion
    containingListDelimiter: isInDelimitedList ? "\n\n" : undefined,
```

no?

pokey (commented):
lgtm with some comments

src/processTargets.ts
AndreasArvidsson: That's why more eyes are better :)

src/processTargets.ts
AndreasArvidsson: Yes we should probably remove that if statement

src/processTargets.ts
AndreasArvidsson: Agreed

src/processTargets.ts
AndreasArvidsson: Agreed is just left over from the line context that I copied

Issue #12, opened 2021-06-23
Fix bug with two adjacent tokens


pokey:
Only occurs if you pick a range from the second token to the first

PR #129, opened 2021-07-21
Fixed bug with two adjacent tokens

AndreasArvidsson:
Issue #12


pokey (approved):
Haven't quite wrapped my head around what this one does for overlapping ranges, but looks like it will fix the bug and won't affect non-overlapping ranges so let's ship it

Closes #12

PR #131, opened 2021-07-21
Added support to exclude start or end token in ranges

AndreasArvidsson:
For the `air till bat` commands


src/inferFullTargets.ts
```diff
@@ -269,6 +269,8 @@ export function inferSingleNonListTarget(
       );
       return {
         type: "range",
+        excludeStart: !!target.excludeStart,
+        excludeEnd: !!target.excludeEnd,
```
pokey, line 273:
```suggestion
        excludeStart: target.excludeStart ?? false,
        excludeEnd: target.excludeEnd ?? false,
```

src/processTargets.ts
```diff
@@ -106,6 +128,28 @@ function processSingleRangeTarget(
   });
 }
 
+function getRangePosition(
+  target: TypedSelection,
+  isStart: boolean,
+  exclude: boolean
+) {
```
pokey, line 135: This one could use a rename + docstring.  Not clear to me what it does / what params mean

src/processTargets.ts
```diff
@@ -106,6 +128,28 @@ function processSingleRangeTarget(
...
+  const selection = target.selection.selection;
+  if (exclude) {
+    const outerSelection = target!.selectionContext.outerSelection;
+    const delimiterPosition = isStart
+      ? target!.selectionContext.trailingDelimiterRange?.end
+      : target!.selectionContext.leadingDelimiterRange?.start;
```
pokey, line 141: Let's def get some tests in #59 for this one

src/processTargets.ts
```diff
@@ -106,6 +128,28 @@ function processSingleRangeTarget(
...
+function getRangePosition(
+  target: TypedSelection,
+  isStart: boolean,
+  exclude: boolean
+) {
+  const selection = target.selection.selection;
+  if (exclude) {
+    const outerSelection = target!.selectionContext.outerSelection;
+    const delimiterPosition = isStart
+      ? target!.selectionContext.trailingDelimiterRange?.end
```
pokey, line 140: I don't think you need the `!` on this line or next

pokey (commented):
Looks good with some minor comments.  This range code is def complex haha

PR #136, opened 2021-07-24
A possible approach

pokey:
I gotta sign off for tonight but feel free to carry this one if you think they're nice abstractions

If you like it but won't have a chance I can try to pick it up tomorrow

### DIFF
--- a/src/CalculateChanges.ts
+++ b/src/CalculateChanges.ts
@@ -78,7 +78,7 @@ function updateSelectionInfoMatrix(
 
 /**
  * Takes a list of selections and future edits, and then the client can call
- * the computeChangedOffsets function after applying the edits to get the
+ * the calculateUpdatedSelections function after applying the edits to get the
  * original selections adjusted for the effect of the given edits
  */
 export default class SelectionUpdater {
--- a/src/CalculateChanges.ts
+++ b/src/CalculateChanges.ts
@@ -9,18 +9,18 @@ import {
 } from "vscode";
 import { Edit } from "./Types";
 
-interface MatrixSelection {
+interface SelectionInfo {
   range: Range;
   isReversed: boolean;
   startOffset: number;
   endOffset: number;
 }
 
-function selectionsToMatrix(
+function selectionsToSelectionInfos(
   document: TextDocument,
-  selections: Selection[][]
-): MatrixSelection[][] {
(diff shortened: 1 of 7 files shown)


src/CommandAction.ts
```diff
@@ -42,15 +42,12 @@ export default class CommandAction implements Action {
...
-
-          await commands.executeCommand(this.command);
-
           const [updatedOriginalSelections, updatedTargetSelections] =
-            await promise;
+            await callFunctionAndUpdateSelections(
+              editor,
+              () => commands.executeCommand(this.command),
+              [originalSelections, targetSelections]
+            );
```
pokey, line 50: WDYT of this one?  I didn't implement it but it should be trivial given what you've already built

src/actions/delete.ts
```diff
@@ -32,15 +32,9 @@ export default class Delete implements Action {
...
-          [targets.map((target) => target.selection.selection)],
-          edits
-        );
-
-        await performDocumentEdits(editor, edits);
-
-        const [updatedSelections] = calculateChanges.calculateSelections();
+        const [updatedSelections] = await performDocumentEdits(editor, edits, [
+          targets.map((target) => target.selection.selection),
+        ]);
```
pokey, line 37: was thinking of just baking this stuff into `performDocumentEdits` so that we don't have this multi-step process.  I already implemented it lmk what you think

AndreasArvidsson: Sounds like a good solution.

src/CommandAction.ts
AndreasArvidsson: Unsure. The function itself should work but I don't know if we're going to use it any in any more places. If it is only used in CommandAction I would say that we don't need it.

src/CalculateChanges.ts
```diff
@@ -30,56 +30,72 @@ function selectionsToMatrix(
...
   });
 }
 
-export default class CalculateChanges {
-  document: TextDocument;
-  matrix: MatrixSelection[][];
-  contentChanges: TextDocumentContentChangeEvent[];
+/**
+ * Takes a list of selections and future edits, and then the client can call
+ * the computeChangedOffsets function after applying the edits to get the
```
AndreasArvidsson, line 81: The function is called `calculateUpdatedSelections`

PR #137, opened 2021-07-24
C# support

sterlind:
Requires `vscode-parse-tree` 0.5.0.

I've tested this by hand, but I wasn't sure about how to write unit tests, or if language bindings are actually unit tested.


pokey (approved):
looks great!

PR #143, opened 2021-07-25
Don't do inference from end to start

### LINKED ISSUE #8: Infer type of first swap argument
To be consistent with the way we infer the type of a start of a range based on the type of the end of the range, we could infer the type of the first swap argument from the type of the second argument. For example, "swap with funk gust", can infer that the first argument is the function containing the cursor.

AndreasArvidsson:
`take air past line bat` should not take line for air
`take air past state bat` should not take state for air

With this behavior now consistent we probably can just close #8


AndreasArvidsson:
@brxck @sterlind
What do you guys think about this? @pokey and me have a little bit of different view on this.

I'm of the opinion that you should never do inference from back to front while pokey want it in the case of missing start mark.
eg. `take past air` 
My thoughts is that that is just syntactic sugar for `take this past air` so it's still inference from end to start.

pokey:
So @AndreasArvidsson you're arguing that "take past line air" should select from cursor to the end of line containing 'a', rather than from beginning of current line to end of line containing 'a'? So in your version you'd say "take line past air"? I do feel that while we're using "past", this version is confusing, because it sounds like we want the line after "air". But if we switch to "through" when conformer number bugs are fixed, it's a bit clearer: "take line through air". 

Given that the code is simpler this way, I won't push too strongly against it if people feel it's intuitive

brxck:
I think Andreas's suggestion lines up with what I would intuitively expect. 

"through" would be a nice variation as I've caught myself trying to say it before.

pokey:
Ok cool let's get a couple tests in to make sure I can't sneak my old grammar in here and then merge it in!

AndreasArvidsson:
@pokey Tests added!

pokey (commented):
This looks great. Shall we add "take past line air" and "take line past air" before merging? Maybe I missed those?

AndreasArvidsson:
@pokey Last two test added. Go ahead and merge.

pokey (approved):

PR #151, opened 2021-07-26
Declarative language definition

AndreasArvidsson:
* New declarative syntax for languages
    - [Documentation](https://github.com/pokey/cursorless-vscode/blob/language/src/languages/PATTERN_DOCS.md)
* Standardized scope modifier for array and map `items`
* Added scope modifier `regex`
* Support `arg` in plain JavaScript functions
* Support `item` and `map` / `list` for deconstructed lists and objects.
* Support `funk` on default export function
* Support `string` for template literal strings
* Support `name` for function argument identifier
* Improved error message about missing/unimplemented language or scope types.
* Added language support for java
* Added language support for jsx


src/Types.ts
```diff
@@ -48,21 +48,27 @@ export type ScopeType =
   | "arrowFunction"
   | "class"
   | "className"
+  | "collectionItem"
+  | "collectionKey"
   | "comment"
   | "dictionary"
+  | "xmlElement"
```
pokey, line 55: Presumably this applies to react elements as well. Is there a more generic term?

src/Types.ts
```diff
@@ -48,21 +48,27 @@ export type ScopeType =
...
-  | "pair"
-  | "pairKey"
+  | "regex"
   | "statement"
   | "string"
   | "type"
-  | "value";
+  | "value"
+  | "xmlAttribute"
+  | "xmlElement"
```
pokey, line 68: Looks like this is duplicate of above?

src/Types.ts
```diff
@@ -282,21 +288,26 @@ export interface DecorationColorSetting {
...
  * Returns the desired relative of the provided node.
  * Returns null if matching node not found.
  **/
-export type NodeFinder = (node: SyntaxNode) => SyntaxNode | null;
+export type NodeFinder = (
+  node: SyntaxNode,
+  selection?: vscode.Selection
+) => SyntaxNode | null;
 
 /** Returns a selection for a given SyntaxNode */
```
pokey, line 295: I guess this should now be "one or more selections"?

src/languages/java.ts
```diff
@@ -0,0 +1,76 @@
...
+  class: "class_declaration",
+  className: "class_declaration[name]",
+  namedFunction: ["method_declaration", "constructor_declaration"],
+  functionName: [
+    "method_declaration.identifier!",
+    "constructor_declaration.identifier!",
+  ],
+  ifStatement: "if_statement",
+  string: "string_literal",
+  name: ["*[declarator.name]", "*[name]", "formal_parameter.identifier!"],
```
pokey, line 54: What about 
```suggestion
  name: ["*[declarator][name]", "*[name]", "formal_parameter.identifier!"],
```

I think reserving dot for child is clearer

pokey (commented):
Just had very quick look and left some comments. Will have another look later. Looks like the new syntax is a win!

src/Types.ts
AndreasArvidsson: Yes because jsx is like html a super set of xml so I think the name is fitting.

src/languages/java.ts
AndreasArvidsson: But it is a child. Is(name) the child node of the declarator node

pokey: Right but "name" is the field name not the child type, as a dot usually indicates, no?

AndreasArvidsson: In both cases not just mean a child to a parent. In one case its type in one the case it field names but both are parent child. If you feel strongly about it I can try to fix a regular expression for your format.

pokey: So square brackets and dots always mean children, right? But square bracket indicates a field name, and dot indicates a type name. So I'd stick to that convention if that's the case

AndreasArvidsson: Correct

AndreasArvidsson: Okay get I'll try to get that work

AndreasArvidsson: Fixed!

pokey: Awesome! Looks like it simplified the code a bit too so that's always a good sign

AndreasArvidsson: Yea if you don't count all the time it took to write all those regular expressions ;)

PR #168, opened 2021-07-31
Add surrounding pair modifier (#3)

maciejklimek:
This PR implements the matching modifier.


src/test/suite/fixtures/recorded/surroundingPair/clearMatching14.yml
```diff
@@ -0,0 +1,23 @@
...
+      modifier: {type: surroundingPair, delimiter: null, delimitersOnly: false}
+  extraArgs: []
+marks: {}
+initialState:
+  documentContents: "{const [bongo, baz] = [foo, bar]}"
+  selections:
+    - anchor: {line: 0, character: 17}
+      active: {line: 0, character: 20}
+finalState:
+  documentContents: "{const [] = [foo, bar]}"
```
pokey, line 16: This is the only one I'm not in love with; it really only looks at start of selection, so if end of selection is outside of pair it will still count as containing.  Would be more work to get this case working so I'm not sure it's worth the trouble

pokey (approved):

PR #169, opened 2021-08-01
Converted test recorder for bulk recordings

AndreasArvidsson:


src/extension.ts
```diff
@@ -79,21 +79,33 @@ export async function activate(context: vscode.ExtensionContext) {
...
   const cursorlessRecordTestCaseDisposable = vscode.commands.registerCommand(
     "cursorless.recordTestCase",
     async () => {
-      console.log("Recording test case for next command");
-      testCaseRecorder.start();
+      if (testCaseRecorder.active) {
+        vscode.window.showInformationMessage("Stopped recording test cases");
+        testCaseRecorder.stop();
+      } else {
+        if (await testCaseRecorder.start()) {
```
pokey, line 90: If I've understood code correctly, isn't it an error if `start` returns false? Why not just let it throw an exception instead?

src/TestCaseRecorder.ts
```diff
@@ -128,32 +96,35 @@ export class TestCaseRecorder {
...
+      this.fixtureSubdirectory!
     );
 
     if (!fs.existsSync(targetDirectory)) {
       fs.mkdirSync(targetDirectory);
     }
 
-    this.outPath = path.join(targetDirectory, `${filename}.yml`);
-    return this.outPath;
+    const filename = camelize(testCase.spokenForm);
```
pokey, line 116: Yeah I'd been debating this one. The only problem is that there may be several tests with the same cursorless command. For example, `take funk` is prob what I'd use to test all the different typescript function matchers (eg top-level, class member function, etc). But maybe you have a stop-gap here I've missed?

pokey (commented):
I think this is mostly a win but see my two comments

src/extension.ts
AndreasArvidsson: That's one option. My thought was that that the user may have changed his mind. you don't really start before you have actually selected a folder

src/TestCaseRecorder.ts
AndreasArvidsson: I was thinking that maybe just number them.

src/extension.ts
pokey: Ok I didn't review too closely just seemed like there were lots of falses thrown around where an exception might make more sense eg user is in wrong directory. I think for case where user presses escape agreed we can just silently bail

src/TestCaseRecorder.ts
pokey: Maybe simplest is just to append a number if test with that name exists? Prob not worth the trouble of coming up with better test names, tho I guess we could add an option if we think it would be useful

pokey: Ha ok sounds like we're on the same page 😊

AndreasArvidsson: Yup!

src/extension.ts
AndreasArvidsson: I can add some exceptions for in those cases.

src/test/suite/fixtures/recorded/actions/bringMade.yml
```diff
@@ -0,0 +1,33 @@
...
+      position: contents
+      modifier: {type: identity}
+      insideOutsideType: inside
+  extraArgs: []
+marks:
+  default.m:
+    start: {line: 0, character: 6}
+    end: {line: 0, character: 21}
+initialState:
+  documentContents: "const myVariableValue = \"hello\";\r\n"
```
pokey, line 20: Hmm. So I don't think it's worth re-recording everything, but I believe if you don't use line feeds these will get turned into block output and be a bit nicer to read. At least I think that's what's happening here. Check out the example yamls

src/test/suite/fixtures/recorded/actions/drinkCap.yml
```diff
@@ -0,0 +1,27 @@
...
+  default.c:
+    start: {line: 0, character: 0}
+    end: {line: 0, character: 5}
+initialState:
+  documentContents: "const myVariableValue = \"hello\";\r\n"
+  selections:
+    - anchor: {line: 1, character: 0}
+      active: {line: 1, character: 0}
+finalState:
+  documentContents: "\r\nconst myVariableValue = \"hello\";\r\n"
```
pokey, line 19: Does VSCode auto-detect line ending type? If not then this will fail in some cases I believe

pokey (approved):

PR #171, opened 2021-08-01
Updates to actions and modifiers

AndreasArvidsson:
* Added actions:
    - `call`
    - `sort`
    - `reverse`
    - `replace` both with text and numeric range
    - `reformat`
    - `phones`
    - `paste`
    

* Added modifiers:
    - `head`
    - `tail`
    - `lineNumber`

* Added mark: `source`
* Infer cursor token on modifier `word` and `char`
* Improved decoration on multiple line actions
* Improver delimiter ranges on tokens and sub tokens
* Use built in action for copy and past
* Added tests


src/test/suite/fixtures/recorded/actions/roundWrapVest.yml
```diff
@@ -0,0 +1,30 @@
...
+    const value_hello_str = "Hello world";
+  selections:
+    - anchor: {line: 1, character: 21}
+      active: {line: 1, character: 21}
+finalState:
+  documentContents: |
+
+    const (value_hello_str) = "Hello world";
+  selections:
+    - anchor: {line: 1, character: 22}
```
pokey, line 25: Shouldn't this be 23? Because two chars were inserted to the left of cursor?

AndreasArvidsson: This is on purpose. I made special considerations for the right hand insert in the wrap command. If the right hand insertion is at the cursor don't move the cursor. This is we actually can wrap the cursor and have the curses stay inside the wrapt area afterwards.
`round wrap this` => `(|)`
`round wrap vest` => `(value|)`

pokey:
Did some git surgery and this stuff is all merged in now

PR #172, opened 2021-08-01
Test on PR and push to master

brxck:
This runs tests for linux, mac, and windows on every PR and push to master.

Tests occasionally flake out, it seems like adding a 100ms wait after decoration set up solves this. I've run the tests a few times successfully now. The timeout could be smaller, but things are pretty bad on the macos runner which is pretty dang slow. I'm guessing because of this:
```
Unable to create basic Accelerated OpenGL renderer.

Core Image is now using the software OpenGL renderer. This will be slow.
```

[From what I've been able to read](https://github.com/actions/virtual-environments/issues?q=is%3Aissue+sort%3Aupdated-desc+macos+opengl+is%3Aclosed), it doesn't seem like we're going to be able to do much about that...?

### DIFF
--- a/src/test/suite/recorded.test.ts
+++ b/src/test/suite/recorded.test.ts
@@ -33,7 +33,7 @@ suite("recorded test cases", async function () {
     __dirname,
     "../../../src/test/suite/fixtures/recorded"
   );
-  const files = await walkFilesSync(directory);
+  const files = walkFilesSync(directory);
 
   teardown(() => {
     sinon.restore();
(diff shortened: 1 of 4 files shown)


pokey (approved):

src/test/suite/recorded.test.ts
```diff
@@ -82,6 +82,7 @@ suite("recorded test cases", async function () {
 
       // Wait for cursorless to set up decorations
       cursorlessApi.addDecorations();
+      await new Promise((resolve) => setTimeout(resolve, 100));
```
pokey, line 85: this shouldn't be necessary; let's figure out how to remove it in a follow-up

AndreasArvidsson:
@brxck Is it only mac that needs the sleep time? If that's the case we could probably do something like this.
```
if (os.platform() === "darwin") {
        await new Promise((resolve) => setTimeout(resolve, 100));
 }
```

PR #177, opened 2021-08-04
Added documentation for test recorder

AndreasArvidsson:


pokey (commented):
I would normally think to put both those docs in the `docs/`, directory, but I don't feel strongly. We should either put them there and have a simple link readme from here, or make sure to link here from there. If the latter, we could prob merge this one on and I can add a link from docs in a follow up PR

Btw does the test case runner try to run this readme as a test case? I thought it just walked all files regardless of extension but could be wrong

AndreasArvidsson:
I thought of that and I agree with you so I moved everything to the docs folder now.

pokey (approved):

PR #178, opened 2021-08-04
Added tests for head, tail and line numbers

AndreasArvidsson:


src/test/suite/fixtures/recorded/headTail/takeHead.yml
```diff
@@ -0,0 +1,31 @@
...
+      active: {line: 1, character: 9}
+finalState:
+  documentContents: |
+
+    const value = "Hello world";
+
+    const value = "Hello world";
+  selections:
+    - anchor: {line: 1, character: 9}
+      active: {line: 1, character: 0}
```
pokey, line 27: Do we want these reversed? I can see the logic, but I'd imagine it's more common to want it forward, so would prob make sense to have that be the default

src/test/suite/fixtures/recorded/lineNumbers/takeRowTwoPastDownThree.yml
```diff
@@ -0,0 +1,35 @@
...
+      active: {line: 0, character: 0}
+finalState:
+  documentContents: |
+
+    const value = "Hello world";
+
+    const value = "Hello world";
+  selections:
+    - anchor: {line: 1, character: 0}
+      active: {line: 3, character: 28}
```
pokey, line 31: So "down three" is a mark, not a modifier? I guess fine for now but I'd argue that long term we want it as a modifier

pokey (approved):
This LGTM but we should potentially file issues for those two behaviours I call out below if you agree they should change

src/test/suite/fixtures/recorded/headTail/takeHead.yml
AndreasArvidsson: I really like them like they are to be honest. Its define as past start of line after all.

src/test/suite/fixtures/recorded/lineNumbers/takeRowTwoPastDownThree.yml
AndreasArvidsson: Yes this is just a stopgap intill we have the full "next" syntax working.

src/test/suite/fixtures/recorded/headTail/takeHead.yml
AndreasArvidsson: And to be real these modifers are mostly usefull for chuck or clear. With a take it doesn't really matter where the active is because its gonna get overewritten if you insert anything.

pokey: Fair enough; if they're working well for you this way and it's simpler let's leave them!

PR #180, opened 2021-08-05
Treat line numbers as proper marks

AndreasArvidsson:


src/test/suite/fixtures/recorded/lineNumbers/takeRowFour.yml
```diff
@@ -5,7 +5,7 @@ command:
   partialTargets:
     - type: primitive
       selectionType: line
-      modifier:
```
pokey, line 8: shall we add a test case that uses it like a proper mark?  Eg "take funk row five"

src/test/suite/fixtures/recorded/lineNumbers/takeRowFour.yml
```diff
@@ -5,7 +5,7 @@ command:
   partialTargets:
     - type: primitive
       selectionType: line
```
pokey, line 7: So this `selectionType` comes from talon side?  Probably ok but came as a bit of a surprise.  I wonder if we could infer that info somewhere in the extension, either during inference or processing

pokey (commented):
LGTM with a couple minor comments

src/test/suite/fixtures/recorded/lineNumbers/takeRowFour.yml
AndreasArvidsson: We don't really have that for the other marks. There is no "take funk that" or similar. But it is treated like the other marks in the code so there really is no difference. If we should add more test for marks in general is a different question.

src/test/suite/fixtures/recorded/lineNumbers/takeRowFour.yml
AndreasArvidsson: We could enter it extension side if we want. But that is true for several other commands as well. This solution has the upside that we don't add additional code if I don't actually need to. Instead we reuse the functionality of the line selector and the mark actually only selects the first character of the correct line.

src/test/suite/fixtures/recorded/lineNumbers/takeRowFour.yml
pokey: I think I'd like to have that kind of test, but if you think it's better to do those in one pass that's fine too; we don't need to block this PR on that

AndreasArvidsson: Exactly I think that we probably should do proper tests for all the marks as a separate thing and not just this one specifically.

pokey (approved):

PR #194, opened 2021-08-06
Updated how line decorations are shown

AndreasArvidsson:
* Line decorations are now always shown for the correct line
* Decorations for line/paragraph with position before/after uses non line decoration. 
    - This means that the decorations aren't shown for empty lines
* Before line and paragraph are adjusted to not include indentations on line with text
    - `chuck before line` keeps indentation
* Fixed line decoration on float command

### DIFF
--- a/src/editDisplayUtils.ts
+++ b/src/editDisplayUtils.ts
@@ -63,14 +63,14 @@ export default async function displayPendingEditDecorations(
           const { start, end } = selection.selection.selection;
           const startLine = document.lineAt(start);
           const hasLeadingLine = start.character === startLine.range.end.character;
-          if (end.character === 0 && (!leadingLine || start.character === 0)) {
+          if (end.character === 0 && (!hasLeadingLine || start.character === 0)) {
             // NB: We move end up one line because it is at beginning of
             // next line
             return selection.selection.selection.with({
               end: end.translate(-1),
             });
           }
-          if (leadingLine) {
+          if (hasLeadingLine) {
             // NB: We move start down one line because it is at end of
             // previous line
             return selection.selection.selection.with({
--- a/src/editDisplayUtils.ts
+++ b/src/editDisplayUtils.ts
@@ -59,23 +59,23 @@ export default async function displayPendingEditDecorations(
       selections
         .filter((selection) => isLineSelectionType(selection.selectionType))
         .map((selection) => {
-          const start = selection.selection.selection.start;
-          const startLine = selection.selection.editor.document.lineAt(start);
-          if (start.character === startLine.range.end.character) {
-            return selection.selection.selection.with(
-              // NB: We move start down one line because it is at end of
-              // previous line
-              selection.selection.selection.start.translate(1),
-              undefined
-            );
+          const { document } = selection.selection.editor;
+          const { start, end } = selection.selection.selection;
+          const startLine = document.lineAt(start);
+          const leadingLine = start.character === startLine.range.end.character;
+          if (end.character === 0 && (!leadingLine || start.character === 0)) {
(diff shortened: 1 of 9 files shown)


src/editDisplayUtils.ts
```diff
@@ -59,23 +59,23 @@ export default async function displayPendingEditDecorations(
...
-            return selection.selection.selection.with(
-              // NB: We move start down one line because it is at end of
-              // previous line
-              selection.selection.selection.start.translate(1),
-              undefined
-            );
+          const { document } = selection.selection.editor;
+          const { start, end } = selection.selection.selection;
+          const startLine = document.lineAt(start);
+          const leadingLine = start.character === startLine.range.end.character;
```
pokey, line 65: I'd rename to
```suggestion
          const hasLeadingLine = start.character === startLine.range.end.character;
```

src/editDisplayUtils.ts
```diff
@@ -59,23 +59,23 @@ export default async function displayPendingEditDecorations(
...
+          if (end.character === 0 && (!leadingLine || start.character === 0)) {
```
pokey, line 66: My understanding is that start and end will always be at the start or end of a line. This if start isn't at the end of a line (ie `!hasLeadingLine`), then it must be at the beginning of a line. So isn't this equivalent to

```suggestion
          if (end.character === 0) {
```

AndreasArvidsson: No start can be end of line and vise versa due to delimiter ranges.

pokey: Also, I don't think this code will work if start of selection at the end of a line with content, and end of selection is beginning of next line, which is empty. In that case, I believe it will highlight the line with content, but effectively delete the empty line below. Ie try "chuck after line air" where "air" is on a line with content that has an empty line below, like we tried earlier. Haven't tried this myself tho so could be wrong

AndreasArvidsson: Everything seems to work when I try it out.

pokey: > No start can be end of line and vise versa due to delimiter ranges. 

Right but start will always be either the beginning or end of a line so I believe my statement holds

Also see my other comment above

AndreasArvidsson: The code now works if I do the changes you request it doesn't any more when removing the last line. Please try it out for yourself.

pokey: Ok yes I had the logic slightly off.  But this code is broken in a couple of cases:

1. In the following file:

        hello world
        
        
        hello

    Saying "chuck after block whale" highlights final "hello" line when it shouldn't.

2. In the following file:

        hello world
        hello

    Saying "chuck after line drum" highlights final "hello" line when it shouldn't.

I think the fundamental issue here is that "after line" probably shouldn't be treated like a line selection type because it is really targeting a sub-line delimiter range

AndreasArvidsson: After is something different and the code doest really support it properly. But chuck line should work fine now.

src/processTargets.ts
```diff
@@ -18,6 +18,7 @@ import {
 import { performInsideOutsideAdjustment } from "./performInsideOutsideAdjustment";
```
pokey, line 18: Looks like there are some changes to behaviour here so would be great to add a test or two to capture what's changed

pokey (commented):
Looks good but please add tests for behaviour change

src/processTargets.ts
AndreasArvidsson: Tests added

pokey (approved):

PR #203, opened 2021-08-07
Fix breakpoints

### LINKED ISSUE #195: Breakpoint action not working on mac
Not sure exactly why we'll need to investigate

pokey:
Closes #195

### DIFF
--- a/src/actions/SetBreakpoint.ts
+++ b/src/actions/SetBreakpoint.ts
@@ -3,25 +3,18 @@ import {
   ActionPreferences,
   ActionReturnValue,
   Graph,
+  SelectionWithEditor,
   TypedSelection,
 } from "../Types";
-import {
-  SourceBreakpoint,
-  Location,
-  debug,
-  Uri,
-  Range,
-  Breakpoint,
-} from "vscode";
+import { SourceBreakpoint, Location, debug, Range, Breakpoint } from "vscode";
 import displayPendingEditDecorations from "../editDisplayUtils";
-import { isEqual, range, uniqWith } from "lodash";
 
-function getBreakpoint(uri: Uri, line: number) {
+function getBreakpoint(location: Location) {
   return debug.breakpoints.find((breakpoint) => {
     if (breakpoint instanceof SourceBreakpoint) {
       return (
-        breakpoint.location.uri.toString() === uri.toString() &&
-        breakpoint.location.range.start.line === line
+        breakpoint.location.uri.toString() === location.uri.toString() &&
+        breakpoint.location.range.isEqual(location.range)
       );
     }
     return false;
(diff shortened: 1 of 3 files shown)


pokey:
I am worried that removal will be a bit brittle. I wonder if we should have a "clear breakpoint" action that removes all breakpoints in a given range

src/actions/SetBreakpoint.ts
```diff
@@ -44,29 +37,19 @@ export default class SetBreakpoint implements Action {
...
-
     const toAdd: Breakpoint[] = [];
     const toRemove: Breakpoint[] = [];
 
-    lines.forEach(({ uri, line }) => {
-      const existing = getBreakpoint(uri, line);
+    targets.forEach((target) => {
+      const location = new Location(
+        target.selection.editor.document.uri,
+        target.selection.selection.start
```
AndreasArvidsson, line 46: If you want to be consistent with how vscode does it internally all breakpoints should be added to character 0.
With your current changes you can have multiple breakpoints on the same line.

After some testing I've came to a conclusion that it's only the active part of the selection that determines the line and then the character is zero. `new Position(target.selection.selection.active.line, 0)` is probably better solution.

pokey: @AndreasArvidsson That's not my understanding. See https://code.visualstudio.com/docs/editor/debugging#_inline-breakpoints

AndreasArvidsson: Thats the result of issuing the breakpoint command and printing the list. Always char 0. And how could you have multiple breakpoints on the same line? 

Thats why my code only checked line and didnt care at all about char.

pokey: According to the link above:

> Inline breakpoints will only be hit when the execution reaches the column associated with the inline breakpoint. This is particularly useful when debugging minified code which contains multiple statements in a single line.
>
> An inline breakpoint can be set using ⇧F9 or through the context menu during a debug session. Inline breakpoints are shown inline in the editor.
>
> Inline breakpoints can also have conditions. Editing multiple breakpoints on a line is possible through the context menu in the editor's left margin.

But mainly it was that your version didn't work for me for some reason.  Does this new code work on your box?  If so I think I'd be inclined to support these inline breakpoints given that the code is simpler and vscode supports them out of the box

AndreasArvidsson: This has several edge cases doesn't work for me

1. You must know which token the breakpoint is added to otherwise we just add another instead of removing it
2. Breakpoints added using the built in breakpoint command can't be removed if there is not an token on the first character

I wouldn't say that the code is simpler because you use character instead of zero.

pokey: > This has several edge cases that doesn't work for me
> 
> 1. You must know which token the breakpoint is added to otherwise we just add another instead of removing it

VSCode displays a red dot inline where the breakpoint is added so you can easily see which token it is

> 2. Breakpoints added using the built in breakpoint command can't be removed if there is not an token on the first character

I think this and the above could be solved by having "break add" and "break clear" as separate commands.  Then you could do "break clear line air".  I guess if the common case is using lines, we could have this action have a modifier preference of "line", and user can use "token" to override, so that "break clear air" deletes any breakpoints on the line "air"

> 
> I have no idea how you would even use multiple breakpoints on the same line. For me a breakpoint is one per line.

It breaks when it gets to that column.  If it's a long line, that can be helpful

AndreasArvidsson: Then I must ask have you ever used two breakpoints on one line? Having to manually clear the breakpoints is not a good user experience in my mind and I think we should treat breakpoints as we do with comments because that is how people normally use them I believe. If you like we can always have a setting but I really think that one breakpoint per line should be the default.

The modifier token or line wouldn't really help with that since they will only change the start character if you don't want the action to actually be aware of selection types which no other action is as far as I know.

pokey: Ok I tested the mainline breakpoint action again and it seems to work fine for me now.  I guess must have been something flakey in VSCode's breakpoint impl.  Let's leave existing for now as I agree inline breakpoints is prob more of a niche requirement.  If things act weird again we can revisit, tho I'm not convinced anything I did here actually fixed the problem; I think more likely was restarting VSCode that helped 😂

AndreasArvidsson: I have found that right after starting the debug session the list of breakpoints is empty for the first call. Subsequent calls work. I have no idea why it this way.

PR #227, opened 2021-08-13
Check ancestors for every

### LINKED ISSUE #220: "take every key" and "take every value" are broken
For example:

```typescript
const someMap = {
  key1: "hello",
  key2: "there",
};
```

Saying "take every key one" or "take every value one" results in error `Require at least one target with this action`

AndreasArvidsson:
fixes #220

Added a few tests

### DIFF
--- a/src/languages/index.ts
+++ b/src/languages/index.ts
@@ -59,19 +59,38 @@ function matcherIncludeSiblings(matcher: NodeMatcher): NodeMatcher {
     selection: SelectionWithEditor,
     node: SyntaxNode
   ): NodeMatcherValue[] | null => {
-    const matches = matcher(selection, node);
+    let matches = matcher(selection, node);
     if (matches == null) {
       return null;
     }
-    return matches
-      .flatMap((match) =>
-        match.node.parent!.namedChildren.flatMap((sibling) =>
-          matcher(
-            selectionWithEditorFromRange(selection, match.selection.selection),
-            sibling
-          )
-        )
+    matches = matches.flatMap((match) =>
+      findNearestContainingAncestorNode(
+        match.node,
+        selectionWithEditorFromRange(selection, match.selection.selection),
+        matcher
       )
-      .filter((match) => match != null) as NodeMatcherValue[];
+    ) as NodeMatcherValue[];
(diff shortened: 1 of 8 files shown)


src/test/suite/fixtures/recorded/languages/typescript/takeEveryItem5.yml
```diff
@@ -0,0 +1,45 @@
...
+    };
+  selections:
+    - anchor: {line: 1, character: 4}
+      active: {line: 1, character: 17}
+    - anchor: {line: 2, character: 4}
+      active: {line: 2, character: 17}
+    - anchor: {line: 1, character: 4}
+      active: {line: 1, character: 17}
+    - anchor: {line: 2, character: 4}
+      active: {line: 2, character: 17}
```
pokey, line 35: Is this really what we want? Appears selections are duplicated

src/languages/index.ts
```diff
@@ -59,19 +59,38 @@ function matcherIncludeSiblings(matcher: NodeMatcher): NodeMatcher {
...
-      .filter((match) => match != null) as NodeMatcherValue[];
+    ) as NodeMatcherValue[];
+    if (matches.length > 0) {
+      return matches;
+    }
+    return null;
   };
 }
+
+function findNearestContainingAncestorNode(
```
pokey, line 80: I'd rename and add a docstring, because the name is a bit misleading to me. Maybe `iterateNearestIterableAncestor` or something?

pokey (commented):
Overall looks good with a few comments. Also might add one test with nesting. Eg `foo(bar(baz, bongo), bazman)`

src/test/suite/fixtures/recorded/languages/typescript/takeEveryItem5.yml
AndreasArvidsson: Agreed,  but has actually nothing to do with this pr. transformSelection/getNodeMatcher gets called for every selection. But might be a good idea to actually filter out duplicate selections at the end of the pipeline. You should probably put that on the task list for the new grammar.

pokey (approved):

PR #229, opened 2021-08-14
Bugfixes: past token, past end of, subtoken out of range, sort tokens

### LINKED ISSUE #198: Rename `xmlAttribute` scope type to `attribute`
So we can use it more generally than just for xml attributes, eg in C++

### LINKED ISSUE #200: Containing token inference doesn't work next to single-character token
When there are two tokens next to each other and you use a subword modifier, cursorless will assume that you are referring to the token to the right of your cursor. However, if the token to the right of your cursor is a single character, then this will not be what you want. It might make more sense to assume the user is referring to token to the left in this case. For example

```typescript
fooBar;
```

If the cursor is before the `;`, the command "take first word" doesn't work.

### LINKED ISSUE #201: Improve error messages for sub token modifiers
For example, calling "take third word" on `helloWorld` results in the error message `Cannot read property 'start' of undefined`.

### LINKED ISSUE #209: "Chuck past end of T" removes trailing delimiter
"end of" is not supposed to include delimiter; that would be expected behaviour for "chuck past after T"

### LINKED ISSUE #221: "token" modifier doesn't work with ranges
For example,

```
hello there
```

With a cursor in the middle of "hello", saying "take token past trap" should select both words, but instead starts from current cursor position rather than expanding to start token.

### LINKED ISSUE #222: "Take past end of token" doesn't work
Position cursor in the middle of a token and say "take past end of token". Observe that nothing happens

AndreasArvidsson:
fixes #221 
fixes #222
fixes #201
fixes #209 
fixes #200 
fixes #198


src/inferFullTargets.ts
```diff
@@ -377,6 +380,7 @@ export function inferRangeEndTarget(
 
   const mark =
     target.mark ??
+    (target.selectionType === "token" ? CURSOR_MARK_TOKEN : null) ??
```
pokey, line 383: This is not quite right. It won't work for "take second char trap past end of token", which should refer to the end of "token trap" rather than "token this", to be consistent with eg "take trap past end of line"

src/NavigationMap.ts
```diff
@@ -65,8 +65,36 @@ export default class NavigationMap {
   }
 
   public getTokenForRange(range: Range) {
-    return Object.values(this.map).find(
+    const matches = Object.values(this.map).filter(
       (token) => token.range.intersection(range) != null
     );
+
+    // If multiple matches sort and take the first
```
pokey, line 72: I would expand to containing tokens, not just take the first.  Eg if anchor and active fall in the middle of tokens, return beginning / end of anchor to beginning / end of active, depending whether reversed

src/test/suite/fixtures/recorded/surroundingPair/takeFirstChar.yml
```diff
@@ -0,0 +1,24 @@
+spokenForm: take first char
```
pokey, line 1: I think this file is in the wrong directory, no?

src/test/suite/fixtures/recorded/surroundingPair/takeFirstChar5.yml
```diff
@@ -0,0 +1,24 @@
...
+marks: {}
+initialState:
+  documentContents: //**
+  selections:
+    - anchor: {line: 0, character: 2}
+      active: {line: 0, character: 2}
+finalState:
+  documentContents: //**
+  selections:
+    - anchor: {line: 0, character: 0}
```
pokey, line 19: I might have expected it to pick `**`, tho this one is def a corner case

pokey:
Ok nice fixes; left a few comments tho.  Also would be good to add a test with same initial state as "take token past trap" (eg cursor in middle of first token), but say "take past trap" to make sure it doesn't expand to first token.  Could see us messing that one up down the road if there's no test 😅

pokey (commented):
Ok nice fixes; left a few comments tho. Also would be good to add a test with same initial state as "take token past trap" (eg cursor in middle of first token), but say "take past trap" to make sure it doesn't expand to first token. Could see us messing that one up down the road if there's no test 😅

src/NavigationMap.ts
AndreasArvidsson: This has nothing to do with range it's just if the cursor is between two tokens then we have to select one.

src/test/suite/fixtures/recorded/surroundingPair/takeFirstChar.yml
AndreasArvidsson: Correct. Eye tracker error
I'm going to add the directory name to the message

src/test/suite/fixtures/recorded/surroundingPair/takeFirstChar5.yml
AndreasArvidsson: If all equal it chooses the one to the left. Of course we could continue to tweak these corner cases but I think this should work for all practical cases.

src/inferFullTargets.ts
AndreasArvidsson: I think combining sub token with a range is just incorrect. Sub token should be defined as a single token.

src/NavigationMap.ts
pokey: Yes but if your selection spans across two tokens, "take token" should expand to the containing tokens.  This code will cause it to select only the first token

src/test/suite/fixtures/recorded/surroundingPair/takeFirstChar5.yml
pokey: Yeah kind of a toss up.  I'd assume it picks token it's at the start of, but I don't think there's much of a convention here.  This decision is probably fine

src/inferFullTargets.ts
pokey: - [ ] add test

pokey (approved):
Pushed a couple changes but otherwise looks good to me!

PR #234, opened 2021-08-14
Added action insert

### LINKED ISSUE #219: Implement delimiter generators for "change"
Would be "pair X", eg "change pair square pair round"

Impl would send an object `{pair: "parentheses"}` and then extension side would do division etc

AndreasArvidsson:
fixes #219

### DIFF
--- a/src/actions/Insert.ts
+++ b/src/actions/Insert.ts
@@ -6,7 +6,7 @@ import {
   TextGenerator,
   TypedSelection,
 } from "../typings/Types";
-import displayPendingEditDecorations from "../util/editDisplayUtils";
+import { displayPendingEditDecorationsForSelection } from "../util/editDisplayUtils";
 import { runForEachEditor } from "../util/targetUtils";
 import { flatten, zip } from "lodash";
 import { maybeAddDelimiter } from "../util/getTextWithPossibleDelimiter";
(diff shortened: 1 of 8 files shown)


pokey:
A few things I noticed in testing:

- The count is in order of the selections, rather than the order of the selections in the file.  This is logical, and consistent with "sort", but potentially a bit surprising.  I wonder if there should be a command to normalize / sort your selection list by position in document
- The count starts from 0 by default.  I'd be tempted to start from 1?
- The selection order thing is even more confusing for surrounding pairs, where you'll get the delimiters facing the wrong way if your selection list is backwards

Also, in terms of insert vs replace, I think there is a meaningful distinction, but not sure we totally respect that here.  If I have something selected, what should "insert" do?  If I have an empty selection, what should replace do?  Maybe we should merge replace and insert, and then use positional modifier + selection length (eg empty or not) to determine what to do

AndreasArvidsson:
* This is consistent with sort, bring and so on. But might of course not be what you expected.
* That might be more common so I agree
* Yeah the problem is that we don't treat delimiters differently then any other replacement

Yes the naming is problematic.

pokey:
Ok after discussion with @AndreasArvidsson we've decided the following:

- `insert` should probably just be merged with `replace`
- changing surrounding pair should be its own command (see #270)

so this PR just adds count, and needs some tweaking to merge with `replace`, so probably not worth the work right now

PR #239, opened 2021-08-16
Retry tests and remove sleep

### LINKED ISSUE #173: Fix nondeterministic test failures
Let's start by just retrying tests 3 times if they fail.  This approach is terrible, because it swallows all nondeterministic bugs, but at least we can start requiring CI passes, unlike now where they're disabled because they fail

_Originally posted by @pokey in https://github.com/pokey/cursorless-vscode/pull/172#r680547202_

### LINKED ISSUE #175: Set decoration sleep timeout to 0 during tests
Should make tests much faster. I think the easiest approach is probably an environment variable?

pokey:
Depends on https://github.com/pokey/vscode-parse-tree/pull/8

Closes #175
Closes #173

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -23,6 +23,9 @@
 			"name": "Extension Tests",
 			"type": "extensionHost",
 			"request": "launch",
+			"env": {
+				"CURSORLESS_TEST": "true"
+			},
 			"args": [
 				"--disable-extension",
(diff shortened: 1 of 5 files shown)


pokey:
Let's open a follow-up issue to try to actually fix the non-deterministic failures

AndreasArvidsson (approved):

PR #246, opened 2021-08-20
Add cell selection type

pokey:
Depends on https://github.com/pokey/cursorless-talon/pull/63

### DIFF
--- a/src/test/suite/fixtures/recorded/selectionTypes/drinkCell.yml
+++ b/src/test/suite/fixtures/recorded/selectionTypes/drinkCell.yml
@@ -1,0 +1,26 @@
+spokenForm: drink cell
+languageId: python
+command:
+  actionName: editNewLineAbove
+  partialTargets:
+    - {type: primitive, selectionType: notebookCell}
+  extraArgs: []
+marks: {}
+initialState:
+  documentContents: |-
+    # %%
+    print("hello")
+  selections:
+    - anchor: {line: 1, character: 12}
+      active: {line: 1, character: 12}
+finalState:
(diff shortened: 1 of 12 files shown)


src/test/suite/fixtures/recorded/selectionTypes/drinkCell.yml
```diff
@@ -0,0 +1,26 @@
...
+  documentContents: |-
+    # %%
+    print("hello")
+  selections:
+    - anchor: {line: 1, character: 12}
+      active: {line: 1, character: 12}
+finalState:
+  documentContents: |-
+    # %%
+    print("hello")
```
pokey, line 19: whoops.  looks like (shockingly) the `jupyter.insertCellAbove` command returns a promise that resolves before the action is actually complete.  awesome.  so do we keep the test to just make sure nothing errors, delete the test, or insert some random sleep? cc/ @AndreasArvidsson @brxck

AndreasArvidsson (dismissed):

src/test/suite/fixtures/recorded/selectionTypes/drinkCell.yml
AndreasArvidsson: Wtf they dont ubderstand how a promise work do they?
I dont really care since we cant actually test it.

brxck: Yikes. If it were up to me I would leave the test as is with a note. I don't think it's worth it to accommodate with a sleep, we're never going to be 100% tests anyhow.

PR #248, opened 2021-08-23
Renames for csv overrides PR

pokey:
Many changes to support the new names in csv overrides PR.  This PR also does the following:

- Until we properly support them, we throw an exception for subtoken exclusions.  Probably not worth supporting properly until we generalize the "next" / "nth" thing in next-gen cursorless engine
- Adds a script to transform recorded test cases in bulk

Todo:
- [x] Throw proper error for unknown action from canonicalize 
- [x] Remove old extra actions and use canonicalization instead


src/typings/Types.ts
```diff
@@ -113,6 +113,8 @@ export interface SubTokenModifier {
   pieceType: SubTokenType;
   anchor: number;
   active: number;
+  includeAnchor?: boolean;
+  includeActive?: boolean;
```
pokey, line 117: making these optional for now for backwards compatibility.  Also, I went with "include" instead of "exclude" because I prefer positive booleans, otherwise it's a double negative when they're false which is a bit more cognitive load

my one hesitation with using "include" instead of "exclude" is that while these are optional, we'll be defaulting to `true`, which is always a bit jarring cc/ @AndreasArvidsson

AndreasArvidsson: I will always prefer that optional booleans default to false. That why they were defined as excluding.

pokey: Yes but I would prefer to make these not optional. The reason they're optional is for backwards compatibility. Hmm. I'm not seeing a way to make them backwards compatible while making them positive. Any ideas?

AndreasArvidsson: In Infer full targets just do.
Includeachor = target.includeanchor ?? true

pokey: Wait isn't that defaulting to `true`?

AndreasArvidsson: Yes and that is how it is today

src/canonicalizeActionName.ts
```diff
@@ -0,0 +1,32 @@
+import { ActionType } from "./typings/Types";
+
+const actionAliasToCanonicalName: Record<string, ActionType> = {
```
pokey, line 3: @AndreasArvidsson for now I'm just mapping from new to old.  Should probably switch over the action names here eventually, but not sure it's worth the effort for this initial PR.  Fwiw, the new action names are not actually tested when using this approach

AndreasArvidsson (approved):

PR #255, opened 2021-09-01
Fix copy when focus is not on editor

pokey:
Unfortunately, we aren't able to set focus back to where it was before, eg a search box.  Maybe this argues for switching back to our own copy implementation?

### DIFF
--- a/src/actions/CommandAction.ts
+++ b/src/actions/CommandAction.ts
@@ -35,14 +35,14 @@ export default class CommandAction implements Action {
       await runOnTargetsForEachEditor<SelectionWithEditor[]>(
         targets,
         async (editor, targets) => {
+          const originalSelections = editor.selections;
+
           const targetSelections = targets.map(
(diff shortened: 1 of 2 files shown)


src/actions/CommandAction.ts
```diff
@@ -32,18 +35,14 @@ export default class CommandAction implements Action {
...
-
-          const originalSelections = editor.selections;
-
           const targetSelections = targets.map(
             (target) => target.selection.selection
           );
 
-          editor.selections = targetSelections;
+          // For command to the work we have to have the correct editor focused
+          await setSelectionsAndFocusEditor(editor, targetSelections, false);
```
pokey, line 43: When setting selections and focusing editor it is important to first set the selections and then focus the editor. This was fixed in the `setSelectionsAndFocusEditor` function but because it wasn't used here we didn't get that fix

AndreasArvidsson (approved):

PR #256, opened 2021-09-01
Add a bunch of new hats

pokey:
image

Can easily enable / disable colors and shapes in settings: 

image

see also https://github.com/pokey/cursorless-talon/pull/71

- [x] Add documentation for new settings
- [ ] Align on which hats to enable by default


package.json
```diff
@@ -182,6 +182,11 @@
 						"light": "#19a1d6",
 						"highContrast": "#19a1d6"
 					}
+				},
+				"cursorless.individualHatAdjustments": {
```
pokey, line 186: a bit on the fence about having this setting.  I added it because it was really useful while prototyping rather than restarting the extension every time

src/core/Decorations.ts
```diff
@@ -131,6 +190,42 @@ export default class Decorations {
     );
   }
 
+  private constructHatStyleMap() {
+    const hatStyleSettings = vscode.workspace
+      .getConfiguration("cursorless")
+      .get<HatStyleSettingMap>("hatStyles")!;
+
+    hatStyleSettings.default = { penalty: 0, enabled: true };
```
pokey, line 198: Do we want to allow user to configure default color and shape `penalty` / `enabled`?  If so, note that settings setup will have to change because they'll both want to use the setting `cursorless.hatStyles.default`

AndreasArvidsson (approved):

PR #258, opened 2021-09-03
Fix "take every value" in python

### LINKED ISSUE #257: "take every value" is broken in python
For example, in the following map:

```python
{
    "foo": "bar",
    "baz": "bongo",
}
```

Saying "take every value" will only select the given value.

pokey:
Fixes #257 

Mixed feelings about this solution.  There is no way to exclude two types, for example.

The alternative would be to explicitly list the types that we want to allow user to use "value" field with.  There are about 7

### DIFF
--- a/src/languages/python.ts
+++ b/src/languages/python.ts
@@ -55,7 +55,7 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
   className: "class_definition[name]",
   namedFunction: "decorated_definition?.function_definition",
   functionName: "function_definition[name]",
-  type: ["function_definition[return_type]", "*[type]"],
+  type: valueMatcher("function_definition[return_type]", "*[type]"),
   name: [
     "assignment[left]",
     "typed_parameter.identifier!",
--- a/src/languages/python.ts
+++ b/src/languages/python.ts
@@ -63,7 +63,7 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
     "*[name]",
   ],
   collectionItem: argumentMatcher(...dictionaryTypes, ...listTypes),
-  value: valueMatcher("assignment[right]", "*[value]"),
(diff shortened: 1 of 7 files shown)


src/languages/python.ts
```diff
@@ -55,15 +55,15 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
   className: "class_definition[name]",
   namedFunction: "decorated_definition?.function_definition",
   functionName: "function_definition[name]",
-  type: ["function_definition[return_type]", "*[type]"],
+  type: valueMatcher("function_definition[return_type]", "*[type]"),
```
pokey, line 58: So that leading `:` gets deleted when using "chuck type"

src/test/suite/fixtures/recorded/languages/python/chuckValue.yml
```diff
@@ -0,0 +1,23 @@
...
+      modifier: {type: containingScope, scopeType: value, includeSiblings: false}
+  extraArgs: []
+marks: {}
+initialState:
+  documentContents: "foo: str = \"hello\""
+  selections:
+    - anchor: {line: 0, character: 2}
+      active: {line: 0, character: 2}
+finalState:
+  documentContents: "foo: str "
```
pokey, line 16: We should prob clean up this trailing string too, but prob best to leave that one for object-oriented targets refactor

AndreasArvidsson:
I'm unsure. I think this could be a reoccurring problem in different languages. One solution could be that we do an ancestral crawl and just pick the highest match. Maybe should have dialogue about this before doing anything?

pokey:
> I'm unsure. I think this could be a reoccurring problem in different languages. One solution could be that we do an ancestral crawl and just pick the highest match. Maybe should have dialogue about this before doing anything?

not sure i follow; yes let's discuss

AndreasArvidsson (approved):

PR #260, opened 2021-09-06
Add suffixedMatcher; use for keys

pokey:
Allows saying "chuck key" and having the trailing `:` cleaned up

### DIFF
--- a/src/util/nodeSelectors.ts
+++ b/src/util/nodeSelectors.ts
@@ -40,11 +40,11 @@ export function selectWithLeadingDelimiter(
   editor: TextEditor,
   node: SyntaxNode
 ): SelectionWithContext {
-  const leadingDelimiterToken = node.previousSibling;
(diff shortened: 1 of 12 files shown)


src/util/nodeSelectors.ts
```diff
@@ -40,11 +40,11 @@ export function selectWithLeadingDelimiter(
   editor: TextEditor,
   node: SyntaxNode
 ): SelectionWithContext {
-  const leadingDelimiterToken = node.previousSibling;
+  const leadingNonDelimiterToken = node.previousSibling?.previousSibling;
```
pokey, line 43: Note that we now use the end of the token before the delimiter rather than the start of the delimiter.  This way, for example, if you have "foo = bar" and say "chuck value", you won't get the trailing space after `foo`.  See changed test case above (`chuckValue`)

AndreasArvidsson (approved):

PR #263, opened 2021-09-06
Rename purple to pink

pokey:
"pink" will become the default term for this color in the next few weeks (see https://github.com/pokey/cursorless-talon/issues/66), and we will change the default color to be more obviously pink.  Hoping to reduce cognitive dissonance for colors

See also https://github.com/pokey/cursorless-talon/pull/75

### DIFF
--- a/package.json
+++ b/package.json
@@ -165,8 +165,8 @@
 						"highContrast": "#f0b800"
 					}
 				},
-				"cursorless.colors.purple": {
-					"description": "Color to use for purple symbols",
+				"cursorless.colors.pink": {
+					"description": "Color to use for pink symbols",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -165,8 +165,8 @@
 						"highContrast": "#f0b800"
 					}
 				},
-				"cursorless.colors.purple": {
-					"description": "Color to use for purple symbols",
+				"cursorless.colors.pink": {
```
pokey, line 168: this will break settings for people who have a custom purple.  was thinking of just pinging on slack before releasing?

AndreasArvidsson (approved):

PR #268, opened 2021-09-08
Fix color tokens

### LINKED ISSUE #266: Treat hex colors as single tokens
- Do we want to include the `#` or not?
- Tempted to just have special case for them
- Note that VSCode on double click will select the color without the hash

pokey:
Closes #266

### DIFF
--- a/src/core/tokenizer.ts
+++ b/src/core/tokenizer.ts
@@ -35,15 +35,15 @@ const REPEATABLE_SYMBOLS_REGEX = REPEATABLE_SYMBOLS.map(escapeRegExp)
   .map((s) => `${s}+`)
   .join("|");
 const FIXED_TOKENS_REGEX = FIXED_TOKENS.map(escapeRegExp).join("|");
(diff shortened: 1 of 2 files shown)


AndreasArvidsson:
Did you really need to change the numbers regex?

pokey:
> Did you really need to change the numbers regex?

Yep!

```
  1) tokenizer
       #11aaaa:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

  [
    '#',
+   '11',
+   'aaaa'
-   '11aaaa'
  ]
      + expected - actual

       [
         "#"
      -  "11"
      -  "aaaa"
      +  "11aaaa"
       ]
```

Notice that I changed the order of the regex pieces so that number is now before token.  So if number can match a digit string, it will break up tokens that start with digits.  And digit strings now are matched by the token regex anyway

AndreasArvidsson (approved):

Issue #269, opened 2021-09-09
Do one more pass at shapes and colors


pokey:
- [ ] `eye`: slightly wider pupil?
- [ ] `frame`: with a bit more paint
- [ ] `curve`: slightly thicker
- [ ] Tweak light colors
- [x] Maybe make yellow darker
- [x] New hole
- [x] Maybe make pink brighter
- [x] Try not to clip descenders from line above
- [x] Tweak alignment

Issue #78, opened 2021-07-07
Decorations don't scale with editor zoom


brxck:
I'm on 5271ce3f1c2d80fded9917826b4153d270351f1e and Ubuntu 20.10. 

VS Code info:

```
Version: 1.57.1
Commit: 507ce72a4466fbb27b715c3722558bb15afa9f48
Date: 2021-06-17T13:26:56.255Z
Electron: 12.0.7
Chrome: 89.0.4389.128
Node.js: 14.16.0
V8: 8.9.255.25-electron.0
OS: Linux x64 5.8.0-59-generic
```

#### No Zoom

![no zoom](https://user-images.githubusercontent.com/28877984/124815744-7a5e2280-df1c-11eb-98e1-fd16d0bdfce3.png)

#### 5x Zoom

![5x zoom](https://user-images.githubusercontent.com/28877984/124815783-86e27b00-df1c-11eb-863e-e1b9ef0d8ba0.png)

brxck:
This is the case on my new machine as well 😢

pokey:
Weird! I believe that zoom level is a setting, so we could maybe read zoom level and adjust based on that.

Strange that it only happens for some VSCode setups. Does it happen on a clean VSCode instance with no extensions installed?

We could have a setting to toggle this correction and then mention it in known issues

pokey:
@brxck when you get a minute, would be amazing if you could check whether https://github.com/pokey/cursorless-vscode/pull/273 fixes this one

brxck:
@pokey unfortunately not 😦 

![Screenshot from 2021-09-11 10-08-24](https://user-images.githubusercontent.com/28877984/132955725-d407a91d-0763-40da-9376-24cba1ffd860.png)

![Screenshot from 2021-09-11 10-08-45](https://user-images.githubusercontent.com/28877984/132955727-a8171696-4a13-4a52-854c-fc710357763f.png)

AndreasArvidsson:
Is this still a problem?

brxck:
Yep, same behavior. Let me know if I can provide any other info/help!

<img width="356" alt="image" src="https://user-images.githubusercontent.com/28877984/145690180-970ad3ea-7675-4dbb-b943-fbe4189fbcaa.png">

AndreasArvidsson:
This is so strange. We have tested this on the windows, linux, and mac and it works on all three except for on your computer.

brxck:
[...]

PR #273, opened 2021-09-11
Fix SVG alignment and click box

### LINKED ISSUE #92: Fix decoration appearing after GitLens Current Line Blame
This happens if another plugin tries to put something at the end of the line, eg GitLens's [Current Line Blame](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens#current-line-blame-).  Notice how the final gray dot is all the way off to the right:

image

This could probably be fixed by using the svg `before` instead of `after`

### LINKED ISSUE #110: SVG hat obstructs click target area
If the user clicks on a letter with a hat over it, the cursor will end up to the right of the letter, unless the user clicks significantly to the left of the letter

pokey:
I adjusted the measurements to fix the math bug for my setup, but unfortunately different users will be affected proportional to the square of the ratio of their character width to the ceiling of the width, so they'll unfortunately need to do their own tweaking if they've adjusted their defaults

image

Note that the hats have moved up very slightly, but it is minor, and we will tweak these heights in #269 anyway

Fixes #110
Fixes #92 

And possibly #78

### DIFF
--- a/src/core/Decorations.ts
+++ b/src/core/Decorations.ts
@@ -138,9 +138,6 @@ export default class Decorations {
       const { color, shape } = this.hatStyleMap[styleName];
       const { svg, svgWidthPx, svgHeightPx } = hatSvgMap[shape];
 
-      const spanWidthPx =
-        svgWidthPx + (fontMeasurements.characterWidth - svgWidthPx) / 2;
-
       const colorSetting = vscode.workspace
         .getConfiguration("cursorless.colors")
         .get<DecorationColorSetting>(color)!;
@@ -150,7 +147,7 @@ export default class Decorations {
         decoration: vscode.window.createTextEditorDecorationType({
           rangeBehavior: vscode.DecorationRangeBehavior.ClosedClosed,
           light: {
-            after: {
+            before: {
               contentIconPath: this.constructColoredSvgDataUri(
                 svg,
                 colorSetting.light
@@ -158,16 +155,16 @@ export default class Decorations {
             },
           },
           dark: {
-            after: {
+            before: {
               contentIconPath: this.constructColoredSvgDataUri(
                 svg,
                 colorSetting.dark
               ),
             },
           },
-          after: {
-            margin: `-${svgHeightPx}px 0 0 -${spanWidthPx}px`,
-            width: `${spanWidthPx}px`,
+          before: {
+            margin: `-${svgHeightPx}px -${svgWidthPx}px 0 0`,
+            width: `${svgWidthPx}px`,
             height: `${svgHeightPx}px`,
           },
         }),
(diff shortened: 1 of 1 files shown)


src/core/Decorations.ts
```diff
@@ -150,24 +147,24 @@ export default class Decorations {
...
                 svg,
                 colorSetting.dark
               ),
             },
           },
-          after: {
-            margin: `-${svgHeightPx}px 0 0 -${spanWidthPx}px`,
-            width: `${spanWidthPx}px`,
+          before: {
+            margin: `-${svgHeightPx}px -${svgWidthPx}px 0 0`,
```
pokey, line 166: No need to play games with span width anymore because we now align to the left side of the char

src/core/Decorations.ts
```diff
@@ -150,24 +147,24 @@ export default class Decorations {
...
+            width: `${svgWidthPx}px`,
```
pokey, line 167: No need to play games with span width anymore because we now align to the left side of the char

src/core/Decorations.ts
```diff
@@ -261,25 +258,24 @@ export default class Decorations {
...
-    const svgWidthPx = Math.ceil(fontMeasurements.characterWidth);
-    const svgHeightPx =
-      fontMeasurements.characterHeight + hatHeightPx + hatVerticalOffset;
+    const svgWidthPx = Math.ceil(characterWidth);
+    const svgHeightPx = characterHeight + hatHeightPx + hatVerticalOffset;
 
     const newViewBoxWidth =
-      ((originalViewBoxWidth / hatWidthToCharacterWidthRatio) *
-        fontMeasurements.characterWidth) /
-      svgWidthPx;
```
pokey, line 280: This was the dumb mistake.  We need to adjust from char width to svg width, not the other way around 😅

src/core/Decorations.ts
```diff
@@ -261,25 +258,24 @@ export default class Decorations {
...
+      ((originalViewBoxWidth / hatWidthToCharacterWidthRatio) * svgWidthPx) /
+      characterWidth;
     const newViewBoxHeight = (newViewBoxWidth * svgHeightPx) / svgWidthPx;
-    const newViewBoxX = -(newViewBoxWidth - originalViewBoxWidth) / 2;
+    const newViewBoxX =
+      (-(characterWidth - hatWidthPx) * (newViewBoxWidth / svgWidthPx)) / 2;
```
pokey, line 278: Since we're left-aligned, we basically need to shift right by `(characterWidth - hatWidthPx) / 2`, translated to svg view box space (`newViewBoxWidth / svgWidthPx`)

AndreasArvidsson (approved):

PR #274, opened 2021-09-13
Added triple quote to regular expression

AndreasArvidsson:


pokey (commented):
Looks great but mind adding an empty pair of single / double quotes?

pokey (approved):

PR #275, opened 2021-09-13
Added support for unicode characters in regular expression

### LINKED ISSUE #252: Regular expression splits words with swedish characters
Our identifiers today only supports [a-zA-Z0-9_] which is not a problem in programming when writing drafts it is. 

decorations swedish

AndreasArvidsson:
fixes #252

### DIFF
--- a/src/core/tokenizer.ts
+++ b/src/core/tokenizer.ts
@@ -36,7 +36,7 @@ const REPEATABLE_SYMBOLS_REGEX = REPEATABLE_SYMBOLS.map(escapeRegExp)
   .map((s) => `${s}+`)
   .join("|");
 const FIXED_TOKENS_REGEX = FIXED_TOKENS.map(escapeRegExp).join("|");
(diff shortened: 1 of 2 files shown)


pokey (approved):
This looks good, but still gonna be pretty painful with cursorless putting hats over letters you can't say sometimes no? I'd be tempted to include the feature that strips the accents in this PR, but if this is a serious usability win for Swedish as is go ahead and merge

If you do merge as is I'd file a follow-up issue for the accent stripping. 

Btw any idea where I put the link for the accent stripping code? Could have sworn I added it to the issue but can't seem to find it

AndreasArvidsson:
That regex all read the decorates unicode characters today; they get caught by the symbol regex. I therefore don't think treating them as identifiers is a larger problem. Personally I now have voice commands to reference those characters and is going to give it a try. 

New issue created regarding as setting to strip them instead
https://github.com/pokey/cursorless-vscode/issues/276

PR #278, opened 2021-09-19
Put identifier before repeatable symbols

AndreasArvidsson:
Since repeatable symbols was before identifier `_a` was split into two tokens


pokey (commented):
Looks good but could use a test

AndreasArvidsson:
Test added

pokey (approved):

PR #280, opened 2021-09-20
Invert color settings

### LINKED ISSUE #271: Invert color settings
So that they're grouped by mode (eg `dark` / `light`) rather than grouped by color

pokey:
Closes #271


AndreasArvidsson:
Maybe we should hide high contrast for now since it isn't actually working?

pokey:
> Maybe we should hide high contrast for now since it isn't actually working?

done

AndreasArvidsson (approved):

PR #296, opened 2021-10-15
Snippet wrap

### LINKED ISSUE #21: Support "wrap with snippet"
VSCode has support for snippets that can take a selection.  We'd like to use this to enable things such as "wrap state fine with snip try catch"

pokey:
tryWrapFine

Closes #21

- [x] Switch to `cursorless.wrappers.scopeTypes` for settings
- [x] Switch to use language-specific settings instead of language mapping to simplify implementation and make it easier to configure
- [x] Allow snippets to signal their preferred scope type
- [x] Support user snippets?  If so, we'd prob also want a way to signal what their preferred scope type is
- [x] Return `that` mark

### DIFF
--- a/schemas/cursorless-snippets.json
+++ b/schemas/cursorless-snippets.json
@@ -49,6 +49,10 @@
                         "wrapperScopeType": {
                             "$ref": "#/$defs/scopeType",
                             "description": "Default to this scope type when wrapping a target without scope type specified"
+                        },
+                        "description": {
+                            "type": "string",
+                            "description": "Description of the snippet variable"
                         }
                     }
                 }
--- a/src/typings/snippet.ts
+++ b/src/typings/snippet.ts
@@ -22,6 +22,11 @@ export interface SnippetVariable {
    * specified.
    */
   wrapperScopeType?: ScopeType;
+
+  /**
+   * Description of the snippet variable
+   */
+  description?: string;
 }
 
 export interface Snippet {


pokey:
@AndreasArvidsson looks like this is failing because snippet inserts `\r\n` on windows.  Any ideas how to fix?  Should we set line ending for all of our tests unless they have a special flag in the yaml?

package.json
```diff
@@ -400,9 +400,100 @@
 							"verticalOffset": 0
 						}
 					}
+				},
+				"cursorless.wrapperSnippetPreferences": {
+					"description": "Allows snippets to signal what their default scope types etcetera should be",
```
pokey, line 405: Down the road we may allow signalling things like `selectionType`, etc, so I figured let's have this be a map

package.json
```diff
@@ -400,9 +400,100 @@
...
+						"cursorless.wrapper.ifElseStatementElseBranch": {
+							"scopeType": "statement"
+						}
+					},
+					"additionalProperties": {
+						"type": "object",
+						"properties": {
+							"scopeType": {
+								"type": "string",
+								"enum": [
```
pokey, line 427: This enum will get stale, but I don't think it will cause any serious problems, and the autocomplete will be nice

snippets/cpp.json
```diff
@@ -0,0 +1,26 @@
+{
```
pokey, line 1: Decided to duplicate these snippets between languages as I'd imagine they'll start to diverge.  but easy to just use one file as well

src/actions/BringMoveSwap.ts
```diff
@@ -31,7 +31,7 @@ interface MarkEntry {
 }
 
 class BringMoveSwap implements Action {
-  targetPreferences: ActionPreferences[] = [
+  getTargetPreferences: () => ActionPreferences[] = () => [
```
pokey, line 34: We now have a function to give action a chance to see args

src/actions/WrapWithSnippet.ts
```diff
@@ -0,0 +1,95 @@
...
+
+interface SnippetPreferences {
+  scopeType?: ScopeType;
+}
+
+type UserLanguageSnippetMap = Record<string, SnippetPreferences>;
+
+export default class WrapWithSnippet implements Action {
+  getTargetPreferences(
+    partialTargets: PartialTarget[],
```
pokey, line 23: We have this arg here so that below FIXME can be implemented in the future.  Mixed feelings about letting action see this thing but not sure how else to let it do something like figure out which editor a target is referring to

src/test/suite/fixtures/recorded/languages/cpp/tryCatchWrapThis2.yml
```diff
@@ -0,0 +1,47 @@
+spokenForm: try catch wrap this
```
pokey, line 1: Test for multiple cursors

src/test/suite/recorded.test.ts
```diff
@@ -59,6 +60,12 @@ async function runTest(file: string) {
   });
   const editor = await vscode.window.showTextDocument(document);
 
+  if (!fixture.initialState.documentContents.includes("\n")) {
```
pokey, line 63: This seemed reasonable to me and doesn't break your line ending tests

src/test/suite/recorded.test.ts
```diff
@@ -118,15 +125,20 @@ async function runTest(file: string) {
...
   );
 
-  assert.deepStrictEqual(
-    resultState,
-    fixture.finalState,
-    "Unexpected final state"
-  );
+  if (process.env.CURSORLESS_TEST_UPDATE_FIXTURES == "true") {
+    const outputFixture = { ...fixture, finalState: resultState, returnValue };
+    await fsp.writeFile(file, serialize(outputFixture));
```
pokey, line 130: Easy way to update snapshots

.vscode/launch.json
```diff
@@ -36,6 +36,25 @@
 				"${workspaceFolder}/out/test/**/*.js"
 			],
 			"preLaunchTask": "${defaultBuildTask}"
+		},
+		{
+			"name": "Update fixtures",
```
pokey, line 41: Easy way to update snapshots

package.json
```diff
@@ -400,9 +400,100 @@
 							"verticalOffset": 0
 						}
 					}
+				},
+				"cursorless.wrapperSnippetPreferences": {
```
pokey, line 404: - [ ] have a `cursorless.snippets` setting with just a `defaultScopeType` property.

src/actions/Sort.ts
```diff
@@ -7,7 +7,7 @@ import {
 } from "../typings/Types";
 
 export class Sort implements Action {
-  targetPreferences: ActionPreferences[] = [{ insideOutsideType: "inside" }];
+  getTargetPreferences: () => ActionPreferences[] = () => [{ insideOutsideType: "inside" }];
```
pokey, line 10: - [ ] see if i can get rid of type here

src/actions/WrapWithSnippet.ts
pokey: - [ ] pass in context object with callbacks instead

pokey (commented):
- [ ] remove prefixes so just `ifStatement`
- [ ] double-check that can say "if wrap token air"

PR #306, opened 2021-10-31
Improved update selections

### LINKED ISSUE #138: Improve and unify selection updating behaviour to handle overlapping ranges
## The problem
The code in `updateSelections.ts` that actually updates the selections is a bit picky about what it will update.  It works if all edits either

- have no overlap with the given selections, or
- have a range that is identical to one of the selections

However, if, for example, `change.range.start` is not before `selection.range.start`, but is before `selection.range.end`, and not the exact same range, the selections will silently not be updated

One example where the current logic breaks: if the `commentLines` action is applied with a target that is the entire line, the `that` mark ends up not getting moved to the right

In addition, we do not use this new selection updating code in the `NavigationMap`.  

## The solution
We would like to do the following:

### Improve the selection updating code
There are lots of ways we could get fancy, but let's start with something simple.  The approach is to take each selection, and split it into `start` and `end` positions.  Then for each of these positions, if the end of the `change` range is  before or equal to the position, we shift it.  If the change range contains the position, we can have an option passed in by the caller to determine what to do, either expand to include it, contract to exclude it, or throw an exception / remove the selection.  Could maybe just steal the `OpenOpen` / `OpenClosed` etc stuff that VSCode does with [decorations](https://code.visualstudio.com/api/references/vscode-api#DecorationRangeBehavior)

This solution at least allows us to handle changes that are contained entirely within the selection.

- [ ] Improve selection updating code

### Use selection updating code in the navigation map

- [ ] Use selection updating code in the navigation map

### LINKED ISSUE #298: Automatic token expansion doesn't work if token is offscreen
### Steps to reproduce

1. Put your cursor in the middle of a token
2. Scroll the token off the screen
3. Say "chuck first word"

### Root cause

We [use](https://github.com/pokey/cursorless-vscode/blob/d31239dddf73887e5d58e07ac0dd736dcb36718f/src/processTargets/processMark.ts#L46) the navigation map to get tokens and the navigation map only has tokens that are on screen.

### Proposed solution

Tokenise the line containing the start of the range and the line containing the end of the range and use those tokens

pokey:
This pull request is a complete rework of the way that we update ranges in response to document changes.  There is now a `rangeUpdater` component on the `graph` with which you can register ranges that you'd like it to keep up-to-date.

Closes #138
Closes #298

- [x] Add tests


src/test/suite/fixtures/recorded/actions/bringVest.yml
```diff
@@ -12,17 +12,17 @@ command:
       modifier: {type: identity}
       insideOutsideType: inside
   extraArgs: []
-marks:
```
pokey, line 15: Marks are now a property of the initial state, because they can and do change in response to a command.  Before this fact didn't matter because we only checked marks at the beginning.  We now test navigation map updating behaviour, so we also store how the marks look after we run a test

src/test/suite/fixtures/recorded/navigationMap/bringPointAfterFirstCarWhaleTakeWhale.yml
```diff
@@ -0,0 +1,44 @@
...
+  selections:
+    - anchor: {line: 0, character: 14}
+      active: {line: 0, character: 14}
+  marks:
+    default..:
+      start: {line: 0, character: 5}
+      end: {line: 0, character: 6}
+    default.w:
+      start: {line: 0, character: 7}
+      end: {line: 0, character: 13}
```
pokey, line 37: Notice how we just include the `.` when it's in the middle, even though it technically is breaking up the token

src/test/suite/fixtures/recorded/navigationMap/bringPointToEndOfWhaleTakeWhale.yml
```diff
@@ -0,0 +1,43 @@
...
+  selections:
+    - anchor: {line: 0, character: 14}
+      active: {line: 0, character: 14}
+  marks:
+    default..:
+      start: {line: 0, character: 5}
+      end: {line: 0, character: 6}
+    default.w:
+      start: {line: 0, character: 7}
+      end: {line: 0, character: 12}
```
pokey, line 36: But if we add `.` to the end of the token we don't include it

src/test/suite/fixtures/recorded/navigationMap/chuckHarpPastAirTakeWhale.yml
```diff
@@ -0,0 +1,49 @@
...
+  marks:
+    default.h:
+      start: {line: 0, character: 0}
+      end: {line: 0, character: 0}
+    default.a:
+      start: {line: 0, character: 0}
+      end: {line: 0, character: 0}
+    default.w:
+      start: {line: 0, character: 0}
+      end: {line: 0, character: 0}
```
pokey, line 45: Notice how clobbering the range just leaves the mark at the beginning of the deleted range

src/test/suite/fixtures/recorded/navigationMap/chuckThirdCarHarpPastSecondCarWhaleTakeWhale.yml
```diff
@@ -0,0 +1,47 @@
...
+  selections:
+    - anchor: {line: 0, character: 14}
+      active: {line: 0, character: 14}
+  marks:
+    default.h:
+      start: {line: 0, character: 0}
+      end: {line: 0, character: 2}
+    default.w:
+      start: {line: 0, character: 2}
+      end: {line: 0, character: 5}
```
pokey, line 43: This behaviour is what we agreed upon in discussion: although technically the whale token has joined with the harp token, we preserve only the part of whale that remains; we don't join it with harp

src/test/suite/fixtures/recorded/navigationMap/moveFourthCarHarpPastSecondCarWhaleToEndOfWhaleTakeWhale.yml
```diff
@@ -0,0 +1,54 @@
...
+  selections:
+    - anchor: {line: 0, character: 20}
+      active: {line: 0, character: 20}
+  marks:
+    default.h:
+      start: {line: 0, character: 0}
+      end: {line: 0, character: 3}
+    default.w:
+      start: {line: 0, character: 3}
+      end: {line: 0, character: 8}
```
pokey, line 47: This test is kinda bananas, but basically we've deleted first 2 characters of whale, then added stuff with a space in it to the end of whale.  We preserve those last 3 chars of whale and pick up the test that was added until the space on the end! 😅

src/test/suite/fixtures/recorded/updateSelections/bringFine.yml
```diff
@@ -0,0 +1,35 @@
+spokenForm: bring fine
```
pokey, line 1: The tests in this directory are all selection based, because we now control cursor in response to a bring.  For example, in this one we check that the selection is moved to the right

src/test/suite/fixtures/recorded/updateSelections/bringFineAfterThis2.yml
```diff
@@ -0,0 +1,32 @@
...
+      active: {line: 0, character: 9}
+  marks:
+    default.f:
+      start: {line: 0, character: 0}
+      end: {line: 0, character: 3}
+finalState:
+  documentContents: foo world foo
+  selections:
+    - anchor: {line: 0, character: 9}
+      active: {line: 0, character: 9}
```
pokey, line 25: I think this behaviour is kinda cool.  If we say "bring air after this", then the cursor doesn't move the way it usually does when you say "bring air"

AndreasArvidsson (approved):

PR #322, opened 2021-11-11
Execute arbitrary code action

pokey:


pokey:
It's probably worth extraction https://github.com/pokey/cursorless-vscode/blob/068343199ce26a0f3e12ba42d05332f30dfbe4f1/src/actions/ExecuteCodeAction.ts#L43-L66 into a command on a separate repo, but otherwise this behaviour is easily replicated using the new run vscode command action

PR #337, opened 2021-11-28
Removed additional highlight before scroll

AndreasArvidsson:


pokey (approved):

pokey:
@AndreasArvidsson is that CI failure the same thing we've seen locally?

AndreasArvidsson:
> @AndreasArvidsson is that CI failure the same thing we've seen locally?

Yes

PR #339, opened 2021-11-28
Join sources on single destination

AndreasArvidsson:


pokey (commented):
Tbh I find the impl a bit confusing, but I'm going to rework this code soon anyways so happy to leave it as long as it has decent test coverage.  But would be good to add tests for the following:
- A couple "before" examples
- Another delimited scope type, eg "arg"

AndreasArvidsson:
The only thing I changed was to join the text on multiple sources into one destination
Nothing in this pr has changed behavior of before and after. We could probably add a couple of tests anyway.

pokey:
> The only thing I changed was to join the text on multiple sources into one destination Nothing in this pr has changed behavior of before and after. We could probably add a couple of tests anyway.

Yeah I mean adding tests where you bring multiple before.  I could see that getting broken somehow where we end up adding two delimiters.  And same adding tests where we bring multiple to a target that is an arg.  Also prob useful "bring air and bat to arg cap"; I think currently your tests just do after and cursor

AndreasArvidsson:
I understand in completely agree. Tests added

src/test/suite/fixtures/recorded/actions/bringAirAndBatAndCapToAfterItemEach.yml
```diff
@@ -0,0 +1,62 @@
...
+    default.e:
+      start: {line: 4, character: 16}
+      end: {line: 4, character: 17}
+finalState:
+  documentContents: |
+    a
+    b
+    c
+
+    const values = [e, a, b, c]
```
pokey, line 48: this is actually pretty cool 😊

pokey (approved):
This is great.  I get the impression these extra tests might have found a bug 😊

src/test/suite/fixtures/recorded/actions/bringAirAndBatAndCapToAfterItemEach.yml
AndreasArvidsson: > This is great. I get the impression these extra tests might have found a bug 😊

Maybe

AndreasArvidsson: It is really cool!

Issue #309, opened 2021-11-03
Add scope types "trailing" / "leading"


pokey:
Main use case being  "chuck trailing" / "chuck leading" /  "clear trailing" / "clear leading".  

Need this in order to merge in the "move air after bat" branch

- [ ] Also look for tests that have a single target with non-`null` `position` to see if they should change to using `trailing` / `leading` instead.  Esp tests introduced in #340

AndrewDant:
I'm not clear on what these scope types are meant to represent. Leading/trailing whitespace?

PR #340, opened 2021-11-29
Updated leading and trailing delimiters for before and after positions

AndreasArvidsson:
Delimiter was not aggressive enough for positions before/after. Many times `chuck before token` would fail. 
Updated delimiter code and added additional tests.

### DIFF
--- a/src/processTargets/processSelectionType.ts
+++ b/src/processTargets/processSelectionType.ts
@@ -213,10 +213,16 @@ function getTokenSelectionContext(
         : null;
   }
 
-  const isInDelimitedList =
-    (leadingDelimiterRange != null || trailingDelimiterRange != null) &&
-    (leadingDelimiterRange != null || start.character === 0) &&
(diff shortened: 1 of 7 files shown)


pokey:
I thought we decided to replace these with their own modifiers in https://github.com/pokey/cursorless-vscode/issues/309

Is this code useful for anything other than "chuck after"? Conversely, can we easily use this code for "chuck trailing"?

Also hesitant to add a bunch of unit tests for something we're shortly removing

AndreasArvidsson:
We will but I need this one right now because it's bugging the hell out of me. Much of the same logic can be reused of course and I think I change like three lines. You always likes when I had lot of tests :D

pokey:
> We will but I need this one right now because it's bugging the hell out of me. Much of the same logic can be reused of course and I think I change like three lines. You always likes when I had lot of tests :D

yeah i'm not going to spend too much time arguing against tests 😄.  I've added a note to update these tests once #309 is implemented

PR #342, opened 2021-11-29
Added specific typescript type selector

### LINKED ISSUE #281: "chuck type" leaves trailing colon
For example:

```typescript
const foo: number = 0;
```

Saying "chuck type" leaves the following:

```typescript
const foo: = 0;
```

AndreasArvidsson:
fixes #281


src/languages/typescript.ts
```diff
@@ -80,6 +86,25 @@ function valueMatcher() {
...
+      : null;
+
+  return {
+    ...simpleSelectionExtractor(editor, node),
+    context: {
+      leadingDelimiterRange,
+    },
+  };
+}
+
```
pokey, line 107: How come the default one didn't work?

pokey (commented):
Looks good but please add tests. I'd say at least:

- [x] return type,
- [x] variable declaration type,
- [x] parameter type, and
- [ ] chucking a whole interface.

src/languages/typescript.ts
AndreasArvidsson: The default implementation assumes that the name/key is a previous sibling to the type. That is true for python but not typescript according to the tree sitter.
I have some ideas for rewriting that entire thing to work for both when I have the time.

pokey: Ah right ok. Fine with having it here for now

AndreasArvidsson:
Tests added

src/test/suite/fixtures/recorded/languages/typescript/chuckTypeSoon2.yml
```diff
@@ -0,0 +1,33 @@
+spokenForm: chuck type soon
```
pokey, line 1: Ah this is nice but I actually meant chucking the entire interface, which can also be referred to as "type", just to make sure that wasn't broken here as it has no leading `:`.  But don't remove this test; it's nice too 😊

AndreasArvidsson: I didn't know that. Test added.

Is just optional leading delimiter. But always good have tests.

pokey (approved):
Ok great thanks for fixing this one; been bugging me for a while 😅

PR #347, opened 2021-11-30
Delimiter selector updated

AndreasArvidsson:
* Delimiter selector made more robust and usable for all languages
* Added a lot of language specific play/test files. They have very little in them right now but the purpose is that each language should have a file with all the common structures making it easier to test while developing.


src/languages/cpp.ts
```diff
@@ -1,109 +1,112 @@
-import {
```
pokey, line 1: Not sure what happened with this diff. Did you change line endings or something? Same with all the other language parse tree pattern files files

pokey (commented):
I like this idea. A few things

- [x] Looks like the line endings got messed up or something? Strange diffs
- [ ] Looks like a few junk files appeared, eg `tmp.tsx` and a couple empty files. The markdown one also looks strange
- [ ] Please add tests for the things you fixed
- [ ] I'd be tempted to put all the files, including hat-test, into a `data/playground` directory. I think "test" is probably a confusing name because they're not actually tests. Prob my fault for calling it "hat-test". I think I'd maybe go `data/playground/hats.txt`, `data/playground/typescript.ts`, etc?

src/languages/cpp.ts
AndreasArvidsson: I'm trying to switch to LF and something rent wrong.

AndreasArvidsson:
Now I think everything is fixed :)

src/test/suite/fixtures/recorded/languages/java/chuckTypeNear.yml
```diff
@@ -0,0 +1,37 @@
...
+    - anchor: {line: 3, character: 1}
+      active: {line: 3, character: 1}
+  marks:
+    default.n:
+      start: {line: 1, character: 24}
+      end: {line: 1, character: 28}
+finalState:
+  documentContents: |-
+    public class Java {
+        private Java( name) {
```
pokey, line 27: This one is a bit annoying but probably out of scope for this PR

src/util/nodeSelectors.ts
```diff
@@ -75,47 +75,63 @@ export function unwrapSelectionExtractor(
...
-      : null;
-
-  return {
-    ...simpleSelectionExtractor(editor, node),
-    context: {
-      trailingDelimiterRange,
-    },
+        );
+      } else {
+        trailingDelimiterRange = makeRangeFromPositions(
```
pokey, line 123: What if we walked up parents until we find one with a previous sibling and then ask for end of previous sibling there? Not sure it's worth the effort but might make white space a bit cleaner

pokey (approved):
This is really nice.  Left a couple minor suggestions

src/languages/csharp.ts
```diff
@@ -213,33 +205,27 @@ const getMapMatchers = {
...
     makeDelimitedSelector("(", ")")
   ),
-  namedFunction: typeMatcher(...NAMED_FUNCTION_TYPES),
-  functionName: composedMatcher([
-    typedNodeFinder(...NAMED_FUNCTION_TYPES),
-    getNameNode,
-  ]),
-  comment: matcher(typedNodeFinder("comment")),
-  regularExpression: matcher(typedNodeFinder("regex")),
-  type: matcher(findTypeNode, selectWithLeadingDelimiter),
```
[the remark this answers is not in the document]
AndreasArvidsson: c# don't have any delimiter for their type. Just like java it's just a space. One could argue that we could include the space as a delimiter but we don't do that for java or c++ so I just wanted to be consistent. And that delimiter is trailing not leading so the previous implementation makes no sense.

src/test/suite/fixtures/recorded/languages/java/chuckTypeNear.yml
AndreasArvidsson: One way we could avoid that would be to include a trailing space as a delimiter. But I don't think spaces are tokens so we going to have to do that text wise without the tree sitter.

src/util/nodeSelectors.ts
AndreasArvidsson: We can't try that

src/test/suite/fixtures/recorded/languages/java/chuckTypeNear.yml
pokey: Yeah I think in the future we could find a way to let token spacing treatment handle this one. Happy to call this out of scope for this PR

src/util/nodeSelectors.ts
pokey: Did you mean to say "we _can_ try that?"

AndreasArvidsson: I don't think we have to. So far no language has need of this with the current implementation.

src/test/suite/fixtures/recorded/languages/java/chuckTypeNear.yml
AndreasArvidsson: This problem is now fixed :)

src/languages/csharp.ts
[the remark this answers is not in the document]
AndreasArvidsson: New selector removes but white space and also used for c#

PR #348, opened 2021-12-01
Added action remove selection

### LINKED ISSUE #95: Add "remove cursor" action
When using multiple cursors this would allow the user to remove a cursor that is contained in the given target

Could call it "give" (opposite of "take")

AndreasArvidsson:
Closes #95


src/actions/RemoveSelection.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  async run([targets]: [TypedSelection[]]): Promise<ActionReturnValue> {
+    await runOnTargetsForEachEditor(targets, async (editor, targets) => {
+      editor.selections = editor.selections.filter(
+        (selection) =>
+          !targets.find(
+            (target) =>
+              target.selection.selection.intersection(selection) &&
+              !target.selection.selection.start.isEqual(selection.end) &&
+              !target.selection.selection.end.isEqual(selection.start)
```
pokey, line 31: I believe you're checking that the selection has non empty overlap? Can't you just ask the intersection whether it is empty? If not, I'd add a comment to clarify what this is doing

pokey (approved):
Looks good! Left one minor comment. The tests look really good. I see you have an example where the target strictly contains the selection. In that vein, I'd add a test for the converse:

- [ ] Test where selection strictly contains target

src/actions/RemoveSelection.ts
AndreasArvidsson: Well you could do that. . . if it wasn't in the middle of the night :p
I have updated it now :)

src/actions/Deselect.ts
```diff
@@ -0,0 +1,41 @@
...
+      // Remove selections with a non-empty intersection
+      const newSelections = editor.selections.filter(
+        (selection) =>
+          !targets.find((target) => {
+            const intersection =
+              target.selection.selection.intersection(selection);
+            return intersection && !intersection.isEmpty;
+          })
+      );
+      // The editor requires at least one selection. Keep "primary" selection active
```
pokey, line 31: Shouldn't we just let this be an error?

src/actions/Deselect.ts
```diff
@@ -0,0 +1,41 @@
...
+      editor.selections = newSelections.length
+        ? newSelections
+        : [new Selection(editor.selection.active, editor.selection.active)];
+    });
+
+    return {
+      thatMark: targets.map((target) => target.selection),
```
pokey, line 38: Do you think the `that` mark would be more useful if it were the selections that got dropped? We could probably just leave it for now and see how it works in practice but just a thought

src/test/suite/fixtures/recorded/actions/giveDrum.yml
```diff
@@ -0,0 +1,28 @@
+spokenForm: give drum
```
pokey, line 1: Nice thorough tests 👍

src/actions/Deselect.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  async run([targets]: [TypedSelection[]]): Promise<ActionReturnValue> {
+    await runOnTargetsForEachEditor(targets, async (editor, targets) => {
+      // Remove selections with a non-empty intersection
+      const newSelections = editor.selections.filter(
+        (selection) =>
+          !targets.find((target) => {
+            const intersection =
+              target.selection.selection.intersection(selection);
+            return intersection && !intersection.isEmpty;
```
pokey, line 28: Ok I had one more thought: what happens if selection is empty, eg just a cursor? In that case there's no way to deselect it right? Maybe we should allow intersection to be empty if selection was empty to begin with?

src/actions/Deselect.ts
```diff
@@ -0,0 +1,41 @@
...
+  constructor(private graph: Graph) {
+    this.run = this.run.bind(this);
+  }
+
+  async run([targets]: [TypedSelection[]]): Promise<ActionReturnValue> {
+    await runOnTargetsForEachEditor(targets, async (editor, targets) => {
+      // Remove selections with a non-empty intersection
+      const newSelections = editor.selections.filter(
+        (selection) =>
+          !targets.find((target) => {
```
pokey, line 25: Total nit here sorry, but

```suggestion
          !targets.some((target) => {
```

pokey (commented):
Ok nice tests thanks. Left a few more comments. I think I found a case we're missing

src/actions/Deselect.ts
AndreasArvidsson: We could if you prefer that. If you have a really large selection just getting that out of the way with that the select command might be nice. It basically turns into an escape in that case

src/actions/Deselect.ts
AndreasArvidsson: I think we can leave it like this and try it out. In most cases that is token you actually referenced. But I do agree that could potentially be useful. 

Maybe we can use the source mark?

src/actions/Deselect.ts
AndreasArvidsson: That is a thought
you could just do a range over it but I get what you mean

AndreasArvidsson: Implemented and it works quite well. You can reference a token to remove empty selections on both sides of it or you can combine with before and after to just remove on a specific side.

src/actions/Deselect.ts
pokey: Oh that's kinda cool i guess.  Feels logically incorrect but useful enough to let it slide 😊

src/actions/Deselect.ts
pokey: Hmm not sure source mark is right here but interesting idea.  Agreed tho let's just see if we ever actually want a way to refer to the removed selections in practice

PR #351, opened 2021-12-02
Column selection type

### LINKED ISSUE #345: Modifier that allows you to put cursor on column in every row between two targets
```
foo bar
baz bongo
hey world
```

could say "take column from bat to whale" to put cursors before each second word

AndreasArvidsson:
Selection type to create multiple selections in a column between two targets instead of a continuous range

Right now doesn't do anything without a ranged target.

fixes #345

### DIFF
--- a/src/processTargets/index.ts
+++ b/src/processTargets/index.ts
@@ -63,7 +63,7 @@ function processRangeTarget(
   const anchorTargets = processPrimitiveTarget(context, target.anchor);
   const activeTargets = processPrimitiveTarget(context, target.active);
 
-  return zip(anchorTargets, activeTargets).map(
+  return zip(anchorTargets, activeTargets).flatMap(
     ([anchorTarget, activeTarget]) => {
       if (anchorTarget == null || activeTarget == null) {
         throw new Error("anchorTargets and activeTargets lengths don't match");
@@ -92,6 +92,38 @@ function processRangeTarget(
         !isForward,
         !target.excludeActive
       );
+
+      if (
+        anchorTarget.selectionType === "column" ||
+        activeTarget.selectionType === "column"
+      ) {
+        const start = isForward ? anchor : active;
+        const end = isForward ? active : anchor;
+        const tokens = context.navigationMap
+          .getTokens()
+          .filter(
+            (token) =>
+              token.displayLine >= start.line && token.displayLine <= end.line
+          );
+        tokens.sort((a, b) => a.offsets.start - b.offsets.start);
+        console.log(tokens);
+        const index = tokens.findIndex((token) =>
+          token.range.start.isEqual(start)
+        );
+        console.log("index", index);
+        let line = start.line;
+        let column = 0;
+        tokens.forEach((token) => {
+          if (token.displayLine !== line) {
+            line++;
+            column = 0;
+          }
+          if (column === index) {
+            console.log(token.text);
+          }
+          column++;
+        });
+      }
 
       const outerAnchor = target.excludeAnchor
         ? null
--- a/src/processTargets/index.ts
+++ b/src/processTargets/index.ts
@@ -171,17 +171,19 @@ function processColumnTarget(
     !isForward,
     !target.excludeActive
   );
-  const startLine = isForward ? anchorLine : activeLine;
-  const endLine = isForward ? activeLine : anchorLine;
   const anchorSelection = anchorTarget.selection.selection;
-  const startCharacter = anchorSelection.start.character;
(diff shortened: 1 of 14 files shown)


src/processTargets/index.ts
```diff
@@ -142,6 +155,45 @@ function processRangeTarget(
...
+    !target.excludeActive
+  );
+  const startLine = isForward ? anchorLine : activeLine;
+  const endLine = isForward ? activeLine : anchorLine;
+  const anchorSelection = anchorTarget.selection.selection;
+  const startCharacter = anchorSelection.start.character;
+  const endCharacter = anchorSelection.end.character;
+  const results: TypedSelection[] = [];
+
+  for (let i = startLine; i <= endLine; ++i) {
```
pokey, line 181: So I think to be the most consistent with what the user said, we should probably add the selections in order from anchor to active, rather than start to end.  Going to be a bit confusing when they try to sort a backwards list of selections, but it is consistent with the kind of thing we do elsewhere

src/processTargets/index.ts
```diff
@@ -142,6 +155,45 @@ function processRangeTarget(
...
+    results.push({
+      selection: {
+        selection: new Selection(i, startCharacter, i, endCharacter),
```
pokey, line 184: I think most consistent would be to use the anchor and active of the anchor selection rather than the start and end, so that the generated selections have the same direction has the anchor

src/test/suite/fixtures/recorded/selectionTypes/postColumnWhaleUntilEach.yml
```diff
@@ -0,0 +1,53 @@
...
+  documentContents: |
+    hello world !
+    hi there
+    Second things to say
+    and I'm here
+  selections:
+    - anchor: {line: 0, character: 11}
+      active: {line: 0, character: 11}
+    - anchor: {line: 1, character: 8}
+      active: {line: 1, character: 8}
```
pokey, line 43: Oh man this really confused me for a while 😅.  We should figure out some way to add comments to our recorded tests.  I see now that this is testing that it clamps to line end; nice test

pokey (commented):
This is super cool!  Left a few minor comments but otherwise looks good

src/processTargets/index.ts
```diff
@@ -142,6 +155,45 @@ function processRangeTarget(
...
+        editor: anchorTarget.selection.editor,
+      },
+      selectionType: "column",
+      selectionContext: {},
+      insideOutsideType: anchorTarget.insideOutsideType,
+      position: "contents",
```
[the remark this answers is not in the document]
AndreasArvidsson: Is the same code as we used for the normal range. The column modifier just doesn't change it. Please try it out without column and see if this is a problem.

src/test/suite/fixtures/recorded/selectionTypes/postColumnWhaleUntilEach.yml
AndreasArvidsson: Yeah some test becomes quite esoteric. But you probably had reason for recording it in the first place :D

src/processTargets/index.ts
AndreasArvidsson: done

src/processTargets/index.ts
AndreasArvidsson: done.

src/processTargets/index.ts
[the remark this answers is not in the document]
AndreasArvidsson: Just tried it and bringing after a range doesn't work. With or without the column modifier

src/processTargets/index.ts
```diff
@@ -142,6 +155,51 @@ function processRangeTarget(
...
+  const activeLine = targetToLineLimitPosition(
+    activeTarget,
+    !isForward,
+    !target.excludeActive
+  );
+  const anchorSelection = anchorTarget.selection.selection;
+  const delta = isForward ? 1 : -1;
+  const results: TypedSelection[] = [];
+
+  for (let i = anchorLine; true; i += delta) {
```
pokey, line 178: This would be a bit cleaner imo with a loadash range + map but not worth rewriting; maybe something to play with next time

pokey (approved):
Good stuff!

src/processTargets/index.ts
AndreasArvidsson: I must admit that I'm not a big fan of lodash. But you're probably right in this case and I did think about it.

PR #352, opened 2021-12-02
Implemented strict here selection type

AndreasArvidsson:
- [x] Switch strict here to instead just have a field `isImplicit`, and no other fields than `type`
- [x] Switch "just" to use `toRawTarget` modifier
- [x] Run test upgrade for strict here
- [x] Upgrade "just" tests (don't need backward compatibility stuff because it's unreleased)
- [x] [WON'T DO] Add backwards compatibility test for strict here


pokey (approved):
Ok I'm happy with this one but @AndreasArvidsson this one is kinda my PR now so I think you're the one that should review it 😅

PR #353, opened 2021-12-02
Change column modified to range vertical connector

AndreasArvidsson:


src/core/inferFullTargets.ts
```diff
@@ -75,6 +75,7 @@ function inferRangeTarget(
     type: "range",
     excludeAnchor: target.excludeStart ?? false,
     excludeActive: target.excludeEnd ?? false,
+    rangeType: target.rangeType,
```
pokey, line 78: Should we default this one to `"horizontal"` or `"standard"` or `"contiguous"` or something like that?

src/typings/Types.ts
```diff
@@ -220,8 +221,11 @@ export interface RangeTarget {
   active: PrimitiveTarget;
   excludeAnchor: boolean;
   excludeActive: boolean;
+  rangeType?: RangeType;
 }
 
+export type RangeType = "vertical";
```
pokey, line 227: as mentioned above, I'd be tempted to add another option and then make it required in the full `RangeTarget`

pokey (commented):
LGTM with minor tweaks

src/test/suite/fixtures/recorded/compoundTargets/bringAirToAfterBatVerticalPastFine.yml
```diff
@@ -1,5 +1,5 @@
-spokenForm: bring air to after column bat past fine
-languageId: markdown
+spokenForm: bring air to after bat vertical past fine
+languageId: typescript
```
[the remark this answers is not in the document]
AndreasArvidsson: Deni think we should do a cleanup because all tests I have added are probably on typescript and the majority of them don't require a language server. All take pre post clean actions, testing ranges and more. Should we do a cleanup pr maybe?

src/core/inferFullTargets.ts
AndreasArvidsson: We could do that if you prefer. The problem is that all existing tests need to be updated and I don't know how to do that yet.

AndreasArvidsson: Agreed we can default it in the inference stage

AndreasArvidsson:
@pokey The format of the test has changed and I can't figure it out. Can you solve the conflicts?

src/typings/Types.ts
```diff
@@ -220,8 +221,11 @@ export interface RangeTarget {
   active: PrimitiveTarget;
   excludeAnchor: boolean;
   excludeActive: boolean;
+  rangeType: RangeType;
 }
 
+export type RangeType = "vertical" | "continuous";
```
pokey, line 227: I think it's worth a comment to indicate what vertical / continuous mean

pokey (approved):
LGTM with minor comment

src/typings/Types.ts
AndreasArvidsson: done

AndreasArvidsson:
@pokey More merge conflicts due to change in test format

PR #318, opened 2021-11-08
Hat snapshots

pokey:
### Description

Allows voice engine to emit a signal to indicate start of phrase, which causes the hat token map to be frozen during the course of all commands issued during the given phrase.  The ranges will be updated as the document changes, but the same hat names will refer to the same logical document token during the course of the phrase.  This feature allows users to issue longer voice commands without worrying about the hats shifting during the course of phrase execution

### Related issues

- https://github.com/pokey/cursorless-talon/pull/101
- https://github.com/knausj85/knausj_talon/pull/676
- https://github.com/pokey/command-server/pull/10

### Todo
- [x] Add transformation that can upgrade old tests
- [x] Improve documentation for transform script
- [x] Record video walkthrough of using transform script
- [x] Add code links to hat snapshots doc (will do after merge)
- [x] Figure out how to mock prePhrase signal
- [x] Don't try to touch signal from talon if communication dir doesn't exist
- [x] Handle case where user has this version of cursorless-talon but old command-client; need to fall back to not using snapshot?  Do this talon side or extension side?
- [x] Check age of snapshot if it is requested. If too old, show error message and don't use snapshot
- [x] Switch `signals` to `inboundSignals`
- [x] Remove `getNamedSignal` from command-server API (here and in extension)
- [x] Fix issue with toggle decorations
- [x] Figure out why seeing duplicate hats after clone line then undo
- [x] Figure out promise rejection undefined

PR #356, opened 2021-12-02
Fix hats in local tests

pokey:
Ignores hat enablement settings so that tests don't break locally if you've changed your hat enablement


pokey:
Fixed in #318

PR #359, opened 2021-12-05
Disabled statement block on typescript

AndreasArvidsson:
At the moment a `statement` in typescript will just captured a block/body of if statements and loops. Disabling statement block makes the ancestral crawl find the entire statement instead.

### DIFF
--- a/src/languages/typescript.ts
+++ b/src/languages/typescript.ts
@@ -46,7 +46,7 @@ const STATEMENT_TYPES = [
   "lexical_declaration",
   "module",
   "return_statement",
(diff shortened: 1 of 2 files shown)


pokey:
Weird.  So this issue only arises if you are within the body of the if statement but not within one of its containing statements, correct?  Trying to figure out why I've never encountered it before

AndreasArvidsson:
You have to be inside the curly braces. In that case you will only select inside the braces and not the entire statement.

pokey (approved):
good stuff

PR #365, opened 2021-12-07
Add rewrap action

### LINKED ISSUE #270: Support "round rewrap" action

pokey:
Adds support for `rewrap` action which can be used to change surrounding paired delimiters

Closes #338 
Closes #270 

See also https://github.com/pokey/cursorless-talon/pull/120

squareRepackHarp


src/processTargets/processSelectionType.ts
```diff
@@ -226,11 +226,11 @@ function getTokenSelectionContext(
   }
 
   return {
+    ...selectionContext,
```
pokey, line 229: This way `processToken` won't clobber things it doesn't know about

src/test/suite/fixtures/recorded/surroundingPair/textual/chuckPairHarp.yml
```diff
@@ -0,0 +1,29 @@
+languageId: plaintext
```
pokey, line 1: Testing that `processToken` still causes whitespace cleanup for surrounding pair scopes

AndreasArvidsson (approved):

PR #368, opened 2021-12-07
Added a rebounce delay setting

AndreasArvidsson:


src/core/HatAllocator.ts
```diff
@@ -72,11 +71,15 @@ export class HatAllocator {
       clearTimeout(this.timeoutHandle);
     }
 
+    const decorationDebounceDelay = vscode.workspace
+      .getConfiguration("cursorless")
+      .get<number>("decorationDebounceDelay")!;
```
pokey, line 76: Do you think there's much of a performance impact to reading this one every time the document changes?  The proper way to do this would be to set in constructor and then listen for settings changes, but I can't decide whether it's worth the effort.  I'll leave it to you to decide whether to do that or just leave this one as is

pokey (approved):
Looks good with minor comment

src/core/HatAllocator.ts
AndreasArvidsson: Might actually be that vscode internally caches settings if the file hasn't updated

PR #370, opened 2021-12-07
Fix JSON surrounding pairs

### LINKED ISSUE #355: Fix Typescript text-based surrounding pairs
In the latest version of the tree-sitter parser, strings contain string fragments, so the existing `stringMatcher` won't detect them as strings, and thus won't use the text-based parser.  Thus, the following will fail

```typescript
"(hello)"
```

If we say "take round harp".  We need to be a bit more liberal in https://github.com/pokey/cursorless-vscode/blob/0c394d787e3dc88b7f6e629b86789fe76deaf93f/src/processTargets/modifiers/surroundingPair/index.ts#L71

I'm not sure we want to look at all ancestors, or potentially even parents, because they we might avoid parse tree when we're in an interpolation scope within a template string, (eg `` `foo ${(bar)}` ``)

We might need to allow languages to define a `isInStringFragment` function that will default to just using the string matcher, but in typescript will also add `string_fragment` type

Try running Cursorless unit tests with https://github.com/pokey/vscode-parse-tree/pull/12 installed to see the failure

### LINKED ISSUE #364: Text-based surrounding pair doesn't work in json strings
Eg in the following:

```json
"(hello)"
```

Saying "take round harp" doesn't work.  Could be related to #355

pokey:
Closes #364
Closes #355 

- [x] Fix Java tests
- [x] Fix #355 
- [x] Error nodes should be treated as text fragments
- [x] Allow `<` if it has sibling that is `>`?
  - [x] Or maybe allow it if it is first child, similar to the way we do for quotes?


src/languages/getNodeMatcher.ts
```diff
@@ -0,0 +1,99 @@
+import { SyntaxNode } from "web-tree-sitter";
```
pokey, line 1: Moved stuff out of `index.ts`

src/processTargets/modifiers/surroundingPair/findSurroundingPairParseTreeBased.ts
```diff
@@ -204,12 +204,10 @@ function findSurroundingPairContainedInNode(
...
 
           // Distinguish between a greater-than sign and an angle bracket by
-          // looking at its parent type
+          // looking at its position within its parent node.
           if (
             delimiterInfo.delimiter === "angleBrackets" &&
-            !ALLOWABLE_ANGLE_BRACKET_PARENTS.includes(
-              delimiterNode.parent?.type!
-            )
+            inferDelimiterSide(delimiterNode) !== delimiterInfo.side
```
pokey, line 210: I think this approach will be more robust

src/test/suite/fixtures/recorded/surroundingPair/parseTree/json/clearRound2.yml
```diff
@@ -0,0 +1,23 @@
+languageId: json
+command:
+  version: 0
+  spokenForm: clear round
+  action: clearAndSetSelection
+  targets:
+    - type: primitive
+      modifier: {type: surroundingPair, delimiter: parentheses}
+initialState:
+  documentContents: // (hello)
```
pokey, line 10: Note that we support comments in json now because we fallback to text-based if we're in a parse-tree error node

AndreasArvidsson (approved):

PR #372, opened 2021-12-09
Include generic in java type definition

AndreasArvidsson:


src/test/suite/fixtures/recorded/languages/java/takeTypeBlueLook.yml
```diff
@@ -0,0 +1,39 @@
+languageId: java
```
pokey, line 1: Not worth rewriting all these tests, but I thought we had aligned on "clear" instead of "take" to make these tests easier to read. Makes it a bit painful having to manually count offsets to figure out what the test is doing

src/test/suite/fixtures/recorded/languages/java/takeTypeGust2.yml
```diff
@@ -0,0 +1,39 @@
...
+  marks:
+    default.g:
+      start: {line: 3, character: 42}
+      end: {line: 3, character: 48}
+finalState:
+  documentContents: |-
+    public class MyClass {
+        private MyClass () {
+            Map<String, String> map = new HashMap<String, String>();
+            List<String> list = new ArrayList<String>();
```
pokey, line 30: I don't think this behaviour is consistent with typescript. I believe in typescript it would select `List<String>` here. Can you double check that? I'd argue for consistency across languages to the extent that it makes sense

Are you arguing to have this behaviour in Java but the other in Typescript? Or to have this behaviour in both Java and Typescript?

In my mind the semantics of "type" are that it takes the type annotation of the current scope. I recognise that it is already a bit overloaded though because it can be used to refer to a type / interface declaration

src/test/suite/fixtures/recorded/languages/java/takeTypeBlueLook.yml
AndreasArvidsson: You're completely correct.

src/test/suite/fixtures/recorded/languages/java/takeTypeGust2.yml
AndreasArvidsson: I would like to have this behavior in both java and typescript. I have never used a generic class in typescript yet, but doing a test I believe it's incorrect in typescript as it is implemented now.

Exactly. The current scope. If your selection is in the value/new expression there,s a type there as well. And I think it's very useful being able to take both types.

src/test/suite/fixtures/recorded/languages/java/takeTypeBlueLook.yml
pokey: Fwiw I'm planning a rework of the tests to make modifier tests clearer / more compact, and then the "clear" / "take" thing won't matter.  Not sure if I've already filed an issue but I will at some point

src/test/suite/fixtures/recorded/languages/java/takeTypeGust2.yml
pokey: I would lean towards today's typescript behaviour, but I'm getting the impression you feel more strongly than I do, so I'll yield 😊.  Mind either changing typescript or at least filing an issue to do so in the future?

src/test/suite/fixtures/recorded/languages/typescript/clearTypeWhale.yml
```diff
@@ -0,0 +1,27 @@
+languageId: typescript
```
pokey, line 1: This test case fails, but I believe it should pass

src/util/nodeSelectors.ts
```diff
@@ -54,6 +54,23 @@ export function simpleSelectionExtractor(
   };
 }
 
+export function pairSelectionExtractor(
```
pokey, line 57: Docstring would be great here as it's not obvious to me what this function is supposed to do based on its name

src/test/suite/fixtures/recorded/languages/typescript/clearTypeOx.yml
```diff
@@ -0,0 +1,27 @@
+languageId: typescript
```
pokey, line 1: This test case fails, but I believe it should pass

src/languages/typescript.ts
```diff
@@ -69,12 +73,46 @@ const getTags = (selection: SelectionWithEditor, node: SyntaxNode) => {
   return startTag != null && endTag != null ? startTag.concat(endTag) : null;
 };
 
-const findTypeNode = (node: SyntaxNode) => {
-  const typeAnnotationNode = node.children.find((child) =>
-    ["type_annotation", "opting_type_annotation"].includes(child.type)
-  );
-  return typeAnnotationNode?.lastChild ?? null;
-};
+function typeMatcher(): NodeMatcher {
```
pokey, line 76: As indicated by test case I added, I believe there's a bug in this function

src/test/suite/fixtures/recorded/languages/java/takeTypeBlueLook.yml
AndreasArvidsson: than that sounds really useful

src/util/nodeSelectors.ts
AndreasArvidsson: done

src/test/suite/fixtures/recorded/languages/typescript/clearTypeOx.yml
AndreasArvidsson: I agree. Updated

src/languages/typescript.ts
```diff
@@ -69,12 +73,50 @@ const getTags = (selection: SelectionWithEditor, node: SyntaxNode) => {
...
+      node.type !== "new" &&
+      node.type !== "arguments"
+    ) {
+      const identifierNode = node.parent.children.find(
+        (n) => n.type === "identifier"
+      );
+      const argsNode = node.parent.children.find(
+        (n) => n.type === "type_arguments"
+      );
+      if (identifierNode && argsNode) {
```
pokey, line 90: How come we require a type arguments node?  Shouldn't "take type" still work if you are just in `new Foo()`?  I think it is more consistent to just take the full constructor no matter what it is

AndreasArvidsson: In that case it will do the same thing as `take token` which is fine. 
Done and done

pokey (approved):

PR #375, opened 2021-12-09
Preliminary clojure support

### LINKED ISSUE #373: Basic clojure support
- [x] List
- [x] Map

pokey:
Adds some basic Clojure support, including a few scope types.  Can prob merge without adding others, as this PR also makes sure that we have support for paired matchers, eg "take pair", "take round", "take curly", etc

Closes #373

- [x] Handle TODOs in code (and add tests for them)
- [x] Add some docstrings
- [x] Support "funk" (eg just a list where child 0 is "defn"?)


src/languages/clojure.ts
```diff
@@ -0,0 +1,17 @@
...
+  map: "map_lit",
+
+  // A list is either a vector literal or a quoted list literal
+  list: ["vec_lit", "quoting_lit.list_lit"],
+
+  string: "str_lit",
+
+  // A function call is a list literal which is not quoted
+  functionCall: "~quoting_lit.list_lit!",
+};
```
pokey, line 15: @sdorminey @somanythings Here's an initial set of supported scope types.  I couldn't figure out what to do with keys / values, because they seem to be treated identically by the parse tree 🤔.  I didn't add other things as well because I think it's better to have someone who really knows the language involved in that step

src/test/suite/fixtures/recorded/languages/clojure/clearCall.yml
```diff
@@ -0,0 +1,23 @@
+languageId: clojure
```
pokey, line 1: Hopefully these test cases are at least somewhat readable if you're unfamiliar with them 😅.  Basically they're all of the form "clear X", where "X" is a scope type such as "call", "list", "pair", etc.  The idea is that it deletes the matched scope, so you can see what it referred to.  

You can see the spoken form on line 4, and then check out the initial document contents on line 10, followed by the final document contents on line 16.  So you can see here that "call" (eg function call), matches a parenthesized expression that's unquoted

src/test/suite/fixtures/recorded/languages/clojure/clearList.yml
```diff
@@ -0,0 +1,23 @@
+languageId: clojure
```
pokey, line 1: Note how quoted lists are treated as lists, but unquoted are not; they're treated as function calls

src/test/suite/fixtures/recorded/surroundingPair/parseTree/clojure/clearBound.yml
```diff
@@ -0,0 +1,27 @@
+languageId: clojure
```
pokey, line 1: Checking that we can remove parens, tho in reality you'd prob say "chuck bound" instead, to remove without moving cursor.  You'll prob also want to try eg "box repack pair" to change parens to square brackets

src/languages/clojure.ts
somanythings: yeah it's entirely based on the number of items in the map it needs to be even.
a slight bugbear of mine in keyboard editing as well that I've not figured out. :shrug: the calva, will end up slurping more than I want too

pokey: Ok cool.  We can solve this one fairly easily in cursorless

AndreasArvidsson (approved):

src/languages/clojure.ts
pokey: Ok so fairly easily turned out to be wrong 😅.  But it works now.  Can now say "key", "value", and "item"

PR #376, opened 2021-12-09
 Added support for user define custom actions

AndreasArvidsson:


src/actions/CommandAction.ts
```diff
@@ -22,18 +22,22 @@ export default class CommandAction implements Action {
...
     { insideOutsideType: "inside" },
   ];
+  private command: string;
   private ensureSingleEditor: boolean;
 
   constructor(
     private graph: Graph,
-    private command: string,
-    { ensureSingleEditor = false } = {}
+    { command = "", ensureSingleEditor = false } = {}
```
pokey, line 30: I'd be tempted to use `null` instead of empty string here

src/actions/CommandAction.ts
```diff
@@ -69,7 +73,7 @@ export default class CommandAction implements Action {
 
   async run(
     [targets]: [TypedSelection[]],
-    { showDecorations = true } = {}
+    { command = "", showDecorations = true } = {}
```
pokey, line 76: As above, I'd probably just make this one optional or nullable and then use `??` below

src/actions/index.ts
```diff
@@ -59,6 +60,7 @@ class Actions implements ActionRecord {
   replaceWithTarget = new Bring(this.graph);
   reverseTargets = new Reverse(this.graph);
   rewrapWithPairedDelimiter = new Rewrap(this.graph);
+  runCommandOnSelection = new RunCommandOnSelection(this.graph);
```
pokey, line 63: I believe you should add this one to the `Action` type.  Would be nice if there were some way to make it error when you forget to do so 🤔

Also as mentioned in other PR, I think I'd tweak this name

pokey (commented):
Looks good with some minor comments

pokey (commented):
Looks good, but could use a couple tests for the new `executeCommand`, esp exercising a couple args such as `restoreSelection`

AndreasArvidsson:
Paste and copy use restore selection so already tested that way. I can add some more if you like?

AndreasArvidsson:
I added a test for a custom command. More than that I don't know how to do. We do not support different arguments in custom command csv yet and none of the makeshift actions do anything that our test recorder can use.

pokey (approved):

PR #377, opened 2021-12-11
Reworked decorations debounce delay with special exception for scrolling

AndreasArvidsson:
Removed single setting for debounce delay and replaced with two different default values. 
60Hz for all decorations except visible ranges at 25Hz to improve performance and remove annoyance when scrolling.


src/core/HatAllocator.ts
```diff
@@ -37,15 +39,27 @@ export class HatAllocator {
...
+      // An Event which fires when the selection in an editor has changed.
       vscode.window.onDidChangeTextEditorSelection(
-        this.addDecorationsDebounced
+        this.addDecorationsDebouncedDefault
       ),
-      vscode.workspace.onDidChangeTextDocument(this.addDecorationsDebounced)
+      // An Event which fires when the visible ranges of an editor has changed.
+      vscode.window.onDidChangeTextEditorVisibleRanges(() =>
+        // Increase debounced delay to stop the decorations moving when scrolling. 25Hz
+        this.addDecorationsDebounced(40)
```
pokey, line 61: Move to constant at top?

src/core/HatAllocator.ts
```diff
@@ -66,25 +80,25 @@ export class HatAllocator {
     }
   }
 
-  addDecorationsDebounced() {
+  // Use default debounced delay. 60Hz
+  addDecorationsDebouncedDefault() {
+    this.addDecorationsDebounced(16);
```
pokey, line 85: Move to constant at top?

src/core/HatAllocator.ts
```diff
@@ -37,15 +39,27 @@ export class HatAllocator {
         this.toggleDecorations
       ),
 
-      vscode.window.onDidChangeTextEditorVisibleRanges(
-        this.addDecorationsDebounced
+      // An Event which fires when the active editor has changed. Note that the event also fires when the active editor changes to undefined.
```
pokey, line 42: Liking all the comments here; helpful

pokey (commented):
Looks good but I'd like to have a quick play with it before we merge

src/core/HatAllocator.ts
AndreasArvidsson: Idea is that if more events want to deviate from the default we can add a different delay for each single event. A constant at the top would make more sense if we actually reused it multiple times.

src/core/HatAllocator.ts
AndreasArvidsson: I mean it's only every used at one place. What would the constant improve?

src/core/HatAllocator.ts
AndreasArvidsson: I thought so too. Some of the names aren't that clear.

src/core/HatAllocator.ts
pokey: I'm not a fan of magic numbers in code, but I guess I don't feel strongly here as you have a comment right above it explaining the magic number

src/core/HatAllocator.ts
pokey: As below, I have a philosophical objection but I think it's prob fine here as well

pokey:
Ok just wired it up.  Def nice to get that 16ms thing, but I think 40ms is a bit fast for scroll, as there are a bunch of refreshes as my momentum scrolling slows at the end.  Maybe we could bump it up a little bit, or make it a setting?

src/core/HatAllocator.ts
AndreasArvidsson: I agree that unnamed and undocument that numbers is not a good thing. In this case I think we're fine due to the comment.

pokey (approved):

PR #379, opened 2021-12-11
Add padding top if not set already

AndreasArvidsson:
If the user has not set a padding top add it. Prevent decoration on first row getting cut off. 

I found no way of actually setting a default value. Now we check if the user has changed the value and if not set it for them.

fixes: https://github.com/pokey/cursorless-talon/issues/93


src/extension.ts
```diff
@@ -8,11 +8,14 @@ import { TestCaseRecorder } from "./testUtil/TestCaseRecorder";
 import { getCommandServerApi, getParseTreeApi } from "./util/getExtensionApi";
 import isTesting from "./testUtil/isTesting";
 import CommandRunner from "./core/commandRunner/CommandRunner";
+import updateMissingSetting from "./util/updateMissingSetting";
 
 export async function activate(context: vscode.ExtensionContext) {
   const { getNodeAtLocation } = await getParseTreeApi();
   const commandServerApi = await getCommandServerApi();
 
+  updateMissingSetting("editor.padding.top", 5);
```
pokey, line 17: Hesitant to add to this file, as we're currently trying to remove as much as we can. Maybe this belongs in init function of `Decorations`?

pokey (commented):
LGTM tho I wanna quick test this one locally to see if that 5px is bothersome at all

src/extension.ts
AndreasArvidsson: Even though this helps with decoration problem something as core as updating the user settings I still believe belongs in this file. In the case we have to add another setting it's nice to have all the changes we make to the user setting in one place.

AndreasArvidsson:
@pokey On windows all the snippet tests are now failing for some reason. And only on windows of course

src/extension.ts
pokey: Ok, I don't feel super strongly so let's leave it here

pokey:
> @pokey On windows all the snippet tests are now failing for some reason. And only on windows of course

hmm yeah I noticed that on one of my branches too.  This must have just started, because I was able to merge a branch in earlier today.  Strange.  Does it work locally on Windows?

pokey:
Do you think we could use [this new feature](https://code.visualstudio.com/updates/v1_63#_configuration-defaults-overrides) from 1.63.0 to do this one more simply?  Just saw it in the release notes.  We'd need to set min vscode version to 1.63.0, which is maybe a bit aggressive 4 days after it was released?

AndreasArvidsson:
I do think that solution is better but wouldn't that require every user have the latest version of vscode? I don't know if everyone always running automatic updates.
At least think which should roll out one solution. I will let you choose which one.

pokey:
> I do think that solution is better but wouldn't that require every user have the latest version of vscode? I don't know if everyone always running automatic updates. At least think which should roll out one solution. I will let you choose which one.

As long as calling `paddingConfig.update` is not persistent, I'd lean towards avoiding any new APIs at least for another week or two.  Alternately, we could leave this PR out of the 0.24.0 release.  But if it is non-persistent, then I think I'd be fine shipping this one, esp as it's a fairly modest code change.  I'm not seeing a big downside

AndreasArvidsson:
The implementation in this pr is permanent. We actually updating the users settings.json

pokey:
Ah ok right. In that case I'd lean towards reworking this one to use the new api, but waiting to merge until after 0.24.4 is released. Make sense?

AndreasArvidsson:
I think so

pokey:
> I think so

Shall we close this one then?  I don't think there will be any code reused in the new implementation

PR #380, opened 2021-12-11
Only show debug log in development mode

AndreasArvidsson:


src/extension.ts
```diff
@@ -2,16 +2,19 @@ import * as vscode from "vscode";
...
 import { TestCaseRecorder } from "./testUtil/TestCaseRecorder";
 import { getCommandServerApi, getParseTreeApi } from "./util/getExtensionApi";
 import isTesting from "./testUtil/isTesting";
 import CommandRunner from "./core/commandRunner/CommandRunner";
+import { ExtensionMode } from "vscode";
 
 export async function activate(context: vscode.ExtensionContext) {
   const { getNodeAtLocation } = await getParseTreeApi();
   const commandServerApi = await getCommandServerApi();
+  const isModeDevelop = context.extensionMode === ExtensionMode.Development;
```
pokey, line 16: I like to be able to have this output active even not in develop mode.  I'd advocate for a setting and command to toggle debug output on or off

src/test/suite/recorded.test.ts
```diff
@@ -43,7 +41,6 @@ suite("recorded test cases", async function () {
     "../../../src/test/suite/fixtures/recorded"
   );
   const files = walkFilesSync(directory);
-  enableDebugLog(false);
```
pokey, line 46: So during tests, `context.extensionMode !== ExtensionMode.Development`?

pokey (commented):
As mentioned in review comment, I'd be hesitant to remove the debug logs without some way to re-enable them.  They can be useful, eg for users to send us information, and I do use them when not in development mode sometimes as well to see parse tree, exact command, etc

src/test/suite/recorded.test.ts
AndreasArvidsson: Correct. There are three modes: production, development and test

src/extension.ts
AndreasArvidsson: Is it enough with one or the other or do we need both?

pokey: I would lean towards both, because I will personally have them on all the time, and in other situations I'd just ask users to turn it on to check something.  It's what we have for decorations, where there's a setting that determines whether they're shown by default, and then a command to quickly toggle them on / off.

But I don't feel strongly.  Just having one or the other is prob good enough if you think it's too much noise to have both

AndreasArvidsson: I think that if you're going to be running it all the time and we want something simple for the users to turn on in debugging purposes a setting is probably the more appropriate. 
I assume this setting will only be active in production mode? It shouldn't enable debugging in test for example.

AndreasArvidsson: Added a setting called debug

src/util/debug.ts
```diff
@@ -1,16 +1,73 @@
-import * as vscode from "vscode";
```
pokey, line 1: This should to be a graph component.  I'm not sure it makes sense to do this refactor unless you're going to take it all the way to a being a graph component.  I'd lean towards just leaving things as they were if you're not comfortable with making a graph component

AndreasArvidsson: I could easily make it a graph component but decided not to. Nothing in this file needs to be accessed anywhere in the system. Who didn't make sense to me to actually expose it like that.

pokey: Because it has its own dependencies, so it should them from the graph rather than the hard-coded stuff in extension.  See for example how other graph components access the extension context, other apis, register themselves as disposables, etc

AndreasArvidsson: Okay I will update it

src/core/Debug.ts
```diff
@@ -0,0 +1,121 @@
...
+import { Graph } from "../typings/Types";
+
+const originalDebugLog = console.debug;
+const disabledDebugLog = () => {};
+
+console.debug = disabledDebugLog;
+
+export default class Debug {
+  getNodeAtLocation?: any;
+  disposable?: Disposable;
```
pokey, line 19:
```suggestion
  private getNodeAtLocation?: any;
  private disposable?: Disposable;
```

src/core/Debug.ts
```diff
@@ -0,0 +1,121 @@
...
+        this.enableDebugLog();
+        break;
+      // Test mode. Always disable.
+      case ExtensionMode.Test:
+        this.disableDebugLog();
+        break;
+      // Production mode. Enable based on user setting.
+      case ExtensionMode.Production:
+        this.evaluateSetting();
+        workspace.onDidChangeConfiguration(this.evaluateSetting);
```
pokey, line 38: don't forget to add this one to disposable

src/core/Debug.ts
```diff
@@ -0,0 +1,121 @@
...
+        break;
+    }
+  }
+
+  init(getNodeAtLocation: any) {
```
pokey, line 43: I'd prefer if this one came from the graph like we do with command server api, but not a showstopper

src/core/Debug.ts
```diff
@@ -0,0 +1,121 @@
...
+  window,
+  workspace,
+} from "vscode";
+import { SyntaxNode, TreeCursor } from "web-tree-sitter";
+import { Graph } from "../typings/Types";
+
+const originalDebugLog = console.debug;
+const disabledDebugLog = () => {};
+
+console.debug = disabledDebugLog;
```
pokey, line 15: Why do you set this here?

src/core/Debug.ts
```diff
@@ -0,0 +1,121 @@
...
+  private enableDebugLog() {
+    console.debug = originalDebugLog;
+    this.disposable = window.onDidChangeTextEditorSelection(
+      this.logBranchTypes()
+    );
+  }
+
+  private disableDebugLog() {
+    console.debug = disabledDebugLog;
+    this.dispose();
```
pokey, line 62: I wouldn't call dispose here, because when you add the other disposable for listening to config, you'll end up removing that as well

src/core/Debug.ts
```diff
@@ -0,0 +1,121 @@
...
+      .get<boolean>("debug")!;
+    if (debugEnabled) {
+      this.enableDebugLog();
+    } else {
+      this.disableDebugLog();
+    }
+  }
+
+  private logBranchTypes() {
+    return (event: TextEditorSelectionChangeEvent) => {
```
pokey, line 77: Why do you return a function here?  Why not just use the method itself (after binding it)?

src/core/Debug.ts
```diff
@@ -0,0 +1,121 @@
...
+      }
+
+      const leafText = ancestors[ancestors.length - 1].text
+        .replace(/\s+/g, " ")
+        .substring(0, 100);
+      console.debug(">".repeat(ancestors.length), `"${leafText}"`);
+    };
+  }
+
+  private print(cursor: TreeCursor, depth: number) {
```
pokey, line 116:
```suggestion
  private printCursorLocationInfo(cursor: TreeCursor, depth: number) {
```

src/extension.ts
```diff
@@ -95,12 +95,7 @@ export async function activate(context: vscode.ExtensionContext) {
     }
   }
 
-  context.subscriptions.push(
-    cursorlessRecordTestCaseDisposable,
-    vscode.window.onDidChangeTextEditorSelection(
-      logBranchTypes(getNodeAtLocation)
-    )
-  );
+  context.subscriptions.push(cursorlessRecordTestCaseDisposable);
```
pokey, line 98: Just 1 to go! 😊

pokey (commented):
Ok that looks a lot better, thanks.  Left a few minor comments

src/core/Debug.ts
AndreasArvidsson: I looked have you did in decoration.ts and you did not add onDidChangeConfiguration too disposable. Should I update both?

src/core/Debug.ts
AndreasArvidsson: To remove debugs from things that are initialized before this. The tree sitter for example

src/core/Debug.ts
AndreasArvidsson: Of course I'm going to have to change that in that case

src/core/Debug.ts
AndreasArvidsson: oh, that was how it was implemented before and I just copied it. I can clean it up

src/core/Debug.ts
pokey: Hmm. But what if you actually want to see those things?

src/core/Debug.ts
pokey: Sure, but I think in general prob best to avoid calling dispose unless you're actually disposing of the object

src/core/Debug.ts
AndreasArvidsson: Fixed that

pokey (commented):
Ok the last sticking points for me are:

- What happens if you want to see the initial debug output?  I guess maybe the answer is that if you really want that debug output you run extension in debugger? Mixed feelings
- Worth confirming that this doesn't remove debug output from other extensions

src/core/Debug.ts
AndreasArvidsson: What if you don't? I have now removed at line

AndreasArvidsson:
* Initial debug output is now enabled again. My view on it was that if you actually wanted dialogs you to use debug mode. I would not care to not even have a setting.
* I don't know another extension with debug logs so I can't really test that. Guess I have to make one ;)

AndreasArvidsson:
I have just tested with another extension and we will disable debug logs for all extensions as it seems. The way around this is for cursorless to use its own logging function which we can disable.

pokey:
> I have just tested with another extension and we will disable debug logs for all extensions as it seems. The way around this is for cursorless to use its own logging function which we can disable.

Hmm.  Yeah I'm definitely not in love with that.  That's crazy that there's so little sand boxing between extensions.  But does it disable it only when you call the other extension's function from cursorless, or always for that extension?

AndreasArvidsson:
I didn't call it from cursorless and still the debug log was disabled.

AndreasArvidsson:
Now we have our own debug logger which we can disable without affecting other extensions

src/core/Debug.ts
```diff
@@ -0,0 +1,130 @@
...
+  window,
+  workspace,
+} from "vscode";
+import { SyntaxNode, TreeCursor } from "web-tree-sitter";
+import { Graph } from "../typings/Types";
+
+const originalDebugLog = console.debug;
+const disabledDebugLog = () => {};
+
+export const log = {
```
pokey, line 15: Why not just have this be an attribute of the debug graph component?  That seems more natural to me

AndreasArvidsson: Because it's not available in every single file. Tying everything down to the graph structure is...cumbersome. But I will update it

src/core/Debug.ts
```diff
@@ -0,0 +1,128 @@
...
+    this.active = true;
+    this.disposableSelection = window.onDidChangeTextEditorSelection(
+      this.logBranchTypes
+    );
+  }
+
+  private disableDebugLog() {
+    this.active = false;
+    if (this.disposableSelection) {
+      this.disposableSelection.dispose();
```
pokey, line 70: do we need to set this to null?  or is it idempotent?

src/core/commandRunner/CommandRunner.ts
```diff
@@ -110,8 +110,8 @@ export default class CommandRunner {
       this.testCaseRecorder.commandErrorHook();
       const err = e as Error;
       vscode.window.showErrorMessage(err.message);
-      console.debug(err.message);
-      console.debug(err.stack);
+      this.graph.debug.log(err.message);
+      this.graph.debug.log(err.stack);
```
pokey, line 114: I'm tempted to make these just `console.error`, because that way we can catch the stack trace the first time an error occurs.  WDYT?

pokey (approved):
Ok I left two more minor comments (as standalone comments above) but otherwise this looks great!

src/core/Debug.ts
AndreasArvidsson: You shouldn't be able to end up in disable twice without having enabled in between. I tried toggling it back and forward without any problem.

src/core/commandRunner/CommandRunner.ts
AndreasArvidsson: I was also thinking that.

src/core/Debug.ts
pokey: no but the issue is that if the extension gets deactivated it will get called twice.  probably not a big deal, but just being fastidious

pokey: eg if you enable log, then disable it, then deactivate extension

Issue #160, opened 2021-07-28
Test a command's effect on visible ranges


brxck:
Currently we're storing visible ranges during test case recording, but we are not asserting anything about them. 

This will be interesting because visible ranges depend on the size of the editor, and because the tests will run headlessly in CI.

### To do

- [x] Test folding, esp #401 
- [x] Test scrolling as described in https://github.com/pokey/cursorless-vscode/issues/160#issuecomment-1004807911

brxck:
> [In headless Linux CI machines xvfb is required to run VS Code](https://code.visualstudio.com/api/working-with-extensions/continuous-integration#github-actions)

💭 [xvfb](https://linux.die.net/man/1/xvfb) allows us to specify a screen size...

pokey:
Ooh that's cool @brxck. So sounds like we might be able to do this one more easily on Linux. Then I guess we'd check OS and only check that scrolling happened on Linux, but on all OSes at least check that it didn't break?  Tho one problem is that different screen size might impact navigation map because cursorless only decorates visible ranges. So it might use hats in different places if it's trying to decorate more / less tokens. In current tests all files fit on screen so we don't have that issue. So we might need to get fancier bout navigation map if we impl this one. Eg actually modify it rather than just asserting it matches our expectations. (Which seems to be flakey for some reason anyway at this point 😅)

pokey:
Thinking about this one a bit more, I don't think we need to worry bout screen size.  We can test scrolling as follows.  Have a file like below:

```
hello
there
```

And then say "crown trap".  The "hello" should scroll off the top of the screen, and thus not be part of visible ranges

AndreasArvidsson:
@pokey Should this just be one of unit test or do we want to add is functionality to the test recorder? Your easy examples above would probably work but if we validate them for every single test in the test recorder there is potential for problems of course.

PR #382, opened 2021-12-11
Group by document and not editor

### LINKED ISSUE #316: Cannot issue `move` command from one editor to the other if they are both on the same document

AndreasArvidsson:
fixes #316


src/util/targetUtils.ts
```diff
@@ -29,10 +29,14 @@ export async function runForEachEditor<T, U>(
...
 ): Promise<U[]> {
+  // Actually group by document and not editor. If the same document is open in multiple editors we want to perform all actions in one editor or an concurrency error will occur.
+  const getDocument = (target: T) => getEditor(target).document;
+  const editorMap = groupBy(targets, getDocument);
   return await Promise.all(
-    Array.from(groupBy(targets, getEditor), async ([editor, editorTargets]) =>
-      func(editor, editorTargets)
-    )
+    Array.from(editorMap.values(), async (editorTargets) => {
+      const editor = getEditor(editorTargets[0]);
```
pokey, line 37:
```suggestion
      # Just pick any editor with the given document open; doesn't matter which
      const editor = getEditor(editorTargets[0]);
```

pokey (commented):
Looks good, but please add a test

pokey:
To my knowledge, we have no tests that operate on two editors with the same document, which is what this PR is supposed to fix.  Would be good to have a test for that

AndreasArvidsson:
To my knowledge we can't do tests like that. The code changes are used by almost every single action so we have regression testing. The issue with multiple editors for the same document is tested manually and I have no idea how to write a test for that.

pokey:
Yeah you can't use test case recorder but it shouldn't be too bad.  Have a look at https://github.com/pokey/cursorless-vscode/blob/14749f4f42b30fe8c7f7c9906ce79138982b0b2e/src/test/suite/backwardCompatibility.test.ts#L21-L36

AndreasArvidsson:
The only way to open the same document in multiple editors is to actually open a file on disk. A bit problematic on running unit tests isn't it?

pokey:
> The only way to open the same document in multiple editors is to actually open a file on disk. A bit problematic on running unit tests isn't it?

Can just make a temp file

AndreasArvidsson:
Of course true. I might have a look at that when I have some time over.

pokey:
I'm not sure I want us to get in the habit of merging PRs without tests. I made an exception for #401 because it required #160, which is out of scope for that PR. Not sure I see why a test is out of scope for this PR

AndreasArvidsson:
@pokey Test added. I took special care to make sure that the hats are now in different editors of the same document.

src/test/suite/groupByDocument.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  const hat1 = hatMap
+    .getEntries()
+    .find(([, token]) => token.editor === editor1 && token.text === "hello");
+  const hat2 = hatMap
+    .getEntries()
+    .find(([, token]) => token.editor === editor2 && token.text === "world");
+
+  const [color1, char1] = hat1![0].split(".");
+  const [color2, char2] = hat2![0].split(".");
```
pokey, line 44: **nit**: Please use [`HatTokenMap.splitKey`](https://github.com/pokey/cursorless-vscode/blob/6cb9dea7a6baacd9a7245b67ada7593270941850/src/core/HatTokenMap.ts#L57)

pokey (commented):
Awesome test thank you!  Left a tiny nit but otherwise looks great

pokey (approved):
looks good!

PR #383, opened 2021-12-12
Updated settings with format on save

AndreasArvidsson:


pokey:
Hmm. Mixed feelings about imposing this setting on contributors. I personally think everyone should have it on, but that feels like a personal choice

I think what we could do is make it a pre-commit hook and put it in CI. Tbh I actually thought it was already in CI but I have noticed that some non-prettier stuff has slipped through

AndreasArvidsson:
Well since you added formatting as a step in the ci I think we already force it on them?

pokey:
Well we force it to be formatted before it merges, but whether it formats on save feels like more of a personal decision. Perhaps some people would prefer to write a bunch of code and then format afterwards 🤷

But I do think CI / pre-commit hook make sense

PR #384, opened 2021-12-12
Fixed selection bug and added source mark to copy lines

AndreasArvidsson:
* Since is replaced was false empty selections at the end was incorrectly moved
* Added source mark


src/actions/CopyLines.ts
```diff
@@ -46,14 +46,20 @@ class CopyLines implements Action {
...
+        : new Range(range.start, range.start);
       return {
-        editor,
-        range: newRange,
-        text,
+        edit: {
+          editor,
+          range: newRange,
+          text,
+          isReplace: true,
```
pokey, line 59: I would think we only want it to be a replace of we're cloning up. If we clone down and cursor is at beginning of line won't this line cause the cursor to end up at the beginning of the upper line, rather than the lower one?

src/test/suite/fixtures/recorded/actions/clearSource.yml
```diff
@@ -0,0 +1,30 @@
+languageId: plaintext
```
pokey, line 1: This test is just testing "source", right? I'd put it in the `marks` dir

src/test/suite/fixtures/recorded/actions/clearSource2.yml
```diff
@@ -0,0 +1,29 @@
+languageId: plaintext
```
pokey, line 1: Not sure how this test differs from the other "source" test?

src/test/suite/fixtures/recorded/actions/clearThat.yml
```diff
@@ -0,0 +1,29 @@
+languageId: plaintext
```
pokey, line 1: Surely we already have a test for the "that" mark? If not then I'd move this one into `marks` like the "source" test

src/test/suite/fixtures/recorded/actions/clearThat2.yml
```diff
@@ -0,0 +1,30 @@
+languageId: plaintext
```
pokey, line 1: As above not sure how this differs from other "that" test

src/test/suite/fixtures/recorded/actions/cloneUpVest2.yml
```diff
@@ -0,0 +1,31 @@
+languageId: plaintext
```
pokey, line 1: This test looks identical to the other "clone up vest". What am I missing?

src/test/suite/fixtures/recorded/actions/cloneVest2.yml
```diff
@@ -0,0 +1,31 @@
+languageId: plaintext
```
pokey, line 1: Not sure how this one differs from the other "clone vest"

pokey (approved):
Looks good with some minor comments. Worth adding a test for "clone up" / "clone down" with cursor at start of line, because I believe there's a bug in "clone up" with cursor at start of line

src/test/suite/fixtures/recorded/actions/clearSource.yml
AndreasArvidsson: Well it's specifically testing the source implementation of this specific action. If you want any move it

src/test/suite/fixtures/recorded/actions/clearSource2.yml
AndreasArvidsson: One test clone up the other one tests clone down. The implementation of source is different

src/test/suite/fixtures/recorded/actions/clearThat.yml
AndreasArvidsson: comment about source

src/test/suite/fixtures/recorded/actions/clearThat2.yml
AndreasArvidsson: comment about source

src/test/suite/fixtures/recorded/actions/cloneUpVest2.yml
AndreasArvidsson: It probably is. There is no way to redo this without making another recording

src/test/suite/fixtures/recorded/actions/cloneVest2.yml
AndreasArvidsson: See above

src/test/suite/fixtures/recorded/actions/clearSource.yml
pokey: Not really. The test for the action includes the "source" output already

src/test/suite/fixtures/recorded/actions/clearSource2.yml
pokey: Again, I'd argue that the "source" output of the test itself captures that. I'd argue we just need one or two tests to check that it is possible to use "source" as a mark

src/test/suite/fixtures/recorded/actions/cloneUpVest2.yml
pokey: So the plan is to delete this duplicate test then?

src/test/suite/fixtures/recorded/actions/clearSource.yml
AndreasArvidsson: So it does. I will remove these tests

src/test/suite/fixtures/recorded/actions/cloneUpVest2.yml
AndreasArvidsson: Yes

src/test/suite/fixtures/recorded/actions/clearSource.yml
pokey: Worth a quick double check to make sure that we have at least one test for the "source" mark in `marks` but otherwise yeah I'd just remove these

PR #385, opened 2021-12-12
Added html language implementation

AndreasArvidsson:


src/test/suite/fixtures/recorded/languages/html/clearStartTag.yml
```diff
@@ -0,0 +1,29 @@
+languageId: html
```
pokey, line 1: I wonder if "clear" should leave the angles. Can't decide

src/test/suite/fixtures/recorded/languages/html/takeBoundsBatt.yml
```diff
@@ -0,0 +1,31 @@
...
+command:
+  version: 1
+  spokenForm: take bounds batt
+  action: setSelection
+  targets:
+    - type: primitive
+      modifier: {type: surroundingPair, delimiter: any, delimiterInclusion: excludeInterior}
+      mark: {type: decoratedSymbol, symbolColor: default, character: b}
+initialState:
+  documentContents: "<html>\r\n    <body>\r\n        <hr/>\r\n    </body>\r\n</html>"
```
pokey, line 11: Mind fixing line endings and switching to "clear"?

pokey (approved):
This looks good with a couple minor comments.

- [x] Also would be good to add a test to make sure that plaintext tag angle pair works the same as in html

src/test/suite/fixtures/recorded/languages/html/clearStartTag.yml
AndreasArvidsson: Take and clear always have the same range so far. I don't want to change that

src/test/suite/fixtures/recorded/languages/html/takeBoundsBatt.yml
AndreasArvidsson: I swear those line endings are going to kill me. I keep changing them and changing them but vscode has no way of actually forcing line endings.

AndreasArvidsson:
I added a plaintext version as well
Clear bounds does not work correctly in plaintext
`<hr/>` => `hr`
`</body>` => `/body`

src/processTargets/modifiers/surroundingPair/delimiterMaps.ts
```diff
@@ -9,7 +9,7 @@ export const delimiterToText: Record<
   SimpleSurroundingPairName,
   [IndividualDelimiterText, IndividualDelimiterText]
 > = {
-  angleBrackets: ["<", ">"],
+  angleBrackets: [["<", "</"], [">", "/>"]],
```
pokey, line 12: maybe try

```suggestion
  angleBrackets: [["</", "<"], [">", "/>"]],
```

pokey:
> I added a plaintext version as well Clear bounds does not work correctly in plaintext `<hr/>` => `hr` `</body>` => `/body`

yeah was worried about that.  what happens if you switch the order like [this](https://github.com/pokey/cursorless-vscode/pull/385/files#r770011028)?

src/test/suite/fixtures/recorded/surroundingPair/html/clearBoundsOx2.yml
```diff
@@ -0,0 +1,41 @@
+languageId: plaintext
```
pokey, line 1: Ok looks great. One more nit: would be good to put the plaintext tests in the other dir of text-based surrounding pair tests

pokey (approved):
Ok looks great. I left one more thing. Also thought of one more quick test case:

- [ ] test having parens inside a string to make sure it properly reverts to text-based when inside a string
- [ ] Also parens inside a comment

src/languages/html.ts
```diff
@@ -0,0 +1,52 @@
...
+  comment: "comment",
+};
+
+export const patternMatchers = createPatternMatchers(nodeMatchers);
+
+export function stringTextFragmentExtractor(
+  node: SyntaxNode,
+  selection: SelectionWithEditor
+) {
+  if (node.type === "attribute_value") {
```
pokey, line 47: Will an attribute value always be a string?

pokey (commented):
Ok looks good with one last question

src/languages/html.ts
AndreasArvidsson: No. But it's the only type that can contain other matching pairs

pokey: Ok so this is a hack, then, right?  Because there might be things that are children that are not strings?  It's ok if it's a hack, we should just document that in a comment

pokey: You could also just check node parent type as well maybe?

pokey: Or check that it has first and last child of type `"`?

AndreasArvidsson: I wouldn't say it's a hack. It is a type that can contain text based matching pairs. But I can't contain any children in the tree sitter as far as I know.

PR #386, opened 2021-12-12
Added support for user colors

AndreasArvidsson:
Added nine configurable colors for the user. I'm going to experiment with more colors so I had need of this myself.


pokey (commented):
Seems ok to me with minor comments.  What's left to do here?

src/core/Decorations.ts
```diff
@@ -193,7 +193,7 @@ export default class Decorations {
 
     // So that unit tests don't fail locally if you have some colors disabled
     const activeHatColors = isTesting()
-      ? HAT_COLORS
+      ? HAT_COLORS.filter((color) => !color.startsWith("user"))
```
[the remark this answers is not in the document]
AndreasArvidsson: Nothing fancy but it will do the work  :D

AndreasArvidsson:
> Seems ok to me with minor comments. What's left to do here?

Nothing really. I wasn't sure that this was the most elegant solution. But it is working already and I can't really see any immediate downsides with it except for it being kind of la sledgehammer approach with a lot of copy paste.

AndreasArvidsson:
You know what, I like this. Running with an extra colors seems to work fine. Feel free to merge this.

data/colorTest/index.html
```diff
@@ -0,0 +1,10 @@
+<html>
```
pokey, line 1: I think these two files are starting to abuse the notion of "data". To me, if the code in these files will be executed, then it is not data. The other files are data in my mind because we're not ever treating them as code to be executed

I'm not sure where to put them, though. Maybe we should just remove them?

package.json
```diff
@@ -150,6 +150,33 @@
 						},
 						"yellow": {
 							"type": "string"
+						},
```
pokey, line 153: Ok so one possibility for an alternate implementation is to not have any requirements for what the color names can be. Gives more flexibility, but might just make things more confusing for the user? Can't decide

pokey (commented):
Left a couple more comments

data/colorTest/index.html
AndreasArvidsson: I think the playground files are really useful as well as this new one. I agreed that a term data is not applicable

package.json
AndreasArvidsson: You mean just allow any key value in this map? Personally I would rather set my own names than have incorrect names and we can always ship et with default names of course that the user can change?

pokey: Shall we close this one then?  That's gonna be a completely different implementation

AndreasArvidsson: I want to try other solution first.

AndreasArvidsson: I don't think this implementation without any fixed names are hard the problem is that without any predefined names we can't have default settings. There will be no way for them to edit the color or enable the setting without editing the json file.

AndreasArvidsson: I've done some testing and as soon as you set `additionalProperties` to true you lose all settings gui for that. All the default values and existing settings are only accessible in the json file. 

As I see it we have two options:
1. Default names. The pros is that you can change color and enable them from the setting gui. The downside is that you can't select your own names. 
2. The user selects the names. Upside is that the user can choose their own names that match the color selection better. The downside is that you can't change enablement or color setting without editing the json file. The user can also spell the the color wrong in the csv files and we have no whitelist to reference against. 

I'm actually leaning towards the implementation in this pr. But we might want to reduce the number of colors to three or something. What do you think?

src/core/constants.ts
```diff
@@ -7,6 +7,15 @@ export const HAT_COLORS = [
...
   "yellow",
+  "userColor1",
+  "userColor2",
+  "userColor3",
+  "userColor4",
+  "userColor5",
+  "userColor6",
+  "userColor7",
+  "userColor8",
+  "userColor9",
```
pokey, line 18:
```suggestion
```

pokey (commented):
Looks good with a couple remaining comments

AndreasArvidsson:
I don't think it's that important so in that case I will just remove it.

pokey (approved):
looks good!

pokey (commented):
Sorry I should have been clearer.  The cursorless identifiers should remain `userColor1` and `userColor2`.  I was merely proposing default spoken forms of "navy" and "salmon" ("apricot" is fine too)

PR #390, opened 2021-12-14
Added default configuration for padding top

AndreasArvidsson:
I think you can just go ahead and merge this. Our required vscode version is still the same and it's still working if you have the correct version.


pokey:
I only seem to need 2px to avoid clipping.  Do you need a full 5px?

AndreasArvidsson:
Then we do 2px

pokey (approved):

PR #391, opened 2021-12-14
Adding Language Support for Go

trace-andreason:
I haven't recorded any test cases yet, but everything seems to be working. The only thing that is missing is `class` and `class name` as go does not have classes.


src/languages/go.ts
```diff
@@ -0,0 +1,74 @@
...
+// `curl https://raw.githubusercontent.com/tree-sitter/tree-sitter-go/master/src/node-types.json | jq '[.[] | select(.type == "_statement" or .type == "_simple_statement") | .subtypes[].type]'`
+const STATEMENT_TYPES = [
+  "_expression",
+  "assignment_statement",
+  "dec_statement",
+  "inc_statement",
+  "send_statement",
+  "short_var_declaration",
+  "_simple_statement",
+  "block",
```
pokey, line 29: I think we want to drop this one.  @AndreasArvidsson just fixed a bug in Typescript where if you are in an empty if statement, saying "take state" will take only the curly brackets when you really want the whole if statement.  It's worth recording a test case like the following:

```go
if true {
    |
}
```

where `|` is your cursor position, then saying "clear state".  It should result in an empty file

```suggestion
```

src/languages/go.ts
```diff
@@ -0,0 +1,74 @@
...
+  functionCall: "call_expression",
+  comment: "comment",
+  namedFunction: ["function_declaration", "method_declaration"],
+  type: cascadingMatcher(
+    patternMatcher("pointer_type"),
+    patternMatcher("qualified_type"),
+    patternMatcher("type_identifier"),
+    patternMatcher("function_declaration[result]"),
+    patternMatcher("method_declaration[result]"),
+  ),
```
pokey, line 64: Curious why you can't just do the following:

```suggestion
  type: [
    "pointer_type",
    "qualified_type",
    "type_identifier",
    "function_declaration[result]",
    "method_declaration[result]",
  ],
```

src/languages/go.ts
```diff
@@ -0,0 +1,74 @@
...
+  functionName: ["function_declaration[name]", "method_declaration[name]"],
+  anonymousFunction: "func_literal",
+  condition: conditionMatcher("*[condition]"),
+  argumentOrParameter: argumentMatcher("argument_list", "parameter_list"),
+  collectionItem: ["keyed_element.*", "element.*"],
```
pokey, line 69: How come you don't want
```suggestion
  collectionItem: ["keyed_element", "element"],
```

src/util/nodeSelectors.ts
```diff
@@ -54,6 +54,22 @@ export function simpleSelectionExtractor(
   };
 }
 
+export function selectNamedChild(childIdx: number) {
```
pokey, line 57:
```suggestion
export function namedChildAtIndexSelector(childIdx: number) {
```

src/util/nodeSelectors.ts
```diff
@@ -54,6 +54,22 @@ export function simpleSelectionExtractor(
...
+  return function (editor: TextEditor, node: SyntaxNode): SelectionWithContext {
+    var child = node.namedChild(childIdx);
+    if (child == null) {
+      child = node;
+    }
```
pokey, line 62: Why do we revert to the parent if we can't find a child at the given index?

src/util/nodeMatchers.ts
```diff
@@ -78,6 +79,13 @@ export function conditionMatcher(...patterns: string[]): NodeMatcher {
   return matcher(patternFinder(...patterns), conditionSelectionExtractor);
 }
 
+export function childMatcher(patterns: string[], child: number): NodeMatcher {
```
pokey, line 82:
```suggestion
export function childAtIndexMatcher(patterns: string[], childIndex: number): NodeMatcher {
```

pokey (commented):
Looks really good!  Left a few comments.  Let me know if you have any trouble with the test case recorder 😊

src/languages/go.ts
trace-andreason: Yeah the current behavior is wrong, and looks like removing `block` fixes it, thanks!

src/languages/go.ts
trace-andreason: I think I can, I was just seeing if the behavior was at all different. The type selection behavior doesn't work _exactly_ like I want it to, but one could argue that the way it is makes more sense.

src/languages/go.ts
trace-andreason: :+1:

src/util/nodeSelectors.ts
trace-andreason: I didn't really know what to do here because it looks like all the nodeSelectors always return a selection. Can I return `null`?

src/util/nodeSelectors.ts
trace-andreason: :+1:

src/util/nodeSelectors.ts
AndreasArvidsson: This is the wrong place to do that. The node finder can return a node or null. The purpose of the selector is to extract the correct range from at give node.

src/util/nodeMatchers.ts
trace-andreason: :+1:

src/util/nodeSelectors.ts
trace-andreason: @AndreasArvidsson I guess my dilemma here is, not every node has a child right? So what selection should I return in that case? Open to any suggestions on how this should be structured

AndreasArvidsson: No there are leaf nodes without any children. The selector should be called with a correct node from the beginning. Find the correct node before calling the selector. You should probably add this logic to childAtIndexMatcher.

AndreasArvidsson: Let me know if you have any problem with this and I will give you a hand. but basically you have to check that the node return by patternFinder actually has that child. Then pass the child itself to the selector. If you don't need any specific logic in the selector you can use the simple one which is default in the matcher.

trace-andreason: @AndreasArvidsson thanks! I think I understand what you are suggesting now

trace-andreason: @AndreasArvidsson fixed this, thanks for this suggestion!

AndreasArvidsson: Now it's looking like you are not using the patterns argument? This means that it will match any node at the correct index no mother of the parent type. I would probably solve with something like this:
```ts
export function childAtIndexMatcher(patterns: string[], childIdx: number): NodeMatcher {
  const finder = patternFinder(...patterns);
  return matcher(
    (node: SyntaxNode) => finder(node)?.namedChild(childIdx) ?? null
  );
}
```

trace-andreason: @AndreasArvidsson used your suggestion

src/test/suite/fixtures/recorded/languages/go/clearState.yml
```diff
@@ -0,0 +1,23 @@
+languageId: go
+command:
+  version: 0
+  spokenForm: clear state
+  action: clearAndSetSelection
+  targets:
+    - type: primitive
+      modifier: {type: containingScope, scopeType: statement, includeSiblings: false}
+initialState:
+  documentContents: "if true {\n\t\n}"
```
pokey, line 10: It's a bit odd that you have these newlines.  Mind running your tests through the [autoformatter](https://github.com/pokey/cursorless-vscode/blob/main/docs/contributing/test-case-recorder.md#autoformatting)?

```sh
yarn run compile && node ./out/scripts/transformRecordedTests/index.js
```

AndreasArvidsson: I formatted like that on purpose

AndreasArvidsson: Shouldn't we just add the autoformatter as the part of the test recorder?

pokey: The test recorder does output stuff formatted correctly, so I'm not sure how the test ended up this way

trace-andreason: @pokey I ran the automated formatter, but it didn't seem to have any affect :thinking:

pokey: Hmm yeah I wonder if it's because the string includes tabs?  It's supposed to look more like this https://github.com/pokey/cursorless-vscode/blob/8d5e372ed494d996d382ffd450165d2c474e94e6/src/test/suite/fixtures/recorded/languages/typescript/clearFunk.yml#L10-L15

pokey: I guess it's fine.  Probably not worth rewriting your tests.  Just maybe for the future try to use spaces

pokey (commented):
Wow 👏!  Fantastic first PR 😄

There are a few more tests I'd love to see:

- [x] "chuck arg" when there are two args in the arg list, to ensure that comma is properly cleaned up on removed arg
- [x] "chuck arg" when there is one arg in the arg list, just to make sure delimiter code doesn't get tripped up
- [x] "take every key" to ensure that sibling expansion works properly for keys
- [ ] "take every item" to ensure that sibling expansion works properly for items
- [ ] "take key" where the cursor is in the key itself
- [ ] "take value" where the cursor is in the value itself
- [ ] "take round" with a document like so: `"(hello)"`, and cursor in the middle of `hello`.  Will only work if you're on `develop` branch of cursorless-talon
- [ ] "clear if state"
- [ ] "clear condition"

Also, in other languages, we're able to say "take value" to select the value of a return statement.  Would that be useful in Go in your opinion?

pokey:
Also, would be great if you could file a PR against https://github.com/pokey/vscode-parse-tree for the parse tree support

trace-andreason:
@pokey isn't go already in there?
https://github.com/pokey/vscode-parse-tree/blob/main/src/extension.ts#L20

pokey:
> @pokey isn't go already in there?
> 
> https://github.com/pokey/vscode-parse-tree/blob/main/src/extension.ts#L20

Ohhh ok. That's how you did it 😂. Forgot go was already supported by that extension

trace-andreason:
@pokey ok, added almost all those tests. I didn't add these two:

"take round" with a document like so: "(hello)", and cursor in the middle of hello. Will only work if you're on develop branch of cursorless-talon
 "clear condition"

Do I have to add the first? I based my branch on main.

For the second, I can't get the `condition` action to register in talon, I'm not sure why.

pokey (approved):

pokey:
> @pokey ok, added almost all those tests. I didn't add these two:
> 
> "take round" with a document like so: "(hello)", and cursor in the middle of hello. Will only work if you're on develop branch of cursorless-talon "clear condition"
> 
> Do I have to add the first? I based my branch on main.
> 
> For the second, I can't get the `condition` action to register in talon, I'm not sure why.

no worries; I just added them myself.  Thanks for all the great work!  This should be released in the next couple days

trace-andreason:
Thanks @pokey and @AndreasArvidsson! I'm planning to do rust next!

AndreasArvidsson:
Good work!  :)

PR #394, opened 2021-12-16
Allow to fold single line

### LINKED ISSUE #39: Fold action not working properly with multiple list elements
When calling `"fold every element"` on [this list](https://github.com/pokey/cursorless-vscode/blob/974c91ba2d5341c35b29918a78f9dfc1164788ae/package.json#L44-L144), it folds the entire list, but if you call `"take every element"` followed by `"fold this"`, it works fine

AndreasArvidsson:
I could never figure out why none of the fold or unfold commands worked for me until I looked at the code. Apparently single line targets are filtered away. Why? The most obvious use case for fold is `fold air` and I can't see any reason why that shouldn't work. 

Also made it so you can fold in a different editor/across splits

fixes #39


pokey:
The problem is that if you say "fold every item", and one of the items is a single line, it will fold the whole map.  Sorry I should have left a note in a comment

pokey:
see https://github.com/pokey/cursorless-vscode/issues/39

pokey (approved):
Good work!

PR #395, opened 2021-12-16
Added support for child indexes in the pattern finder

### LINKED ISSUE #393: Add new index-based pattern matchers
- [x] `foo[0]`: Select the first named child of a node with type `foo`
- [ ] `foo[bar][0]`: Select the first child at field `bar` of a node with type `foo`

### LINKED ISSUE #481: Modifier funk doesn't work on the name of arrow functions
const myFunk = () => {};

`take funk cap` and `take funk made` fails

AndreasArvidsson:
We can now do patterns with child indexes:
```js
key: "pair[0]"
```
fixes #393 

Also fixed a serious bug in the pattern finder
fixes #481


pokey (commented):
This one looks good, but please add a test.  Either a proper unit test, or if easier, just rewriting  an existing test to use this mechanism.

Also, this PR is only a partial fix for #393, so we'll want to leave that one open or file a new issue for the second half

AndreasArvidsson:
For now nothing is used in this mechanism as for as I know. And writing a unit test might be hard because we need to create a tree sitter tree. 

It should fulfill1both user cases of #393

pokey:
> For now nothing is used in this mechanism as for as I know. And writing a unit test might be hard because we need to create a tree sitter tree.

What about https://github.com/pokey/cursorless-vscode/blob/14749f4f42b30fe8c7f7c9906ce79138982b0b2e/src/languages/go.ts#L73-L75?

> 
> It should fulfill1both user cases of #393

I'm not sure I believe that it will satisfy requirement 2 of #393, because my understanding is that we currently don't support having multiple children with the same field name.  Anyways, if it does support requirement 2 then it should be easy to change the sub matcher here https://github.com/pokey/cursorless-vscode/blob/14749f4f42b30fe8c7f7c9906ce79138982b0b2e/src/languages/clojure.ts#L124 to be `patternMatcher("*[value][1]"]`

AndreasArvidsson:
1. Yes we have a user case for it but nothing is using it right now. We could of course make it so it uses the new matcher.
2. That is not how I interpreted the issue. Each bracket is evaluated separately in a chain.

pokey:
> 1. Yes we have a user case for it but nothing is using it right now. We could of course make it so it uses the new matcher.

I'm arguing that changing that code to use this matcher would be an easy way to test it

> 2. That is not how I interpreted the issue. Each bracket is evaluated separately in a chain.

Ah, hmm.  I see now that my syntax proposal in that issue is ambiguous.  We probably do want `foo[bar][0]` to be selecting the first child of the child of `foo` at field `bar`, rather than the first child of `foo` at field `bar`, where we expect `bar` to be a field that allows multiple children.  We'll have to come up with a new syntax for the latter, which we use in Clojure.

AndreasArvidsson:
@pokey Could we merge this pr and still have the issue about the new foo[bar.1] syntax open? To avoid a horrendous merge conflict I had to do a quite important bugfix together with the updates already made here. I would really like to get those merged as soon as possible.

pokey:
Yep that's fine but see above about an easy way to get this PR tested

AndreasArvidsson:
I have actually no idea how to do that. The implementation enclosure needs the entire issue feature. I don't think we actually have a need for foo[0] syntax that I could leverage right now?

AndreasArvidsson:
Go updated to use new index based pattern matching

pokey (approved):
Looks good!

PR #304, opened 2021-10-26
Support insertSnippet action

pokey:
Adds support for the following type of command:

- "snip funk"
- "snip funk hello world"
- "snip funk after air"

Still todo:

- [x] Final check on cpu usage when no custom snippets
- [x] wait for #306 
- [x] fix "snip funk before this"
- [x] Test "snip if after air and bat"
- [x] Test "snip if after air and before bat"
- [x] Test "snip if after air and before bat and cap"
- [x] remove undo stop for delimiter insertion
- [x] fix `that` mark for "insert snippet after"
- [x] Fix bug with "snip funk before this" with multiple cursors in same statement (+test)
- [x] Fix bug with "pour this" with multiple cursors in same statement (+test)
- [x] test multiple placeholders of same variable for wrap, with it as wrap var
- [x] test multiple placeholders of same variable for wrap, with it not as wrap var
- [x] test multiple placeholders of same variable for insert, with it as substitution var
- [x] test multiple placeholders of same variable for insert, with it not as substitution var
- [x] Only show error messages in active window
- [x] Add docs

PR #398, opened 2021-12-17
Fix inside-outside for implicit targets

pokey:
Fixes bug introduced with new implicit target implementation.  Doesn't break anything here, but caused a breakage on #304 so I've added a [test](https://github.com/pokey/cursorless-vscode/blob/14c17682b130746f1ddcbe30402b2d30762c6851/src/test/suite/fixtures/recorded/actions/snipIf.yml) there

### DIFF
--- a/src/core/commandRunner/CommandRunner.ts
+++ b/src/core/commandRunner/CommandRunner.ts
@@ -66,10 +66,8 @@ export default class CommandRunner {
         action.getTargetPreferences(...extraArgs)
       );
 
-      if (this.graph.debug.active) {
-        this.graph.debug.log("Full targets:");
-        this.graph.debug.log(JSON.stringify(targets, null, 3));
-      }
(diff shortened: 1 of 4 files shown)


src/core/commandRunner/CommandRunner.ts
```diff
@@ -66,6 +66,9 @@ export default class CommandRunner {
         action.getTargetPreferences(...extraArgs)
       );
 
+      this.graph.debug.log("Full targets:");
```
AndreasArvidsson, line 69: Do you mind adding a (debug.active) around these? Seems unnecessary to run stringify if the result is not going to be used

pokey:
Ok 66 failing tests 😅.  This PR needs more thought; think I may just try to fix the insertSnippet branch another way

AndreasArvidsson (approved):

PR #399, opened 2021-12-17
Scala support

SCdF:
Basic Scala support. Tested with Scala 2.x syntax, idk enough about 3.x to know how it fares.

Relies on: https://github.com/pokey/vscode-parse-tree/pull/16


SCdF:
(putting this out there so people know it's happening, also because I can't run tests locally in WSL so this gives me CI)

src/test/suite/fixtures/recorded/languages/scala/takeClass.yml
```diff
@@ -0,0 +1,29 @@
+languageId: scala
```
pokey, line 1: One quick thing: we generally prefer "clear" instead of "take" for these tests, as it makes the test cases easier to read

SCdF: done!

src/languages/scala.ts
```diff
@@ -0,0 +1,100 @@
...
+  conditionMatcher,
+  trailingMatcher,
+  chainedMatcher,
+} from '../util/nodeMatchers';
+import { NodeMatcherAlternative, ScopeType } from '../typings/Types';
+
+// Generated by the following command:
+// > curl https://raw.githubusercontent.com/tree-sitter/tree-sitter-scala/master/src/node-types.json | jq '[.[] | select(.type == "_definition" or .type == "_expression" or .type == "_pattern") | .subtypes[].type]'
+const STATEMENT_TYPES = [
+  // FIXME: is this right? What is a "statement" in a language where most traditional statements are expressions
```
pokey, line 14: As you suggest here, this list is very likely not what you want 😄. Notice for example that it has "identifier", so it would rarely expand very far. 

If you do want to support "statement" for Scala, I'd maybe start with just declaration types?

Fwiw I didn't define any statement scope type when I did clojure

SCdF: Yeah I'm going to swing back around to this one once I'm more familar with what the master branch generates

SCdF:
Also @pokey I'm guessing the tests are failing ("do not support plaintext") is because the version of vscode-parse-tree that's out there doesn't yet support scala?

src/languages/scala.ts
SCdF: I have decided to just not implement it for now. The more I use the scala bindings the more I can work out where they are weak and adjust accordingly

src/languages/scala.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  ifStatement: 'if_expression',
+
+  string: ['interpolated_string_expression', 'string'],
+  comment: 'comment',
+
+  // lists basic definition is just a function call to a constructor, eg List(1,2,3,4)
+  // MISSING: fancy list style: val foo = 1 :: (2 :: (3 :: Nil)) // List(1,2,3)
+  list: 'call_expression',
+  map: 'call_expression',
```
pokey, line 24: Hmm these seem a bit generic.  Maybe it would be better to check the name of the function that is being called to ensure it is `List` or `Map`?  You'll have to write a custom node finder, but you just check type of the parent node, then access the function call node and check `node.text`.  Lmk if you want to pair on this one.

If you do get it working, it's worth adding a test case where you have a list, one of whose elements is a function call, to ensure that when you say "clear list" within that function call you get the outer list rather than just the call

src/languages/scala.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  // list.size(), does not count foo.size (field_expression), or foo size (postfix_expression)
+  functionCall: 'call_expression',
+  namedFunction: 'function_definition',
+  // Do we want to consider partial functions as lambdas? eg
```
pokey, line 29: I would say "yes", but I have never used Scala so maybe there's some reason not to?

src/languages/scala.ts
```diff
@@ -0,0 +1,64 @@
...
+  name: ['*[name]', '*[pattern]'],
+  functionName: 'function_definition[name]',
+
+  // MISSING: function/class/trait generics
+  type: leadingMatcher([
+    'generic_type.type_identifier',
+    'type_identifier',
+    'type_parameters',
+  ], [':']),
+  value: leadingMatcher(['*[value]', '*[default_value]'], ['=']),
```
pokey, line 47: nice 👍

src/languages/scala.ts
```diff
@@ -0,0 +1,64 @@
...
+  condition: conditionMatcher('*[condition]'),
+
+  // Pulled from the complete list that isn't implemented above
+
+  // collectionItem: "???"
+  // collectionKey: "???",
```
pokey, line 53: Don't you want to support `"AL" -> "Alabama"` expressions?  For these, "key" would be `"AL"`, "value" would be `"Alabama"`, and "item" would be the pair.

`collectionItem` is also supposed to include list elements, tho since those are just function calls you would basically use an argument matcher and check that the parent is a `List`.  For fancy stuff like this, you might check out the [Clojure implementation](https://github.com/pokey/cursorless-vscode/blob/main/src/languages/clojure.ts).  That one took tons of custom matchers to get working

src/languages/scala.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  // regularExpression: "???",
+
+  // attribute: "???",
+  // xmlElement: "???",
```
pokey, line 58: Doesn't scala support inline html elements?

src/test/suite/fixtures/recorded/languages/scala/chuckTypeLook.yml
```diff
@@ -0,0 +1,29 @@
+languageId: scala
```
pokey, line 1: nice test 👍

src/test/suite/fixtures/recorded/languages/scala/chuckTypeRed.yml
```diff
@@ -0,0 +1,33 @@
+languageId: scala
```
pokey, line 1: We usually define type to expand to the entire type expression, rather then getting trapped in a subcomponent of the type, so I'd be inclined to change the matcher

src/test/suite/fixtures/recorded/languages/scala/chuckTypeRed2.yml
```diff
@@ -0,0 +1,33 @@
+languageId: scala
```
pokey, line 1: same with this one

src/test/suite/fixtures/recorded/languages/scala/chuckTypeSun2.yml
```diff
@@ -0,0 +1,29 @@
+languageId: scala
```
pokey, line 1: same here

docs/contributing/adding-a-new-language.md
```diff
@@ -11,15 +11,20 @@ for how to add support for a new parser
 
 ## 2. Define parse tree patterns in Cursorless
 
+Minimum changes that each language needs:
+
+- new file in `/src/languages/<yourlanguage>.ts`. Take a look at [existing languages](../../src/languages) as a base. At its core you're implementing your language's version of the `nodeMatchers` const, mapping scope types found in [`Types.ts:ScopeType`](../../src/typings/Types.ts) with matching expressions that align with the parse tree output.
+- new entry in [`getNodeMatcher.ts:languageMatchers`](../../src/languages/getNodeMatcher.ts), importing your new file above
+- new entry in [`constants.ts`](../../src/languages/constants.ts)
+- new text fragment extractor (default is likely fine) in [`getTextFragmentExtractor.ts:textFragmentExtractors`](../../src/languages/getTextFragmentExtractor.ts)
+
```
pokey, line 20: 👍

pokey (commented):
Really good stuff!  Left some inline comments.  Also, it would be great to see the following tests:

- [x] Tests for "type" where the mark is not in the type itself.  I don't know scala enough to say what situations would be comprehensive, but here are a couple that jump out at me:
   - [x] Have the mark in the value / body of a function and say "clear type" to see that it clears the return type of the function
   - [x] Have the mark in the name of a function call argument and say "clear type" to see that it clears the type of the function call argument
- [x] Similarly, tests for "value" where the mark is not in the value itself, eg it's in the name / type of the argument or assignment
- [x] A check for "condition" where the mark is not in the condition itself
- [x] "take round" with a document like so: `"(hello)"`, and cursor in the middle of `hello` to check that the text fragment extractor is working

src/test/runTest.ts
```diff
@@ -7,7 +7,11 @@ import {
   downloadAndUnzipVSCode,
 } from "vscode-test";
 
-const extensionDependencies = ["pokey.parse-tree", "ms-toolsai.jupyter"];
+const extensionDependencies = [
+  "pokey.parse-tree",
+  "ms-toolsai.jupyter",
+  "scalameta.metals",
```
pokey, line 13: In case you're curious, this line was what fixed CI.  The problem is that without this extension, VSCode doesn't know about the `scala` language id, so when our test runner creates a language with that language id, it just ignores us and treats it as `plaintext`, hence the error message where cursorless tried to process scope types for `plaintext`

SCdF:
On types, given:

```scala
class Example(foo: String) {
  def str(bar: String): String = foo + bar
}
```

We get:

```
(compilation_unit [0, 0] - [3, 0]
  (class_definition [0, 0] - [2, 1]
    name: (identifier [0, 6] - [0, 13])
    class_parameters: (class_parameters [0, 13] - [0, 26]
      (class_parameter [0, 14] - [0, 25]
        name: (identifier [0, 14] - [0, 17])
        type: (type_identifier [0, 19] - [0, 25])))
    body: (template_body [0, 27] - [2, 1]
      (function_definition [1, 2] - [1, 42]
        name: (identifier [1, 6] - [1, 9])
        parameters: (parameters [1, 9] - [1, 22]
          (parameter [1, 10] - [1, 21]
            name: (identifier [1, 10] - [1, 13])
            type: (type_identifier [1, 15] - [1, 21])))
        return_type: (type_identifier [1, 24] - [1, 30])
        body: (infix_expression [1, 33] - [1, 42]
          left: (identifier [1, 33] - [1, 36])
          operator: (operator_identifier [1, 37] - [1, 38])
          right: (identifier [1, 39] - [1, 42]))))))
```

From what you're saying it looks like we want to walk up the tree to find a sibling type, and take the parent. For example, if your cursor was in `class_parameter.name` (foo) and you said "take type" it should find `class_parameter.type` (String) and then match `class_parameter` (foo: String)?

How would I write this? And how would I merge this with the leading matcher I'm already using that I obviously do not want to use in this case.

pokey:
> On types, given:
> 
> ```scala
> class Example(foo: String) {
>   def str(bar: String): String = foo + bar
> }
> ```
> 
> We get:
> 
> ```
> (compilation_unit [0, 0] - [3, 0]
>   (class_definition [0, 0] - [2, 1]
>     name: (identifier [0, 6] - [0, 13])
>     class_parameters: (class_parameters [0, 13] - [0, 26]
>       (class_parameter [0, 14] - [0, 25]
>         name: (identifier [0, 14] - [0, 17])
>         type: (type_identifier [0, 19] - [0, 25])))
>     body: (template_body [0, 27] - [2, 1]
>       (function_definition [1, 2] - [1, 42]
>         name: (identifier [1, 6] - [1, 9])
>         parameters: (parameters [1, 9] - [1, 22]
>           (parameter [1, 10] - [1, 21]
>             name: (identifier [1, 10] - [1, 13])
>             type: (type_identifier [1, 15] - [1, 21])))
>         return_type: (type_identifier [1, 24] - [1, 30])
>         body: (infix_expression [1, 33] - [1, 42]
>           left: (identifier [1, 33] - [1, 36])
>           operator: (operator_identifier [1, 37] - [1, 38])
>           right: (identifier [1, 39] - [1, 42]))))))
> ```
> 
> From what you're saying it looks like we want to walk up the tree to find a sibling type, and take the parent. For example, if your cursor was in `class_parameter.name` (foo) and you said "take type" it should find `class_parameter.type` (String) and then match `class_parameter` (foo: String)?
> 
> How would I write this? And how would I merge this with the leading matcher I'm already using that I obviously do not want to use in this case.

Pretty sure you can just define a pattern matcher `"*[type]"`

pokey:
> On "take round", this does not work ("TypeError: Cannot read property 'endPosition' of undefined"). Any hints as to how to fix this?

Yeah probably scala isn't creating child nodes for quotes.  I'd try to steal Java's text fragment extractor

src/languages/scala.ts
SCdF: For now I'm going to drop these, as users can use "call" just as effectively

SCdF:
OK I added code / tests for types, and used the hacked text fragment extractor and it looks to work correctly

src/languages/scala.ts
SCdF: Looking into this, they're not identifiable in the current tree setter implementation, or possibly ever. Dropping the comment for now, I guess I'll see how much it annoys me.

src/languages/scala.ts
SCdF: I think I'd have to write something custom for that: `("AL" -> "Alabama")` comes out as:
```
          arguments: (arguments [1, 19] - [1, 38]
            (infix_expression [1, 20] - [1, 37]
              left: (string [1, 20] - [1, 24])
              operator: (operator_identifier [1, 25] - [1, 27])
              right: (string [1, 28] - [1, 37]))))))))
```

So I'd need to take the left or right of an operator, when that operator === '->'

src/languages/scala.ts
SCdF: It does (well, XML), but it's deprecated and will be dropped in 3.x, to be replaced with string formatting (I think, I don't know much about 3.x). I'm not sure how much people use it. The tree sitter implemention also errors with it.

src/languages/scala.ts
```diff
@@ -0,0 +1,66 @@
...
+  name: ['*[name]', '*[pattern]'],
+  functionName: 'function_definition[name]',
+
+  // MISSING: function/class/trait generics
+  type: leadingMatcher([
+    '*[type]',
+    '*[return_type]',
+    'generic_type.type_identifier',
+    'type_identifier',
+    'type_parameters',
```
pokey, line 42: What happens if you remove these three lines?

src/languages/scala.ts
pokey: Yeah that looks like it should be doable, but a bit custom.  I think I'd be ok with shipping without it for now but filing an issue to capture all the missing things.  I believe cursorless will generate a helpful message so that users can just see it hasn't been implemented yet

src/languages/scala.ts
```diff
@@ -0,0 +1,82 @@
...
+  comment: 'comment',
+
+  // list.size(), does not count foo.size (field_expression), or foo size (postfix_expression)
+  functionCall: 'call_expression',
+  namedFunction: 'function_definition',
+  anonymousFunction: 'lambda_expression',
+
+  argumentOrParameter: argumentMatcher('arguments', 'parameters', 'class_parameters', 'bindings'),
+
+  name: ['*[name]', '*[pattern]'],
```
pokey, line 28: interesting.  we don't do this in typescript, but maybe we should 🤔 cc/ @AndreasArvidsson

src/languages/scala.ts
```diff
@@ -0,0 +1,82 @@
...
+    'val_declaration[type]',
+    'var_definition[type]',
+    'var_declaration[type]',
+    'type_definition',
+    'extends_clause[type]',
+    'class_parameter[type]',
+    'parameter[type]',
+    'function_definition[return_type]',
+    'typed_pattern[type]',
+    'binding[type]',
```
pokey, line 46: nice work 👍.  Might add a comment listing a few examples that we've left out on purpose so someone doesn't try to just change it to `*[type]` in the future.  Your tests would catch that I believe so that's good

src/languages/scala.ts
```diff
@@ -0,0 +1,82 @@
...
+  value: leadingMatcher(['*[value]', '*[default_value]', 'type_definition[type]'], ['=']),
+  condition: conditionMatcher('*[condition]'),
+
+  // Scala features unsupported in Cursorless terminology
+  //  - Pattern matching
+
+  // Cursorless terminology not yet supported in this Scala implementation
+  /*
+    lists and maps basic definition are just function calls to constructors, eg List(1,2,3,4)
+    These types are also basically arbitrary, so we can't really hard-code them
```
pokey, line 58: Why can't we hard-code them?  Not saying you need to do it in this PR, but this comment implies that we can't do it, whereas I thought we were just punting

src/languages/scala.ts
pokey: btw I'd capture your example here into the follow-up issue for remaining scala features; it's quite helpful

SCdF:
Raised https://github.com/pokey/cursorless-vscode/issues/434 to talk about / track pattern matching

pokey (approved):
This looks great! 👏👏

src/languages/scala.ts
SCdF: They are just classes, not a syntax / language construct like `[]` in JS or whatever. Scala comes with [a bunch of them](https://docs.scala-lang.org/overviews/collections-2.13/overview.html) but people will also use other random libraries as well.

pokey: Interesting.  Almost feels like we need LSP support for this one?  Let's just call it out in the missing language features issue and discuss there

pokey (approved):

src/languages/scala.ts
AndreasArvidsson: What is it we don't do in typescript? The current implementation looks like this
```ts
name: [
    "*[name]",
    "optional_parameter.identifier!",
    "required_parameter.identifier!",
  ]
```

pokey: Oh yeah just tested and looks like it works.  The example was

```typescript
const {foo, bar} = baz;
```

Looks like saying "take name" already selects `{foo, bar}`, which is what I was proposing, so 👍

PR #401, opened 2021-12-17
Re enable fold single line

### LINKED ISSUE #72: fold/unfold requires funk modifier
None of these work:
`fold air`
`fold this`

You have to say:
`folk funk air`
`fold funk`

AndreasArvidsson:
Re enabled folding of single line. To "solve" the problem where `fold every state` will fold the parent if one of the states is a single line I added a constraint that you can't combine multi and single line targets. In that case just used the multiline targets. Multiple single line targets is fine though.

fixes #72

### DIFF
--- a/src/actions/Fold.ts
+++ b/src/actions/Fold.ts
@@ -29,16 +29,29 @@ class FoldAction implements Action {
       await focusEditor(editor);
     }
 
+    const singleLineTargets = targets.filter(
+      (target) => target.selection.selection.isSingleLine
+    );
+    const multiLineTargets = targets.filter(
+      (target) => !target.selection.selection.isSingleLine
+    );
+    // Don't mix multi and single line targets.
+    // This is probably the result of an "every" command
+    // and folding the single line targets will fold the parent as well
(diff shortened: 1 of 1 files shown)


pokey:
Tempted to introduce visible ranges to recorded tests for this one. As long as the file fits within a reasonable sized window I think it should be fairly straightforward

AndreasArvidsson:
> Tempted to introduce visible ranges to recorded tests for this one. As long as the file fits within a reasonable sized window I think it should be fairly straightforward

I think that should probably be fine

pokey (approved):
Ok I'm going to approve this one without tests because adding support for visible range testing is probably out of scope for this PR.  But it would be good to add that as a fast follow, because this folding stuff is getting a bit more complex than I'm comfortable not having tests for 😅

@AndreasArvidsson rather than requesting tests on this PR, I'll assign https://github.com/pokey/cursorless-vscode/issues/160 to you to do after this one is merged in so that we don't clutter this PR.  I think it actually should be fairly straightforward.  But if you don't have capacity no worries feel free to assign it to me

PR #402, opened 2021-12-17
Added asynchronous implementation of command action

AndreasArvidsson:


src/actions/CommandAction.ts
```diff
@@ -67,24 +69,45 @@ export default class CommandAction implements Action {
...
-            await callFunctionAndUpdateSelections(
-              this.graph.rangeUpdater,
-              () => commands.executeCommand(args.command!, args.commandArgs),
-              editor.document,
-              [originalSelections, targetSelections]
-            );
+          const commandPromise = commands.executeCommand(
+            args.command!,
+            args.commandArgs
+          );
```
pokey, line 75: I don't think you want to start the command before you call the track selections function. If it does an update before it yields you may miss the update. Could just be getting lucky that it hasn't hit you here

pokey (commented):
Haven't reviewed in depth but left one thought

src/actions/CommandAction.ts
AndreasArvidsson: Got it

AndreasArvidsson: I actually had reservations about that. Updated now.

src/actions/CommandAction.ts
```diff
@@ -67,24 +69,45 @@ export default class CommandAction implements Action {
...
-              this.graph.rangeUpdater,
-              () => commands.executeCommand(args.command!, args.commandArgs),
-              editor.document,
-              [originalSelections, targetSelections]
-            );
+          if (args.awaitCommand) {
+            const [updatedOriginalSelections, updatedTargetSelections] =
+              await callFunctionAndUpdateSelections(
+                this.graph.rangeUpdater,
+                () => commands.executeCommand(args.command!, args.commandArgs),
```
pokey, line 76: Just noticed:

- Command should never be undefined, right?  We should make it required if so so we can avoid the `!`
- We should spread `commandArgs` because some commands take more than one arg.  If this will break anything we should fix that too

```suggestion
                () => commands.executeCommand(args.command, ...args.commandArgs),
```

src/actions/CommandAction.ts
```diff
@@ -67,24 +69,45 @@ export default class CommandAction implements Action {
...
+
+            return updatedTargetSelections.map((selection) => ({
+              editor,
+              selection,
+            }));
+          }
+
+          const commandPromise = commands.executeCommand(
+            args.command!,
+            args.commandArgs
```
pokey, line 97: Here too
```suggestion
            args.command,
            ...args.commandArgs
```

pokey (commented):
I thought we decided to always await here, but not await talon-side if the command is long-running. What changed?

AndreasArvidsson:
> I thought we decided to always await here, but not await talon-side if the command is long-running. What changed?

That is the opposite of the impression I had. You specifically said that sleep time post command was still useful since you waited while the command was executing. If we don't wait talon side the post command sleep in vscode does nothing?

src/actions/CommandAction.ts
AndreasArvidsson: * It is. All arguments are available through the constructor or the run method. In that case they have to be optional. 
* Spreading the command argument creates a compilation error. The current implementation works fine with multiple arguments.

pokey (approved):

src/actions/CommandAction.ts
```diff
@@ -95,31 +86,42 @@ export default class CommandAction implements Action {
...
+      defaultOptions,
+      this.options,
+      options
+    );
+
+    if (partialOptions.command == null) {
+      throw Error("Command id must be specified");
+    }
+
+    const actualOptions = partialOptions as Required<CommandOptions>;
```
AndreasArvidsson, line 102: Required was a smart move. Didn't think about that one.

pokey: Yeah I love all those type transformations that typescript has

AndreasArvidsson: That type system in typescript is way powerful.

PR #403, opened 2021-12-17
Hide decorations for selections that are larger than the viewport

AndreasArvidsson:
Just work around so that `crown file` don't flash entire screen. Large functions that cover most of the viewport will have a similar effect but now at least we can get rid of those cases when the entire viewport would flash.


pokey (commented):
Not sure I love the idea of having no visual feedback. Maybe we could just make the color less dramatic?

AndreasArvidsson:
You still get the visual feedback of the window actually scrolling. I think that decorating the entire viewport gives us no extra information. The point of the decoration is for the user to get a feedback on which part of the text was actually affected/referenced. This is useful because you might not always know the bounds of a statement. If we just highlight the entire viewport you have no information about the exact range of the selection/reference and I then would argue that flashing the entire viewport is just irritating and gives the user no relevant feedback.

pokey:
That's fair. I'll pull it to get a feel for it

pokey (commented):
I tested this one locally but it doesn't work.  I said "crown file", and it still highlighted everything

AndreasArvidsson:
It only excludes selections that are larger than the viewport. The reason for this is that if you have a small file that maybe only contains a single function it still might be a good idea to be able to highlight the entire document. You need to have a file with more rows than you can fit in the viewport/editor for this fix to take effect.

pokey:
> It only excludes selections that are larger than the viewport. The reason for this is that if you have a small file that maybe only contains a single function it still might be a good idea to be able to highlight the entire document. You need to have a file with more rows than you can fit in the viewport/editor for this fix to take effect.

I tested it on a file that had more rows than fit in my viewport, and it still highlighted it

AndreasArvidsson:
Where you scrolled down? If your selection are matching the start or the end of the viewport it will still show.
Opening a large file and scrolling down will not highlight for me.

pokey:
But you check if the start matches after you've already scrolled, so you will always be at the start.  I'm surprised it works for you

AndreasArvidsson:
Worked for me. I will have another look at it tomorrow.

AndreasArvidsson:
Have another look at it now

src/actions/Scroll.ts
```diff
@@ -49,8 +47,23 @@ class Scroll implements Action {
...
+    const decorationSelections = targets
+      .map((target) => target.selection)
+      .filter((selection) => {
+        const visibleRanges = selection.editor.visibleRanges;
+        const start = visibleRanges[0].start;
+        const end = visibleRanges[visibleRanges.length - 1].end;
+        // Don't show decorations for selections that are larger than the visible range
+        return (
+          selection.selection.start.isAfter(start) ||
+          selection.selection.end.isBefore(end) ||
```
pokey, line 59: I believe you want to use line numbers here instead of just a raw before or after. This doesn't work correctly if the target is indented

AndreasArvidsson: You are correct. Code updated.

pokey (approved):

PR #408, opened 2021-12-20
Fix hat map jitter

pokey:
Change timeout for all hat map redecorations to 40ms because it was too jittery at 16ms.  Was slowing down dictation for me, and rapidly pressing the up arrow also caused the hats to flail

### DIFF
--- a/src/core/HatAllocator.ts
+++ b/src/core/HatAllocator.ts
@@ -24,7 +24,6 @@ export class HatAllocator {
       .get<boolean>("showOnStart")!;
 
     this.addDecorationsDebounced = this.addDecorationsDebounced.bind(this);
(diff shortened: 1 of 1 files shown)


src/core/HatAllocator.ts
```diff
@@ -22,14 +24,13 @@ export class HatAllocator {
       .get<boolean>("showOnStart")!;
 
     this.addDecorationsDebounced = this.addDecorationsDebounced.bind(this);
-    this.addDecorationsDebouncedDefault =
-      this.addDecorationsDebouncedDefault.bind(this);
+    this.addDecorationsDebounced = this.addDecorationsDebounced.bind(this);
```
pokey, line 27: Whoops looks like this one was a rename error
```suggestion
```

AndreasArvidsson (approved):

PR #417, opened 2021-12-27
Keyboard interface

pokey:
Enables cursorless to be used by keyboard

### DIFF
--- a/docs/internal/keyboard-version.md
+++ b/docs/internal/keyboard-version.md
@@ -1,0 +1,64 @@
+### Without modifier keys
+
+- `tda` => "take air" (ie "take default air")
+- `tga` => "take green air"
+- ? `tfda` => "take fox air" (ie "take fox default air")
+  - or `tsfda` => "take fox air" (ie "take shape fox default air")
+- ? `tfga` => "take fox green air"
+- `tfda` => "take funk air"
+- `tft` => "take funk" (ie "take funk this")
+- ? `tf<pause>` => "take funk"
+- `tldadbl` -> "take air and bat"
+  - Note that the `l` is used both to start and end a list
+- `trdadb` -> "take air past bat"
+  - Note that the range ends on its own because you've defined the start and end
+- `tlrdadbdcl` -> "take air past bat and cap"
+
+### To do / think about
+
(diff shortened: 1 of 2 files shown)


ym-han:
one quick thought about extending this to keyboard: it'd be good if the way that this is implemented is compatible with a steno setup (see, e.g., https://www.openstenoproject.org/plover/). I don't know offhand how plover does its thing; just wanted to note that it'd be good to look into it -- I'd be happy to do the necessary research once we start thinking about what abstractions to use. The reason why it'd be nice if it's compatible with steno is because if you use steno, you can basically type (they call it 'stroke') _the same things_ that a voiced cursorless user would say (steno allows you to do this  sustainably and efficiently). Consequently, **you wouldn't need to learn new keybindings if you already know steno and the cursorless voice bindings.** (This goes back to the appeal of having a unified interface that Andreas was talking about.)

I think steno users who don't already use Talon would find cursorless attractive as well --- I've seen steno users talking about how they have to use normal, non-steno typing often than they'd like when navigating code.

It might also not be too hard for those who don't already know steno to learn just enough to use cursorless with steno, given how there aren't too many unique words.

pokey:
@ym-han Yes I remember discussing this idea with someone at some point. The issue they pointed out was accuracy: talon is designed to try its best to map your utterance to the allowable grammar, whereas I'm not sure that's how plover works

But a POC probably wouldn't be too hard to make: you could just pass text from Plover to talon's `mimic` function

ym-han:
> @ym-han Yes I remember discussing this idea with someone at some point. The issue they pointed out was accuracy: talon is designed to try its best to map your utterance to the allowable grammar, whereas I'm not sure that's how plover works
> 
> But a POC probably wouldn't be too hard to make: you could just pass text from Plover to talon's `mimic` function

passing text from Plover to Talon's mimic function sounds like a great idea (and honestly probably robust enough). I'll have to learn some steno before it'll make sense for me to work on this, but it's on my to-do-eventually list now :P (assuming of course that no one else gets to it first)

pokey:
Closing for now in favour of #710; feel free to reopen if anyone wants to work on it

PR #419, opened 2022-01-01
Update description and keywords

pokey:
Also bumps version to 0.24.2


AndreasArvidsson:
\- "description": "Don't let the cursor slow you down"

Noo it's the end of an era! :D

AndreasArvidsson (approved):

PR #430, opened 2022-01-05
Cross-cell set selection

### LINKED ISSUE #420: Support moving cursor between cells in a notebook
Currently if you say `"take air"` and `"air"` is in a cell that doesn't contain your cursor, the cursor will not move to the new cell

pokey:
Fixes #420


src/test/suite/crossCellsSetSelection.test.ts
```diff
@@ -0,0 +1,54 @@
...
+});
+
+async function runTest() {
+  const graph = (await getCursorlessApi()).graph!;
+
+  await openNewNotebookEditor(['"hello"', '"world"']);
+
+  // FIXME: There seems to be some timing issue when you create a notebook
+  // editor
+  await sleep(1000);
```
pokey, line 27: This is a pain, but seems to be the only way to get CI to pass reliably 😕

AndreasArvidsson (approved):

PR #471, opened 2022-01-13
Simpler dollar pairs

pokey:
Don't include `$` in pairs for eg `$(foo)`


AndreasArvidsson (approved):

pokey:
Hmm ok this one is failing because in typescript the `${` is reported as a single token with type `${`.  So we'd need to do something typescript-specific to make this one work.  Hmm.  Also fwiw VSCode considers the dollar to be part of the leading delimiter of the pair:

image

So maybe we should match VSCode's behaviour here?

AndreasArvidsson:
We could of course do a work around where we find the `${` token and then step one character to the right

pokey:
I think this one needs more thought

PR #472, opened 2022-01-13
Make matching pair work inside regex in typescript

AndreasArvidsson:


pokey (commented):
Looks good but please add test

pokey (approved):

PR #480, opened 2022-01-14
Implemented selection type non white space sequence

### LINKED ISSUE #330: Add `nonWhitespaceSequence` scope type
We can call it "paint", and it will expand to include all non-whitespace characters adjacent to selection.  Can proceed by:

1. Get line containing end of selection
2. Move end forward until just before the first whitespace character you encounter
3. Get line containing start of selection
4. Move start back until just after the first whitespace character you encounter

Question: what to do if one of the endpoints is in the middle of whitespace?  Which of the following do we do:

- Leave the endpoint untouched
- Move it forward until the end of the next non-whitespace block
- Move it backward to the end of the preceding non-whitespace block

AndreasArvidsson:
fixes #330


src/processTargets/processSelectionType.ts
```diff
@@ -173,6 +175,78 @@ function processParagraph(
...
+  const { selectionType, insideOutsideType, position, modifier } = target;
+  const { document } = selection.editor;
+  const { start, end } = selection.selection;
+  let newSelection;
+
+  let startLine;
+  let i = start.line;
+  do {
+    startLine = document.lineAt(i++);
+  } while (startLine.isEmptyOrWhitespace && i <= end.line);
```
pokey, line 192: Why can't you just use `start.line` directly?

src/test/suite/fixtures/recorded/selectionTypes/takePaint.yml
```diff
@@ -0,0 +1,26 @@
+languageId: plaintext
```
pokey, line 1: Based on our discussion, this one should be an error, no?

src/test/suite/fixtures/recorded/selectionTypes/clearPaint.yml
```diff
@@ -0,0 +1,28 @@
+languageId: plaintext
```
pokey, line 1: I don't understand this test. Why is this not an error?

src/processTargets/processSelectionType.ts
```diff
@@ -173,6 +175,78 @@ function processParagraph(
...
+
+  // Can't find any non whitespace characters in the selection
+  if (startLine.isEmptyOrWhitespace) {
```
pokey, line 195: This implementation seems overly complicated. What's wrong with the approach proposed in the issue description?

pokey (commented):
Thanks for taking the initiative on this one, but not sure bout the impl. I was expecting it to end up a lot simpler. Is there some corner case I didn't think of?

src/typings/Types.ts
```diff
@@ -186,7 +186,12 @@ export type Modifier =
 
 export type SelectionType =
   //   | "character" Not implemented
-  "token" | "line" | "notebookCell" | "paragraph" | "document";
+  | "token"
+  | "line"
+  | "notebookCell"
+  | "paragraph"
+  | "document"
+  | "nonWhitespaceSequence";
```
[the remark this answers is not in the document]
AndreasArvidsson: Exactly according to the implementation we have today selection types are more based around text syntax

src/processTargets/processSelectionType.ts
AndreasArvidsson: Because that one might be empty. This solves the corner case where one of the edges are in an empty part of the document

src/test/suite/fixtures/recorded/selectionTypes/clearPaint.yml
AndreasArvidsson: Why would it? Are referring to my issue where all plain text tests fail?

src/processTargets/processSelectionType.ts
AndreasArvidsson: I would say it follows the issue exactly. . The implementation needs to be this complex to get that functionality

src/processTargets/processSelectionType.ts
pokey: If it's empty, then the start is not part of a non-white space sequence so that should be an error

src/typings/Types.ts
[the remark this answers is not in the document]
pokey: Ok that's prob fine for now then. That distinction is gonna be removed soon anyways

src/test/suite/fixtures/recorded/selectionTypes/clearPaint.yml
pokey: See above; the end of selection is not contained in a non-white space sequence, so should be an error

src/processTargets/processSelectionType.ts
AndreasArvidsson: I've implemented this port
```
Move it forward until the end of the next non-whitespace block
Move it backward to the end of the preceding non-whitespace block
```
I think that makes total sense and the implementation as it is right now is really useful in my opinion

pokey: Yes but we discussed yesterday and decided to throw an error if an endpoint is not touching a non-whitespace sequence

pokey: Also I don't think "take paint file" will work today, will it? Can't specify two selection types

pokey: And "take paint file" violates the semantics of saying a scope type, which implies smallest containing instance of that scope type, not largest contained instance

AndreasArvidsson: That's misunderstanding I said that we should throw an error if it was in complete white space. that the ends are is fine I think.

It will work with new grammar. It's already implemented and is useful for the future.

AndreasArvidsson: We already have sub word modifier. I see no reason why modify can't both grow and shrink. 
A is now paint selects all content the current selection touches.

pokey: So would you argue that if your selection is larger than a function, it should shrink to the function? What happens if it is larger than one function but contained in another? Just seems a bit ambiguous to allow "funk" to both shrink and expand. To me the semantics of `containingScope` are that it expands

AndreasArvidsson: But I haven't implemented this as a containing scope. It is a selection type.

Containing scopes by their name is a bit different. But definitely to select all functions that your selection touches would probably be useful.

AndreasArvidsson: But I'm not married to the solution so if you hate it in can change you to something simpler

pokey: In the future there will be no distinction between containing scope and selection type so I'd just treat this as if it were selection type

I guess I'm not opposed to changing the semantics of containing scope to just be every instance of that scope that is touched by the input, just want to make sure we're consistent

I think that's actually pretty close to what we do with "token", right?

AndreasArvidsson: I know but in this future we don't have to stick to the name containing scope if we don't want to. 

Yes we token we are growing but we don't shrink. But the question is if we shouldn't treat all the scope modifiers like this instead? I think we have to have a proper discussion about it

AndreasArvidsson: To give some context the reason I think this fits particularly well with paint is that we already allow start and end to be in different sequences. Something we don't do containing scopes. If we already decided that this one is going to be treated differently I think allowing it to shrink is totally fine even if we don't implement it on the others scopes. I think we ide can have it as wildcard with its own rules or it should adhere to the same as containing scopes where both ends of the selection has to be inside and an error is thrown if it's not.

pokey: We could def discuss, but fwiw I'm happy to just agree to change the semantics of today's `containingScope` modifier to be "any instance of the scope type touching the input to the modifier", which would necessitate your impl here, and then we could just merge.  That way we don't need to block on having a discussion.  But if you wanna talk it through before we merge I'm of course happy to do that

The one catch is that we'll then want to change impl of "round", "funk", etc to be consistent with these semantics.  Probably not urgent because it's kind of a corner case but just a thought

We could just merge this one, then file an issue to make sure to discuss / think through the semantics of containing scope (or whatever we rename it to to reflect that it's really "touching scope")

AndreasArvidsson: I want to have a quick discussion about it before. I'm available before today's meeting if you are?

pokey: As discussed, let's expand start cursor to containing sequence and end cursor to containing sequence, but never shrink, just throwing an error of either end is not contained in a non-whitespace sequence

AndreasArvidsson: done

PR #486, opened 2022-01-16
Use generic notebook insert cell command instead of jupyter specific

AndreasArvidsson:


pokey:
How come you closed this one?  Because it breaks the text-based setup?

AndreasArvidsson:
I don't think so

pokey:
Can you not just do something like this and then check if it's null? https://github.com/cursorless-dev/cursorless-vscode/blob/04fcbfeaa362564d13f8bea5932675204055d76f/src/util/setSelectionsAndFocusEditor.ts#L54-L56

AndreasArvidsson:
Not a pretty solution but might work. Okay it's back on the list.

AndreasArvidsson:
@pokey Turned out quite nicely actually

pokey (commented):
Yes looks great!  Test please 😊

AndreasArvidsson:
How do you propose we make tests for a notebook?

pokey:
> How do you propose we make tests for a notebook?

like this https://github.com/cursorless-dev/cursorless-vscode/blob/main/src/test/suite/crossCellsSetSelection.test.ts

AndreasArvidsson:
Thanks test is now added.

pokey:
I made the checks a bit more thorough and did some quick refactoring while I was there https://github.com/cursorless-dev/cursorless-vscode/pull/486/files/a37d5474c5cfbd9360f0d077162390c1ca7e3e25..4b02b6039fe6d57ad8ed25eadf906f1e016a6e27

AndreasArvidsson:
Good changes

pokey (approved):

PR #490, opened 2022-01-17
Fixed bug where line breakpoint would use incorrectly character

AndreasArvidsson:
Line level breakpoints are always on character 0. This new implementation is consistent with the built in vscode breakpoint functions in regard to line and inline breakpoints.

### DIFF
--- a/src/test/suite/breakpoints.test.ts
+++ b/src/test/suite/breakpoints.test.ts
@@ -129,7 +129,7 @@ async function breakpointTokenHarpRemove() {
     ),
   ]);
 
-  assert.deepStrictEqual(vscode.debug.breakpoints.length, 1);
+  assert.deepStrictEqual(vscode.debug.breakpoints.length, 2);
 
   await vscode.commands.executeCommand(
     "cursorless.command",
--- a/src/test/suite/breakpoints.test.ts
+++ b/src/test/suite/breakpoints.test.ts
@@ -1,0 +1,88 @@
+import * as assert from "assert";
+import * as vscode from "vscode";
+import * as sinon from "sinon";
+import { getCursorlessApi } from "../../util/getExtensionApi";
+import { openNewEditor } from "../openNewEditor";
+
+suite("breakpoints", async function () {
+  this.timeout("100s");
+  this.retries(3);
+
+  teardown(() => {
+    sinon.restore();
+  });
+
+  test("breakpoint harp", breakpointHarp);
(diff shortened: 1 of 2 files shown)


pokey (commented):
This looks good but I am _definitely_ going to break this by accident when I implement object-oriented targets 😅.  It shouldn't be too bad to add a test to capture this fix, should it?

AndreasArvidsson:
Adding break points to the test is definitely something we want to do but I do think it's an pr in itself

pokey:
I don't think it should be too hard.  Just a manual test like the one you did for #382, no?  Ie run the add breakpoint command, then list editor breakpoints.  I'd think the actual content (ignoring test setup boilerplate) would be like 10–15 lines.  But maybe I'm missing something?

AndreasArvidsson:
Ok I was thinking line breaks for the test recorder. A one of can of course be made.

pokey:
> Ok I was thinking line breaks for the test recorder. A one of can of course be made.

No that seems like overkill. Just like one to test happy path and one to test the thing you just fixed is probably plenty. Can't imagine we'll be recording lots of breakpoint tests

AndreasArvidsson:
Tests added

src/test/suite/breakpoints.test.ts
```diff
@@ -0,0 +1,88 @@
...
+});
+
+async function breakpointHarp() {
+  const graph = (await getCursorlessApi()).graph!;
+
+  await openNewEditor("  hello");
+
+  await graph.hatTokenMap.addDecorations();
+
+  removeBreakpoints();
```
pokey, line 30: should we toss this line into `teardown` above?

pokey (commented):
Nice test!  I believe your code is also careful to make sure it doesn't move the end of the target to the start of the line when it is doing a removal, so might be good to add a quick test to ensure that using "break air" for removal works properly, lest someone (me probably 😅) moves that [if statement](https://github.com/pokey/cursorless-vscode/pull/490/files#diff-d889ddecfcfb955719c394b94aab31a71ba7506164747b7421d82282b261081eR60-R62) upwards in an effort to "simplify".  Make sense?

src/test/suite/breakpoints.test.ts
AndreasArvidsson: It has to be no breakpoints before the test it set upnot tear down in that case

AndreasArvidsson:
Removal tests added

Issue #17, opened 2021-06-23
Add continuous delivery


pokey:
See:
- https://code.visualstudio.com/api/working-with-extensions/continuous-integration
- https://github.com/microsoft/vscode-test/blob/main/sample/azure-pipelines.yml

- [ ] add black checker for python code
- [ ] add isort for python code
- [ ] use version bump script `bump major --tag --push` from [bump](https://www.npmjs.com/package/version-bump-prompt) as an action like in [semver-release-action](https://github.com/zwaldowski/semver-release-action#advanced)
   - [ ] including auto-update release notes with link to PR, date, etc.  (use plop.js?)
- [ ] Dismiss stale reviews
- [ ] Add code owners?

pokey:
https://github.com/github-changelog-generator/github-changelog-generator

PR #507, opened 2022-01-25
Publish extension on merge to main

pokey:
This is the first step in our moved to continuous delivery (#17). This PR adds a Github action which automatically releases the extension every time we merge to main. It uses a version number which is automatically generated from the commit number. 

We will move to automatically deploying the talon files once we moved to a mono repo. It is safe to just begin with the extension publication because we are careful about backwards compatibility in this direction

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -1,0 +1,31 @@
+name: Deploy
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    environment: production
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 16
+          cache: yarn
+      - run: yarn install --frozen-lockfile
+      - run: yarn run compile
+      - run: yarn run prepare-for-extension-publish
+      - name: Publish to Open VSX Registry
+        uses: HaaLeo/publish-vscode-extension@v1
+        id: publishToOpenVSX
+        with:
+          pat: ${{ secrets.OPEN_VSX_TOKEN }}
+      - name: Publish to Visual Studio Marketplace
+        uses: HaaLeo/publish-vscode-extension@v1
+        with:
+          pat: ${{ secrets.VS_MARKETPLACE_TOKEN }}
+          registryUrl: https://marketplace.visualstudio.com
+          extensionFile: ${{ steps.publishToOpenVSX.outputs.vsixPath }}
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -1,13 +1,15 @@
 name: Deploy
 
 on:
-  push:
-    branches:
-      - main
+  workflow_run:
+    workflows: [Run Tests]
+    types: [completed]
+    branches: [main]
 
 jobs:
   deploy:
     runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
     environment: production
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    environment: production
```
pokey, line 11: We're using [GitHub environments](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) here so that only runs on `main` have access to the secrets they'll need for extension publication

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 16
+          cache: yarn
+      - run: yarn install --frozen-lockfile
+      - run: yarn run compile
+      - run: yarn run prepare-for-extension-publish
```
pokey, line 20: Note this step, where we update the extension version just locally on CI machine before publication

src/scripts/prepareForExtensionPublish.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Prepares the directory for extension publication. Does the following:
+ *
+ * 1. Changes the package version so that the patch number is the number of
+ *    commits on the current branch
+ * 2. Writes a file called `build-info.json` for provenance
+ */
+async function main() {
+  const { major, minor } = semver.parse(process.env.npm_package_version)!;
+
+  const commitCount = await runCommand("git rev-list --count head");
```
auscompgeek, line 18: Should this be spelt `HEAD`?

pokey:
Maybe I should only run this workflow if tests pass?  Could do something like [this](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow).  Seems a bit like overkill because we already have branch protection, but probably good practice

src/scripts/prepareForExtensionPublish.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const newVersion = `${major}.${minor}.${commitCount}`;
+
+  await runCommand(`npm --no-git-tag-version version ${newVersion}`);
+
+  // These are automatically set for Github actions
+  // See https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
+  const repository = process.env["GITHUB_REPOSITORY"];
+  const runId = process.env["GITHUB_RUN_ID"];
```
auscompgeek, line 27: Would it be worth asserting these exist?

src/scripts/prepareForExtensionPublish.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Strips leading and trailing whitespace from string
+ *
+ * From https://stackoverflow.com/a/1418059
+ * @param str The string to process
+ * @returns The stripped string
+ */
+function strip(str: string): string {
+  return str.replace(/^\s+|\s+$/g, "");
```
auscompgeek, line 46: I think you can replace this with `String#trim`?

auscompgeek (commented):
On mobile right now. I can take another look once at my laptop but mostly LGTM.

src/scripts/prepareForExtensionPublish.ts
pokey: Wow i was really misled by stack overflow there 😅.  Thanks for the sanity check

src/scripts/prepareForExtensionPublish.ts
pokey: Works lowercase too, but agreed `HEAD` is more canonical

AndreasArvidsson (approved):

PR #509, opened 2022-01-26
Continuous delivery take II

pokey:
Attempt to fix issues with previous attempt at continuous delivery

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -13,6 +13,8 @@ jobs:
     environment: production
     steps:
       - uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yaml
```diff
@@ -13,6 +13,8 @@ jobs:
     environment: production
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
pokey, line 17: We now do a deep checkout so that we get the real number of commits in the branch. Will also need this for the subtree push in an upcoming PR

AndreasArvidsson (approved):

PR #511, opened 2022-01-27
Attempt to fix subrepo push

pokey:
Looks like we need to set a username and email for git before we can push

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -52,5 +52,7 @@ jobs:
           path: git-subrepo
       - run: |
           source ../git-subrepo/.rc
+          git config user.name github-actions
+          git config user.email github-actions@github.com
           git subrepo push cursorless-talon
         working-directory: main


AndreasArvidsson (approved):

.github/workflows/deploy.yaml
```diff
@@ -52,5 +52,7 @@ jobs:
           path: git-subrepo
       - run: |
           source ../git-subrepo/.rc
+          git config user.name github-actions
+          git config user.email github-actions@github.com
```
auscompgeek, line 56:
```suggestion
          git config user.email action@github.com
```
I believe is the email for @actions-user

pokey: I just was following https://github.com/actions/checkout#push-a-commit-using-the-built-in-token

pokey: maybe that's out of date?  I wouldn't be shocked 😅

auscompgeek: :shrug:

PR #513, opened 2022-01-27
Limit before/after to certain contexts

### LINKED ISSUE #464: Have syntactic sugar for "bring to before", "bring to after", etc
We'd like to be able to say "bring air before bat", "move air after bat", etc, rather than "bring air **_to_** before bat", "move air **_to_** after bat", etc.

I wonder if we also want to support "move after bat" as shorthand for "move this to after bat".  Note that this behaviour won't come for free with the above change, because saying "move after bat" is actually supported today, but it doesn't do what you might expect: it will take the empty range after the `"bat"` token and insert it at the current position.  So I believe it will be equivalent to "chuck this"!  Haven't test that hypothesis tho 😊

We might consider being stricter about where user can use positional modifiers like "after" and "before", because you can / will be able to recover their functionality more naturally in other ways, so maybe they should just be allowed in specific places.  For example
- "take air past before bat" => "take air till bat"
- "move air to before bat" => "move air before bat"
- "take before bat" => "pre bat"
- "take line before bat" will use the special construct `"<SCOPE_TYPE> before"`
- "move this before bat" => "move before bat"

Not sure bout the last paragraph; just a thought.  Could be useful to have fully compositional positional modifiers

pokey:
And support cleaner "move after" syntax
Fixes #464

And also fix "to after" / "to before" on "bring" / "move"

Note: currently based off `insert-snippet` because it changes the grammar to be consistent with removing "to" elsewhere

Closes https://github.com/pokey/cursorless-vscode/issues/464

- [ ] Requires https://github.com/pokey/cursorless-vscode/issues/309 because otherwise we lose that functionality, which today can be done using "chuck after" / "chuck before"
- [ ] Fix cheatsheet

### DIFF
--- a/cursorless-talon/src/actions/actions.py
+++ b/cursorless-talon/src/actions/actions.py
@@ -4,7 +4,11 @@ from ..csv_overrides import init_csv_and_watch_changes
 from .actions_callback import callback_action_defaults, callback_action_map
 from .actions_custom import custom_action_defaults
 from .actions_makeshift import makeshift_action_defaults, makeshift_action_map
-from .actions_simple import simple_action_defaults
+from .actions_simple import (
+    no_wait_actions,
+    positional_action_defaults,
+    simple_action_defaults,
+)
 
 mod = Module()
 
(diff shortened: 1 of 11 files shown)


pokey:
Incorporated into #672

PR #517, opened 2022-01-29
Added no unused rule to lint

AndreasArvidsson:
By default unused variables are off. This means that we don't get proper intellisense when developing and we run the risk of compiling with unused variables intact.

This is actually a fix(read workaround) for the issue of unused variables having lower opacity that affects our hats/decorations. Disabling the vscode setting for "unused" and relying on the language server/linter gives us the same information without affecting the hats.


pokey:
Mind adding an issue description? Not exactly sure what this PR is supposed to accomplish. Ie is the rule off by default? Error by default?

AndreasArvidsson:
sure

pokey (commented):
Ok I guess I'm ok with this one in principle, but it generates a ton of log spam; our CI now has 45 warnings.  I'd argue that we should fix the warnings if we're going to turn them on, otherwise we run the risk of having so many warnings that we just ignore them all together

In some cases, I'm not sure I agree with the warnings, especially in the case of function / constructor parameters, even when they're named `_`

AndreasArvidsson:
I agree with that we should clean up all warnings
`_` is just at python thing

brxck:
For what it's worth, this rule can be configured to ignore variables with the underscore prefix! [See here.](https://eslint.org/docs/rules/no-unused-vars#argsignorepattern)

(Hi! Just fired up talon for the first time in while and I'm lurking the repo...)

pokey:
> For what it's worth, this rule can be configured to ignore variables with the underscore prefix! [See here.](https://eslint.org/docs/rules/no-unused-vars#argsignorepattern)
> 
> (Hi! Just fired up talon for the first time in while and I'm lurking the repo...)

🙌 he's back!! 😊

Nice find.  @AndreasArvidsson wdyt?

AndreasArvidsson:
Sounds like a good solution

AndreasArvidsson:
@brxck Thanks for the heads up about the ignore pattern

@pokey No more warnings. Since I'm literally touching half of the files in the repository please merge this as quick as possible to avoid future merge conflicts :D

src/processTargets/modifiers/processModifier.ts
```diff
@@ -55,12 +54,8 @@ export default function (
       break;
 
     case "toRawSelection":
-      result = processRawSelectionModifier(context, selection, modifier);
+      result = processRawSelectionModifier(context, selection);
       break;
-
-    default:
-      // Make sure we haven't missed any cases
-      const neverCheck: never = modifier;
```
pokey, line 63: Why have you removed this?  It is enables the type checker to enforce that we haven't missed a case.  Maybe just rename it to `_neverCheck`?

src/processTargets/processMark.ts
```diff
@@ -134,10 +133,7 @@ function processDecoratedSymbol(
   context: ProcessedTargetsContext,
   mark: DecoratedSymbol
 ) {
-  const token = context.hatTokenMap.getToken(
-    mark.symbolColor,
-    mark.character
-  );
+  const token = context.hatTokenMap.getToken(mark.symbolColor, mark.character);
```
pokey, line 136: Btw is this stuff enforced as of this PR?  If not, we really have to get these linting checks into CI 😅

src/typings/generic.ts
```diff
@@ -7,6 +7,7 @@ type ObjectKeys<T> = T extends object
   ? string[]
   : never;
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
pokey, line 10: How come we need this here?

pokey (commented):
LGTM. For the most part I think this PR is nice cleanup.

Is it actually enforced in CI, though?

To be honest I'm not in love with having the rule on parameters, because they're often there in order to conform to a particular signature (eg accepting a `graph`), but I acknowledge that this rule did catch some genuine extra arguments, so maybe forcing the programmer to prepend with `_` to acknowledge that they're doing it on purpose is a decent compromise.

src/processTargets/modifiers/processModifier.ts
AndreasArvidsson: I don't understand this syntax at all. To my eyes you're just creating unused variable with no purpose. Could you point me to some documentation?

src/typings/generic.ts
AndreasArvidsson: Another piece of code I don't understand how it works. This file exports nothing but is still needed or we get compilation errors.

src/processTargets/modifiers/processModifier.ts
pokey: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#exhaustiveness-checking

src/processTargets/processMark.ts
AndreasArvidsson: I haven't touched anything relating to formatting

src/typings/generic.ts
pokey: See the link at the top of the file

src/processTargets/processMark.ts
pokey: Ok.  Yeah we gotta get these checks into CI so we don't keep having all these polluted diffs.  I'll see if we have an issue for it

AndreasArvidsson: Sounds great

src/typings/generic.ts
pokey: yeah i think this is fine

pokey:
I think we're good to go, except one last thing: is this check actually enforced in CI?  If not, then this effort is all a bit of a waste of time, I think

AndreasArvidsson:
I don't know the ci as you do. I think it should be enforced but if not it really helps my own work flow at least.

pokey:
Ok no worries I'll set that up in a separate PR

AndreasArvidsson:
Sounds good to me

pokey:
https://github.com/cursorless-dev/cursorless-vscode/issues/558

pokey (approved):

PR #520, opened 2022-01-29
Added open link action

### LINKED ISSUE #519: Add action for following links in vscode
Add a cursorless action for vscode's `editor.action.openLink` to follow links. Should this ship as a default action in cursorless?

AndreasArvidsson:
fixes #519


pokey (commented):
As discussed, the plan is to do the following:

- [ ] Deprecate "define" (tho not sure that's actually strictly necessary?)
- [ ] Add a new action "follow", which will use [`vscode.executeLinkProvider`](https://code.visualstudio.com/api/references/commands#:~:text=vscode.executeLinkProvider) to determine if there are any links under the selection.  If there is a link, and it is external, it will use [`openExternal`](https://code.visualstudio.com/api/references/vscode-api#env.openExternal) to open the URI.  Otherwise, eg if it starts with `file://`, it will open / reveal the given location in the editor.  If links at the location, I guess it will just try to show definition?

AndreasArvidsson:
@pokey The new follow action is now implemented. Should be leave define for now?

AndreasArvidsson:
Second thought. Should we re implement the new scope link/url using `vscode.executeLinkProvider` instead of a regex? The upside would be that cursorless an vscode has the same definition of what is a link and not.

pokey:
> Second thought. Should we re implement the new scope link/url using `vscode.executeLinkProvider` instead of a regex? The upside would be that cursorless an vscode has the same definition of what is a link and not. 
> 
> And if we think that vscode definition is lacking we could just contribute our own link provider. 
> 
> 

Hmm that's a cool idea. Do we get the source range from that api?

AndreasArvidsson:
> Hmm that's a cool idea. Do we get the source range from that api?

Yes

AndreasArvidsson:
I guess we have to determine two things
1. Should we deprecated define or not?
2. Should we update the link scope?

src/actions/Follow.ts
```diff
@@ -0,0 +1,70 @@
...
+    };
+  }
+
+  private async findLink(target: TypedSelection) {
+    const links = <DocumentLink[]>(
+      await commands.executeCommand(
+        "vscode.executeLinkProvider",
+        target.selection.editor.document.uri
+      )
+    );
```
pokey, line 51: wow super annoying we need to get links for the entire document rather than just at a location.  Seems unnecessarily expensive if you're in a big document 🤔

src/actions/Follow.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  private async openUri(uri: Uri) {
+    switch (uri.scheme) {
+      case "http":
+      case "https":
+        await env.openExternal(uri);
+        break;
+      case "file":
+        await window.showTextDocument(uri);
```
pokey, line 64: If we're not able to get a link to a specific line, then that's a problem.  Or do these URIs somehow link to a specific line? If not, maybe we should fall back to the makeshift `editor.action.revealDefinition` implementation if we detect a file URI?

pokey:
> I guess we have to determine two things
> 
> 1. Should we deprecated define or not?

I'm really on the fence on this one.  Got any leaning?

> 2. Should we update the link scope?

Tempted to 🤔.  I guess these document links are a bit more general then URI links, but I don't suppose there would a conflict arising from that fact

src/actions/Follow.ts
pokey: Oh yeah not saying it's your fault; just lamenting the api

src/actions/Follow.ts
AndreasArvidsson: A uri does not include a line number. But definitions are not returned as a line
`file://a.md` is the kind of things that will be returned as a file link

pokey: Ah ok so definitions don't get returned from `executeLinkProvider`?  If not, then I think this approach is reasonable

pokey:
I mean they're more general in the sense that any extension could register a link provider.  For example, do paths within markdown show up from the link provider?

AndreasArvidsson:
> I mean they're more general in the sense that any extension could register a link provider. For example, do paths within markdown show up from the link provider?

If they start with HTTPS http or file yes. But any provider can of course extend that.

src/actions/Follow.ts
AndreasArvidsson: Correct

src/actions/Follow.ts
AndreasArvidsson: I didn't take it that way either. Just a bad api
I haven't actually tried to give it a second argument. Who knows it might actually work even though it's not in the documentation?

pokey: Unfortunately, [no](https://github.com/microsoft/vscode/blob/54dec24a4d1e67dc6cb7fae0e6119dc84e0fab92/src/vs/editor/contrib/links/browser/getLinks.ts#L165) 😕

AndreasArvidsson: Doesn't really surprise me

cursorless-talon/src/actions/actions_simple.py
```diff
@@ -20,6 +20,7 @@
     "extract": "extractVariable",
     "float": "insertEmptyLineAfter",
     "fold": "foldRegion",
+    "follow": "follow",
```
pokey, line 23: Should we rename the action to `followLink`?  Might be a bit clearer in the cheatsheet, without making it so specific that we shoot ourselves in the foot down the road

pokey (commented):
Ok after discussion in comments, I think this approach seems fine, and we can leave "define" for now.  It would be good to add a test of the following, though:

- [ ] Calling "follow" when cursor is on a definition.  Could try something like the following:

    ```typescript
    const foo = "hello";
    const bar = foo;
    ```

    and then scroll down to hide the first line, and call "follow" on the visible `foo` and use your previous scroll testing technique to ensure it revealed the first line
- [ ] Test a `file://` URI to see that it opens a document

I don't think we want to try any external URIs 😄

cursorless-talon/src/actions/actions_simple.py
pokey: Kind of?  It is following a link to somewhere else.  I would argue that our usage of "link" here is short for "hyperlink", which is defined as

> a link from a hypertext document to another location, activated by clicking on a highlighted word or image

Given that we're just mimicking the default VSCode ctrl-click behaviour, I think this definition applies?  Idk, I also think if someone just sees "follow" they'll be confused

AndreasArvidsson: You and I know that but I would be surprised if people would consider that a link in general
Do you want follow link I have no objections

pokey: I'd lean towards switching to `followLink` if you don't have a strong preference, just because if I saw "follow", I'm not sure I'd know what it was supposed to do.  I certainly wouldn't be likely to assume that it would reveal definition, which is the reason that we would use it instead of `followLink`, right?

AndreasArvidsson: followLink it is

AndreasArvidsson:
And we have tests!

src/test/suite/followLink.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        type: "primitive",
+        mark: {
+          type: "cursor",
+        },
+      },
+    ]
+  );
+
+  const editor = vscode.window.activeTextEditor;
+  assert.equal(editor?.document?.uri?.scheme, "file");
```
pokey, line 71: why not check that the uri is equal to `link`?  I mean prob not necessary given you're checking file contents, but I don't see why not to check the full uri

src/test/suite/followLink.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  );
+
+  assert.equal(editor.visibleRanges[0].start.line, 0);
+}
+
+async function followLink() {
+  const filename = getFixturePath("helloWorld.txt");
+  const link =
+    os.platform() === "win32" ? `file:///${filename}` : `file://${filename}`;
+  await openNewEditor(link);
```
pokey, line 54: might be worth a quick comment here and/or renaming the link variable, otherwise it looks a like you're opening the link, rather than creating a file whose text is the link

src/test/suite/getFixturePaths.ts
```diff
@@ -0,0 +1,18 @@
+import * as path from "path";
```
pokey, line 1: This looks fine, but you'll want to move it into the new `test/util` directory and you'll need to resolve the merge conflict because, as mentioned yesterday, I also touched the fixture enumeration code in your other PR

pokey (commented):
Basically good to go imo, but see a couple minor comments below

src/test/suite/followLink.test.ts
AndreasArvidsson: The reason for this is that we need to handle path separators for both windows and linux. Doable of course but I think this solution is good enough.

pokey: Got it. Yeah probably fine given you're checking file content

pokey (approved):

PR #529, opened 2022-01-31
Add url scope

alexhwoods:
Related to issue https://github.com/pokey/cursorless-vscode/issues/426

- [x] Yes, I have added tests

I wouldn't be surprised if you guys want me to take another approach and not touch the non-whitespace scope. I would be perfectly fine with that. I just wanted you guys to see the idea so we can have the debate before we decide.

I recommend you go through the commits one by one; that will tell the most coherent story.

### DIFF
--- a/cursorless-talon/src/modifiers/containing_scope.py
+++ b/cursorless-talon/src/modifiers/containing_scope.py
@@ -58,6 +58,7 @@ selection_types = {
     "file": "document",
     "line": "line",
     "paint": "nonWhitespaceSequence",
+    "link": "url",
     "token": "token",
 }
 
--- a/src/processTargets/processSelectionType.ts
+++ b/src/processTargets/processSelectionType.ts
@@ -40,6 +40,13 @@ export default function (
         selection,
         selectionContext
       );
+    case "url":
+      return processRegexDefinedScope(
+        /(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g, // taken from https://regexr.com/3e6m0
+        target,
+        selection,
+        selectionContext
+      );
   }
 }
 
--- a/src/processTargets/processSelectionType.ts
+++ b/src/processTargets/processSelectionType.ts
@@ -16,6 +16,10 @@ import {
 } from "../typings/Types";
 import { getDocumentRange } from "../util/range";
 
+// taken from https://regexr.com/3e6m0
+const URL_REGEX =
+  /(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g;
+
 export default function (
   context: ProcessedTargetsContext,
   target: PrimitiveTarget,
@@ -42,7 +46,7 @@ export default function (
       );
     case "url":
       return processRegexDefinedScope(
-        /(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g, // taken from https://regexr.com/3e6m0
+        URL_REGEX,
         target,
         selection,
         selectionContext
--- a/src/processTargets/processSelectionType.ts
+++ b/src/processTargets/processSelectionType.ts
@@ -34,7 +34,12 @@ export default function (
     case "paragraph":
       return processParagraph(target, selection, selectionContext);
     case "nonWhitespaceSequence":
-      return processNonWhitespaceSequence(target, selection, selectionContext);
+      return processRegexDefinedScope(
+        /\S+/g,
+        target,
+        selection,
+        selectionContext
+      );
   }
 }
 
@@ -175,7 +180,8 @@ function processParagraph(
   };
 }
 
-function processNonWhitespaceSequence(
+function processRegexDefinedScope(
+  regex: RegExp,
   target: PrimitiveTarget,
   selection: SelectionWithEditor,
   selectionContext: SelectionContext
@@ -184,7 +190,7 @@ function processNonWhitespaceSequence(
 
   const getMatch = (position: Position) => {
     const line = selection.editor.document.lineAt(position);
-    const result = [...line.text.matchAll(/\S+/g)]
+    const result = [...line.text.matchAll(regex)]
       .map(
         (match) =>
           new Range(
@@ -196,7 +202,7 @@ function processNonWhitespaceSequence(
       )
       .find((range) => range.contains(position));
     if (result == null) {
-      throw new Error("Couldn't find containing non whitespace sequence");
+      throw new Error(`Cannot find sequence defined by regex: ${regex}`);
     }
     return result;
   };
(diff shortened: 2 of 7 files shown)


src/processTargets/processSelectionType.ts
```diff
@@ -196,7 +202,7 @@ function processNonWhitespaceSequence(
       )
       .find((range) => range.contains(position));
     if (result == null) {
-      throw new Error("Couldn't find containing non whitespace sequence");
+      throw new Error(`Cannot find sequence defined by regex: ${regex}`);
     }
     return result;
   };
```
alexhwoods, line 215: I'm not totally sure if there's functionality specific to the non-whitespace scope that I'm ignoring when making this generic. Definitely consider that in the review

AndreasArvidsson:
I'm definitely in favor of this solution. This enables us to add user defined scopes in the future as they can just contribute the regex.

src/processTargets/processSelectionType.ts
```diff
@@ -40,6 +40,13 @@ export default function (
         selection,
         selectionContext
       );
+    case "url":
+      return processRegexDefinedScope(
+        /(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g, // taken from https://regexr.com/3e6m0
```
alexhwoods, line 45: Note that the specific regex I chose requires a top-level domain, so urls like `http://localhost:3000` would not be found.

We could definitely refactor that, either now or later. Tbh it would just take me a long time to write that regex 😅

alexhwoods:
Also looks like I broke the black linting somehow, not sure how I managed to do that
```
Run psf/black@stable
Run # Exists since using github.action_path + path to main script doesn't work because bash
--- cursorless-talon/src/vendor/jstyleson.py	2022-01-31 13:20:01.512373 +0000
+++ cursorless-talon/src/vendor/jstyleson.py	2022-01-31 13:20:07.380250 +0000
@@ -103,11 +103,11 @@
         normal = True
         for i in range(former_index, len(json_str)):
would reformat cursorless-talon/src/vendor/jstyleson.py
             result_str[i] = ""
 
     # Show respect to original input if we are in python2
-    return ("" if isinstance(json_str, str) else u"").join(result_str)
+    return ("" if isinstance(json_str, str) else "").join(result_str)
 
 
 # There may be performance suffer backtracking the last comma
 def _remove_last_comma(str_list, before_index):
     i = before_index - 1

Oh no! 💥 💔 💥
1 file would be reformatted, 42 files would be left unchanged.
```

Could that be one of the fixtures?

cursorless-talon/src/modifiers/containing_scope.py
```diff
@@ -58,6 +58,7 @@ def cursorless_containing_scope(m) -> str:
     "file": "document",
     "line": "line",
     "paint": "nonWhitespaceSequence",
+    "link": "url",
```
alexhwoods, line 61: This is just the first thing that came to mind. The name for this scope is totally up for debate

AndreasArvidsson: Link is what comes to mind for me as well.

src/processTargets/processSelectionType.ts
AndreasArvidsson: Preferably those should work as well but it's not a deal breaker for first version.

src/processTargets/processSelectionType.ts
AndreasArvidsson: Not really. I think this should work fine.

alexhwoods:
One other thing I wanted to mention - I think it would be a good idea to write something about how to get set up to contribute, especially now we have a monorepo.

Here's the workflow I did:
1. Clone the repo
2. Delete the `cursorless-talon` directory I had in my talon user directory, and make a symlink to the one in the newly cloned directory.
```
ln -s .talon/user/cursorless-talon ~/Desktop/cursorless-vscode/cursorless-talon
```
3. Open `cursorless-vscode` with vscode, and use `Run Extension` and `Extension Tests`.


Also, if we could find a way to limit the number of tests that run at a given time that would be really helpful. As a hacky solution, I deleted most of the tests while I was working, other than the ones relevant to stuff I was touching.

cursorless-talon/src/modifiers/containing_scope.py
alexhwoods: Cool. To state the obvious `url` is a voice coding disaster haha

pokey:
Looks good to me! For future, we try to keep the test docs minimal, and as @AndreasArvidsson we try to use "clear" instead of take. Not a showstopper tho

And yeah I do like magic regexes hoisted to the top 😊. Not too bad in this case tho because the context makes it super clear, so I wouldn't block the PR on those grounds if you like this flavour better here

- [ ] Would be good to add a test where start of selection is in one regex and end is in another

Great stuff tho!

pokey:
> Also looks like I broke the black linting somehow, not sure how I managed to do that
> 
> ```
> 
> Run psf/black@stable
> 
> Run # Exists since using github.action_path + path to main script doesn't work because bash
> 
> --- cursorless-talon/src/vendor/jstyleson.py	2022-01-31 13:20:01.512373 +0000
> 
> +++ cursorless-talon/src/vendor/jstyleson.py	2022-01-31 13:20:07.380250 +0000
> 
> @@ -103,11 +103,11 @@
> 
>          normal = True
> 
>          for i in range(former_index, len(json_str)):
> 
> would reformat cursorless-talon/src/vendor/jstyleson.py
> 
>              result_str[i] = ""
> 
>  
> 
>      # Show respect to original input if we are in python2
> 
> -    return ("" if isinstance(json_str, str) else u"").join(result_str)
> 
> +    return ("" if isinstance(json_str, str) else "").join(result_str)
> 
>  
> 
>  
> 
>  # There may be performance suffer backtracking the last comma
> 
>  def _remove_last_comma(str_list, before_index):
> 
>      i = before_index - 1
> 
> 
> 
> Oh no! 💥 💔 💥
> 
> 1 file would be reformatted, 42 files would be left unchanged.
> 
> ```
> 
> 
> 
> Could that be one of the fixtures?

Maybe black changed their rules around empty Unicode strings? I'd just run black locally and commit 🤷

pokey:
> One other thing I wanted to mention - I think it would be a good idea to write something about how to get set up to contribute, especially now we have a monorepo.
> 
> 
> 
> Here's the workflow I did:
> 
> 1. Clone the repo
> 
> 2. Delete the `cursorless-talon` directory I had in my talon user directory, and make a symlink to the one in the newly cloned directory.
> 
> ```
> 
> ln -s .talon/user/cursorless-talon ~/Desktop/cursorless-vscode/cursorless-talon
> 
> ```
> 
> 3. Open `cursorless-vscode` with vscode, and use `Run Extension` and `Extension Tests`.
> 

Yeah that's my workflow too. Feel free to paste that somewhere 👍

> 
> 
> 
> 
> Also, if we could find a way to limit the number of tests that run at a given time that would be really helpful. As a hacky solution, I deleted most of the tests while I was working, other than the ones relevant to stuff I was touching.

https://github.com/pokey/cursorless-vscode/issues/293

pokey (approved):
Looks great!  Not worth changing in this PR, but usually we only use "chuck" when we are trying to test removal directly; otherwise we use "clear". We usually have "clear" tests for everything, and "chuck" tests when removal is interesting, eg removing commas.  But just fyi for future tests; this PR looks great to me

PR #530, opened 2022-01-31
Added support to split on dots in cheat sheet

### LINKED ISSUE #469: Fix regex in cheatsheet to handle dots
This is particularly important when the user has a custom command because the vscode commands use dots to separate

AndreasArvidsson:
fixes #469

### DIFF
--- a/cursorless-talon/src/cheatsheet/cheat_sheet.py
+++ b/cursorless-talon/src/cheatsheet/cheat_sheet.py
@@ -161,7 +161,7 @@ class CheatSheet:
         self.draw_items(canvas, items)
 
     def draw_multicolumn_section(
-        self, canvas, items, column_names: str, scopes_limit=24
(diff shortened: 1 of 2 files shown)


AndreasArvidsson:
@pokey Shouldn't the vendor files be excluded from the linting test?

pokey:
We could, but given it's such a trivial diff I'd be tempted to just run black on it for now 🤷

pokey (approved):

AndreasArvidsson:
I already have. Running black on the file locally gives zero changes.

pokey:
maybe need to update your black?  my hypothesis is that it's due to an update to black

pokey:
looks like the black thing is fixed on #529 so maybe let's get that one in and then you can just rebase onto it

AndreasArvidsson:
> looks like the black thing is fixed on #529 so maybe let's get that one in and then you can just rebase onto it

Sounds good

Issue #210, opened 2021-08-10
Switch targets to object-oriented


pokey:
Target is an object with the following methods
- `getContentSelection` (used for selection, get text, highlight for reference)
- `getRemovalDecorationRange`
- `constructChangeEdit` (takes positional modifier and text)
- `constructInsertionEdit`
- `constructRemovalEdit`

The target will also have the following, but somehow hide it before action sees it because it shouldn't be used there. It's mostly for downstream pipeline stages, eg "after":

- Trailing delimiter range
- Leading delimiter range

Positional modifier is just a pipeline stage

AndreasArvidsson:
* Remove inside/outside as proper modifiers
* Add start of/end of as proper modifiers
* Create a draft target type used in the pipeline. Gets converted into a function only finalized target at the end.

pokey:
I think we prob potentially want edit constructors instead of ranges, eg for camel-case, no?

pokey:
I think we prob potentially want edit constructors instead of ranges, eg for camel-case, no?

pokey:
I think we prob want `constructInsertionEdit` to take a list of texts, so that we can handle #184

AndreasArvidsson:
We can do that

PR #531, opened 2022-01-31
Updated delete action with support for overlapping ranges

### LINKED ISSUE #523: `chuck every arg` doesn't work
https://user-images.githubusercontent.com/7907782/151707097-9e8c34d3-6fab-49d2-95f0-2c97cad64891.mov

AndreasArvidsson:
fixes #523 

And yes it needs to be this complicated. For example if we have three arguments `a, b, c` and you try to remove `b` and `c` they will both select the same comma. Hence the error about overlapping ranges. But we can't just merge both of these ranges because we also want to remove the first comma. The algorithm right now is that we are running a loop that merges overlapping ranges and then recalculates outside selection with leading and trailing delimiters. We repeat this until we have no more overlaps.

## Checklist

- [x] I have added [tests](https://github.com/pokey/cursorless-vscode/blob/main/docs/contributing/test-case-recorder.md)


pokey (commented):
Hmm.  This PR adds a lot of complexity, and doesn't fully solve the problem.  For example, it doesn't fix "move", or "carve".  I think this solution tries to fix the problem too late in the pipeline.  It needs to happen during the inside-outside adjustment, by looking at the whole set of targets at that point.  By the time we've added the delimiters twice, any subsequent code is just trying to clean up a mess of our own making.

Unfortunately, I don't think we should try solve this problem within the current inside-outside framework.  Doing so will just make things harder for us as we move to a proper object-oriented / `removalRange`-based solution.  

All that being said, I do like the fact that this PR enables us to add tests for this case, and the complex logic is at least fairly well contained.  I'd be tempted to do one of the following:

### Option 1

- [ ] fold all of this code into a single file in order to further isolate it, and add a comment at the top saying that we intend to clean things up when we implement #210 
- [ ] call this function from "move" and "carve"

### Option 2

Close this PR for now, and revisit once we have proper object-oriented targets

AndreasArvidsson:
It does solve carve/cut but not move. 
Option one is no big problem and I see no reason why we should wait.

AndreasArvidsson:
Then its not really a mock but absolutely we could actually update the clipboard before each test.

pokey:
I don't think we need to do it before each test; just the tests where the `.yml` has a clipboard field.  The test case recorder only uses that field when recording actions that use the clipboard

Would be good to leave a comment there that we should switch back to mock once we have https://github.com/cursorless-dev/cursorless-vscode/issues/128#issuecomment-1024389906

AndreasArvidsson:
Exactly. If we don't have a clipboard content we of course don't update it.

On the other hand the tests are working now and if we are planning to switch back soon any way we could just leave them as they are?

pokey:
But they're working because you changed the order in which they run?  That feels very brittle

AndreasArvidsson:
They are now prefixed with `clipboard1` and `clipboard2`. Shouldn't be a problem for now and once we have the new implementation we can just remove those prefixes.

pokey:
This just feels too brittle to me.  Tests should be independent.  It's a trivial change, right?  I added a checkbox to #559 to make sure we don't forget

src/actions/BringMoveSwap.ts
```diff
@@ -21,6 +21,7 @@ import {
   getSelectionInfo,
```
pokey, line 21: See above unused variable warnings

pokey (commented):
Ok looks good other than fixing brittle tests and a minor comment about removing warnings

AndreasArvidsson:
Done and done

PR #532, opened 2022-01-31
Added proper support for += operator

### LINKED ISSUE #521: Python value scope type doesn't work with an increment operator
```python
content += raw[current_index:match.start()]
```

Saying `"take value whale"` doesn't work

AndreasArvidsson:
fixes #521

### DIFF
--- a/src/languages/java.ts
+++ b/src/languages/java.ts
@@ -51,7 +51,12 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
   list: "array_initializer",
   functionCall: "method_invocation",
   map: "block",
-  name: ["*[declarator][name]", "*[name]", "formal_parameter.identifier!"],
+  name: [
+    "*[declarator][name]",
+    "assignment_expression[left]",
+    "*[name]",
+    "formal_parameter.identifier!",
+  ],
   namedFunction: ["method_declaration", "constructor_declaration"],
   type: trailingMatcher([
     "generic_type.type_arguments.type_identifier",
@@ -66,7 +71,10 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
     "method_declaration.identifier!",
     "constructor_declaration.identifier!",
   ],
-  value: leadingMatcher(["*[declarator][value]", "*[value]"], ["="]),
+  value: leadingMatcher(
+    ["*[declarator][value]", "assignment_expression[right]", "*[value]"],
+    ["="]
+  ),
   condition: conditionMatcher("*[condition]"),
   collectionItem: argumentMatcher("array_initializer"),
   argumentOrParameter: argumentMatcher("formal_parameters", "argument_list"),
--- a/src/languages/python.ts
+++ b/src/languages/python.ts
@@ -60,16 +60,23 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
   namedFunction: "decorated_definition?.function_definition",
   functionName: "function_definition[name]",
   condition: conditionMatcher("*[condition]"),
-  type: leadingMatcher(["function_definition[return_type]", "*[type]"], [":", "->"]),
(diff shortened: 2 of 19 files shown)


src/languages/java.ts
```diff
@@ -66,7 +71,10 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
     "method_declaration.identifier!",
     "constructor_declaration.identifier!",
   ],
-  value: leadingMatcher(["*[declarator][value]", "*[value]"], ["="]),
+  value: leadingMatcher(
+    ["*[declarator][value]", "assignment_expression[right]", "*[value]"],
+    ["="]
```
pokey, line 76: Do we want to add `+=` etc here?

src/languages/typescript.ts
```diff
@@ -131,6 +131,7 @@ function typeMatcher(): NodeMatcher {
 function valueMatcher() {
   const pFinder = patternFinder(
     "assignment_expression[right]",
+    "augmented_assignment_expression[right]",
```
pokey, line 134: As above, should we add `+=` to the leading delimiters?

src/languages/python.ts
```diff
@@ -60,16 +60,23 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
...
     "typed_parameter.identifier!",
     "parameters.identifier!",
     "*[name]",
   ],
   collectionItem: argumentMatcher(...dictionaryTypes, ...listTypes),
   value: cascadingMatcher(
-    leadingMatcher(["assignment[right]", "~subscript[value]"], ["=", ":"]),
+    leadingMatcher(
+      ["assignment[right]", "augmented_assignment[right]", "~subscript[value]"],
+      ["=", ":"]
```
pokey, line 78: As above re `+=` etc

src/test/suite/fixtures/recorded/languages/cpp/clearName.yml
```diff
@@ -0,0 +1,23 @@
...
+    - type: primitive
+      modifier: {type: containingScope, scopeType: name, includeSiblings: false}
+initialState:
+  documentContents: a += "world"
+  selections:
+    - anchor: {line: 0, character: 6}
+      active: {line: 0, character: 6}
+  marks: {}
+finalState:
+  documentContents: " += \"world\""
```
pokey, line 16: Side note: man yaml is so weird. Imagine how painful it would have been to figure out that we need quotes here but not on the `documentContents` above. Cc/ @somanythings

src/test/suite/fixtures/recorded/languages/csharp/clearName.yml
```diff
@@ -0,0 +1,23 @@
+languageId: csharp
```
pokey, line 1: C sharp tests 🙌

pokey (approved):
Oh yeah awesome thanks for taking this one down 👍👍. Left a couple minor nits.

src/languages/java.ts
pokey: On second thought, did you leave it out because it doesn't really make sense? I could see that. Tho I guess you might use it if the increment is a sub expression? That's legal in some languages I think. Eg 

```
if (a += 2) {...}
```

Then you might say "chuck two" and want just the `a`

AndreasArvidsson: I'm unsure about this. I think that we probably should support identifiers in more types of expression if that's the case

src/test/suite/fixtures/recorded/languages/cpp/clearName.yml
AndreasArvidsson: Yeah yikes

src/languages/java.ts
AndreasArvidsson: I create an issue for it. Will have a proper look when I have time
https://github.com/pokey/cursorless-vscode/issues/533https://github.com/pokey/cursorless-vscode/issues/533

PR #536, opened 2022-02-01
Add link wrap gif

pokey:


AndreasArvidsson:
Should we really add a gif to a feature that is not in vscode? I do believe this is one of your custom snippets?

pokey:
Yeah I had that thought too, but people keep asking whether Cursorless is useful for things other than code so I thought this one was a nice illustration of Cursorless for markdown. Plus I think it's kinda cool

Shall I just add the snippet to Cursorless core?

AndreasArvidsson:
I definitely think that some work down examples are in order. But examples that the user country produce are probably not good ux. 

Maybe. We should probably have a proper discussion about snippets. I know you have a lot of thoughts and ideas about them but we haven't really had a proper discussion about it.

pokey:
So I think the best argument against including this GIF in the README is that wrapper snippets are still marked as experimental.  Forgot about that detail 😅

Given that they are experimental, I don't think there's much harm in adding "link wrap" snippet to cursorless, though.

What if I removed the GIF from the README, but added it to the experimental docs for wrapper snippets and added `"link wrap"` snippet to core?  I think it's a nice experimental snippet and nicely illustrates the power of Cursorless for editing non-code text

AndreasArvidsson:
I think that sounds fine

pokey:
Done

AndreasArvidsson (approved):

PR #539, opened 2022-02-02
Added tests for visible ranges(scroll, fold)

### LINKED ISSUE #160: Test a command's effect on visible ranges

AndreasArvidsson:
fixes #160
## Checklist

- [x] I have added [tests](https://github.com/pokey/cursorless-vscode/blob/main/docs/contributing/test-case-recorder.md)

### DIFF
--- a/src/test/suite/fold.test.ts
+++ b/src/test/suite/fold.test.ts
@@ -19,7 +19,6 @@ suite("fold", async function () {
 async function foldMade() {
   const graph = (await getCursorlessApi()).graph!;
   const editor = await openNewEditor("function myFunk() {\n\n}", "typescript");
-  await graph.hatTokenMap.addDecorations();
 
   await vscode.commands.executeCommand(
     "cursorless.command",
@@ -29,9 +28,7 @@ async function foldMade() {
       {
         type: "primitive",
         mark: {
-          type: "decoratedSymbol",
-          symbolColor: "default",
-          character: "m",
+          type: "cursor",
         },
       },
     ]
@@ -50,7 +47,6 @@ async function unfoldMade() {
   await vscode.commands.executeCommand("editor.fold", {
     selectionLines: [0],
   });
-  await graph.hatTokenMap.addDecorations();
 
   assert.equal(editor.visibleRanges.length, 2);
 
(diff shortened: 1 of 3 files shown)


src/test/suite/fold.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    "fold made",
+    "fold",
+    [
+      {
+        type: "primitive",
+        mark: {
+          type: "decoratedSymbol",
+          symbolColor: "default",
+          character: "m",
+        },
```
pokey, line 35: The problem with using a mark here is that this test will break if we change our hat allocation algorithm, whereas for the recorded tests we capture the required allocations.  Maybe the easiest thing would be to just change this test to use the current selection, like we do for some of our [other manual tests](https://github.com/cursorless-dev/cursorless-vscode/blob/30bc0f2ad9ed0d070a256965913afb1e1a675557/src/test/suite/backwardCompatibility.test.ts#L23)?

src/test/suite/scroll.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    "crown whale",
+    "scrollToTop",
+    [
+      {
+        type: "primitive",
+        mark: {
+          type: "decoratedSymbol",
+          symbolColor: "default",
+          character: "w",
+        },
```
pokey, line 35: as above, maybe we should just use cursor mark to avoid hat allocation fragility?

pokey (dismissed):
This looks really nice; thanks for following up on this stuff!  I left one minor comment.  Not a showstopper, as we have the same problem in a couple other manual tests, but seems easy to fix and might save us some headaches down the road 🤷‍♂️

But yeah re your comment on the issue, I think it's fine to handle these as one-offs rather than adding support to test case recorder

src/test/suite/fold.test.ts
AndreasArvidsson: done

pokey (approved):
Nice 😊

PR #573, opened 2022-02-21
LaTeX support

du-da:
## What

Adds support for the `LaTeX` programming language

## Checklist


src/languages/latex.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  collectionItem: "enum_item",
+
+  comment: ["block_comment", "line_comment"],
+
+  // TODO: use other scope(s)? adding individual scopes is probably too much? (part, chapter, section, subsection, subsubsection, paragraph, subparagraph)
+  xmlStartTag: SECTIONING,
+
+  // TODO: eventually add scope instead of reusing xmlElement
+  xmlElement: ENVIRONMENTS,
```
pokey, line 168: I am happy for you to be liberal in adding scopes.  This language doesn't really fit into the paradigm of the existing implemented languages.  We'll soon be moving to a more flexible scope model.  So for now just add as many as you want.  If it gets out of hand we can trim it down

src/languages/latex.ts
```diff
@@ -0,0 +1,171 @@
...
+  "listing_environment",
+  "minted_environment",
+  "pycode_environment",
+];
+
+const sectioningText = SECTIONING.map((s) => `${s}[text]`);
+const sectioningCommand = SECTIONING.map((s) => `${s}[command]`);
+const commandXGroup = GROUPS.map((group) =>
+  COMMANDS.map((cmd) => `${cmd}.${group}!`)
+).flat();
```
pokey, line 92: Hmm.  I guess this one is fine?  Feels a bit strange to cross-product these patterns.  I believe you could do this one using [`findPossiblyWrappedNode`](https://github.com/cursorless-dev/cursorless-vscode/blob/a85c1e1625a87d1eea6ea99586ec9d78a09c643e/src/util/nodeFinders.ts#L100), passing in the parent finder and the child finder.  I think that's better?  cc/ @AndreasArvidsson

src/languages/latex.ts
```diff
@@ -0,0 +1,171 @@
...
+
+function unwrapParenExtractor(
```
pokey, line 94: @AndreasArvidsson do we not already have something like this?

src/languages/latex.ts
```diff
@@ -0,0 +1,171 @@
...
+  return {
+    selection: new Selection(
+      editor.document.positionAt(startIndex),
+      editor.document.positionAt(endIndex)
+    ),
+    context: {},
+  };
+}
+
+function extendToNamedSiblingIfExists(
```
pokey, line 121: Could you not use [`trailingMatcher`](https://github.com/cursorless-dev/cursorless-vscode/blob/main/src/util/nodeMatchers.ts#L112) here?

pokey (commented):
Super excited to see this one underway!  Great start.  Left a few comments

Some tests will really help me get a flavour for how these scope types look in practice.  The [test case recorder](https://github.com/cursorless-dev/cursorless-vscode/blob/main/docs/contributing/test-case-recorder.md) makes this process really quite painless; I promise 😊

src/languages/latex.ts
AndreasArvidsson: I don't think we do anything like this since we always check type and seldom text in our finders

src/languages/latex.ts
pokey: Thought about it a bit more, and I may throw together a quick finder for this purpose.  Stay tuned

pokey: Ok here it is https://github.com/cursorless-dev/cursorless-vscode/pull/581.  Here's an example usage:

https://github.com/Will-Sommers/cursorless-vscode/pull/1/files#diff-92bb04e3bc2a809b7086ef3b90b2ace5ef95e73095ff438fc316a40d8e2a496aR109-R116

src/languages/latex.ts
du-da: From my understanding, trailingMatcher would not be of use here. I want the scope to refer to the respective node-pair as target for any action, not just for deletion.

The Treesitter grammar does [not expose](https://github.com/latex-lsp/tree-sitter-latex/blob/1666e9780de42a31b1376d32fbe8332e8dd5850f/grammar.js#L177) a common parent node for (only) this pair. Instead, the parent node contains the overall section (including its content, subsections, ..).

src/test/suite/fixtures/recorded/languages/latex/changeName.yml
```diff
@@ -0,0 +1,25 @@
+languageId: latex
```
pokey, line 1: Nice.  As mentioned above, would be cool if this one worked from within the section itself as well

pokey, line 1: Nice.  As mentioned above, would be cool if this one worked from within the section itself as well

cursorless-talon/src/modifiers/containing_scope.py
```diff
@@ -43,6 +43,15 @@
...
     "end tag": "xmlEndTag",
+    # LaTeX
+    "command": "latexCommand",
+    "environment": "latexEnvironment",
+    "part": "latexPart",
+    "chapter": "latexChapter",
+    "subsection": "latexSubsection",
+    "subsubsection": "latexSubSubSection",
+    "paragraph": "latexParagraph",
+    "subparagraph": "latexSubParagraph",
```
pokey, line 54: I would be inclined to remove the `latex` prefix from all of these.  We have found that it can be helpful to be a bit less verbose for these scope types so that they can be used elsewhere

cursorless-talon/src/modifiers/containing_scope.py
```diff
@@ -43,6 +43,15 @@
     "tags": "xmlBothTags",
     "start tag": "xmlStartTag",
     "end tag": "xmlEndTag",
+    # LaTeX
+    "command": "latexCommand",
```
pokey, line 47: As mentioned elsewhere, I think I'd be inclined to remove this one and just use `functionCall`

src/test/suite/fixtures/recorded/languages/latex/changeType.yml
```diff
@@ -0,0 +1,25 @@
+languageId: latex
```
pokey, line 1: Huh.  Not obvious to me that this one would be "type".  I think in the future we'll support something like "callee", ie the function that is being called in a function call expression.  Could introduce something like that now and use it here?

src/test/suite/fixtures/recorded/languages/latex/changeArg.yml
```diff
@@ -0,0 +1,27 @@
+languageId: latex
```
pokey, line 1: Would be good to support / test "chuck arg" as well, which should remove the curly braces

pokey (commented):
Looking really good!  I left a few more comments.  Also, one more thing:

- [ ] Isn't there a version of arguments that uses `[]` instead of `{}`?  I believe it's for optional arguments?  If so, it's prob worth supporting / testing that, including a "change" and "chuck" version, the latter which would remove the `[]` as well

du-da:
@pokey Thank you for your feedback. Your suggestions should now be reflected in the changes. (I hope I haven't missed anything.)

Regarding the brackets following a command (now: call):
This is due to the generated parse tree (see related [issue](https://github.com/latex-lsp/tree-sitter-latex/issues/42)).
For a lot of the commands the generic command rule (see: [vsc-parse-tree dependency](https://github.com/latex-lsp/tree-sitter-latex/blob/1666e9780de42a31b1376d32fbe8332e8dd5850f/grammar.js#L560) / [most recent](https://github.com/latex-lsp/tree-sitter-latex/blob/e550c23486b166cf93d7f063c96306b568737985/grammar.js#L579) version) used in the tree-sitter grammar is sufficient for parsing. Others, however, require individually defined rules, as the generic one would not be able to parse them or they have some "special purpose", eg providing structure ([section](https://github.com/latex-lsp/tree-sitter-latex/blob/e550c23486b166cf93d7f063c96306b568737985/grammar.js#L199), subsection,...).
The problem with the generic rule is that it has no fixed argument count/order and "absorbs" subsequent pairs into the command node.

cursorless-talon/src/modifiers/containing_scope.py
```diff
@@ -13,6 +13,7 @@
     "arg": "argumentOrParameter",
     "attribute": "attribute",
     "call": "functionCall",
+    "callee": "functionCallee",
```
pokey, line 16: This is the first step of #645.  What do we want to use for the identifier and default spoken form?  I'd lean towards

```
    "callee": "functionCallFunction",
```

@AndrewDant feel free to implement this scope type for other languages as part of closing #645.  You can either wait for this PR to merge or just implement the couple relevant lines; the actual definition is fairly simple

cc/ @AndreasArvidsson @wenkokke @AndrewDant

src/languages/latex.ts
```diff
@@ -0,0 +1,210 @@
...
+      unwrapGroupParens
+    ),
+    matcher(
+      patternFinder(...["begin[name]", "end[name]"].concat(sectioningText)),
+      unwrapGroupParens
+    )
+  ),
+
+  functionCall: cascadingMatcher(
+    matcher(patternFinder(...COMMANDS.concat(["begin", "end"]))),
```
pokey, line 188: Can you not just do
```suggestion
    matcher(patternFinder(...COMMANDS, "begin", "end")),
```

src/test/suite/fixtures/recorded/languages/latex/changeCallee.yml
```diff
@@ -0,0 +1,23 @@
...
+    - type: primitive
+      modifier: {type: containingScope, scopeType: functionCallee, includeSiblings: false}
+initialState:
+  documentContents: \href{https://some.url}{some text} \LaTeX
+  selections:
+    - anchor: {line: 0, character: 3}
+      active: {line: 0, character: 3}
+  marks: {}
+finalState:
+  documentContents: "{https://some.url}{some text} \\LaTeX"
```
pokey, line 16: Genuine question here: do we want to keep the initial `\`?  Ie would there ever be a situation where we want to change the command and it wouldn't end up starting with a `\`?  Another way to look at it is: is the `\` considered part of the command name, or is it part of the overall call structure

pokey (commented):
This is so cool! 😍

Out of curiosity, have you been using this one much locally?  How does it feel?

cursorless-talon/src/modifiers/containing_scope.py
AndrewDant: I'm not a huge fan of functionCallFunction. Not intuitively distinct from "call" and "functionCall", it could cause some confusion for someone trying to learn the code or debug some json.

AndrewDant: I've come around on callee though.

pokey: Maybe `functionCallCallee`?

pokey:
@du-da I just wanted to follow up on this one.  I think it's getting really close!

src/typings/targetDescriptor.types.ts
```diff
@@ -106,6 +106,14 @@ export type SimpleScopeTypeType =
   | "xmlElement"
   | "xmlEndTag"
   | "xmlStartTag"
+  // Latex scope types
+  | "part"
+  | "chapter"
+  | "subsection"
+  | "subSubSection"
+  | "paragraph"
```
pokey, line 114: Was hoping to just take this PR to the finish line myself, but got stuck on this one.  The Cursorless scope type "block" (sequence of non-empty lines) internally uses the identifier `paragraph` as well, so this causes a conflict.  I think it will be surprising if the behaviour of “block” changes in LaTeX, especially since empty lines are actually a valid way to create paragraphs.

I decided to call this one `namedParagraph`

src/test/suite/fixtures/recorded/languages/latex/changeCallee.yml
pokey: Let's leave this one for now

pokey (approved):
Looks good!

src/languages/latex.ts
```diff
@@ -0,0 +1,210 @@
...
+    context: {
+      leadingDelimiterRange: new Range(
+        editor.document.positionAt(node.startIndex),
+        editor.document.positionAt(contentStartIndex - 1)
+      ),
+    },
+  };
+}
+
+function sectionNameFinder(node: SyntaxNode) {
```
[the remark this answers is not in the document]
pokey: Tried it; seemed to work 🤷

pokey:
Ok I think this one is good to go!  @du-da @AndreasArvidsson I'll give you 24 hours to object then I'll merge it in

src/languages/latex.ts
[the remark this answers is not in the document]
du-da: Yeah, you are right. 
I can't remember why I introduced this finder. I must have used it for debugging, but after I fixed the actual fault (ordering of the `SECTIONING` array), I didn't bother to rethink/clean up what I did.

PR #578, opened 2022-02-23
Add support for Ruby

Will-Sommers:
## What

Adds support for the `ruby` programming language

## Checklist

- [x]  Added a test for `"change round"` inside a string, eg `"hello (there)"`
- [ ] Supported` "type"` both for type annotations (eg `foo: string`) and declarations (eg `interface Foo {}`) (and added tests for this behaviour 😊) 
- [x]  "change" test for "arg"
- [x] "change" test for "item"
- [x] "change every item"
- [x] "change" test for "name"
- [x] "chuck value" test for a value in a map, making sure it cleans up the :
- [x] "arg" tests that target function-call arguments, rather than just method parameters
- [x] Tests for "state"
- [x] Tests for "lambda"
- [x] A couple more "name" tests (I think you might need to fix your matcher for these; see comment below)
  - [x] referring to function name (using "take name", not "take funk name")
  - [x] referring to class name (using "take name", not "take class name")


Will-Sommers:
👋 Heyo

Here's a few notes from the PR. I need to clean up the `TODO` in the `ruby.ts` file. 

- Look into pairs within strings, eg. `take pair` "(hello)"
- "chuck (every|) arg" doesn't work for function calls
- Block comments are edited in an odd way, if there is an =end within the string e.g.
```
=begin
words comment =end
=end 
```
The second `=end` should be highlighted as well since the first `=end` is not at column position 0, which is the only valid ending position for a block comment.

src/test/suite/fixtures/recorded/languages/ruby/changeKey2.yml
```diff
@@ -0,0 +1,23 @@
+languageId: ruby
+command:
+  version: 1
+  spokenForm: change key
+  action: clearAndSetSelection
+  targets:
+    - type: primitive
+      modifier: {type: containingScope, scopeType: collectionKey, includeSiblings: false}
+initialState:
+  documentContents: "h = {\"1\": \"one\", \"2\":   \"two\"}"
```
Will-Sommers, line 10: I'm not sure why this was recorded with a wrapped `"` and the one above it wasn't.

src/test/suite/fixtures/recorded/languages/ruby/takeAfterKey.yml
```diff
@@ -0,0 +1,24 @@
+languageId: ruby
+command:
```
Will-Sommers, line 2: Going to change all of these `take` to `chuck`

src/test/suite/fixtures/recorded/languages/ruby/changeKey2.yml
pokey: No clue. The vagaries of yaml syntax are not for us mere mortals to understand 😅. Cc/ @somanythings

src/test/suite/fixtures/recorded/languages/ruby/takeAfterKey.yml
pokey: Assuming you mean replace "take after key" with "chuck key"? I wouldn't advise using "take after" / "chuck after". It's somewhat of an accident that those work imo

For other tests that use "take" (not "take after"), I'd replace them with "change" rather than "chuck", but it's not critical; fine to leave them if already recorded

src/test/suite/fixtures/recorded/languages/ruby/changeKey2.yml
Will-Sommers: Got it, I might just re-record and see if I can reproduce it.

src/test/suite/fixtures/recorded/languages/ruby/takeAfterKey.yml
Will-Sommers: Sounds good.

src/languages/ruby.ts
```diff
@@ -0,0 +1,74 @@
...
+    ),
+    name: ["assignment[left]", "operator_assignment[left]"],
+    value: trailingMatcher(
+    [
+      "pair[value]",
+      "assignment[right]",
+      "operator_assignment[right]"
+    ],
+    [":"]
+    ),
```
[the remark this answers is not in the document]
AndreasArvidsson: Not at a moment if you don't want to make your own finder. I want to update the matchers to use operator instead of specific text but that is still in the backlog.

src/languages/ruby.ts
```diff
@@ -0,0 +1,74 @@
...
+} from "../util/nodeMatchers";
+import { NodeMatcherAlternative, ScopeType, SelectionWithEditor } from "../typings/Types";
+import { SyntaxNode } from "web-tree-sitter";
+import { getNodeRange } from "../util/nodeSelectors";
+
+// Generated by the following command:
+// > curl https://raw.githubusercontent.com/tree-sitter/tree-sitter-ruby/master/src/node-types.json \  
+// | jq '[.[] | select(.type == "_statement" or .type == "_simple_statement") | .subtypes[].type]'
+const STATEMENT_TYPES = [
+  "_expression",
```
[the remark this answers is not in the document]
pokey: Ok see https://github.com/Will-Sommers/cursorless-vscode/pull/1

pokey:
> Heyo @pokey, working on this now.
> 
> Regarding the following test case request:
> 
> > "chuck value" test for a value in a map, making sure it cleans up the ":".
> 
> I think that this is likely incorrect behavior. There are two types of Ruby hash syntax:
> 
> ```
> { a: 10 }
> { "a" => 10 }
> ```
> 
> In the first case, the `":"` refers more to the key than the value, making it a symbol. In the second case, I agree that `chuck` should remove the `"=>"`. I can think of cases where we would like to disregard this nuance, for example, changing a hash to a list would be made easier, although we would still have to move the color from the rear to the front of the identifier. Happy to go either way on it.

I'll defer to your knowledge of Ruby here 👍

src/languages/ruby.ts
```diff
@@ -0,0 +1,156 @@
...
+];
+
+// TODO: Add more allowable parent types
+const EXPRESSION_STATEMENT_PARENT_TYPES = ["program", "method"];
+
+const mapTypes = ["hash"];
+const listTypes = ["array", "string_array", "symbol_array"];
+
+const assignmentOperators = ["=", '+=', '-=', '*=', '**=', '/=', '||=', '|=', '&&=', '&=', '%=', '>>=', '<<=', '^='];
+const mapKeyValueSeparators = [":", "=>"];
```
Will-Sommers, line 105: @pokey — I opted to go with terminating the leading `:` as well since Ruby is odd in this case and it would be better to be inline with how other languages behave.

src/languages/ruby.ts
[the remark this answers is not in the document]
Will-Sommers: Heyo @pokey — I've got to confess that I feel like I'm in the deep end on this one, in terms of adding extra `EXPRESSION_STATEMENT_PARENT_TYPES`. 

ex.
```
if true
    a = 1
end
```

output:
```
[program](https://tree-sitter.github.io/tree-sitter/playground#) [0, 0] - [3, 0]
  [if](https://tree-sitter.github.io/tree-sitter/playground#) [0, 0] - [2, 3]
    condition: [true](https://tree-sitter.github.io/tree-sitter/playground#) [0, 3] - [0, 7]
    consequence: [then](https://tree-sitter.github.io/tree-sitter/playground#) [0, 7] - [1, 9]
      [assignment](https://tree-sitter.github.io/tree-sitter/playground#) [1, 4] - [1, 9]
        left: [identifier](https://tree-sitter.github.io/tree-sitter/playground#) [1, 4] - [1, 5]
        right: [integer](https://tree-sitter.github.io/tree-sitter/playground#) [1, 8] - [1, 9]
  ```

In this case, if I was to say `"take state air"`, we would expect that only that statement is highlighted, rather than the broader, if statement?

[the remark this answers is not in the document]
Comment (pokey) [reply src/languages/ruby.ts:24]: Yeah that's right. If I'm reading this right, you need to add `then` as an allowable parent type to make this one work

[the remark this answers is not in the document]
Comment (Will-Sommers) [reply src/languages/ruby.ts]: Cools, added almost all of the common cases here. Reading through the corpus on the ruby-tree-sitter project was a trip/learning experience...

**Comment by Will-Sommers** [comment]
I think this is getting ready to go, happy to take any comments/direction on more tests.

Still not covered/supported
- [ ] Heredocs
- [ ] Accessing value on early returns for Ruby, last line returns have an implicit return
- [ ] Handling `array_pattern` and `hash_pattern` in pattern matching

The 1st and 2nd cases are probably most common. Do you think they are blocking a merge? 

heredoc
```
a = <<~TEXT
content
TEXT

[program] [0, 0] - [4, 0]
  [assignment] [0, 0] - [0, 11]
    left: [identifier] [0, 0] - [0, 1]
    right: [heredoc_beginning] [0, 4] - [0, 11]
  [heredoc_body] [0, 11] - [2, 4]
    [heredoc_content] [0, 11] - [2, 0]
    [heredoc_end] [2, 0] - [2, 4]
```

early return
```
 return foo if true 
[program] [0, 0] - [1, 0]
  [if_modifier] [0, 0] - [0, 18]
    body: [return] [0, 0] - [0, 10]
      [argument_list] [0, 7] - [0, 10]
        [identifier] [0, 7] - [0, 10]
    condition: [true] [0, 14] - [0, 18]
```

array_pattern
```
case expr 
  in 1, *x, 3 
end

[program] [0, 0] - [4, 0]
  [case_match] [0, 0] - [3, 3]
    value: [identifier] [0, 5] - [0, 9]
    clauses: [in_clause] [1, 2] - [1, 14]
      pattern: [array_pattern] [1, 5] - [1, 13]
        [integer] [1, 5] - [1, 6]
        [splat_parameter] [1, 8] - [1, 10]
          name: [identifier] [1, 9] - [1, 10]
        [integer] [1, 12] - [1, 13]
```

File: src/languages/ruby.ts [hunk 6]
```diff
@@ -0,0 +1,169 @@
...
+  ),
+  collectionItem: argumentMatcher(...mapTypes, ...listTypes),
+};
+export const patternMatchers = createPatternMatchers(nodeMatchers);
+
+export function stringTextFragmentExtractor(
+  node: SyntaxNode,
+  _selection: SelectionWithEditor
+) {
+  if (node.type === "string_content") {
```
pokey, line 164: I believe you should add `heredoc_content` to the list of acceptable node types here, and then add a test that you can say "clear round" within a heredoc body

pokey:
> heredoc
> 
> ```
> a = <<~TEXT
> content
> TEXT
> 
> [program] [0, 0] - [4, 0]
>   [assignment] [0, 0] - [0, 11]
>     left: [identifier] [0, 0] - [0, 1]
>     right: [heredoc_beginning] [0, 4] - [0, 11]
>   [heredoc_body] [0, 11] - [2, 4]
>     [heredoc_content] [0, 11] - [2, 0]
>     [heredoc_end] [2, 0] - [2, 4]
> ```

Wow this seems like a really bad parse 😅.  I'd merge as-is but file a follow-up issue here, and probably worth looking into filing an issue against tree-sitter-ruby, or seeing if one is already filed.  Why would `heredoc_beginning` not share a parent with `heredoc_body`?

> 
> early return
> 
> ```
>  return foo if true 
> [program] [0, 0] - [1, 0]
>   [if_modifier] [0, 0] - [0, 18]
>     body: [return] [0, 0] - [0, 10]
>       [argument_list] [0, 7] - [0, 10]
>         [identifier] [0, 7] - [0, 10]
>     condition: [true] [0, 14] - [0, 18]
> ```

Can't you just add a pattern `return.argument_list` to the `value` matcher?

> array_pattern
> 
> ```
> case expr 
>   in 1, *x, 3 
> end
> 
> [program] [0, 0] - [4, 0]
>   [case_match] [0, 0] - [3, 3]
>     value: [identifier] [0, 5] - [0, 9]
>     clauses: [in_clause] [1, 2] - [1, 14]
>       pattern: [array_pattern] [1, 5] - [1, 13]
>         [integer] [1, 5] - [1, 6]
>         [splat_parameter] [1, 8] - [1, 10]
>           name: [identifier] [1, 9] - [1, 10]
>         [integer] [1, 12] - [1, 13]
> ```

Are you using a custom vsode-parse-tree extension build?  I get that same parse on tree-sitter-ruby's master branch, but for some reason in the editor I get an error in the parse

What is it that you're not supporting here btw?  "take list"? "take item"?  The former could just be done by adding `array_pattern` to matcher, no?  For the latter, I think you might be able to use `argumentMatcher("array_pattern")`

src/languages/ruby.ts
```diff
@@ -0,0 +1,169 @@
...
+  "symbol_array",
+  "true",
+  "unary",
+  "unless",
+  "until",
+  "while",
+  "yield",
+];
+
+// TODO: Add more allowable parent types
```
pokey, line 97: Can probably remove this TODO now?

src/test/suite/fixtures/recorded/languages/ruby/changeLambda.yml
```diff
@@ -0,0 +1,25 @@
...
+initialState:
+  documentContents: |
+    hi = -> { puts "Hi!" }
+  selections:
+    - anchor: {line: 0, character: 19}
+      active: {line: 0, character: 19}
+  marks: {}
+finalState:
+  documentContents: |
+    hi = ->
```
pokey, line 18: should the `->` not be included here?

src/test/suite/fixtures/recorded/languages/ruby/changeLambda4.yml
```diff
@@ -0,0 +1,23 @@
...
+    - type: primitive
+      modifier: {type: containingScope, scopeType: anonymousFunction, includeSiblings: false}
+initialState:
+  documentContents: lambda { puts "Hi!" }
+  selections:
+    - anchor: {line: 0, character: 15}
+      active: {line: 0, character: 15}
+  marks: {}
+finalState:
+  documentContents: "lambda "
```
pokey, line 16: I think the `lambda` should be included here

pokey (commented):
Very close!  See comments

src/test/suite/fixtures/recorded/languages/ruby/changeLambda4.yml
Will-Sommers: Hey @pokey, good catch. What is happening here is that both `block - `.each { |x| ... ` and `do_block -`.each do |x| ...` are also defined as lambdas so there is some ambiguity here. I guess ideally, we'd pull the `->` and the `lambda` out in this case. It does seem a bit odd that the behavior is different between a `.each { |x|` where only the block is taken versus the lambda function where everything including the block is taken. I do think the behavior you're specifying though is what we want. 

`lambda` under `anonymousFunction: ["lambda", "block", "do_block"],` isn't doing anything right now, I guess I need to look at how to do a precedence match and if a block is preceded by a lambda, match both rather than just the block.

I'm using the [tree-sitter playground](https://tree-sitter.github.io/tree-sitter/playground) to generate these but here's what a lambda function looks like versus a regular block:

```
hi = -> { puts "Hi!" }
[program] [0, 0] - [1, 0]
  [assignment] [0, 0] - [0, 22]
    left: [identifier] [0, 0] - [0, 2]
    right: [lambda] [0, 5] - [0, 22]
      body: [block] [0, 8] - [0, 22]
        [call] [0, 10] - [0, 20]
          method: [identifier] [0, 10] - [0, 14]
          arguments: [argument_list] [0, 15] - [0, 20]
            [string] [0, 15] - [0, 20]
              [string_content] [0, 16] - [0, 19]
```

vs 

```
h.each do |x|
	puts x
end

[program] [0, 0] - [3, 0]
  [call] [0, 0] - [2, 3]
    receiver: [identifier] [0, 0] - [0, 1]
    method: [identifier] [0, 2] - [0, 6]
    block: [do_block] [0, 7] - [2, 3]
      parameters: [block_parameters] [0, 10] - [0, 13]
        [identifier] [0, 11] - [0, 12]
      [call] [1, 1] - [1, 7]
        method: [identifier] [1, 1] - [1, 5]
        arguments: [argument_list] [1, 6] - [1, 7]
          [identifier] [1, 6] - [1, 7]
```

Will-Sommers:
Heyo @pokey re: `array_pattern`, I am getting an error as well. I'll table it I think until its fixed.

src/test/suite/fixtures/recorded/languages/ruby/changeLambda4.yml
Will-Sommers: Actually got it working for one of them using `lambda?.do_block`

pokey:
> Heyo @pokey re: `array_pattern`, I am getting an error as well. I'll table it I think until its fixed.

Ok just published a new version of vscode-tree-sitter which should fix that error; try upgrading and see if it now works

src/test/suite/fixtures/recorded/languages/ruby/changeLambda4.yml
pokey: Just pushed a fix.  Lmk if that makes sense https://github.com/cursorless-dev/cursorless-vscode/pull/578/commits/3e83c2d34e394373e18afb4c15e8b8304cfd85e2

Will-Sommers: Yep, I think it makes sense. I need to dig into the why we don't need to specify `block` by itself in `cascadingMatcher` but I tested the cases and it looks like it is doing the right thing

pokey: Block only ever appears as a child of a `lambda` or `block` node.  The former is handled by `"lambda"`, the latter by the new `blockMatcher`.  Make sense?

Will-Sommers: Yep yep

pokey: The reason I had to remove it is that otherwise it would have stopped the traversal code before it had a chance to see the parent `call` node

Will-Sommers:
Great — cleaning it up and will get something through this weekend. Thanks for the help.

Will-Sommers:
Heyo @pokey — I'm going to create 2 issues for proper support of `here_doc` as well as the `array_pattern` and `hash_pattern`. I'm not super familiar with the new pattern matching syntax and need to do some more research. There's a bunch of stuff in the ruby-tree-sitter corpus that I need to read more about. Just adding `array_pattern` and `hash_pattern` to the collection types resulted in some weird behavior.

pokey:
> Heyo @pokey — I'm going to create 2 issues for proper support of `here_doc` as well as the `array_pattern` and `hash_pattern`. I'm not super familiar with the new pattern matching syntax and need to do some more research. There's a bunch of stuff in the ruby-tree-sitter corpus that I need to read more about. Just adding `array_pattern` and `hash_pattern` to the collection types resulted in some weird behavior.

Sounds good.  Will look out for those issues and then I think this thing is ready to go—great work!

pokey (approved):

PR #580, opened 2022-02-25
Add rewrap command to docs

Will-Sommers:
This adds the rewrap command to the user docs. As an aside, I had a hard time getting it recognized over wrap and had to switch the command to "switch".

### DIFF
--- a/docs/user/README.md
+++ b/docs/user/README.md
@@ -430,8 +430,8 @@ eg:
 `box wrap blue air`
 Wraps the token containing letter 'a' with a blue hat in square brackets.
 
-The rewrap command can be used to swap a given pair of symbols with another
-- `"curly rewrap <TARGET>"`: wraps the target with curly brackets, replacing the previous paired delimiter
+The rewrap command, mapped to `"repack"` by default, can be used to swap a given pair of symbols with another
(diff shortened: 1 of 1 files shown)


docs/user/README.md
```diff
@@ -430,6 +430,9 @@ eg:
 `box wrap blue air`
 Wraps the token containing letter 'a' with a blue hat in square brackets.
 
+The rewrap command can be used to swap a given pair of symbols with another
+- `"curly rewrap <TARGET>"`: wraps the target with curly brackets, replacing the previous paired delimiter
```
pokey, line 434: I believe the [default](https://github.com/cursorless-dev/cursorless-vscode/blob/b9355460a8e602bf47ec0b5ff1e609ad2f9017b3/cursorless-talon/src/actions/actions.py#L89) is

```suggestion
- `"curly repack <TARGET>"`: wraps the target with curly brackets, replacing the previous paired delimiter
```

pokey: I don't feel strongly about the heading. The internal identifier is rewrap so maybe let's keep that for the heading? Cc/ @AndreasArvidsson

AndreasArvidsson: Agreed, the  identifiers is what I prefer to use since everybody has different spoken forms.

pokey (approved):
Looks good!

PR #581, opened 2022-02-27
Add ancestor chain node finder

pokey:
I couldn't find a good place to test this one, but I applied it in https://github.com/Will-Sommers/cursorless-vscode/pull/1 so it will be tested once that merges


pokey:
Included in #578

PR #594, opened 2022-03-08
Add PHP support

nathanheffley:
## What

Adds support for the `php` programming language

## Checklist


nathanheffley:
Very early progress, mostly just got it to the point where I can record tests. I did have some trouble recording a test for "take if" but realized that that command doesn't work for me even when I'm just using cursorless regularly in languages that should support it so I guess it's an issue on my and I'll have to resolve.

pokey:
> Very early progress, mostly just got it to the point where I can record tests. I did have some trouble recording a test for "take if" but realized that that command doesn't work for me even when I'm just using cursorless regularly in languages that should support it so I guess it's an issue on my and I'll have to resolve.

Should be "take if state"

src/languages/php.ts
```diff
@@ -0,0 +1,94 @@
...
+      "return_statement[0]",
+    ],
+    ["=>"]
+  ),
+
+  collectionKey: trailingMatcher(["array_element_initializer[0]"], ["=>"]),
+  collectionItem: argumentMatcher("array_creation_expression"),
+
+  // "argument" should also be in this list so that you can select args
+  // in a function call, but it doesn't help find the containing argumentOrParameter
```
pokey, line 80: Not sure I follow this comment. Can you elaborate?

src/languages/php.ts
```diff
@@ -0,0 +1,94 @@
...
+  "trait_declaration",
+  "try_statement",
+  "unset_statement",
+  "while_statement"
+];
+
+export const getTypeNode = (node: SyntaxNode) =>
+  node.children.find((child) => child.type === "type") ?? null;
+
+const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
```
pokey, line 52: Don't forget `name`, which should match class name, function name, name in variable declaration, etc

src/languages/php.ts
```diff
@@ -0,0 +1,94 @@
...
+  collectionKey: trailingMatcher(["array_element_initializer[0]"], ["=>"]),
+  collectionItem: argumentMatcher("array_creation_expression"),
+
+  // "argument" should also be in this list so that you can select args
+  // in a function call, but it doesn't help find the containing argumentOrParameter
+  argumentOrParameter: ["simple_parameter", "variadic_parameter"],
+};
+export default createPatternMatchers(nodeMatchers);
+
+export function stringTextFragmentExtractor(
```
pokey, line 85: Did the default matcher we use for most other languages not work?

src/test/suite/fixtures/recorded/languages/php/changeKey.yml
```diff
@@ -0,0 +1,29 @@
+languageId: php
```
pokey, line 1: Worth also having a "key" test where cursor is in the key itself

src/test/suite/fixtures/recorded/languages/php/changeValue.yml
```diff
@@ -0,0 +1,29 @@
+languageId: php
```
pokey, line 1: Worth having a "value" test where cursor is in the key

src/test/suite/fixtures/recorded/languages/php/changeValue2.yml
```diff
@@ -0,0 +1,29 @@
+languageId: php
```
pokey, line 1: Worth having "value" test where cursor is in left-hand side of assignment

src/test/suite/fixtures/recorded/languages/php/chuckArg3.yml
```diff
@@ -0,0 +1,35 @@
...
+    }
+  selections:
+    - anchor: {line: 2, character: 30}
+      active: {line: 2, character: 30}
+  marks: {}
+finalState:
+  documentContents: |-
+    <?php
+
+    function myFunc($value, )
```
pokey, line 25: The comma should be removed here

src/test/suite/fixtures/recorded/languages/php/chuckArgNear2.yml
```diff
@@ -0,0 +1,39 @@
...
+      active: {line: 4, character: 0}
+  marks:
+    default.n:
+      start: {line: 2, character: 28}
+      end: {line: 2, character: 32}
+finalState:
+  documentContents: |-
+    <?php
+
+    function myFunc($value, )
```
pokey, line 29: And here

src/test/suite/fixtures/recorded/languages/php/chuckEveryArg.yml
```diff
@@ -0,0 +1,37 @@
...
+    }
+  selections:
+    - anchor: {line: 2, character: 20}
+      active: {line: 2, character: 20}
+  marks: {}
+finalState:
+  documentContents: |-
+    <?php
+
+    function myFunc(, )
```
pokey, line 25: And here

src/test/suite/fixtures/recorded/languages/php/chuckValue2.yml
```diff
@@ -0,0 +1,29 @@
...
+    $value = 'Hello world';
+  selections:
+    - anchor: {line: 2, character: 7}
+      active: {line: 2, character: 7}
+  marks: {}
+finalState:
+  documentContents: |-
+    <?php
+
+    $value =;
```
pokey, line 22: Should probably remove `=`? Idk if that's not syntactically valid then not sure there's much benefit

src/test/suite/fixtures/recorded/languages/php/chuckValue3.yml
```diff
@@ -0,0 +1,29 @@
...
+    $value += 42;
+  selections:
+    - anchor: {line: 2, character: 6}
+      active: {line: 2, character: 6}
+  marks: {}
+finalState:
+  documentContents: |-
+    <?php
+
+    $value +=;
```
pokey, line 22: Same here

pokey (commented):
Wow looks great! Left a bunch of minor comments

src/languages/php.ts
```diff
@@ -0,0 +1,94 @@
...
+  "return_statement",
+  "switch_statement",
+  "trait_declaration",
+  "try_statement",
+  "unset_statement",
+  "while_statement"
+];
+
+export const getTypeNode = (node: SyntaxNode) =>
+  node.children.find((child) => child.type === "type") ?? null;
```
pokey, line 50: What's this function for?

src/test/suite/fixtures/recorded/languages/php/chuckValue2.yml
nathanheffley: Yeah good call 👍

src/languages/php.ts
nathanheffley: Oops I based my file off of the python language just to get started quicker and this was in there, but I don't know what it's for and must have forgot to remove it.

src/languages/php.ts
nathanheffley: I was just doing this all wrong, another comment of yours helped you fix this, thanks!

src/languages/php.ts
nathanheffley: I almost missed that, thanks!

nathanheffley:
The only thing I'm not 100% satisfied with is that I could not get the type keyword working with typed property declarations on classes. The tree-sitter-php node types seem to suggest that the structure of a property declaration is the same as a simple parameter or variadic parameter (in terms of how to access the type) and cursorless has no problem getting the type for a parameter, but can't find the type when it is on a property declaration.

pokey:
> The only thing I'm not 100% satisfied with is that I could not get the type keyword working with typed property declarations on classes. The [tree-sitter-php files](https://github.com/tree-sitter/tree-sitter-php/blob/master/test/corpus/declarations.txt) seem to suggest that the structure of a property declaration is the same as a simple parameter or variadic parameter (in terms of how to access the type) and cursorless has no problem getting the type for a parameter, but can't find the type when it is on a property declaration.

Presumably you tried `property_declaration[type]`? Could also try just `*[type]`

src/test/suite/fixtures/recorded/languages/php/changeFunk2.yml
```diff
@@ -0,0 +1,31 @@
...
+    };
+  selections:
+    - anchor: {line: 3, character: 0}
+      active: {line: 3, character: 0}
+  marks: {}
+finalState:
+  documentContents: |-
+    <?php
+
+    $myFunc = ;
```
pokey, line 24: In typescript we would include the name as well. This feels like just one way to declare a function

Fwiw this is prob the behaviour I'd expect from "clear lambda"

src/test/suite/fixtures/recorded/languages/php/chuckValue7.yml
```diff
@@ -0,0 +1,29 @@
+languageId: php
```
pokey, line 1: How is this different from chuckValue4?

pokey (commented):
Very close; few more minor ones and I think we can figure out the property type thing too

src/test/suite/fixtures/recorded/languages/php/changeFunk2.yml
nathanheffley: So, moving this to be "clear lambda" and removing the entire declaration would make this more consistent with other languages? I definitely would like to be consistent as much as possible.

pokey: "Clear lambda" should do what this test case does exactly. "Clear funk" should remove the whole declaration. Make sense?

We sort of view this kind of declaration as just another way to create a named function, with the name on the LHS being its name

nathanheffley:
> Presumably you tried `property_declaration[type]`? Could also try just `*[type]`

Yes, I tried the property declaration but it was like I didn't do anything. The wild card works pretty well and I added some tests for two cases I missed, but it still doesn't work for property declarations 😕

src/test/suite/fixtures/recorded/languages/php/chuckType4.yml
```diff
@@ -0,0 +1,29 @@
...
+    (string) $str;
+  selections:
+    - anchor: {line: 2, character: 10}
+      active: {line: 2, character: 10}
+  marks: {}
+finalState:
+  documentContents: |-
+    <?php
+
+    () $str;
```
nathanheffley, line 22: These parens should be checked as well

src/test/suite/fixtures/recorded/languages/php/chuckValue7.yml
pokey: Ah right cool yeah it's good to have both. I'd leave it.

pokey:
> > Presumably you tried `property_declaration[type]`? Could also try just `*[type]`
> 
> 
> 
> Yes, I tried `property_declaration[type]` but it didn't do anything. The wild card works pretty well, and I added some tests for two cases I missed, but it still doesn't work for property declarations 😕 

What does it look like on the [playground](https://tree-sitter.github.io/tree-sitter/playground)?

nathanheffley:
> What does it look like on the [playground](https://tree-sitter.github.io/tree-sitter/playground)?

Oh wow this would have made everything a lot easier 😅 

This is what it looks like in the debugger when I try to record a test case for it. There is an error where I would expect it to say `type`

image

On the playground it looks like what I would expect

image

pokey:
Oh right looks like the version of tree-sitter on master has fixed the bug but it hasn't been released.  Super common.  I just released vscode-tree-sitter 0.16.2, which now uses master branch of tree-sitter-php instead of published version.  Lmk if that fixes it

src/languages/php.ts
pokey: Got it.  It's worth adding a "clear pair" and a "clear inside" test where your cursor is just sitting inside a regular string.  Eg:

```php
<?php

"hello |world";
```

And say "clear inside" and "clear pair"

src/test/suite/fixtures/recorded/languages/php/chuckType4.yml
pokey: Yeah that makes sense to me.  I think you'll need a bit of custom code.  First, you'll want to back away from `*[type]` for this case, so you could change it to `~cast_expression[type]`.  Then, you'll create a separate matcher for `cast_expression`, using a custom extractor, similar to the one we use for `name` in Markdown https://github.com/cursorless-dev/cursorless-vscode/blob/cfbfcb116828cd98b9cb0723a82595f5a212e0ec/src/languages/markdown.ts#L29-L45.  You'll return the range of the `node.childForFieldName("type")` for the content range, and the node itself for the `outerSelection`.  

Lmk if this doesn't make sense and we can do a quick pairing session

nathanheffley: Thank you for the direction, that was exactly what I needed 👍

src/languages/php.ts
nathanheffley: The clear keyword only works for clearing entire lines for me (like "clear ten" clears line ten). I did add tests for "chuck inside and "chuck pair" and made sure it works like when I use those in other languages.

nathanheffley:
> Oh right looks like the version of tree-sitter on master has fixed the bug but it hasn't been released. Super common. I just released vscode-tree-sitter 0.16.2, which now uses master branch of tree-sitter-php instead of published version. Lmk if that fixes it

Yes that worked beautifully, thanks!

src/languages/php.ts
pokey: Btw I can't seem to find those "chuck inside" and "chuck pair" tests.  Mind dropping a link so I can take a quick look?

src/test/suite/fixtures/recorded/languages/php/changeFunk2.yml
nathanheffley: Okay I moved the current functionality to the anonymous function option so this is now "change lambda"
I also added it so that "change funk" will remove the variable name as well.

pokey (approved):
Looks great!

Issue #524, opened 2022-01-31
Add "range" modifier


pokey:
- Creates a single selection that is the union of all input selections
- **Question**: how does this work in the jq model where every modifier only gets to see a single selection?
- **Question**: alternate name? “connected”, “fill”, “flood”, “flooded”?
- **Use case**: have multiple cursors and say “clone range” so that it won’t cause interleaved cloned lines. Could have used when csv editing
- **Use case**: once we have #69, could use it to say "range every", which would be equivalent to #473

PR #597, opened 2022-03-09
Add support for 'all' modifier for items within a collection

Will-Sommers:
## What
This PR addresses [Issue 473](https://github.com/cursorless-dev/cursorless-vscode/issues/473) and adds support for the  `"all"` modifier within a list. 

I'm pretty rusty on the `.ts` types side and also might be missing some context about casting `Range` as `Selection` and setting it. I think there is likely a more elegant way.

## Checklist

- [ ] Update cheatsheet/docs


Will-Sommers:
Fixing/Running tests locally, still needing to close VSCode

cursorless-talon/src/modifiers/containing_scope.py
```diff
@@ -45,18 +45,30 @@
...
-@mod.capture(rule="[every] {user.cursorless_scope_type}")
+@mod.capture(rule="[(every|all)] {user.cursorless_scope_type}")
 def cursorless_containing_scope(m) -> dict[str, dict[str, Any]]:
-    """Expand to containing scope"""
-    return {
-        "modifier": {
-            "type": "containingScope",
-            "scopeType": m.cursorless_scope_type,
-            "includeSiblings": m[0] == "every",
+    """Expand to every scope"""
```
Will-Sommers, line 56: Definitely open to putting this into another file, my python is a little weak as is my understanding of how it interacts with Talon. Just lmk.

cursorless-talon/src/modifiers/containing_scope.py
```diff
@@ -45,18 +45,30 @@
...
 }
 
+select_multiple_modifiers = {
+    "every", 
+    "all"
+}
+
 
-@mod.capture(rule="[every] {user.cursorless_scope_type}")
+@mod.capture(rule="[(every|all)] {user.cursorless_scope_type}")
```
Will-Sommers, line 54: Would be nifty if we could put this into a capture against the set above.

cursorless-talon/src/modifiers/containing_scope.py
```diff
@@ -45,18 +45,27 @@
...
-            "type": "containingScope",
-            "scopeType": m.cursorless_scope_type,
-            "includeSiblings": m[0] == "every",
+    """Expand to every scope"""
+    if m[0] in select_multiple_modifiers:
+        return {
+            "modifier": {
+                "type": "everyScope",
+                "scopeType": m.cursorless_scope_type,
+                "contiguousRange": m[0] == "all",
```
Will-Sommers, line 59: note: re-add check on for items here

cursorless-talon/src/modifiers/containing_scope.py
pokey: You could use a Talon list, no?

Will-Sommers:
Looking at the test failures. A few notes to myself:

- Re-running the tests via the GH UI resulted in `Error: Cannot activate the 'Cursorless' extension because it depends on unknown extension 'pokey.parse-tree'`
- Tests are failing initially because this PR changes change the type from `containingScope` to `everyScope`

pokey:
I'm going to close this one for now, because:

- for "chuck", we can just use "chuck every"
- For other use cases, I think I'd lean towards implementing this one using #524
- This PR has lots of merge conflicts and seems like it will be painful to take to the finish line

Feel free to re-open if you wanna take a crack at it

Issue #43, opened 2021-06-30
Support type / interface fields in typescript


pokey:
### The problem

For example in the following code:

```typescript
interface Foo {
   bar: Bar;
   baz: Baz;
}
```

If your cursor was before `bar`, then "take field" would select `bar: Bar`

### Approach

Look for the `property_signature` node type

A couple options:

- Introduce a new term "prop".  We may also want it to select property accesses (eg `foo.bar`), and class property definitions
- Modify "state" to target this scope type.  I guess technically these aren't considered statements, but that seems a bit subtle to me
- Modify "item" to target this scope type

---
First raised in #24 by @brxck

AndreasArvidsson:
I think this is fixed

pokey:
Really? I don't see a `field` scope type. What am I missing?

AndreasArvidsson:
You just want a synonym for item?

AndreasArvidsson:
Why use a new scope and not just item?

pokey:
Oh does that work?

pokey:
Also, I'm not sure "item" is what I'd use for that. I could see reusing an existing scope type, but I think I'd lean towards "statement", because although technically not a statement, it sure looks and feels a lot like one

For me, "item" is something in a comma-separated list

PR #605, opened 2022-03-16
Add support for css and scss

Will-Sommers:
Trying to round out the webdev stack with css and sass. I understand that a different query framework is top of mind but thought this would help round out some general web-devy use cases. Happy to be responsible for converting this to the scheme query language when the time comes.

CSS:

## What

Adds support for the `CSS` programming language

## Checklist

- [ ] Added a few specific tests that use `"chuck"` instead of `"change"` to test removal behaviour when it's interesting, especially:
  - [ ] `"chuck arg"` with single argument in list
  - N/A ~`"chuck item"` with single argument in list~
  - N/A ~`"chuck item"` with multiple arguments in list~
- N/A ~Supported` "type"` both for type annotations (eg `foo: string`) and declarations (eg `interface Foo {}`) (and added tests for this behaviour 😊)~
- N/A ~Supported` "item"` both for map pairs and list entries (with tests of course)~

SCSS:

## What

Adds support for the `SCSS` programming language

## Checklist

- [ ] Added a few specific tests that use `"chuck"` instead of `"change"` to test removal behaviour when it's interesting, especially:
  - [ ] `"chuck arg"` with single argument in list
  - N/A ~`"chuck item"` with single argument in list~
  - N/A ~`"chuck item"` with multiple arguments in list~
- N/A ~Supported` "type"` both for type annotations (eg `foo: string`) and declarations (eg `interface Foo {}`) (and added tests for this behaviour 😊)~
- N/A ~Supported` "item"` both for map pairs and list entries (with tests of course)~


src/test/suite/fixtures/recorded/languages/css/chuckArgue5.yml
```diff
@@ -0,0 +1,29 @@
+languageId: css
```
Will-Sommers, line 1: I copied these tests over from the SCSS test folder so the names are messed up, nbd? I'm happy to clean it up as well if you feel strongly about it.

src/languages/scss.ts
```diff
@@ -0,0 +1,101 @@
...
+  // Currently data-lang=^"tel" not treated as a statement
+  // a[data-lang^="tel"] {
+  //    color: rgba(0, 255, 0, 0.5);
+  // }
+
+  statement: STATEMENT_TYPES,
+  string: "string_value",
+  functionCall: "call_expression",
+  namedFunction: ["mixin_statement", "function_statement"],
+  // What does `functionName` do?
```
Will-Sommers, line 57: Genuinely curious about this one. Right now `chuck name` is relying on the name matcher below. Curious waht this scope is for.

src/languages/scss.ts
```diff
@@ -0,0 +1,101 @@
...
+  "use_statement",
+  "warn_statement",
+  "while_statement",
+];
+
+const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
+  // map, list not supported in tree-sitter version
+  ifStatement: "if_statement",
+  condition: conditionMatcher("condition"),
+  // Currently data-lang=^"tel" not treated as a statement
```
Will-Sommers, line 48: Happy to drop this but wanted to call it out.

src/languages/scss.ts
```diff
@@ -0,0 +1,101 @@
...
+        "declaration.variable_name!",
+        "mixin_statement.name!",
+        "attribute_selector.attribute_name!"
+      )
+    )
+  ),
+  value: cascadingMatcher(
+    matcher(
+      patternFinder("declaration"),
+      selectChildrenWithExceptions(["property_name", "variable_name"])
```
Will-Sommers, line 75: Basically in both CSS/SCSS, the following line parses to declaration and then everything as a child node. We want to match on everything except for the property name. Here's the [`node-types`](https://github.com/serenadeai/tree-sitter-scss/blob/master/src/node-types.json#L947-L1004) declaration. 
```
  margin: 10px 10px 10px 10px;
```

This code allows for the ability to take a node's children, with or without named exceptions. It is helpful with `namespace_statement` as well below as a namespace import can have multiple nodes as children and `namespace.*!` just matches on the first one. I'm probably complicating this though.

src/languages/scss.ts
```diff
@@ -0,0 +1,101 @@
...
+  // }
+
+  statement: STATEMENT_TYPES,
+  string: "string_value",
+  functionCall: "call_expression",
+  namedFunction: ["mixin_statement", "function_statement"],
+  // What does `functionName` do?
+  functionName: ["mixin_statement.name!", "function_statement.name!"],
+  comment: ["comment", "single_line_comment"],
+  argumentOrParameter: argumentMatcher("arguments", "parameters"),
```
Will-Sommers, line 60: Currently not handling some of the basic shapes correctly. 

`clip-path: ellipse(115px 55px at 50% 40%);`
or 
`clip-path: polygon(50% 0%, 60% 40%, 100% 50%, 60% 60%, 50% 100%, 40% 60%, 0% 50%, 40% 40%);`

src/util/nodeSelectors.ts
```diff
@@ -190,6 +190,27 @@ export function selectWithLeadingDelimiter(...delimiters: string[]) {
   };
 }
 
+export function selectChildrenWithExceptions(excludedItems: string[] = []) {
```
Will-Sommers, line 193: This is still WIP but I wanted to put it up in case there's an easier built-in way to address handling matching all children or matching all children with the exception of `x` type child or `y` type child.

src/languages/scss.ts
pokey: Is this something we should file as a bug?

src/languages/scss.ts
pokey: It's for "funk name". Basically useful if you're inside eg an assignment statement, so then "name" would take the var name, but you instead want to skip up to the function name

pokey: What you've done looks right

src/languages/scss.ts
pokey: Worth filing? Easy fix?

src/test/suite/fixtures/recorded/languages/css/chuckArgue5.yml
pokey: Yeah that's fine

src/test/suite/fixtures/recorded/languages/scss/changeName3.yml
```diff
@@ -0,0 +1,29 @@
...
+      margin: 10px 10px 10px 10px;
+    }
+  selections:
+    - anchor: {line: 1, character: 14}
+      active: {line: 1, character: 14}
+  marks: {}
+finalState:
+  documentContents: |
+    a {
+      : 10px 10px 10px 10px;
```
pokey, line 21: I almost feel like this should be "key", and "name" should be the selector, but I'm really not sure. Usually "name" is the name in a declaration

pokey (commented):
Ok I haven't looked at all the tests yet but figured I'd drop a few comments

src/util/nodeSelectors.ts
pokey: Also, docstring please 😊

src/languages/scss.ts
```diff
@@ -0,0 +1,101 @@
...
+  "namespace_statement",
+  "placeholder",
+  "rule_set",
+  "supports_statement",
+  "use_statement",
+  "warn_statement",
+  "while_statement",
+];
+
+const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
```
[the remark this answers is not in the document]
Will-Sommers: I added selector, pushing it up soon. Used key + Name for lhs of a prop.

src/test/suite/fixtures/recorded/languages/css/changeState3.yml
```diff
@@ -0,0 +1,29 @@
+languageId: css
```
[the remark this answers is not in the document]
Will-Sommers: Done, made it "item". Now state matches a the block rule.

src/languages/scss.ts
Will-Sommers: Fixed

Will-Sommers:
Heyo @pokey — I'm feeling pretty good about the PR, including the tests, with one exception. I don't like the function `childRangeSelector` that much. It "knows too much" about CSS specifically and also the behavior is sort of unclear. Like — what happens when there are named children but not contiguous, what should we do? In cases where we have a "<match> <miss> <match>" I think the behavior is odd(do we want 2 cursors?). This isn't the case in the CSS parses. Basically we're looking for often either:

"Everything after property_name/variable"
"Just these two things next to each other" 
"Every child or a parent as a single selection"

Happy to move it into the scss file itself if you want or could sit on it a little bit and think more.

src/test/suite/fixtures/recorded/languages/css/changeState4.yml
```diff
@@ -0,0 +1,29 @@
+languageId: css
```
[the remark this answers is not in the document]
Will-Sommers: I deleted this one in CSS because it is invalid syntax.

src/test/suite/fixtures/recorded/languages/css/changeItem.yml
```diff
@@ -0,0 +1,29 @@
...
+  version: 1
+  spokenForm: change item
+  action: clearAndSetSelection
+  targets:
+    - type: primitive
+      modifier: {type: containingScope, scopeType: collectionItem, includeSiblings: false}
+initialState:
+  documentContents: |-
+    a {
+      width: calc(100% - 80px);
```
pokey, line 12: @AndreasArvidsson just a quick one: how do you feel about this line as an "item"?  I'd be tempted to use "field" (#43) for this one instead.  WDYT?

src/test/suite/fixtures/recorded/languages/css/changeSelector.yml
```diff
@@ -0,0 +1,29 @@
+languageId: css
```
pokey, line 1: Cool.  Worth also having a test where the cursor is inside the body of the statement rather than inside the selector itself

src/test/suite/fixtures/recorded/languages/scss/changeKey.yml
```diff
@@ -0,0 +1,29 @@
+languageId: scss
```
pokey, line 1: Worth also having one where cursor is inside the value

src/test/suite/fixtures/recorded/languages/scss/changeValue8.yml
```diff
@@ -0,0 +1,29 @@
+languageId: scss
```
pokey, line 1: worth having a test with cursor inside the `@include`

src/test/suite/fixtures/recorded/languages/scss/changeValue9.yml
```diff
@@ -0,0 +1,29 @@
+languageId: scss
```
pokey, line 1: worth also having test with cursor inside `data-lang`

src/test/suite/fixtures/recorded/languages/scss/chuckArgue4.yml
```diff
@@ -0,0 +1,29 @@
+languageId: scss
```
pokey, line 1: Probably worth supporting "name" from cursor anywhere inside `$amount: 100%` to refer to `$amount`.  Also, not sure if you support this one, but cursor anywhere inside `$amount: 100%`, saying "value" should refer to `100%`

src/util/nodeSelectors.ts
```diff
@@ -190,6 +191,30 @@ export function selectWithLeadingDelimiter(...delimiters: string[]) {
   };
 }
 
+/**
+ * Creates a selector for multiple children of a node, useful for contiguous ranges.
```
pokey, line 195:
```suggestion
 * Creates an extractor that returns a contiguous range between children of a node. 
```

src/util/nodeSelectors.ts
```diff
@@ -190,6 +191,30 @@ export function selectWithLeadingDelimiter(...delimiters: string[]) {
...
+/**
+ * Creates a selector for multiple children of a node, useful for contiguous ranges. 
+ * When no arguments are pased, match on every child. 
+ * When items are passed with the inclusion predicate, only match on those items. 
+ * When items are passed with the exclusion predicate, only match on those items. 
+ * @param items A set of types to match against, either for inclusion or exclusion.
+ * @param predicateType Determines wither nodes within the items argument will be included or excluded.
+ * @returns A selection extractor
+ */
+export function childRangeSelector(predicateType: ChildNodeIncludePredicate = "inclusion", items: string[] = []) {
```
pokey, line 203: Hmm. What was wrong with my original proposal of having two args, `typesToExclude` and `typesToInclude`?  Seems clearer

pokey (commented):
Getting there.  Left a few more minor comments

pokey:
> Heyo @pokey — I'm feeling pretty good about the PR, including the tests, with one exception. I don't like the function `childRangeSelector` that much. It "knows too much" about CSS specifically and also the behavior is sort of unclear. Like — what happens when there are named children but not contiguous, what should we do? In cases where we have a " " I think the behavior is odd(do we want 2 cursors?). This isn't the case in the CSS parses. Basically we're looking for often either:
> 
> "Everything after property_name/variable" "Just these two things next to each other" "Every child or a parent as a single selection"
> 
> Happy to move it into the scss file itself if you want or could sit on it a little bit and think more. Edit: Looking at the failures locally as well.

Yeah I agree it's a bit confusing.  Is my assertion from before correct, that we should be able to do this one easily using forthcoming query language?  If so, I think it's ok to have this one for the time being, tho see my comment on the function

src/languages/scss.ts
```diff
@@ -0,0 +1,140 @@
...
+  "use_statement",
+  "warn_statement",
+  "while_statement",
+];
+
+function isArgumentListDelimiter(node: SyntaxNode) {
+  return [",", "(", ")"].includes(node.type) || checkAtPlainValueDelimiter(node);
+};
+
+function checkAtPlainValueDelimiter(node: SyntaxNode) {
```
pokey, line 50: I don't think there's any way to name this one clearly, so it could def use a quick docstring. 

I think for consistency / simplicity, though, I'd name it

```suggestion
function isAtDelimiter(node: SyntaxNode) {
```

src/languages/scss.ts
```diff
@@ -0,0 +1,140 @@
...
+  return (node.type === "plain_value" && node.text === "at");
+}
+
+// Match up until delimiters and including multiple values within delimiters,
+// e.g. repeating-linear-gradient(red, orange 50px)
```
pokey, line 55: Thanks for adding doc, but better to use proper docstring format so that it appears in autocomplete

src/util/nodeSelectors.ts
```diff
@@ -190,6 +190,38 @@ export function selectWithLeadingDelimiter(...delimiters: string[]) {
   };
 }
 
+/**
+ * Creates an extractor that returns a contiguous range between children of a node. 
+ * When no arguments are pased, the function will match on every child node. Pass in either inclusions
+ * or exclusions.
```
pokey, line 196:
```suggestion
 * When no arguments are passed, the function will return a range from the first to the last child node. Pass in either inclusions
 * If an inclusion or exclusion list is passed, we return the first range of children such that every child in the range matches the inclusion / exclusion criteria.
```

pokey (commented):
Looking great! Down to the super minor stuff now

Will-Sommers:
Looks like there was an odd test run for this commit: https://github.com/cursorless-dev/cursorless/pull/605/commits/6400ef2cf96773039c089e8f6a2707f657e493c0

```
 Failed to parse response from https://update.code.visualstudio.com/api/releases/stable as JSON
 ```
 
 Merged and then the tests re-ran and are ✅

pokey (approved):
Looks good!

PR #606, opened 2022-03-16
Support value scope for yield statements in TS and PHP

### LINKED ISSUE #325: Support value scope type for yield statements in typescript

nathanheffley:
Fixes #325 

I also slipped in support for PHP since I missed it when I originally added the language.

## Checklist


pokey (approved):
good stuff!

PR #611, opened 2022-03-18
Attempt to fix docs link

pokey:
## Checklist


pokey:
@bra1nDump any idea why this one is failing?  If I remove the `api/`, we end up with a dead link. See the [deployed site](https://www.cursorless.org/docs/contributing/api); there we don't use `api/`, which makes CI happy, but the link ends up pointing to `https://www.cursorless.org/docs/contributing/classes/core_commandRunner_CommandRunner.CommandRunner#runcommand`.  For some reason the actual link resolution differs from the resolution during this check 🤔

pokey:
Ohhh weird looks like it has to do with trailing slash

PR #612, opened 2022-03-18
Add PHP to snippets

nathanheffley:
Does what it says on the tin; adds PHP as an option to the existing snippets.

## Checklist


pokey (approved):
Nice!  Didn't realise people were actually using wrapper snippets 😅

nathanheffley:
I only realized that they were usable already while perusing the new doc site, I like them a lot! 👍

PR #620, opened 2022-03-24
Change scopeType matchers to rely on tree-sitter style scheme queries

### LINKED ISSUE #616: Change scopeType matchers to rely on tree-sitter style scheme queries
### Background
Currently Cursorless uses a custom pattern definition DSL alongside a set of helper functions in [nodeMatcher.ts](https://github.com/cursorless-dev/cursorless/blob/6c63c210a217549cebd4ff846f4eaff9c23d50f7/src/util/nodeMatchers.ts) to match various scope types, such as `item` within a list or `argue` within a function definition or function invocation. 

The tree-sitter project also provides a DSL, written in Scheme which allows a user to query for patterns within syntax trees. Here's a [link to the docs](https://tree-sitter.github.io/tree-sitter/using-parsers#pattern-matching-with-queries) and an [example usage in JS](https://github.com/tree-sitter/tree-sitter/blob/master/lib/binding_web/test/query-test.js) via the web-tree-sitter project. Each query then is allowed to assign a `name` to a node, such as [`@comment`](https://github.com/tree-sitter/tree-sitter-ruby/blob/master/queries/highlights.scm#L133-L137) or [`@punctuation.bracket`](https://github.com/tree-sitter/tree-sitter-ruby/blob/master/queries/highlights.scm#L133-L137):

```
(comment) @comment

[
  "("
  ")"
  "["
  "]"
  "{"
  "}"
  "%w("
  "%i("
] @punctuation.bracket
```

The name can then be read or [asserted against](https://github.com/tree-sitter/tree-sitter/blob/master/lib/binding_web/test/query-test.js#L63). 

The thought is that moving towards this approach will be more expressive out of the box. Additionally, many other projects including Neovim and Helix rely on queries for syntax highlighting as well as indentation which might help to make the incremental work for adding a new language a little bit simpler, since there are already partial or full definitions to work from.  In particular, Helix already uses these queries for their [textobjects](https://docs.helix-editor.com/usage.html#textobjects), which is a simplified version of Cursorless scope types.  Here's [an example](https://github.com/helix-editor/helix/blob/master/runtime/queries/python/textobjects.scm) of a set of textobject definitions; they exist for several other languages as well

### The Work
- Create a `queries` directory with a subdirectory for each language, eg `queries/python`, etc
- Create a query file for a language that provides queries for some or all of these [`ScopeTypes`](https://github.com/cursorless-dev/cursorless/blob/main/src/typings/Types.ts#L96-L127), placing the file in `queries/<language>/scopeTypes.scm`
- Add the ability to load queries on a per-language basis
- Queries occur on the tree level (`SyntaxNode.Tree.Language`) and so are top down rather than bottom up as cursorless node matchers currently work. 
- Should a query have successful matches, return the smallest range containing the input selection
- Note that for some of the auxiliary definitions listed below (eg `@<scopeType>.searchScope`) we first find a match, and then [search within that range](https://github.com/tree-sitter/tree-sitter/blob/master/lib/binding_web/test/query-test.js#L74-L75)

### The definitions

- Default query tag is just `@<scopeType>`, so eg `@namedFunction`
- In addition, we support a few other queries:
  - `@<scopeType>.removalRange` indicates a different range that should be used for removal
  - `@<scopeType>.domain` indicates that we should first expand to the smallest containing match for this tag and then search for a rooted instance of `@<scopeType>` within this region.  The canonical example for this one is enabling `take value` from within the key in a map: we'd set `@collectionItem.domain` to be the containing pair
  - `@<scopeType>.iterationScope` indicates that when user says `"every <scopeType>"`, we should first expand to the smallest instance of this tag, and then yield all top-level instances of `@<scopeType>` within this range.  Here, top-level means not contained by any other match within the search range.  Also, note that when finding the instances in the range, we should use `@<scopeType>.domain` if it exists.  See below for an explanation
  - `@<scopeType>.interior` is used by `excludeInterior` and `interiorOnly` stages (see #254)

### Migration notes

This will require a replacement of each of the language matcher files with a `scopeTypes.scm` definition. For this reason, we will want to support both paths while the migration occurs.  We can keep doing continuous delivery during migration because every language other than C# is well tested. 

### Questions

- Do we want to change our term `scopeType` to `textObject`?  That is the term used in both nvim tree-sitter, helix, and by [redstart voice](http://redstartsystems.com/human-machine-grammar-dictionary)
- Better term for `@<scopeType>.iterationScope`?
  - `@<scopeType>.parent`?
- How to handle argument lists and collection items?  I have a feeling we'll be repeating `,` stuff for removal ranges a lot.  I wonder if we want to add Toml configuration for languages where we can indicate scopes that should be handled as comma-separated lists.  Along this direction, it's worth thinking about the connection to #357 
- Do we want to support custom queries?  [Here's](https://github.com/nvim-treesitter/nvim-treesitter#adding-queries) how neovim does it
- Do we still want to support the "every" in cases where a scope doesn't explicitly specify a `@<scopeType>.domain`?  We do that today by just iterating the parent.  Might be useful to keep this one as a fallback 🤷‍♂️

#### Why we need to use `@<scopeType>.domain` when searching within `@<scopeType>.iterationScope`

Consider the following case:

```typescript
{
   foo: {
      bar: "baz"
   }
}
```

If the user says `"take every key fine"`, we want to just return `foo`, excluding the nested key `bar`.  In this case `key.iterationScope` is `object` and `key.domain` is `pair`.  If we just looked for instances of `key` within the `object`, we'd get the nested key as well.  However, if we search for top-level `pair` objects we won't, as desired

#### Why `@<scopeType>` must be rooted within `@<scopeType>.domain`

We can actually use the same code example as above:

```typescript
{
   foo: {|
      bar: "baz"
   }
}
```

If the user says "take key" with the cursor at the indicated position (after second opening bracket), we want to select `foo`.  We first expand to the containing `pair`, as that is the definition of `key.domain`.  Then we need to find the `key`.  If we just look for top-level `key`s (ie not contained by other `key`s), we'll end up with both `foo` and `bar`.  If we require that the `key` be rooted within the `pair`, that won't happen

Fwiw, we could possibly instead exclude any `@<scopeType>` matches which are contained within a lower `@<scopeType>.domain`

### Resources

- Helix [documentation](https://docs.helix-editor.com/guides/textobject.html)
- Helix [implementation](https://github.com/helix-editor/helix/blob/3e78b8fdad953a14c9a4f815295dfc6ed75d409f/helix-core/src/textobject.rs#L147-L151)
- Helix [queries](https://github.com/helix-editor/helix/tree/master/runtime/queries)
- nvim-treesitter [documentation](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#parser-configurations)
- nvim-treesitter implementation (might actually be in neovim itself)
- nvim-treesitter [queries](https://github.com/nvim-treesitter/nvim-treesitter/tree/master/queries)
- neovim [documentation](https://github.com/neovim/neovim/blob/master/runtime/doc/treesitter.txt)
- neovim implementation
  - https://github.com/neovim/neovim/blob/master/runtime/lua/vim/treesitter.lua
  - https://github.com/neovim/neovim/tree/master/runtime/lua/vim/treesitter
- neovim [tests](https://github.com/neovim/neovim/tree/master/test/functional/treesitter)

### Addenda:
- The query syntax also allows for the [injection of information](https://github.com/tree-sitter/tree-sitter-html/blob/master/queries/injections.scm#L3) apart from the `name` of a node and using this DSL will likely be the approach used to support [multi-language documents](https://github.com/cursorless-dev/cursorless-vscode/issues/409).
- https://github.com/cursorless-dev/cursorless/issues/685

### Attributions

👋 Big H/T 🎩 to @wenkokke for the original idea

pokey:
Fixes #616

## Checklist

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -41,6 +41,12 @@
 				"reveal": "never"
 			},
 			"group": "build"
+		},
+		{
+			"label": "Parse quarry",
+			"type": "shell",
+			"command": "npx tree-sitter query scopeTypes.scm ts-test.ts",
+			"presentation": {}
 		}
 	]
 }
--- a/queries/typescript/scopeTypes.scm
+++ b/queries/typescript/scopeTypes.scm
@@ -1,0 +1,56 @@
+(object) @map
+
+(arrow_function) @anonymousFunction
+(function) @anonymousFunction
+
+; We could also have used an alternation here, but the above style is clean and
+; seems to be more commonly used.
+; [
+;     (arrow_function)
+;     (function)
+; ] @anonymousFunction 
+
+; The following doesn't work because we can't have a `(` before the field name (`key`)
+; (
+;     (object
+;         (pair
+;             (key: (_) @key
+;             .
+;             ":") @key.removalRange
+;         ) @key.searchScope
+;     ) @key.iterationScope
+; )
+
+; As a workaround, we could allow defining the end of an annotation
+(
+    (object
+        (pair
+            key: (_) @key
+            .
+            ":" @key.removalRange.end
+        ) @key.searchScope
+    ) @key.iterationScope
+)
+
+; We could instead tag the delimiter itself.  Gives us a bit more information,
+; so maybe that's better?
+(
+    (object
+        (pair
+            key: (_) @key
+            .
+            ":" @key.delimiter.trailing)
+        @key.searchScope)
+    @key.iterationScope)
+
+; We could also try setting an attribute.  This approach is likely how we'd
+; handle comma-separated lists, because they'd be a bit painful to define as
+; parse tree patterns
+(
+    (object
+        (pair
+            key: (_) @key
+            (#set! "delimiter" ":")
+        )
+        @key.searchScope)
+    @key.iterationScope)
--- a/queries/typescript/scopeTypes.scm
+++ b/queries/typescript/scopeTypes.scm
@@ -39,9 +39,10 @@
         (pair
             key: (_) @key
             .
-            ":" @key.delimiter.trailing)
-        @key.searchScope)
-    @key.iterationScope)
+            ":" @key.delimiter.trailing
+        ) @key.searchScope
+    ) @key.iterationScope
+)
 
 ; We could also try setting an attribute.  This approach is likely how we'd
 ; handle comma-separated lists, because they'd be a bit painful to define as
@@ -51,6 +52,6 @@
         (pair
             key: (_) @key
             (#set! "delimiter" ":")
-        )
-        @key.searchScope)
-    @key.iterationScope)
+        ) @key.searchScope
+    ) @key.iterationScope
+)
--- a/queries/typescript/scopeTypes.scm
+++ b/queries/typescript/scopeTypes.scm
@@ -8,7 +8,7 @@
 ; [
 ;     (arrow_function)
 ;     (function)
-; ] @anonymousFunction 
+; ] @anonymousFunction
 
 ; The following doesn't work because we can't have a `(` before the field name (`key`)
 ; (


.vscode/tasks.json
```diff
@@ -44,6 +44,12 @@
 				"reveal": "never"
 			},
 			"group": "build"
+		},
+		{
+			"label": "Parse quarry",
+			"type": "shell",
+			"command": "npx tree-sitter query scopeTypes.scm ts-test.ts",
```
pokey, line 51: Couldn't get this one to work but will prob be useful to have a task or launch config to use while hacking.  Maybe `launch.json` is better

pokey:
Fwiw with this file:

```typescript
foo = {
  foo: {
    bar: "baz",
  },
};

bar = () => "baz";
```

Running

```bash
npx tree-sitter query queries/typescript/scopeTypes.scm  ts-test.ts
```

We get the following:

```
ts-test.ts
  pattern: 0
    capture: map, start: (0, 6), end: (4, 1)
  pattern: 5
    capture: key.iterationScope, start: (0, 6), end: (4, 1)
    capture: key.searchScope, start: (1, 2), end: (3, 3)
    capture: 2 - key, start: (1, 2), end: (1, 5), text: `foo`
  pattern: 3
    capture: key.iterationScope, start: (0, 6), end: (4, 1)
    capture: key.searchScope, start: (1, 2), end: (3, 3)
    capture: 2 - key, start: (1, 2), end: (1, 5), text: `foo`
    capture: 3 - key.removalRange.end, start: (1, 5), end: (1, 6), text: `:`
  pattern: 4
    capture: key.iterationScope, start: (0, 6), end: (4, 1)
    capture: key.searchScope, start: (1, 2), end: (3, 3)
    capture: 2 - key, start: (1, 2), end: (1, 5), text: `foo`
    capture: 6 - key.delimiter.trailing, start: (1, 5), end: (1, 6), text: `:`
  pattern: 0
    capture: map, start: (1, 7), end: (3, 3)
  pattern: 5
    capture: key.iterationScope, start: (1, 7), end: (3, 3)
    capture: 4 - key.searchScope, start: (2, 4), end: (2, 14), text: `bar: "baz"`
    capture: 2 - key, start: (2, 4), end: (2, 7), text: `bar`
  pattern: 3
    capture: key.iterationScope, start: (1, 7), end: (3, 3)
    capture: 4 - key.searchScope, start: (2, 4), end: (2, 14), text: `bar: "baz"`
    capture: 2 - key, start: (2, 4), end: (2, 7), text: `bar`
    capture: 3 - key.removalRange.end, start: (2, 7), end: (2, 8), text: `:`
  pattern: 4
    capture: key.iterationScope, start: (1, 7), end: (3, 3)
    capture: 4 - key.searchScope, start: (2, 4), end: (2, 14), text: `bar: "baz"`
    capture: 2 - key, start: (2, 4), end: (2, 7), text: `bar`
    capture: 6 - key.delimiter.trailing, start: (2, 7), end: (2, 8), text: `:`
  pattern: 1
    capture: 1 - anonymousFunction, start: (6, 6), end: (6, 17), text: `() => "baz"`
```

queries/typescript/scopeTypes.scm
```diff
@@ -0,0 +1,57 @@
...
+; ] @anonymousFunction 
+
+; The following doesn't work because we can't have a `(` before the field name (`key`)
+; (
+;     (object
+;         (pair
+;             (key: (_) @key
+;             .
+;             ":") @key.removalRange
+;         ) @key.searchScope
```
pokey, line 20: One thing to note for all our definitions of `key` is that the trailing `,` on the `pair` is not included in the `searchScope`.  We could probably handle that one fairly easily, but I do wonder if we risk people forgetting to do this.  As an alternative, if we had the `searchScope` instead point to the `pair` scope type, then we could potentially get that delimiter information for free

wenkokke: What is `@key.searchScope` meant to do here? Wouldn’t we need to manually match the “,” token with a match, and bind it to, e.g, `@key.removalRange`?

pokey: `key.searchScope` enables us to say "take key" from anywhere within the pair that contains the key, instead of the default searching behaviour which only matches if you're inside the scope itself

pokey: So I tried the obvious thing for `,`:

```
(
    (object
        (
            (pair
                key: (_) @key
                ":" @key.delimiter.trailing
            )
            ","?
        ) @key.searchScope
    ) @key.iterationScope
)
```

And the problem is that for sibling matches, it appears to only return the range of the first sibling, rather than the range of the two siblings.  @Will-Sommers / @wenkokke have you seen this kind of thing?  Do you know if we can get the full range from the web-tree-sitter api?

Here's the output on simplified code.

Query:

```
(((pair) ",") @test1)
(((pair) (pair)) @test2)
(("," (pair)) @test3)
```

Code:

```typescript
foo = { bar: "baz", bongo: "bazman" };
```

Output:

```
    pattern:  0, capture: 0 - test1, start: (0, 8), end: (0, 18), text: `bar: "baz"`
    pattern:  1, capture: 1 - test2, start: (0, 8), end: (0, 18), text: `bar: "baz"`
    pattern:  2, capture: 2 - test3, start: (0, 18), end: (0, 19), text: `,`
```

Will-Sommers: Heyo — I think what we're looking for here is the `[ ]` alternator syntax. This will allow for a match of the first case or the second case.

```
[
        ((pair) ",")
        (pair)
] @pair
```
yields separate match groups for each kv pair. If we add a `*` at the end, it will be grouped together. If we add a tag at the end of the `,` it will be included in the group as a tag. It would normally be excluded as it isn't a named node.

Is this what we want?

```
foo = {
  bar: "baz",
  bongo: "bazman",
  yar: "yaz"
};


(object
    [
        ((pair) "," @delim)
        (pair)
    ]* @pair
)

ts-test.ts
  pattern: 2
    capture: 2 - pair, start: (1, 2), end: (1, 12), text: `bar: "baz"`
    capture: 1 - delim, start: (1, 12), end: (1, 13), text: `,`
    capture: 2 - pair, start: (2, 2), end: (2, 17), text: `bongo: "bazman"`
    capture: 1 - delim, start: (2, 17), end: (2, 18), text: `,`
    capture: 2 - pair, start: (3, 2), end: (3, 12), text: `yar: "yaz"`
```

Will-Sommers: I'm going to open a ticket on the main project to check this, alternative in the breach is to check for the presence of a named `delim` node after a pair/key and add it to the removal list. At least it is in sequential order.

pokey: Interesting.  Fwiw my goal for this specific experiment isn't removal range, but search range.  Ie I want the following to work:

```typescript
{foo: "bar", baz: "bongo"}
```

If your cursor is right after the `,` I'd like to be able to say "take key" and get `foo`.

I guess we could use the same workaround we were using for `removalRange`, where we allow a `.end` tag.  So we could say:

```
(
    (object
         (pair
             key: (_) @key
             ":" @key.delimiter.trailing
         ) @key.searchScope
         ","? @key.searchScope.end
    ) @key.iterationScope
)
```

queries/typescript/scopeTypes.scm
```diff
@@ -0,0 +1,57 @@
...
+)
+
+; We could instead tag the delimiter itself.  Gives us a bit more information,
+; so maybe that's better?
+(
+    (object
+        (pair
+            key: (_) @key
+            .
+            ":" @key.delimiter.trailing
```
pokey, line 42: Advantage of labelling the delimiter itself is that we can then target it with the "trailing" modifier (#309)

Tho we can also do this with the approach below where we use `#set!`

But I think we should prob support all 3 tbh, as they could all be useful for different things

pokey: One question is whether we want to use `@key.delimiter.trailing` here or just `@key.delimiter`

queries/typescript/scopeTypes.scm
Will-Sommers: Got it, got it.

pokey:
Closing this one in favour of #629, but there might be some interesting ideas in here for implementing more complex scope types

PR #624, opened 2022-03-28
implement numeric max penalty setting. Closes #603

### LINKED ISSUE #603: Allow disabling colored shapes
Currently if colors and shapes are both enabled some of the shape hats will also have colors. This allows additional tokens to have hats but results in longer selectors (i.e. "take blue bolt bat"). Additionally it is harder to determine the color of some shapes than others. For example the bolt shape has far less area than the default circle. Therefore I suggest adding a setting to disable colored shapes so that when colors and shapes are both enabled only one or the other will be used at time.

AndrewDant:
Added a new maxPenalty setting for controlling what combinations of colors and hats are allowed.


package.json
```diff
@@ -334,6 +334,11 @@
 					},
 					"additionalProperties": false
 				},
+				"cursorless.maxPenalty": {
```
pokey, line 337:
```suggestion
				"cursorless.maximumHatStylePenalty": {
```

package.json
```diff
@@ -334,6 +334,11 @@
...
+					"description": "The maximum allowed total penalty, equal to the shape penalty plus the colour penalty. Values less than or equal to zero are treated as no maximum.",
```
pokey, line 338:
```suggestion
					"description": "The maximum allowed penalty for a hat style. Any hat style whose penalty is greater than this amount will not be used. A hat style penalty is defined to be the shape penalty plus the colour penalty. Setting this value less than or equal to zero is treated as no maximum.",
```

src/core/Decorations.ts
```diff
@@ -185,6 +185,8 @@ export default class Decorations {
     const colorPenalties = vscode.workspace
       .getConfiguration("cursorless.hatPenalties")
       .get<Record<HatColor, number>>("colors")!;
+      const maxPenalty = vscode.workspace
+    .getConfiguration("cursorless").get<number>("maxPenalty")!;
```
pokey, line 189:
```suggestion
    .getConfiguration("cursorless").get<number>("maximumHatStylePenalty")!;
```

pokey (commented):
Looks fine; left a couple minor comments. Also would be good to run your changes through prettier and Eslint. Those extensions should be suggested when you open cursorless workspace, but lmk if they weren't

pokey (commented):
looks good!  oh yeah don't forget to update [the docs](https://github.com/cursorless-dev/cursorless#extension-settings)

AndrewDant:
Good point, just added that.

AndrewDant:
Closes #603

pokey (approved):
Looks good!

PR #628, opened 2022-04-08
Update docs to clarify that all actions require a target

### LINKED ISSUE #456: Update docs to clarify that all actions require a target
We've had users try to say just "crown" or "square wrap" without a target and wonder why it won't parse.  We could support these commands, but worry that they would pollute the command space when you can just say eg "crown this" instead.  For the time being, we'll

- improve the docs to make it clearer that these actions without targets aren't possible
- modify the cheatsheet so that every action has a `T` after it for consistency

AndrewDant:
Closes #456
image

### DIFF
--- a/cursorless-talon/src/cheatsheet/sections/actions.py
+++ b/cursorless-talon/src/cheatsheet/sections/actions.py
@@ -16,7 +16,7 @@ def get_actions():
         "rewrap",
     ]
     simple_actions = {
-        key: value
+        key + " <T>": value
         for key, value in all_actions.items()
         if value not in multiple_target_action_names
     }
--- a/cursorless-talon/src/cheatsheet/sections/actions.py
+++ b/cursorless-talon/src/cheatsheet/sections/actions.py
@@ -16,7 +16,7 @@ def get_actions():
         "rewrap",
     ]
     simple_actions = {
-        key + " <T>": value
+        f"{key} <T>": value
         for key, value in all_actions.items()
         if value not in multiple_target_action_names
     }
--- a/docs/user/README.md
+++ b/docs/user/README.md
@@ -56,7 +56,7 @@ Note: If you'd like to customize any of the spoken forms, please see the [docume
 
 ## Overview
 
-Every cursorless command consists of an action performed on a target. For example, the command `"chuck blue air"` deletes the token with a blue hat over the `"a"`. In this command, the action is `"chuck"` (delete), and the target is `"blue air"`.
+Every cursorless command consists of an action performed on a target. For example, the command `"chuck blue air"` deletes the token with a blue hat over the `"a"`. In this command, the action is `"chuck"` (delete), and the target is `"blue air"`. There are no actions without at least one target.
 
 ## Targets
 
@@ -138,7 +138,7 @@ minimize syllables.
 
 ##### `"this"`
 
-The word `"this"` can be used as a mark to refer to the current cursor(s) or selection(s). Note that when combined with a modifier, the `"this"` mark can be omitted, and it will be implied.
+The word `"this"` can be used as a mark to refer to the current cursor(s) or selection(s) as a target. Note that when combined with a modifier, the `"this"` mark can be omitted, and it will be implied.
 
 - `chuck this`
 - `take this funk`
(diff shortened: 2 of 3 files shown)


docs/user/README.md
```diff
@@ -467,8 +470,8 @@ Fold the function with the token containing letter 'a' with a blue hat.
 
 Extracts a target as a variable using the VSCode refactor action
 
-- `"extract"`
-- `"extract {TARGET} as hello world"`
+- `"extract <TARGET>"`
+- `"extract  <TARGET> as hello world"`
```
pokey, line 474: Does this one still work? I think we removed this form a while back

docs/user/README.md
```diff
@@ -319,17 +319,17 @@ Selects both the token containing letter 'a' with a blue hat AND the token conta
 
 ## Actions
 
-In any cursorless command the action defines what happens to the given target, for example deleting the target (`"chuck"`) or moving the cursor to select the target (`"take"`).
+In any cursorless command the action defines what happens to the specified target, therefore every action must have a target. For example deleting the target (`"chuck"` used like `"chuck made"`) or moving the cursor to select the target (`"take"`used like `"take made"`).
```
pokey, line 322:
```suggestion
In any cursorless command, the action defines what happens to the specified target. Every command must have a target. For example, deleting the target (`"chuck"`, as in `"chuck air"`) or moving the cursor to select the target (`"take"`, as in `"take air"`).
```

cursorless-talon/src/cheatsheet/sections/actions.py
```diff
@@ -16,7 +16,7 @@ def get_actions():
         "rewrap",
     ]
     simple_actions = {
-        key: value
+        key + " <T>": value
```
pokey, line 19: Please use f-string
```suggestion
        f"{key} <T>": value
```

cursorless-talon/src/cheatsheet/sections/actions.py
```diff
@@ -16,7 +16,7 @@ def get_actions():
         "rewrap",
     ]
     simple_actions = {
-        key: value
```
pokey, line 19: This change is the one that is most controversial. It's ok with me, but there were objections to how much noise it adds to the cheatsheet cc/ @AndreasArvidsson

pokey (commented):
Lots of good improvements here; thanks! I left some comments

docs/user/README.md
AndrewDant: extract T as hello world doesn't seem to work for me. Extract T still does

cursorless-talon/src/cheatsheet/sections/actions.py
AndrewDant: Talked with Tara Roys and a few other users at the cursorless meetup on 4/10 and they all were in favor of adding the T's for clarity.

docs/user/README.md
AndrewDant: Not sure how to verify for certain but extract ___ as ___ doesn't seem to exist.

AndrewDant: I've verified it by reading the ExtractVariable action file. l'll update the read me

pokey:
Here's my original PR for what it's worth https://github.com/cursorless-dev/cursorless-talon/pull/112

pokey (dismissed):

pokey:
Curious why this one is still marked as draft. What's missing?

AndrewDant:
Because I marked it when I first made the request and forgot to unmark it.

pokey:
> Because I marked it when I first made the request and forgot to unmark it.

Cool that was my theory 😊

pokey:
Mind adding a screenshot to the PR description? Would be good to do a quick visual sanity check before I hit that merge button

AndrewDant:
image

Not sure how to edit the description but here's a screenshot.

cursorless-talon/src/cheatsheet/sections/actions.py
AndreasArvidsson: If people prefer it that way: go ahead :)

AndrewDant:
image
I think that looks much better actually.

pokey (approved):

Issue #484, opened 2022-01-15
Update scope type modifier to expand on both ends


AndreasArvidsson:
Right now you can take lines paragraphs and soon non whitespace sequences that expands from both anchor and active of the selection. Currently containing scopes only use start off the selection.

I want both of these to have the same result. (air and bat are in different functions)
`take funk air past bat`
`take air past bat take funk`

I was thinking that we could implement this first thing in the selection/scope modifier by basically calling the specific implementation twice, once for both ends. To day we pass the entire selection to the specific modifier(line, funk, ...). What if we pass in a single position instead?

```js
const startSelection = getSelectionForScope(selection.start, "function"); // function is of course a variable
if (selection.isEmpty() 
  || getNodeAtLocation(selection.start).id === getNodeAtLocation(selection.end).id) {
    return startSelection;
}
const endSelection = getSelectionForScope(selection.end, "function");
return  getSelectionFromPositions(startSelection.start, endSelection.end);
```

- [ ] Add test for https://github.com/pokey/cursorless-vscode/issues/484#issuecomment-1013856057
- [ ] Add test for "funk wrap that" with `that` mark starting in one statement and ending in another (see https://github.com/cursorless-dev/cursorless/issues/883)
- [ ] Add test for "snip funk after that" with `that` mark starting in one function and ending in another (see https://github.com/cursorless-dev/cursorless/issues/883)
- [ ] https://github.com/cursorless-dev/cursorless/issues/901

## _edit (@pokey)_

Let's implement this one in a new `ContainingScope` stage that all containing scopes share.  The algorithm should be as described in https://github.com/cursorless-dev/cursorless/pull/629#issuecomment-1136090441.  The initial implementation will be folded into https://github.com/cursorless-dev/cursorless/pull/629

pokey:
I don't think this will work for "round", even if we decide to make it consistent with "funk". The problem is if the user has `(())`, and says "take round repper", referring to the first `)`. When expanding from the end, it will look right and end up with the whole string, which is incorrect

AndreasArvidsson:
Yes some special handling of matching pairs is fine. Just getting all the containing scope and selection types working as one would be nice.

Issue #436, opened 2022-01-04
structural editing when cursor is right outside the structure


ym-han:
## The problem

Consider something like

```
def test():
    print("a")
```

and suppose the cursor is at the end of the print statement, after the `)`.

Currently `take funk` does not select the function when the cursor is at that position, though it does work if the cursor is anywhere before the final `)`.

_edit (@pokey):_

### Why does this happen?

Cursorless relies on the [Parse tree extension](https://marketplace.visualstudio.com/items?itemName=pokey.parse-tree) to [get the current node](https://github.com/pokey/vscode-parse-tree/blob/a162fb271ff22f6e8ccd62571acfcb98f54e7352/src/extension.ts#L207), and then walks up from there.  The problem is that this function just passes a position to web-tree-sitter's `rootNode.descendantForPosition`, which will always look to the right, so if there is whitespace to the right, it will go way up the tree to find a node that contains the given whitespace.  

##  The solution

We should probably fix this one inside the parse tree extension.  In [`getNodeAtLocation`](https://github.com/pokey/vscode-parse-tree/blob/a162fb271ff22f6e8ccd62571acfcb98f54e7352/src/extension.ts#L207), we will proceed by trying to find an appropriate position to pass to web-tree-sitter, rather than just directly using the given position:

1. If the position is at the end of the file, subtract one from the position, unless the file is empty, in which case just use the position
1. Otherwise, inspect the character to the right of the position
  1. If it is whitespace, subtract one from the position
  2. Otherwise, use position

We then take the position and pass it to `rootNode.descendantForPosition`, like we do today

### Alternate solution

If the above doesn't work for some reason, we could instead always instead first ask for the node from position like today, but then subtract one from position and ask again, and if they overlap, take the smaller one

Will-Sommers:
Heyo, I might pick this one up.

Will-Sommers:
I'm looking at this a bit more and am not sure that the whitespace approach will work for all languages. 

In typescript for example, there is a non-whitespace statement terminator, `;`(e.g. `let d = [2,3,4];`), or a comma `,` in a list or map, e.g.:

```  
    getTree(document: vscode.TextDocument) {
...
    },

    getNodeAtLocation(location: vscode.Location) {
...
    },
```
These terminators/separators could be specified if we want but all of that configuration happens in the other package. 
 
I'm going to hold off on this for a bit and think about it further. I took a cursory look at the alternate approach. If I understand it correctly, what would happen is:

[...]

PR #629, opened 2022-04-08
Introduce tree-sitter queries for syntactic scopes

Will-Sommers:
- Partially addresses #616
- Partially addresses https://github.com/cursorless-dev/cursorless/issues/436
- Depends on https://github.com/cursorless-dev/cursorless/pull/1396

## Todo
- [x] **[DISCUSS]** What to do about fallback `iterationScope`?  That's the only thing that is a regression here.  
- [x] File issues for FIXMEs
- [x] File issue for defining iteration scopes.  Can probably reuse most of the code from the regular scope handler other than creating the target
- [x] File issue to add unit tests for scope handlers
- [x] File issue to add some Python scope types where multiple can end at the same point (due to lack of closing brackets)
- [x] Add test that checks no scope types are duplicated between legacy and new definition, or file issue to add test
- [x] File PR for my 7783da6 (Add support for domain, leading, trailing, interior) https://github.com/cursorless-dev/cursorless/pull/1427
- [x] Look through comments on this thread for anything worth filing / doing
- [x] Open as new PR?
- [x] Remove extraneous test cases
- [x] Double check https://github.com/cursorless-dev/cursorless/pull/629#discussion_r873125256; a lot of those tests we already have for the generic modifier code
- [x] Make sure changes to parse-tree-extension are shipped
- [x] Close https://github.com/cursorless-dev/cursorless/issues/785 if we fix that
- [x] Comment on #484 saying the process has started and providing link to example
- [x] Close https://github.com/cursorless-dev/cursorless/issues/797 if we fix that


pokey:
Shouldn't we use https://github.com/nvim-treesitter/tree-sitter-query instead of scheme?

pokey:
Also worth mentioning https://marketplace.visualstudio.com/items?itemName=jrieken.vscode-tree-sitter-query, which defined a language id for tree sitter queries

queries/ruby/scopeTypes.scm
```diff
@@ -0,0 +1,21 @@
+(comment) @comment
```
Will-Sommers, line 1: This works as does the `@class` tag. I'm still working on the other ones, so still WIP.

queries/typescript/scopeTypes.scm
```diff
@@ -0,0 +1,59 @@
+(object) @map
```
Will-Sommers, line 1: Not sure the status I left this file in, haven't re-tested the definition.

src/languages/getNodeMatcher.ts
```diff
@@ -65,7 +65,7 @@ const languageMatchers: Record<
   markdown,
   php,
   python,
-  ruby, 
+  ruby: queryBasedSpecification("ruby"),
```
Will-Sommers, line 68: Ideally, I would view this as an "overwrite with precedence" for a transition period. 

We could keep the existing scopes and overwrite the ones which are defined within the scheme files. This might require a definition file or a parse of the `.scm` files and then a match against the values in this scopeType->queryName [mapping here](https://github.com/cursorless-dev/cursorless/pull/629/files#diff-f54826399411e7cc52918c4069c886565dd457b9d472d8e40833a0cb2044bd53).

src/util/scopeToKeyword.ts

pokey (commented):
Just publishing a few random comments here

src/languages/getNodeMatcher.ts
Will-Sommers: Yep, sounds good, I agree.

src/util/scopeToKeyword.ts
```diff
@@ -0,0 +1,40 @@
+import { ScopeType } from "../typings/Types";
+
+export type ScopeTypeToKeyword = { [scope in ScopeType]: string };
+
+const scopeToKeyword: ScopeTypeToKeyword = {
```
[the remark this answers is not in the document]
Will-Sommers: Heyo — I thought that it might be nice to have a layer of indirection between the scheme files and the scopeTypes instead of a direct mapping. I'll think about some unlocks that came to mind when writing this an get back to you.

queries/ruby/scopeTypes.scm
```diff
@@ -0,0 +1,21 @@
...
+(class) @class
+
+(hash
+  (pair
+    key: (_) @key)) @map
+
+(method_parameters
+  . (_) @parameter.inner
+  . ","? @_end
+   (#make-range! "parameter.outer" @parameter.inner @_end))
```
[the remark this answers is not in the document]
Will-Sommers: Yep, that makes sense to me. This was just a copy pasta from the textobject neovim.

src/languages/getNodeMatcher.ts
```diff
@@ -65,13 +66,20 @@ const languageMatchers: Record<
...
   typescript,
   typescriptreact: typescript,
   xml: html,
 };
 
+function mergeMatchers(
+  regexMatcher: Record<ScopeType, NodeMatcher>,
+  languageName: string
+): Record<ScopeType, NodeMatcher> {
+  return Object.assign(regexMatcher, queryBasedSpecification(languageName));
```
pokey, line 80: It's worth checking for situations where we've specified the matcher both in the old and new matcher specifications and throwing an error, because otherwise we might override one without remembering to remove the old matcher

src/util/queryNodeMatchers.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export function defaultMatcher(
+  nodeNames: string[] | string,
+  scopeQuery: string,
+  selector: SelectionExtractor = simpleSelectionExtractor
+): NodeMatcher {
+  return (selection: SelectionWithEditor, node: SyntaxNode) => {
+    let pred = Array.isArray(nodeNames) ? nodeNames : [nodeNames];
+    const query = getQuery(node, scopeQuery);
```
pokey, line 21: This is really the result of running the query, no?
```suggestion
    const queryResult = runQuery(node, scopeQuery);
```

src/typings/Types.ts
```diff
@@ -121,6 +121,7 @@ export type ScopeType =
   | "sectionLevelFour"
   | "sectionLevelFive"
   | "sectionLevelSix"
+  | "selector"
```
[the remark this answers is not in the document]
Will-Sommers: Need to rebase, I think

third-party-licenses.csv
```diff
@@ -1,12 +1,4 @@
...
 "immutability-helper@3.1.1","MIT","https://github.com/kolodny/immutability-helper","https://github.com/kolodny/immutability-helper/raw/master/LICENSE","cursorless"
-"mdast-util-find-and-replace@2.1.0","MIT","https://github.com/syntax-tree/mdast-util-find-and-replace","https://github.com/syntax-tree/mdast-util-find-and-replace/raw/master/license","website"
-"prism-react-renderer@1.2.1","MIT","https://github.com/FormidableLabs/prism-react-renderer","https://github.com/FormidableLabs/prism-react-renderer/raw/master/LICENSE","website"
-"react-dom@17.0.2","MIT","https://github.com/facebook/react","https://github.com/facebook/react/raw/master/LICENSE","website"
-"react@17.0.2","MIT","https://github.com/facebook/react","https://github.com/facebook/react/raw/master/LICENSE","website"
-"unist-util-visit@4.1.0","MIT","https://github.com/syntax-tree/unist-util-visit","https://github.com/syntax-tree/unist-util-visit/raw/master/license","website"
\ No newline at end of file
+"lodash@4.17.21","MIT","https://github.com/lodash/lodash","https://github.com/lodash/lodash/raw/master/LICENSE","cursorless"
```
AndrewDant, line 4: I think this change is an automated one by the build system or something. I've seen it pop up during my own changes. Not sure if it should actually be included. @pokey?

src/languages/getNodeMatcher.ts
Will-Sommers: image
Also adding lang name

queries/ruby/scopeTypes.scm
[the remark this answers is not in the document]
Will-Sommers: Will table in favor of moving PR forward, we'll address more complex ranges in subsequent PRs.

src/typings/treeSitter.ts
```diff
@@ -188,6 +188,7 @@ declare module "web-tree-sitter" {
         endPosition?: Point
       ): QueryCapture[];
       predicatesForPattern(patternIndex: number): PredicateResult[];
+      predicates: PredicateResult;
```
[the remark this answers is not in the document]
Will-Sommers: I'm not sure how this file was generated, but the signature includes predicates now. Happy to regen.

src/util/queryNodeMatchers.ts
```diff
@@ -0,0 +1,72 @@
...
+  const language = node.tree.getLanguage() as Language;
+  return language.query(scopeQuery);
+}
+
+export function defaultMatcher(
+  nodeNames: string[] | string,
+  scopeQuery: string,
+  selector: SelectionExtractor = simpleSelectionExtractor
+): NodeMatcher {
+  return (selection: SelectionWithEditor, node: SyntaxNode) => {
```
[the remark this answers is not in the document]
Will-Sommers: I don't think we'll need to do anything around checking for the new pattern matcher in `processScopeType`. I think that we'll just "do the right thing" and return the correct match the first time. We can put removing ancestor chain checking on the "clean up" ticket and remove along side the regex patterns.

src/util/queryNodeMatchers.ts
Will-Sommers: We have it when invoking `defaultMatcher`. I'll ensure its pre-compiled.

src/util/scopeToKeyword.ts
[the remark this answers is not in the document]
Will-Sommers: Got it — thoughts on converting `ScopeType` into an [`enum`](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) type? I want to be able to iterate through it and match against the query file.

src/languages/queryBasedSpecification.ts
```diff
@@ -0,0 +1,39 @@
...
+  return matchers;
+}
+
+function generateMatcher(
+  queries: string,
+  scopeType: ScopeType,
+  matchers: Partial<Record<ScopeType, NodeMatcherAlternative>>
+) {
+  if (queries.match(`@${scopeType}`)) {
+    const searchScopePresent = !!queries.match(`@${scopeType}.searchScope`);
```
pokey, line 23: This is a minor one, but I prefer booleans to start with `is`:

```suggestion
    const isSearchScopePresent = !!queries.match(`@${scopeType}.searchScope`);
```

src/languages/queryBasedSpecification.ts
```diff
@@ -0,0 +1,39 @@
...
+      scopeType,
+      searchScopePresent,
+      queries
+    );
+  }
+}
+
+export default function (languageName: string) {
+  const queries = fs.readFileSync(
+    path.join(__dirname, `../queries/${languageName}/scopeTypes.scm`),
```
pokey, line 34: Would be better to get extension dir from extension context—see how Decorations.ts gets svgs for hats

src/typings/Types.ts
```diff
@@ -92,40 +92,44 @@ export type SurroundingPairName =
...
-  | "sectionLevelFour"
-  | "sectionLevelFive"
-  | "sectionLevelSix"
-  | "selector"
-  | "xmlBothTags"
-  | "xmlElement"
-  | "xmlEndTag"
-  | "xmlStartTag";
+// TODO: Move to constants file;
+export const scopes = [
```
pokey, line 96: I'd actually still call this 
```suggestion
export const scopeTypes = [
```

Because a scope type is a type of scope, where as a scope is a specific instance of that scope type. 

But open to a terminology discussion if that seems odd / confusing

src/typings/Types.ts
```diff
@@ -92,40 +92,44 @@ export type SurroundingPairName =
...
+  "sectionLevelFive",
+  "sectionLevelSix",
+  "selector",
+  "xmlBothTags",
+  "xmlElement",
+  "xmlEndTag",
+  "xmlStartTag",
+] as const;
+
+export type bb = typeof scopes[number];
```
pokey, line 132:
```suggestion
export type ScopeType = typeof scopeTypes[number];
```

src/typings/treeSitter.ts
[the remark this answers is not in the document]
pokey: I don't feel strongly one way or the other

src/util/queryNodeMatchers.ts
```diff
@@ -0,0 +1,147 @@
...
+    );
+
+    if (captures.length === 0) {
+      return null;
+    }
+    let capture: QueryCapture;
+    capture = selectCaptureByRange(captures, selection);
+    if (siblings) {
+      let captures: QueryCapture[];
+      if (searchScopePresent) {
```
pokey, line 42: I think you mean 
```suggestion
      if (iterationScopePresent) {
```

The `searchScope` has nothing to do with iteration. It's used for thinks like the key key-value pairs: the pair is the search scope so that we can select the key from the value

src/util/queryNodeMatchers.ts
```diff
@@ -0,0 +1,147 @@
...
+): NodeMatcher {
+  let query: Query;
+
+  return (
+    selection: SelectionWithEditor,
+    node: SyntaxNode,
+    siblings: boolean | undefined
+  ): NodeMatcherValue[] | null => {
+    if (!query) {
+      query = getQuery(node, scopeQuery);
```
pokey, line 25: I would still prefer to pass in the language rather than the node, to make it clear we don't actually care about the node itself

src/util/queryNodeMatchers.ts
```diff
@@ -0,0 +1,147 @@
...
+
+function selectCaptureByRange(
+  captures: QueryCapture[],
+  selection: SelectionWithEditor
+) {
+  let capture: QueryCapture;
+  captures.forEach((c) => {
+    const range: Range = new Range(
+      new Position(c.node.startPosition.row, c.node.startPosition.column),
+      new Position(c.node.endPosition.row, c.node.endPosition.column)
```
pokey, line 92: I believe there's a function for this in selectors

src/util/queryNodeMatchers.ts
```diff
@@ -0,0 +1,147 @@
...
+    );
+
+    const hasIntersection = selection.selection.intersection(range);
```
pokey, line 95: What if selection is empty? I think adjacent empty selection should count as intersecting

src/util/queryNodeMatchers.ts
```diff
@@ -0,0 +1,147 @@
...
+function extractCaptures(
+  selection: SelectionWithEditor,
+  root: SyntaxNode,
+  query: Query,
+  scopeType: string
+) {
+  const startPoint = generatePointFromSelection(selection, "start");
+  const endPoint = generatePointFromSelection(selection, "end");
+
+  return query.captures(root, startPoint, endPoint).filter((capture) => {
```
pokey, line 79: It's worth having tests with various versions of overlapping selections to test this and your subsequent intersection code

- [ ] empty selection in white space before scope
- [ ] empty selection touching start of scope
- [ ] empty selection contained by scope
- [ ] empty selection touching end of scope
- [ ] empty selection in white space after scope
- [ ] nonempty selection with start and end in white space before scope
- [ ] nonempty selection with start before scope and end touching start of scope
- [ ] nonempty selection with start before scope and end contained by scope
- [ ] Etc etc
- [ ] Cases where start is in one scope and end is in another

pokey (commented):
Haven't read whole PR yet but here's some comments

src/util/queryNodeMatchers.ts
Will-Sommers: Yep, good point. I think that this should be handled via `extractCaptures` however. By this point, we'll not have any captures, if we didn't expand the selection. I do think it makes sense to:

1. Consider making the widest selection within `extractCaptures`. Perhaps expand `startPoint` and `endPoint` to the preceding and following non-whitespace characters. `query.captures(root, startPoint, endPoint)`
2. Rely on the initial selection, as we do above, initially. 
3. Expand the selection outwards, one way at a time, prioritizing the `active` position over the `anchor` position.

I'm debating if I want to act on the third point by calling `selectCaptureByRange` more than once or expanding within this loop.

pokey: ah ok so the tree-sitter function `query.captures(root, startPoint, endPoint)` won't return the node if it is adjacent to an empty `startPoint, endPoint` range?  That's a bummer.  I would lean towards your option 1) above (expanding left and right just past first non-whitespace char)

pokey:
As discussed in meet-up, algorithm is as follows:

1. Split range into start and end position
2. For the start position
   1. Find smallest node that contains or is adjacent to position
   2. If there are two such nodes, pick the one to the right
3. If the end position is contained or adjacent to node returned in previous step, return the node
4. Otherwise run steps 2.1 and 2.2 on the end node
5. Return the range from the start node to the end node, with direction taken from selection

src/util/queryNodeMatchers.ts
Will-Sommers: Heyo @pokey — for these `Error` cases, I'm going to look at units tests for them. Right now, we nullify all tests where an `Error` is thrown in `CommandRunner`.

src/languages/getNodeMatcher.ts
```diff
@@ -41,13 +43,27 @@ export function getNodeMatcher(
     return notSupported;
   }
 
-  if (includeSiblings) {
```
Will-Sommers, line 44: delete

third-party-licenses.csv
```diff
@@ -1,13 +1,11 @@
 "module name","licenses","repository","licenseUrl","parents"
-"@docusaurus/core@2.0.0-beta.17","MIT","https://github.com/facebook/docusaurus","https://github.com/facebook/docusaurus/raw/master/LICENSE","website"
-"@docusaurus/preset-classic@2.0.0-beta.17","MIT","https://github.com/facebook/docusaurus","https://github.com/facebook/docusaurus/raw/master/LICENSE","website"
-"@mdx-js/react@1.6.22","MIT","https://github.com/mdx-js/mdx","https://github.com/mdx-js/mdx/raw/master/license","website"
```
Will-Sommers, line 4: I think I messed up this merge, will look more a bit later.

wenkokke:
I am currently getting compile errors on the head of that branch:
```
$ tsc -p ./
src/processTargets/modifiers/scopeTypeStages/ContainingSyntaxScopeStage.ts:23:1 - error TS1128: Declaration or statement expected.

23 } from "../../../typings/targetDescriptor.types";
   ~

src/processTargets/modifiers/scopeTypeStages/ContainingSyntaxScopeStage.ts:23:3 - error TS1434: Unexpected keyword or identifier.

23 } from "../../../typings/targetDescriptor.types";
     ~~~~


Found 2 errors in the same file, starting at: src/processTargets/modifiers/scopeTypeStages/ContainingSyntaxScopeStage.ts:23
```

src/processTargets/modifiers/scopeTypeStages/ContainingSyntaxScopeStage.ts
```diff
@@ -1,22 +1,26 @@
...
   NodeMatcher,
   ProcessedTargetsContext,
   SelectionWithEditor,
-  SelectionWithEditorWithContext,
+  SelectionWithEditorWithContext
 } from "../../../typings/Types";
 import { selectionWithEditorFromRange } from "../../../util/selectionUtils";
 import { ModifierStage } from "../../PipelineStages.types";
 import ScopeTypeTarget from "../../targets/ScopeTypeTarget";
+} from "../../../typings/targetDescriptor.types";
```
wenkokke, line 23: That's probably an error, since there's no opening `{`

wenkokke:
PR to fix the issues in HEAD: <https://github.com/Will-Sommers/cursorless-vscode/pull/2>

pokey:
> @Will-Sommers What is the rationale behind defining all scope types in a single query?

good question.  Having one big query seems to be the convention that other repos are using; see the links in the description of #616.  I could see performance being superior if we split it up, but it is easier to have one file per language, esp since most queries are fairly small

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/ruby/clearClassName.yml
pokey, line 1: Noticed this test was missing; not really relevant to this PR

packages/cursorless-engine/src/runIntegrationTests.ts
pokey, line 1: I have mixed feelings about this setup, but I wasn't sure how else to check that we don't duplicate definitions between legacy and new.  The test needs to be defined in the engine, because otherwise we'd need to expose lots of stuff to cursorless-vscode-e2e, but it needs a full tree-sitter instance, so can't be run as a unit test.  This file is the compromise I came up with 🤷‍♂️.  We should maybe revise once we have tree-sitter integrated with this repo

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/queryBasedMatchers/clearCall6.yml
pokey, line 1: In retrospect, I don't think I like this one.  I think it should take the smaller one.  Esp eg for something like "arg" or "item": in something like `a(b(c))`, if cursor is after `c`, I'd expect "arg" to take `c`, but it will take `b(c)`

AndreasArvidsson: Totally agree.

pokey: I think that can be follow-up work. Needs to be tweaked in the downstream code eg containing scope. And it's actually not a regression as this PR stands, because only "arg" handles that properly and this PR doesn't use "arg". I'll remove this test case and move it onto a follow-up issue

pokey: Removed and filed in https://github.com/cursorless-dev/cursorless/issues/1424

packages/cursorless-engine/src/languages/LanguageDefinitions.ts
```diff
@@ -0,0 +1,12 @@
+import { ScopeType } from "@cursorless/common";
+import { ScopeHandler } from "../processTargets/modifiers/scopeHandlers/scopeHandler.types";
+
+export interface LanguageDefinitions {
+  get(languageId: string): LanguageDefinition | undefined;
+}
+
+export interface LanguageDefinition {
+  maybeGetLanguageScopeHandler: (
```
pokey, line 9: `getScopeHandler`

packages/cursorless-engine/src/languages/LanguageDefinitions.ts
```diff
@@ -0,0 +1,12 @@
+import { ScopeType } from "@cursorless/common";
+import { ScopeHandler } from "../processTargets/modifiers/scopeHandlers/scopeHandler.types";
+
+export interface LanguageDefinitions {
+  get(languageId: string): LanguageDefinition | undefined;
```
pokey, line 5: languageId: LanguageId

packages/cursorless-engine/src/languages/LanguageDefinitions.ts
pokey, line 1: get rid of these interfaces for simplicity

packages/cursorless-engine/src/languages/LanguageDefinitionsImpl.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * A list of languages which have query definitions.  Note that it's possible
+ * for a language to have some of its scope types defined via queries and the
+ * rest via legacy `nodeMatcher` definitions.  The
+ * {@link LanguageDefinitionImpl} will return `undefined` for any scope types
+ * which are not defined via queries, which will cause the modifier stage to
+ * fall back to the legacy `nodeMatcher` definitions.
+ */
+const languages: LanguageId[] = ["ruby"];
```
pokey, line 37: Remove this

packages/cursorless-engine/src/languages/LanguageDefinitionImpl.ts
```diff
@@ -0,0 +1,38 @@
...
+import { Query } from "web-tree-sitter";
+import { LanguageDefinition } from "./LanguageDefinitions";
+import { LanguageId } from "./constants";
+import { ide } from "../singletons/ide.singleton";
+import { join } from "path";
+import { TreeSitterScopeHandler } from "../processTargets/modifiers/scopeHandlers";
+import { TreeSitter } from "../typings/TreeSitter";
+import { readFileSync } from "fs";
+
+export class LanguageDefinitionImpl implements LanguageDefinition {
```
pokey, line 11: docstring

packages/cursorless-engine/src/languages/LanguageDefinitionImpl.ts
```diff
@@ -0,0 +1,38 @@
...
+  private query!: Query;
```
pokey, line 12: docstring

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  *generateScopeCandidates(
+    editor: TextEditor,
+    position: Position,
+    direction: Direction,
+    hints: ScopeIteratorRequirements,
+  ): Iterable<TargetScope> {
+    const { document } = editor;
+
+    const { start, end } = this.getQueryRange(
```
pokey, line 45: docstring

pokey:
Ok this one is good to go, but unfortunately it breaks "every funk" for Ruby.  I created https://github.com/cursorless-dev/cursorless/pull/1430 to fix that issue.  That PR is ready, so would be great if you could have a look when you get a minute.  I think they should prob merge in together so we don't break "every funk" in Ruby on main

pokey (approved):

PR #619, opened 2022-03-24
Reimplement cheatsheet using html

pokey:
I have left the existing "cursorless help" intact for now, so we can see whether deploy works.  This functionality is all self-contained, and accessible via a new command "cursorless reference"

Fixes #618
Fixes #630
Fixes #454
Fixes #505

### Light mode
image

### Dark mode
image

### Mobile
image

## Todo

- [x] Switch `identifier` to `id`
- [x] Rename `data/cheatsheet`
- [x] Change `spokenForms` to `variations`
- [x] Use camelcase for eg `hat_color` in json

PR #632, opened 2022-04-12
Moved the legend to the top of the cheatsheet to make it more prominent

### LINKED ISSUE #630: The cheatsheet legend should be moved to more prominent position
The legend is currently located at the bottom left of the screen, where it is easily missed or cut off. 
image

I believe it should be moved to the top left so that it is easily visible.

AndrewDant:
image

I've also included the change to move the special marks to the final column for a more accurate picture of how this will look.

Closes #630


pokey:
I think prob better to just wait for #619

PR #633, opened 2022-04-16
add generator expressions to matcher for python argumentOrParameter

### LINKED ISSUE #323: Argument scope type doesn't work in python when argument is a generator expression
In the following code:

```python
" ".join(word for word in word_list)
```

Saying "take arg whale" should select the whole generator expression between the parentheses, because it is the argument to `join`. Instead it finds no argument

AndrewDant:
Closes #323


src/languages/python.ts
```diff
@@ -79,7 +82,10 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
     ),
     patternMatcher("return_statement.~return!")
   ),
-  argumentOrParameter: argumentMatcher("parameters", "argument_list"),
+  argumentOrParameter: cascadingMatcher(
```
Will-Sommers, line 85: Heyo! Cool! 

[Here's the link to the test case recorder](https://www.cursorless.org/docs/contributing/test-case-recorder/). `cursorless record` doesn't recognize well for me, just a heads up. 

A few good tests to record would be: 
- `"chuck arg"` with single argument in list
- `"chuck arg"` with multiple arguments in list, both generator expressions
- `"chuck arg"` with multiple arguments in list, one generator expression one other

I'm not super familiar with python but I assume that there are situations where we can have 2 generator expressions as arguments. If not, disregard that case.

src/languages/python.ts
```diff
@@ -79,7 +82,10 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
...
+    argumentMatcher("parameters", "argument_list"),
+    matcher(patternFinder("generator_expression"), childRangeSelector())
```
pokey, line 87: I get the feeling that this one will match any generator expression. So eg it would match the right-hand side of the following:

```python
foo = (bar for baz in bongo)
```

But that's not actually an argument. So I believe you'll need to look at the parent node to make sure you're in an argument list

Will-Sommers: Since `generator_expression` is a type of argument, coming from call, would it be `call.arguments[generator_expression]`?

https://github.com/tree-sitter/tree-sitter-python/blob/master/src/node-types.json#L776-L785

AndrewDant: Just confirmed this issue.

AndrewDant: Would the best path forward be to write my own finder that essentially extends patternFinder() to filter for patterns with the correct parent node type?

AndrewDant: Or could existing finders such as ancestorChainNodeFinder() perform this functionality if configured correctly?

AndrewDant: Right after asking I started trying it myself and it seems to work better than I expected. I sort of assumed my first thought wouldn't work.

src/languages/python.ts
AndrewDant: Is the idea behind these kinds of tests to record correct behavior and then running the tests will verify the behavior hasn't changed?

AndrewDant: I recorded and pushed a test case for 'take' and 'chuck' for each of the examples you gave. Any others I should add?

src/languages/python.ts
```diff
@@ -79,7 +82,17 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
     ),
     patternMatcher("return_statement.~return!")
   ),
-  argumentOrParameter: argumentMatcher("parameters", "argument_list"),
+  argumentOrParameter: cascadingMatcher(
+    argumentMatcher("parameters", "argument_list"),
+    matcher(
+      ancestorChainNodeFinder(
+        1,
+        patternFinder("call"),
```
Will-Sommers, line 90: Heyo! I think you could do this using the language matching syntax: 

```
matcher(
      patternFinder("call.generator_expression!"),
      childRangeSelector()
    )
```   
What is happening here is — selecting node if the parent is a `call` type and the child is a `generator_expression` type. Without the `"!"`, we would select all of these nodes, however, we just want to select the arguments, not the entire call.

src/languages/python.ts
Will-Sommers: Great!

I think a `chuck every argue` would be great. It looks like the `( )` and "," are cleaned up in multi argument situations but it might be good to have a case where we remove every argument when there are two.

I'll defer to @pokey on whether to chuck or keep the `"take"` tests. From my understanding, they're harder to parse since we need to look at the cursor positions to understand the output, and that's tougher than just looking and seeing that the argument has been deleted. If there isn't something in the contrib page, we should add it as a note for contributions.

pokey: Our policy is as follows:

We like to have "change" tests for everything. It is not the end of the world to use "take" instead of "change", but we do slightly prefer "change" to "take" for the reasons @Will-Sommers mentions above. Generally not worth re-recording tho imo

For scope types where removal is interesting (eg ensuring comma gets removed), we like to also add "chuck" tests. I think this case does warrant "chuck" tests, as you've done

pokey: And agreed a "chuck every" would be nice for good measure

AndrewDant: I've added one "chuck every" test with just a generator expression and one with the generator expression and another argument.

src/languages/python.ts
AndrewDant: I've made this change and verified it still works.

src/languages/python.ts
AndrewDant: Also I'll try to remember the preference for change instead of take for tests.

src/test/suite/fixtures/recorded/languages/python/chuckEveryArgRed.yml
```diff
@@ -0,0 +1,27 @@
+languageId: javascript
```
pokey, line 1: This should probably be 
```suggestion
languageId: python
```

src/test/suite/fixtures/recorded/languages/python/chuckEveryArgRed.yml
```diff
@@ -0,0 +1,27 @@
...
+command:
+  version: 1
+  spokenForm: chuck every arg red
+  action: remove
+  targets:
+    - type: primitive
+      modifier: {type: containingScope, scopeType: argumentOrParameter, includeSiblings: true}
+      mark: {type: decoratedSymbol, symbolColor: default, character: r}
+initialState:
+  documentContents: "\" \".join(word for word in word_list)"
```
pokey, line 11: I guess this one is a good corner case, but would be good to also have a test where there are multiple generator expressions as arguments

src/test/suite/fixtures/recorded/languages/python/chuckEveryArgWhale.yml
```diff
@@ -0,0 +1,27 @@
+languageId: javascript
```
pokey, line 1:
```suggestion
languageId: python
```

pokey (commented):
Almost there!

pokey:
Oh yeah and sorry bout CI; will be fixed once I merge #638

src/test/suite/fixtures/recorded/languages/python/chuckEveryArgRed.yml
AndrewDant: Yeah that's definitely not right

pokey:
pre-commit.ci autofix

src/test/suite/fixtures/recorded/languages/python/chuckEveryArgRed.yml
AndrewDant: Added to new commit.

src/test/suite/fixtures/recorded/languages/python/chuckEveryArgWhale.yml
AndrewDant: I re recorded these tests.

src/test/suite/fixtures/recorded/languages/python/chuckEveryArgRed.yml
AndrewDant: I re recorded these tests.

AndrewDant:
I just pushed the latest round of changes.

pokey (approved):
Good to go!

PR #637, opened 2022-04-20
Add pre-commit

pokey:
Adding pre-commit.  To see what it looks like when we run it, see #638 

See #639 and #640 for a couple minor follow-up tasks


.vscode/settings.json
```diff
@@ -6,9 +6,7 @@
     "search.exclude": {
         "out": true // set this to false to include "out" folder in search results
     },
-    "[typescript]": {
-        "editor.defaultFormatter": "esbenp.prettier-vscode"
-    },
+    "editor.defaultFormatter": "esbenp.prettier-vscode",
```
pokey, line 9: Note that we now use prettier for all file types that it supports

.vscode/extensions.json
```diff
@@ -2,6 +2,7 @@
 	// See http://go.microsoft.com/fwlink/?LinkId=827846
 	// for the documentation about the extensions.json format
 	"recommendations": [
-		"dbaeumer.vscode-eslint"
+		"dbaeumer.vscode-eslint",
+		"esbenp.prettier-vscode"
```
pokey, line 6: This will probably be nice for new contributors

src/core/updateSelections/updateSelections.ts
```diff
@@ -66,7 +66,7 @@ export function getSelectionInfo(
  */
 export function selectionsToSelectionInfos(
   document: TextDocument,
-  selectionMatrix: Selection[][],
+  selectionMatrix: (readonly Selection[])[],
```
pokey, line 69: Looks like our typescript version got bumped when I bumped yarn.lock, so now we need to be stricter about `readonly`

src/util/addDecorationsToEditor.ts
```diff
@@ -20,15 +20,17 @@ export function addDecorationsToEditors(
...
-    editors = vscode.window.visibleTextEditors.filter(
-      (editor) => editor !== vscode.window.activeTextEditor
-    );
-    editors.unshift(vscode.window.activeTextEditor);
+    editors = [
+      vscode.window.activeTextEditor,
+      ...vscode.window.visibleTextEditors.filter(
+        (editor) => editor !== vscode.window.activeTextEditor
+      ),
+    ];
```
pokey, line 33: We can't use `unshift` on a readonly array

package.json
```diff
@@ -516,12 +516,14 @@
...
 		"esbuild": "^0.11.12",
-		"eslint": "^7.15.0",
+		"eslint": "^7.32.0",
+		"eslint-config-prettier": "^8.5.0",
 		"fast-xml-parser": "^3.20.0",
 		"glob": "^7.1.7",
 		"js-yaml": "^4.1.0",
 		"mocha": "^8.1.3",
 		"npm-license-crawler": "^0.2.1",
+		"prettier": "2.6.2",
```
Will-Sommers, line 526: Not a big thing but if you want to add the ability to shift up rather have a strict peg, that would be nice.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,34 @@
+minimum_pre_commit_version: "2.9.0"
```
Will-Sommers, line 1: We should add the pre-commit install process as an install step in the `contributing.md`

.pre-commit-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - id: mixed-line-ending
+      - id: check-merge-conflict
+  - repo: https://github.com/pre-commit/mirrors-eslint
+    rev: "v8.13.0"
+    hooks:
+      - id: eslint
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: "v2.6.2"
+    hooks:
+      - id: prettier
```
Will-Sommers, line 30: I'm trying to grok this but am curious — are we using a `prettier.rc` here or just going with the full defaults?

Will-Sommers:
Looks great to me.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - id: trailing-whitespace
+        # Trailing whitespace breaks yaml files if you use a multiline string
+        # with a line that has trailing white space. Many of our recorded
+        # tests use strings with trailing white space to represent the final
+        # document contents. For example
+        # src/test/suite/fixtures/recorded/languages/ruby/changeCondition.yml
+        exclude: ^src/test/suite/fixtures/recorded/.*/[^/]*\.yml$
+      - id: mixed-line-ending
+      - id: check-merge-conflict
+  - repo: https://github.com/pre-commit/mirrors-eslint
```
Will-Sommers, line 23: One possible alternative would be to pull out eslint and just use prettier. 

The [`"--check"`](https://prettier.io/docs/en/cli.html#--check) flag returns an error code if a change is needed. I don't have an opinion on this but wanted to present the option.

pokey: What's the advantage of pulling out eslint?  I believe this rule will let eslint autofix things if it can

.pre-commit-config.yaml
pokey: Full defaults

.pre-commit-config.yaml
pokey: We can, though fwiw I don't plan to install it personally.  I plan to add it to CI

.vscode/settings.json
AndreasArvidsson: Have you confirmed that it's not trying to use prettier for python for example. "that it supports" it's not something I would rely on to work automatically.

pokey: Yeah I just tried saving and it seems to be using black

.editorconfig
```diff
@@ -0,0 +1,27 @@
...
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+# Matches multiple files with brace expansion notation
+# Set default charset
+[*.{js,py}]
```
auscompgeek, line 13: This repo is mostly TS, right, rather than straight JS?
```suggestion
[*.{ts,py}]
```

.editorconfig
```diff
@@ -0,0 +1,27 @@
...
+charset = utf-8
```
auscompgeek, line 14: Just curious, why limit this default charset to JS/Py?

package.json
auscompgeek: I think it makes sense to pin `prettier` in the package.json to the same version in the pre-commit config. As long as we remember to bump this whenever it gets bumped in the pre-commit config.

.pre-commit-config.yaml
Will-Sommers: Sorry — disregard. I was looking through the prettier docs and misunderstood one point.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+    rev: "v8.13.0"
+    hooks:
+      - id: eslint
```
auscompgeek, line 26: Might need to specify `--fix` to let eslint autofix if that's what you're after.
```suggestion
      - id: eslint
        args: ['--fix']
```

You might need to add `eslint-config-prettier` to `additional_dependencies` as well?

Looking at the [mirrors-eslint readme](https://github.com/pre-commit/mirrors-eslint#using-eslint-with-pre-commit) it looks like you might need to specify the file types for TypeScript?

.editorconfig
pokey: Just stole that from the editorconfig example on their homepage, so had assumed they did it for a reason.  Shall we set it for all files?

package.json
pokey: Hmm it's a bit of a bummer to have to specify prettier version in two places.  I wonder if that's avoidable? 🤔

.pre-commit-config.yaml
pokey: Done thanks!

package.json
pokey: Ok I've now pinned everything in package.json that's specified in the pre-commit config, but that seems not great.  I wonder if there's a better way

Will-Sommers: I'll open an issue on their repo and see what comes of it. I think that duplicating it is the path forward for now.

Will-Sommers: It looks like this is the resolution. The tldr; is to rely on a [repository local hook](https://pre-commit.com/#repository-local-hooks)and throw the check into an npm or yarn script.

https://github.com/pre-commit/pre-commit/issues/945#issuecomment-527234424

pokey: Ok, given that currently `eslint` is already run in CI, the `eslint --fix` isn't buying us a ton, so probably not worth the effort.  Let's remove the `eslint` fixer for now, and wait until we start using it to remove unused imports in #640.  I think it's ok to leave `prettier` pinned until then as well.  It was mainly the pinned eslint typescript parser plugin making me nervous

Sound good?

.pre-commit-config.yaml
pokey: I added a blurb to contributor docs linking to `pre-commit` docs in case people want to learn how to install hooks / run them manually locally

Will-Sommers (dismissed):

.pre-commit-config.yaml
```diff
@@ -0,0 +1,41 @@
...
+exclude: ^src/vendor/
+repos:
+  - repo: https://github.com/Zac-HD/shed
+    rev: 0.9.5
+    hooks:
+      - id: shed
+        # TODO: bump to --py310-plus when Talon moves to Python 3.10.
+        args: [--refactor, --py39-plus]
+        types_or: [python, markdown, rst]
+  - repo: https://github.com/pre-commit/pre-commit-hooks
```
auscompgeek, line 13: Might be a good idea to move pre-commit-hooks before shed? They should be faster I think; failing fast would probably be ideal.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,41 @@
...
+    rev: v4.1.0
+    hooks:
+      - id: check-added-large-files
+      - id: check-case-conflict
+      - id: check-executables-have-shebangs
+      - id: check-merge-conflict
+      - id: check-shebang-scripts-are-executable
+      - id: check-symlinks
```
auscompgeek, line 21:
```suggestion
      - id: check-symlinks
      - id: check-yaml
```

auscompgeek (dismissed):
LGTM. A couple of comments.

.pre-commit-config.yaml
pokey: I'm pretty sure prettier will fail if it finds bad yaml, so that seems redundant to me. I can verify that tho

.pre-commit-config.yaml
pokey: Done.  Fwiw, though, pre-commit doesn't short-circuit, so I'm not sure there's too much benefit

auscompgeek: ah, I suppose you want to run black after flynt anyway

.pre-commit-config.yaml
pokey: Ok confirmed that prettier reports invalid yaml, and the output is prettier than check-yaml

.pre-commit-config.yaml
pokey: Yeah exactly. And as a matter of fact I did find it useful to be able to hit Ctrl-c when running locally if I noticed failure, so there is some benefit.

Thanks for all the helpful comments! I'm happy with how it turned out

PR #638, opened 2022-04-20
Run pre-commit

pokey:
## Checklist


src/test/suite/fixtures/helloWorld.txt
```diff
@@ -1 +1 @@
-hello world
\ No newline at end of file
+hello world
```
pokey, line 1: Note that the added new line here is causing tests to fail; easy fix

auscompgeek:
Maybe add a `.gitattributes` with

```gitattributes
*  text=auto eol=lf
```

which should force CI on Windows to check out the file with Unix line endings.

pokey:
> Maybe add a `.gitattributes` with
> 
> ```gitattributes
> *  text=auto eol=lf
> ```
> 
> which should force CI on Windows to check out the file with Unix line endings.

Ok so first of all TIL so thanks for that 😄

Any objection to just doing https://github.com/cursorless-dev/cursorless/pull/638/files#diff-bcd60fce722e3a07bb2e897fc863f9e0f9baff1e2667a348730b3ee310d85661R72?  Seems simpler, and doesn't really alter the goal of the test, which is to ensure that it is opening the correct document

pokey:
Fwiw I've split this PR into 2 commits.  The first commit consists of manual changes which have some (slight 😊) substance.  The second commit is purely automated and just does the automated pre-commit fixes.  I'll merge these in without squashing, so that the latter can be added to `.git-blame-ignore-revs` without including the manual changes.  cc/ @phillco @auscompgeek

PR #653, opened 2022-04-26
Update to java statements related to switch and block

AndreasArvidsson:
* Was unable too use statement scope on switch statements 
* Removed block statement so if state meant loops and so on can be selected from inside

## Checklist


pokey (approved):
Good stuff!

PR #658, opened 2022-04-26
More docs improvements

pokey:
## Checklist


docs/user/README.md
```diff
@@ -12,48 +12,6 @@ You can get back to these docs by saying `"cursorless instructions"`.
 
 Note: If you'd like to customize any of the spoken forms, please see the [documentation](customization.md).
 
-## Table of contents
```
pokey, line 15: Now that we have a docs hub, the table of contents is in the right sidebar

PR #664, opened 2022-05-03
Add "callee" scope type. 

### LINKED ISSUE #645: Add "callee" scope type
This scope type would select the function being called in a function call, so eg in `foo(bar)`, it would select `foo`.  We'd also use this one for the tag function in Javscript [tagged templates](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates), as proposed in #641.  So saying "take callee" within either of the following lines should take the `foo` on the given line:

```typescript
foo(bar);
foo`hello ${bar}`;
```

### LINKED ISSUE #671: Constructor calls are not captured by the 'call' scope type in several languages.
The pattern matcher for several languages (i.e. java and c#) only matches standard method calls for the 'call' scope type, missing constructor call entirely. For example 'take call' on `new ClassName()` will select nothing and throw an error.

AndrewDant:
Still a long way to go but I thought I would open this up for feedback early. Closes #645
I'm adding Closes #671 on to this since I'm already referencing all the appropriate call types in order to use them for callee.


AndrewDant:
@pokey do we need chuck tests for this scope addition?

pokey:
> 
> 
> @pokey do we need chuck tests for this scope addition?

No I wouldn't say so; removal isn't particularly interesting here

Will-Sommers:
Cool!

AndrewDant:
I will need to go back and test if those languages work for constructors.

AndrewDant:
I got through several languages tonight so this should be ready for review soon. So far I've been doing one test per pattern for each language, but I'm thinking that might be a little low considering some languages only need one pattern. Any guidance on how much to record?

src/languages/typescript.ts
```diff
@@ -188,6 +188,11 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
   regularExpression: "regex",
   className: ["class_declaration[name]", "class[name]"],
   functionCall: ["call_expression", "new_expression"],
+  functionCallee: [
+    "call_expression.identifier!",
+    "call_expression.call_expression!",
+    "new_expression.identifier!",
```
pokey, line 194: as above
```suggestion
    "new_expression[constructor]",
```

(and tests for corner cases, eg `new (foo[0])()`)

src/test/suite/fixtures/recorded/languages/java/changeCallee.yml
```diff
@@ -0,0 +1,23 @@
...
+    - type: primitive
+      modifier: {type: containingScope, scopeType: functionCallee, includeSiblings: false}
+initialState:
+  documentContents: obj.addNumbers(num1, num2);
+  selections:
+    - anchor: {line: 0, character: 24}
+      active: {line: 0, character: 24}
+  marks: {}
+finalState:
+  documentContents: obj.(num1, num2);
```
pokey, line 16: This is wrong imo.  Should select `obj.addNumbers`.  cc/ @AndreasArvidsson tho in case his expectations for Java would be different

src/languages/java.ts
```diff
@@ -53,6 +53,11 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
   anonymousFunction: "lambda_expression",
   list: "array_initializer",
   functionCall: "method_invocation",
+  functionCallee: [
+    "method_invocation.identifier!",
+    "method_invocation[name]",
+    "object_creation_expression.type_identifier!",
```
pokey, line 59: as above, should be
```suggestion
    "object_creation_expression[type]",
```

and worth adding tests to hit at least a few of the cases in https://github.com/tree-sitter/tree-sitter-java/blob/39a11c8330d0b2ad75310456c378396785f79a4e/grammar.js#L1007-L1015

src/test/suite/fixtures/recorded/languages/ruby/takeCallee.yml
```diff
@@ -0,0 +1,23 @@
+languageId: ruby
+command:
+  version: 1
+  spokenForm: take callee
+  action: setSelection
+  targets:
+    - type: primitive
+      modifier: {type: containingScope, scopeType: functionCallee, includeSiblings: false}
+initialState:
+  documentContents: f = comp.(foo, bar)
```
pokey, line 10: what does this syntax do?

src/test/suite/fixtures/recorded/languages/java/changeCallee.yml
AndreasArvidsson: I'm unsure. I don't think I really understand what this scope type is supposed to do.

src/test/suite/fixtures/recorded/languages/java/changeCallee2.yml
```diff
@@ -0,0 +1,23 @@
...
+    - type: primitive
+      modifier: {type: containingScope, scopeType: functionCallee, includeSiblings: false}
+initialState:
+  documentContents: Main obj = new Main();
+  selections:
+    - anchor: {line: 0, character: 20}
+      active: {line: 0, character: 20}
+  marks: {}
+finalState:
+  documentContents: Main obj = new ();
```
AndreasArvidsson, line 16: I'm wondering if I would not expect `new` to be part of the callee

src/test/suite/fixtures/recorded/languages/ruby/takeCallee.yml
AndrewDant: I believe it's a syntax for calling a method returned from another method.
https://www.notonlycode.org/12-ways-to-call-a-method-in-ruby/#:~:text=5%2D7%3A%20using%20%22method%22%20and%20%22call%22

src/languages/typescript.ts
AndrewDant: It looks like you have one extra closing paren there. Also do you want it to select the new as well? It selects correctly other than that.

src/test/suite/fixtures/recorded/languages/java/changeCallee2.yml
AndrewDant: I think we need to make a decision on this because selecting the field as @pokey suggested does not select the new keyword in multiple languages. It does select everything after the 'new ' in cases like `new object.innerClass()`

src/test/suite/fixtures/recorded/languages/python/changeCallee2.yml
```diff
@@ -0,0 +1,23 @@
+languageId: python
+command:
+  version: 1
+  spokenForm: change callee
+  action: clearAndSetSelection
+  targets:
+    - type: primitive
+      modifier: {type: containingScope, scopeType: functionCallee, includeSiblings: false}
+initialState:
+  documentContents: "object.method()\r\n"
```
pokey, line 10: Not sure it's worth re-recording the test, but if you can avoid the `\r\n` by either using unix line endings (`\n`) or just not having new lines at all, the test case will be easier to read as yaml will be able to leave out the quotes

AndrewDant: I went ahead and rerecorded it.

src/test/suite/fixtures/recorded/languages/python/changeCalleeBlueParen.yml
```diff
@@ -0,0 +1,27 @@
+languageId: python
```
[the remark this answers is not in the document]
AndrewDant: It was a duplicate so I've deleted it.

src/util/nodeSelectors.ts
```diff
@@ -200,13 +200,21 @@ export function selectWithLeadingDelimiter(...delimiters: string[]) {
  */
 export function childRangeSelector(
   typesToExclude: string[] = [],
-  typesToInclude: string[] = []
+  typesToInclude: string[] = [],
+  useUnnamedStartAndEnd: boolean = false
```
AndrewDant, line 204: @pokey I'm not sure about this variable name. The prior behavior ignored unnamed nodes at the start or end of the range (but did include any inside the range). This optional parameter keeps the default behavior the same but allows us to include unnamed nodes if desired. Could also be broken up into a separate start and end parameter.

src/languages/java.ts
```diff
@@ -53,6 +53,11 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
   anonymousFunction: "lambda_expression",
   list: "array_initializer",
   functionCall: "method_invocation",
+  functionCallee: [
+    "method_invocation.identifier!",
+    "method_invocation[name]",
```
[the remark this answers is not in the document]
AndrewDant: childRangeSelector() is solving this problem for me with only a slight modification.

src/util/nodeSelectors.ts
```diff
@@ -200,13 +200,16 @@ export function selectWithLeadingDelimiter(...delimiters: string[]) {
...
```
pokey, line 204: I think I'd go with
```suggestion
  includeUnnamedNodes: boolean = false
```

It's a little confusing, but the whole concept behind this function is a little confusing, and I think this naming is consistent with the idea of this function 

Also we'll be getting rid of it soon so it's not that critical

WDYT @Will-Sommers

pokey (commented):
Looking good! I think Python / TS are good to go. I believe what's missing is

- [ ] adding some more thorough tests for the other languages, similar to what you now have for Python / TS
- [ ] Also don't forget about explicit constructor invocations in Java (see comment from before) 😊

src/test/suite/fixtures/recorded/languages/java/changeCallee.yml
AndrewDant: I've removed these bad tests.

src/languages/java.ts
AndrewDant: To be honest I don't know how to interpret this such that I could build test cases from it.

src/util/nodeSelectors.ts
Will-Sommers: This function is a kludge that I should go away with `.scm` parsing. I'm not tied to any naming concerns with it, and agree with you Andrew re: included/excluded. If you look at the method closely, it will also include non-specified nodes within a selection, including "excluded" types since we're just returning between 2 inclusion bounds.

AndrewDant: Yeah haha I ran into that with java since the ';' is included as part of the explicit_constructor_invocation. I think the preferred behaviour might be to stop extending the range once an excluded type is found, but current behaviour is fine for now if we're assuming it's temporary.

src/test/suite/fixtures/recorded/languages/java/changeCallee4.yml
```diff
@@ -0,0 +1,23 @@
+languageId: java
+command:
+  version: 1
+  spokenForm: change callee
+  action: clearAndSetSelection
+  targets:
+    - type: primitive
+      modifier: {type: containingScope, scopeType: functionCallee, includeSiblings: false}
+initialState:
+  documentContents: "public class Test {\r\n    public Test()  {\r\n        super();\r\n    }\r\n}"
```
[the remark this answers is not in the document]
AndreasArvidsson: If you set your line endings to `\n` they will be formatted much better in the yml files.

src/languages/java.ts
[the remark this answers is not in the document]
AndrewDant: This should be resolved by recent updates.

src/test/suite/fixtures/recorded/languages/java/changeCallee4.yml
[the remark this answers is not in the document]
AndrewDant: Okay will make that change. Do I need to rerecord these tests?

src/languages/go.ts
```diff
@@ -43,6 +43,10 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
   string: ["interpreted_string_literal", "raw_string_literal"],
   ifStatement: "if_statement",
   functionCall: "call_expression",
+  functionCallee: [
+    "call_expression.identifier!",
+    "call_expression.call_expression!",
+  ],
```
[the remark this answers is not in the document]
AndrewDant: I've added those two test cases. Go does not have constructors apparently.

src/test/suite/fixtures/recorded/languages/cpp/changeCallee3.yml
```diff
@@ -0,0 +1,23 @@
...
+    - type: primitive
+      modifier: {type: containingScope, scopeType: functionCallee, includeSiblings: false}
+initialState:
+  documentContents: Car carObj1("X5", 1999);
+  selections:
+    - anchor: {line: 0, character: 17}
+      active: {line: 0, character: 17}
+  marks: {}
+finalState:
+  documentContents: Car ("X5", 1999);
```
AndrewDant, line 16: We would probably like this to include the type, but I'm currently not sure how to implement that. The type is in the field declaration[type] but the rest of what we want is in declaration[declarator][declarator]. Simply taking declaration[declarator] would include the arguments list. I'm open to suggestions.

src/test/suite/fixtures/recorded/languages/java/changeCallee4.yml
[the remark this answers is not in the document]
pokey: I think it would be nice to re-record this one. For the single-line tests it's not necessary, but for multi-line it's much easier to read because then yaml won't wrap it in a string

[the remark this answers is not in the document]
AndrewDant: Would it make sense to commit that as a workspace setting so that it's the default for all contributors?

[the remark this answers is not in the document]
AndreasArvidsson: Yes!
`"files.eol": "\n"`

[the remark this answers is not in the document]
AndrewDant: Unfortunately this didn't work well because the workspace opened when debugging does not use the workspace settings from the actual project.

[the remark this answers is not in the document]
AndrewDant: After changing my users level settings it still is not working. I tried restarting the debugger session and deleting then retyping all line breaks to no avail. Not sure what to do about this.

src/languages/java.ts
AndrewDant: @pokey do you mind talking me through how to interpret grammar.js for the sake of creating tests sometime?

src/test/suite/fixtures/recorded/languages/java/changeCallee4.yml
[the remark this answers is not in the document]
AndreasArvidsson: I believe that setting is just for new files. If you already saved files with a a different eol those will persist.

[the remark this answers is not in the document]
AndreasArvidsson: You can change the eol for an entire document. Look in the footer for `CRLF`.

src/languages/java.ts
AndrewDant: Did you mean to reference the tests in the repository under tests/corpus? I'm realizing now I probably should have referenced that. Or did you actually mean the grammar.js file?

AndrewDant:
@pokey I think this is ready for you to look through again. It might be done but I've made enough changes I wouldn't be surprised if there are new issues I missed.

src/test/suite/fixtures/recorded/languages/clojure/changeCallee4.yml
```diff
@@ -0,0 +1,23 @@
+languageId: clojure
```
pokey, line 1: huh this one's a bit of a corner case.  I'd almost expect to get `f` here, but I'm not sure; what you have may be more consistent.  Maybe let's get a couple second opinions cc/ @somanythings @wenkokke

src/test/suite/fixtures/recorded/languages/csharp/changeCallee.yml
```diff
@@ -0,0 +1,23 @@
+languageId: csharp
```
pokey, line 1: @AndreasArvidsson I know you originally were pushing to include `new`, but I believe yesterday we had a use case where you wanted to just target `TestMotorcycle` here, without the `new`.  WDYT?

src/test/suite/fixtures/recorded/languages/java/changeCallee4.yml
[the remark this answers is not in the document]
pokey: would be nice to re-record with `LF` line endings for readability

src/util/nodeSelectors.ts
pokey: Would be good to at least add this parameter to doc string

pokey (commented):
Basically there!  I left a couple questions.  Also,

- [ ] I think we probably want [struct literals](https://yourbasic.org/golang/structs-explained/) for Go

src/test/suite/fixtures/recorded/languages/csharp/changeCallee.yml
AndreasArvidsson: For type I definitely think new should be excluded for callee one could make an argument for both. If you prefer without new I'm happy with that.

pokey:
> @pokey "would be nice to re-record with LF line endings for readability" Should we just enforced that in the test recorder?

nice idea.  yeah let's do that

src/test/suite/fixtures/recorded/languages/csharp/changeCallee.yml
AndrewDant: Would personally prefer it without the new but I could also see the argument either way.

pokey: Ok let's drop the `new` 👍

src/util/nodeSelectors.ts
pokey: Are you using this arg anymore, or was it just for `new`?  If not, I'd just revert this file

pokey (commented):
Nearly there!  I left one minor comment, and let's give 24 hours for a response on Clojure, then I think we're good to merge!

src/util/nodeSelectors.ts
AndrewDant: Reverted.

AndrewDant:
> * [ ]  I think we probably want [struct literals](https://yourbasic.org/golang/structs-explained/) for Go

I'm adding these for both call and callee, can you give me some examples of tests you'd like added? I'm not actually familiar with the language.

pokey:
Looks like this one has some merge conflicts; lmk if you want help resolving

pokey (dismissed):

auscompgeek:
Maybe `call sign` as a nod to radio?

AndrewDant:
I'm surprised to hear that because I haven't had any recognition issues on my end. Maybe `call source` or `caller` would work better.

pokey:
Huh; might be ok to just let it be then. I'll try "call Lee" locally to see if that helps. Conformer can be particular about where you put spaces

@auscompgeek I think "call sign" might be a bit opaque for the default term, but I do quite like it; might take that one for my own csv override 😄

pokey:
Ok I did some more testing, emphasizing the second syllable ("call**ee**"), and it seemed to work ok.  So if it works well for you, probably fine to leave it, and people can remap it if they're having trouble with it.

pokey (approved):
Let's ship it!

PR #670, opened 2022-05-11
Added necessary spacing in talon rules

AndreasArvidsson:
Solves issue with talon's new `DeprecationWarning: Rule needs a space between tokens:`

## Checklist


pokey:
looks like you've just folded this one into #672, so shall we close?

Issue #254, opened 2021-08-31
update inside / outside


pokey:
- "inside" / "outside" / "pair" can be applied to anything that has an interior and scope starters / enders, eg an `if` statement, surrounding pairs, functions, strings, etc.  Note that they do *not* apply to items in delimited sequences, such as args
- separately, we have "leading" / "trailing" that refer to the delimiters before or after items in delimited sequences, eg the commas in an arg list, or whitespace for tokens
- We want "inside" without any scope type to automatically refer to the nearest applicable scope, not including the "space" surrounding pari type
- The above modifiers will just go into a `modifiers.csv` in cursorless-settings
- If you say "chuck inside list" it will remove everything inside list including whitespace padding
- If you say "clear inside list" it will not remove whitespace padding inside
- If you say "chuck leading inside list" it will remove leading padding in the list
- If you say "chuck trailing string" it will remove whitespace after the string
- If you say "chuck list" it will remove list and trailing whitespace

AndreasArvidsson:
I'm gonna take a crack at inside element because that one is so useful

pokey:
Yeah that's fair. Was hoping to wait until #629, but this one is too useful to hold off I think

Prob easiest to just add an `interiorRange` attribute to the legacy [`SelectionContext`](https://github.com/cursorless-dev/cursorless/blob/main/src/typings/Types.ts#L56)?

Once we have #629 we can add weak expansion support that will stop if it hits a scope type that supports inside, so that we can just say "take inside air" for anything inside an element.

AndreasArvidsson:
Exactly the plan

AndreasArvidsson:
Started work in #851

Issue #202, opened 2021-08-07
Improve "that" mark for "bring to after line"


pokey:
In the following file

```
hello
world
```

Calling the following two commands

```
bring line harp to after whale
indent that
```

Results in the following

```
hello
    world
    hello
```

pokey:
I believe we could fix this issue and #185 by not including added delimiter in "that" mark when using "bring to after". Does that seem like reasonable behaviour?

At some point we might want to make it so that the "that" mark includes both an inside and outside range, but I think the above is a reasonable place to start that I believe should be not too bad to implement

PR #672, opened 2022-05-12
Support fully compositional modifiers

### LINKED ISSUE #69: Support fully compositional modifiers
### The problem

Today's cursorless targets are [flat](https://github.com/pokey/cursorless-vscode/blob/0ca8def59ce96a8e8895a41e9c4b983c229db186/src/typings/Types.ts#L229-L237), allowing only a single [modifier](https://github.com/pokey/cursorless-vscode/blob/0ca8def59ce96a8e8895a41e9c4b983c229db186/src/typings/Types.ts#L232). This limitation dramatically restricts the power of cursorless commands. See [What's next](https://github.com/pokey/cursorless-vscode/wiki/What's-next) for the big picture; here we focus on the details of the solution.

### The solution

Instead of being a [flat object](https://github.com/pokey/cursorless-vscode/blob/0ca8def59ce96a8e8895a41e9c4b983c229db186/src/typings/Types.ts#L229-L237) with optional attributes, each of which represent a stage in a makeshift pipeline, `PrimitiveTarget` will simply have a list of pipeline stages. During execution, cursorless will apply each pipeline stage to the output of the one before it. Each pipeline stage output will be a [rich target](https://github.com/cursorless-dev/cursorless-vscode/issues/210), giving the next stage enough information to process the previous output.  The [`mark`](https://github.com/pokey/cursorless-vscode/blob/0ca8def59ce96a8e8895a41e9c4b983c229db186/src/typings/Types.ts#L231) will just become a pipeline stage with zero inputs.

The approach is inspired by [jq](https://stedolan.github.io/jq/). Each pipeline stage in a cursorless target description takes a single input and emits zero or more outputs. The outputs of a modifier stage are passed in one by one to the next stage in the modifier pipeline, and then all outputs from that stage are passed onward, etc. See the [jq language spec](https://github.com/stedolan/jq/wiki/jq-Language-Description#the-jq-language) for a better idea of the paradigm.

Some of our modifiers will be parametrized. For example "every" is a modifier that takes a scope type such as "line", "funk", etc as a parameter, and generates a pipeline stage that yields every instance of the given scope type occurring in each of its inputs. So for example "every line in funk" will create a pipeline that first expands its input to the nearest containing function, and then yields every line in that function

### LINKED ISSUE #199: Support testing decorations
We'd basically do all decorations through a mockable interface and then dump them along with the rest of the yaml. For actions that decorate twice, we could track both, and compare document contents at time of decoration to initial and final state, and then store them as pre- or post-decoration based on that

### LINKED ISSUE #209: "Chuck past end of T" removes trailing delimiter

### LINKED ISSUE #210: Switch targets to object-oriented

### LINKED ISSUE #242: Allow "clone", "pour" etc to operate on other scopes
For example "pour air" would make a new line, "pour token air" would insert space and leave cursor on same line, "pour arg air" would insert comma and leave cursor after. Similarly, "clone air" would clone the line, but "clone token air", "clone arg air" etc would operate with the corresponding delimiters

Using action preference selection type

Springs from #230 

Will rely on #202

Should we also handle "drop" / "float" / "puff" the same way?  So eg "float argue air" would insert a `, ` before the argument

### LINKED ISSUE #309: Add scope types "trailing" / "leading"

### LINKED ISSUE #625: When you say "chuck line target1 between target2" a single empty line sticks around.
All lines between target1 and target2 should be removed. 

In my gif example below “ace” is `}`. 

chuck between

### LINKED ISSUE #675: Use/create lines at the end of a file
It would be great to be able to use a position beyond the last line of a file as a target so I could move paragraphs or code chunks to the very bottom. If it's not possible to use this as a target, it would be nice to be able to "pour" a line to the very bottom so I could move my target by using the row number in two steps rather than three or four. It might also be useful to be able to "pour" to a line number that doesn't exist yet and it would create blank lines until you hit that line number.

### LINKED ISSUE #681: Investigate whether we want "bring to end of" to function like a replace
I would be tempted to do
```suggestion
            destination.position === "after",
```

Because "end of" should prob push things to the right rather than appearing after. What breaks when you do this?

_Originally posted by @pokey in https://github.com/cursorless-dev/cursorless/pull/672#discussion_r875934215_

### LINKED ISSUE #688: Incorrect inside with syntax error
language: typescript
selection: `text`
command: "take inside"
problem: Selection ends before `%`

```
async (editor, targets) => {
    const edits = targets.map((target, i) =>(
    text: i%2 === 0?left:right
    )
}
```

### LINKED ISSUE #693: Add source mark to actions `wrap` and `rewrap`
Use the inner range as that source mark. 
`take that` => selects entire thing with parentheses
`take source` => selects the interior without the parentheses

### LINKED ISSUE #697: The that mark on bring action should not contain delimiter
Today if you bring something after and then say `take that` that delimiter is included

pokey:
Fixes #69
Fixes #675 
Fixes #625
Fixes #688 
Fixes #242 
Fixes #693 
Fixes #199
Fixes #209
Fixes #697
Fixes #210 
Fixes #309 
Fixes #681

Started on #202 - Should we close this one and create a new with the additional parts?
Started on #254 - Let's go through the checkboxes there to see what's left after this one merges

* Support composable modifiers 
     `take token after line`
* Supports every prefix for all scope types
    `take every line`
    `take every token in funk`
* Supports scope specific delimiters for clone, drink, pour
    `drink arg`
    `clone arg`

**Note**: Although this PR was originally opened by @pokey, about 90% of the work ended up being done by @AndreasArvidsson 😊

## Checklist

- [x] Check DFA on public
- [x] Remove pre-extension publication provenance files from VSCode ignore
- [x] Add / ensure there are tests for "drink", "pour", "puff", "float", and "drop" with empty line, empty line with indent, and nonempty line with indent
- [x] Add test for "float" / "puff" / "drop" when line is at beginning / end of file
- [x] Figure out "paste to to" (solution?: in csv_overrides.py, if identifier is our paste action, then remove any trailing `" to"`)
- [x] Support [Mocha's `--grep`](https://mochajs.org/#-grep-regexp-g-regexp) feature for running single test.  prob use env var for simplicity when using launch.json
- [x] Add test for "pour" when target is in opposite split
- [x] Test "float" and "drop" before / after an indented line, ensuring it doesn't add padding
- [x] Add `setSelectionsWithoutFocusingEditor` and use it 
- [x] Deduplicate selections in `setSelections`
- [x] Add a test for "pour line" when it should indent (eg top of function), after we've fixed behaviour
- [x] Add tests for "bring state air to after bat" with new newline insertion
- [x] Fix bug with "drink every arg"
- [x] "bring air before just bat" inserts a space
- [x] "drink just bat" inserts a newline


AndreasArvidsson:
@pokey I have only one thing I'm not completely happy with. 
Surrounding pair modifier has an optional `[{user.cursorless_delimiter_force_direction}]` that you can say things like `left round`. It also supported `inside left`. 

Now when `inside` is a proper and separate modifier direction on that doesn't really make sense. `take inside left funk` what should that do? I think that left and right are properties of the matching pair and not of inside. You could always still combine them `take inside left pair`. I also think that the current syntax with `inside left` is a bit weird and I would want it other way around. My suggestion is to just leave direction for the matching pair and don't have it on the interior modifier. 

https://github.com/pokey/cursorless-vscode/blob/9b7d94d5b99e1b6937a953604b926e26f1231db2/cursorless-talon/src/modifiers/interior_boundary.py#L12

AndreasArvidsson:
@pokey Could we remove this one?
https://github.com/pokey/cursorless-vscode/blob/9b7d94d5b99e1b6937a953604b926e26f1231db2/src/scripts/transformRecordedTests/transformations/updateSurroundingPairTest.ts#L6

pokey:
> @pokey Could we remove this one?
> 
> https://github.com/pokey/cursorless-vscode/blob/9b7d94d5b99e1b6937a953604b926e26f1231db2/src/scripts/transformRecordedTests/transformations/updateSurroundingPairTest.ts#L6

Was just leaving that there as an example. Maybe we should just migrate it to a code snippet in the docs? Is it giving you type errors or something?

AndreasArvidsson:
Yes `DelimiterInclusion` is not a thing on surrounding pair any more. Do you mind moving it or pointing me to where you want me to move it to?

pokey:
> Yes `DelimiterInclusion` is not a thing on surrounding pair any more. Do you mind moving it or pointing me to where you want me to move it to?

https://www.cursorless.org/docs/contributing/test-case-recorder/#custom-transformation

pokey:
> @pokey I have only one thing I'm not completely done with. Surrounding pair modifier has an optional `[{user.cursorless_delimiter_force_direction}]` so that you can say things like `left round`. It also supported `inside left`.
> 
> Now when `inside` is a proper and separate modifier direction on that doesn't really make sense. `take inside left funk` what should that do? I think that left and right are properties of the matching pair and not of inside. You could always still combine them `take inside left pair`. I also think that the current syntax with `inside left` is a bit weird and I would want it other way around. My suggestion is to just leave direction for the matching pair and don't have it on the interior modifier.
> 
> https://github.com/pokey/cursorless-vscode/blob/9b7d94d5b99e1b6937a953604b926e26f1231db2/cursorless-talon/src/modifiers/interior_boundary.py#L12

This makes sense to me.  I believe it's consistent with the [docs](https://www.cursorless.org/docs/#ambiguous-delimiters----etc).  Your [code](https://github.com/cursorless-dev/cursorless/pull/672/files#diff-dbe38a055bfd0b751f461dd4f2b0cec081e34305698fe4480f29291f158c850fR43) looks right

src/processTargets/getModifierStage.ts
```diff
@@ -0,0 +1,73 @@
...
+    case "document":
+      return new DocumentStage(modifier);
+    case "line":
+      return new LineStage(modifier);
+    case "paragraph":
+      return new ParagraphStage(modifier);
+    case "nonWhitespaceSequence":
+      return new NonWhitespaceSequenceStage(modifier);
+    case "url":
+      return new UrlStage(modifier);
```
pokey, line 68: All these class names should be prefixed by `Containing`, eg `ContainingDocumentStage` to make it clear what they actually do.  That will also distinguish them from eg "next paragraph", "first line", etc, which also yield something of the given type

src/processTargets/marks/CursorStage.ts
```diff
@@ -0,0 +1,26 @@
...
+import { MarkStage } from "../PipelineStages.types";
+
+export default class implements MarkStage {
+  constructor(private modifier: CursorMark) {}
+
+  run(): Target[] {
+    if (window.activeTextEditor == null) {
+      return [];
+    }
+    return window.activeTextEditor.selections.map((selection) => {
```
pokey, line 14: As above, why did you remove this info from the context object and get it here?  I'd prefer if the pipeline's only interactions with the environment are via context obj

src/processTargets/marks/DecoratedSymbolStage.ts
```diff
@@ -0,0 +1,32 @@
...
+    const target = {
+      editor: token.editor,
+      contentRange: token.range,
+      isReversed: false,
+    };
+    return [
+      {
+        ...target,
+        ...getTokenContext(target),
+        scopeType: "token",
```
pokey, line 28: how come this one isn't included in the output of `getTokenContext`?

I'm really tempted to switch to the `new Token(...)` setup.  All the object spreading / pojos are feeling like they'll be pretty tough to maintain

src/processTargets/PipelineStages.types.ts
```diff
@@ -0,0 +1,10 @@
+import { Target } from "../typings/target.types";
+import { ProcessedTargetsContext } from "../typings/Types";
+
+export interface MarkStage {
+  run(context: ProcessedTargetsContext): Target[];
+}
+
+export interface ModifierStage {
+  run(context: ProcessedTargetsContext, target?: Target): Target | Target[];
```
pokey, line 9: how come modifiers can return a single target but marks cannot?  seems inconsistent.  I'd lean towards always requiring a list, but at the very least I'd argue for consistency

pokey (commented):
Looking good!  Still more left to review, but figured I'd publish the comments I have so far.

- [ ] Don't forget to migrate the test cases.  Maybe we should do that one at the end to keep the diff from going insane 😅

src/core/commandRunner/CommandRunner.ts
```diff
@@ -94,12 +94,6 @@ export default class CommandRunner {
       }
 
       const processedTargetsContext: ProcessedTargetsContext = {
-        currentSelections:
-          vscode.window.activeTextEditor?.selections.map((selection) => ({
-            selection,
-            editor: vscode.window.activeTextEditor!,
-          })) ?? [],
-        currentEditor: vscode.window.activeTextEditor,
```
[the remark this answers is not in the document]
AndreasArvidsson: They are only used in the cursor mark stage I access them there directly.

src/actions/EditNewLine.ts
```diff
@@ -1,38 +1,27 @@
-import {
```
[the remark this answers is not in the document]
AndreasArvidsson: I do have some code ready for that as soon as the targets already

src/core/IndividualHatMap.ts
```diff
@@ -5,6 +5,7 @@ import HatTokenMap from "./HatTokenMap";
 
 export interface ReadOnlyHatMap {
   getEntries(): [string, Token][];
+  getTokens(): Token[];
```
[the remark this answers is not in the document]
AndreasArvidsson: Get entries is already there. Should that won't be removed as well?

src/processTargets/getModifierStage.ts
AndreasArvidsson: The implementation of containing and every is in the same file. The file represents the scope type

src/processTargets/marks/CursorStage.ts
AndreasArvidsson: My opinion is that passing around objects for no good reason is something that should be avoided. Now it's only accessed when needed which I think is cleaner.

src/processTargets/marks/DecoratedSymbolStage.ts
AndreasArvidsson: Because the token context is used for other types as well. Decorated mark for example

src/processTargets/PipelineStages.types.ts
AndreasArvidsson: The ones that doesn't require a list doesn't user list. I aim for simplicity.

src/processTargets/marks/DecoratedSymbolStage.ts
AndreasArvidsson: I'm going to rename it to something more clear

src/processTargets/marks/CursorStage.ts
AndreasArvidsson: You want me to add it back and I will?

src/core/commandVersionUpgrades/upgradeV1ToV2/upgradeV1ToV2.ts
```diff
@@ -0,0 +1,153 @@
...
+      if (Array.isArray(mod)) {
+        modifiers.push(...mod);
+      } else {
+        modifiers.push(mod);
+      }
+    }
+  }
+
+  if (isImplicit) {
+    modifiers.push({ type: "toRawSelection" });
```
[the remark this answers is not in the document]
AndreasArvidsson: Yes now raw selection is only used if you say "just".

src/core/commandVersionUpgrades/upgradeV1ToV2/upgradeV1ToV2.ts
```diff
@@ -0,0 +1,153 @@
...
+      }
+    }
+  }
+
+  const newMark = mark?.type === "cursorToken" ? undefined : mark;
+
+  return {
+    ...rest,
+    mark: newMark,
+    // Modifiers are processed backwards
```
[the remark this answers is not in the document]
AndreasArvidsson: Because I constructed it in the same order as it was in the old processing stages. I can change it

src/core/commandVersionUpgrades/canonicalizeAndValidateCommand.ts
```diff
@@ -74,11 +93,15 @@ export function validateCommand(
...
+  scopeType: ScopeType,
+  partialTargets: PartialTargetDesc[]
 ) {
-  return getPartialPrimitiveTargets(partialTargets).some(
-    (partialTarget) => partialTarget.selectionType === selectionType
+  return getPartialPrimitiveTargets(partialTargets).some((partialTarget) =>
+    partialTarget.modifiers?.find(
+      (mod) =>
+        (mod.type === "containingScope" || mod.type === "everyScope") &&
+        mod.scopeType === scopeType
```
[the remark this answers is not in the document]
AndreasArvidsson: I think you have a better grasp of the future problem problems related to this so please file an issue.

src/core/commandVersionUpgrades/upgradeV1ToV2/upgradeV1ToV2.ts
```diff
@@ -0,0 +1,153 @@
...
+    modifier,
+    isImplicit,
+    selectionType,
+    position,
+    ...rest
+  } = target;
+  const modifiers: Modifier[] = [];
+
+  if (selectionType) {
+    modifiers.push({ type: "containingScope", scopeType: selectionType });
```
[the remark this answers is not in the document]
AndreasArvidsson: You are indeed correct that they are in wrong order. I have a problem where talon sends selection type token that clobbers our desired modifier if it comes after. I'm working on solution.

src/core/commandRunner/CommandRunner.ts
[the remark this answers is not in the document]
AndreasArvidsson: Restored

src/core/commandVersionUpgrades/upgradeV1ToV2/upgradeV1ToV2.ts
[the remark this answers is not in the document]
AndreasArvidsson: fixed!

src/processTargets/marks/CursorStage.ts
AndreasArvidsson: done

src/processTargets/PipelineStages.types.ts
AndreasArvidsson: But I guess consistency is good as well so I'm updating

src/actions/BringMoveSwap.ts
```diff
@@ -132,12 +118,13 @@ class BringMoveSwap implements Action {
...
-          range: destination.selection.selection,
+          range: destination.contentRange,
           text,
-          editor: destination.selection.editor,
+          editor: destination.editor,
           originalSelection: destination,
           isSource: false,
-          isReplace: destination.position === "after",
+          isReplace:
+            destination.position === "after" || destination.position === "end",
```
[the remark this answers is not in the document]
pokey: sounds good.  opened https://github.com/cursorless-dev/cursorless/issues/681 to track

src/core/IndividualHatMap.ts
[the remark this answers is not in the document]
pokey: `getEntries` is used for testing, so I think we need to keep it.  Given that `getTokens` currently has no use case, I'm not sure it's worth adding.  Easy enough to just revert this file, no?  Would be consistent with our effort to remove exported functions that aren't used I think 😊

src/core/commandVersionUpgrades/canonicalizeAndValidateCommand.ts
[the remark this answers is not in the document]
pokey: filed https://github.com/cursorless-dev/cursorless/issues/682

src/processTargets/modifiers/scopeTypeStages/TokenStage.ts
```diff
@@ -0,0 +1,211 @@
...
+    const { isEmpty } = contentRange;
+    const start = isEmpty
+      ? editor.document.lineAt(contentRange.start).range.start
+      : contentRange.start;
+    const end = isEmpty
+      ? editor.document.lineAt(contentRange.end).range.end
+      : contentRange.end;
+
+    const targets = context.hatTokenMap
+      .getTokens()
```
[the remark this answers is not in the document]
AndreasArvidsson: Got it

src/core/IndividualHatMap.ts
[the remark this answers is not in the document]
AndreasArvidsson: Never mind I found a function you are already using

src/core/inferFullTargets.ts
```diff
@@ -77,112 +81,99 @@ function inferNonListTarget(
...
-    };
+  const previousModifiers = getPreviousModifiers(previousTargets);
+  const implicitModifiers = target.isImplicit
+    ? [{ type: "toRawSelection" } as Modifier]
+    : undefined;
+
+  const modifiers =
+    target.modifiers != null
+      ? target.modifiers
+      : implicitModifiers ??
```
pokey, line 129: can't you just do
```suggestion
    target.modifiers ??
      implicitModifiers ??
```

src/core/inferFullTargets.ts
```diff
@@ -77,112 +81,99 @@ function inferNonListTarget(
...
+        previousModifiers ??
+        actionPreferences?.modifiers ??
+        [];
+
+  // TODO Is this really a good solution?
+  // "bring line to after this" needs to infer line on second target
```
pokey, line 135: Seems good enough for now.  I'd leave the comment about "bring line to after this", though.  Otherwise I'd have no clue what was going on here 😅

src/processTargets/PipelineStages.types.ts
```diff
@@ -0,0 +1,10 @@
+import { Target } from "../typings/target.types";
+import { ProcessedTargetsContext } from "../typings/Types";
+
+export interface MarkStage {
+  run(context: ProcessedTargetsContext): Target[];
+}
+
+export interface ModifierStage {
+  run(context: ProcessedTargetsContext, target?: Target): Target[];
```
pokey, line 9: Why the `?` on the `target` arg?  When would this one be undefined?

src/core/updateSelections/updateSelections.ts
```diff
@@ -76,6 +90,18 @@ export function selectionsToSelectionInfos(
...
 }
 
+function rangesToSelectionInfos(
+  document: TextDocument,
+  rangeMatrix: (readonly Range[])[],
+  rangeBehavior: DecorationRangeBehavior = DecorationRangeBehavior.ClosedClosed
+): FullSelectionInfo[][] {
+  return rangeMatrix.map((ranges) =>
+    ranges.map((range) =>
+      getSelectionInfoInternal(document, range, false, rangeBehavior)
```
pokey, line 100: won't this default to backward ranges because you pass `false` for `isForward`?

src/core/updateSelections/updateSelections.ts
```diff
@@ -180,14 +206,45 @@ export async function performEditsAndUpdateSelections(
...
+    selectionInfoMatrix
+  );
+}
 
-  await performEditsAndUpdateFullSelectionInfos(
+export async function performEditsAndUpdateRanges(
+  rangeUpdater: RangeUpdater,
+  editor: TextEditor,
+  edits: Edit[],
+  originalSelections: (readonly Range[])[]
```
pokey, line 221:
```suggestion
  originalRanges: (readonly Range[])[]
```

src/core/updateSelections/updateSelections.ts
```diff
@@ -180,14 +206,45 @@ export async function performEditsAndUpdateSelections(
...
+) {
```
pokey, line 222: I'd add a return type so that the selections get cast to ranges, no?

pokey (commented):
Ok still haven't reviewed full PR but here's another round of comments

src/actions/EditNewLine.ts
```diff
@@ -41,9 +27,9 @@ class EditNewLine implements Action {
     return getNotebookFromCellDocument(editor.document) != null;
   }
 
-  private getCommand(target: TypedSelection) {
-    if (target.selectionContext.isNotebookCell) {
-      if (this.isNotebookEditor(target.selection.editor)) {
+  private getCommand(target: Target) {
+    if (target.scopeType === "notebookCell") {
```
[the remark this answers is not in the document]
AndreasArvidsson: Same user as above. Definitely

src/actions/Rewrap.ts
```diff
@@ -29,23 +24,28 @@ export default class Rewrap implements Action {
...
-        boundary: boundary.map((edge) =>
-          constructSimpleTypedSelection(target.selection.editor, edge)
+        editor: target.editor,
+        boundary: boundary.map(
+          (edge) =>
+            new BaseTarget({
+              editor: target.editor,
+              contentRange: edge,
+              isReversed: target.isReversed,
+            })
```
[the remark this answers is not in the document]
AndreasArvidsson: Turned out quite nicely didn't it? :)

src/core/commandVersionUpgrades/canonicalizeAndValidateCommand.ts
```diff
@@ -44,28 +69,30 @@ export function canonicalizeAndValidateCommand(
...
-  const extraArgs = inputExtraArgs;
+  while (command.version < LATEST_VERSION) {
+    switch (command.version) {
+      case 0:
+        command = upgradeV0ToV1(command);
+        break;
+      case 1:
+        command = upgradeV1ToV2(command);
+        break;
+    }
```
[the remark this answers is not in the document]
AndreasArvidsson: agreed

src/core/inferFullTargets.ts
```diff
@@ -20,23 +20,26 @@ import {
...
+  if (
+    actionPreferences != null &&
+    targets.length !== actionPreferences.length
+  ) {
     throw new Error("Target length is not equal to action preference length");
   }
 
   return targets.map((target, index) =>
-    inferTarget(target, targets.slice(0, index), actionPreferences[index])
+    inferTarget(target, targets.slice(0, index), actionPreferences?.at(index))
```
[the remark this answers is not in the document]
AndreasArvidsson: :)

src/core/commandVersionUpgrades/upgradeV1ToV2/upgradeV1ToV2.ts
```diff
@@ -0,0 +1,155 @@
...
+        modifiers.push({ type: "position", position: "end" });
+      } else {
+        modifiers.push({ type: "position", position: "after" });
+      }
+    }
+  }
+
+  if (isImplicit) {
+    modifiers.push({ type: "toRawSelection" });
+  }
```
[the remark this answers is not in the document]
AndreasArvidsson: You are technically correct. The very best type of correct there is. 
PS. To be read in a snooty british accent ;)

src/core/inferFullTargets.ts
AndreasArvidsson: Yes. They used to be a length check in there

src/core/inferFullTargets.ts
AndreasArvidsson: I had no idea either when I started to read the original inference code even though I wrote it initially. Comments on this can of things are quite importante

src/processTargets/PipelineStages.types.ts
AndreasArvidsson: true

src/core/updateSelections/updateSelections.ts
AndreasArvidsson: That target itself keeps track of the isForward and what is return here is just a pure range. Not the most prey of codes but I tried to reuse what was already there.

src/core/commandVersionUpgrades/upgradeV1ToV2/upgradeV1ToV2.ts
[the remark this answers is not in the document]
pokey: As of 13:30 BST today, I can now officially take offence to your PS 😄🇬🇧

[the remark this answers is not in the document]
AndreasArvidsson: :D

src/core/inferFullTargets.ts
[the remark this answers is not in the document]
AndreasArvidsson: Btw at can take negativ indices so you dont have to do length-1 kinda of thing anynore :)

src/test/suite/fixtures/recorded/actions/drinkArg.yml
```diff
@@ -0,0 +1,27 @@
...
+  marks: {}
+finalState:
+  documentContents: |
+    function helloWorld(foo: string, , bar: number, baz: string) {}
+  selections:
+    - anchor: {line: 0, character: 33}
+      active: {line: 0, character: 33}
+  thatMark:
+    - anchor: {line: 0, character: 35}
+      active: {line: 0, character: 46}
```
pokey, line 26: hmm a bit on the fence about the "that" mark here.  I'd be tempted to have it be the newly created arg, ie empty selection before the `, `, and in the future it would have the new `, ` in its removal range.  I think that would be consistent with how things work with eg `"bring to after"`: `"that"` refers to the newly created thing

src/test/suite/fixtures/recorded/selectionTypes/bringHarpToAfterFile.yml
```diff
@@ -0,0 +1,37 @@
+languageId: plaintext
```
pokey, line 1: isn't this one identical to `src/test/suite/fixtures/recorded/positions/bringHarpToAfterFile.yml`?

src/test/suite/fixtures/recorded/selectionTypes/bringWhaleToBeforeFile.yml
```diff
@@ -0,0 +1,37 @@
+languageId: plaintext
```
pokey, line 1: as above, identical to `src/test/suite/fixtures/recorded/positions/bringWhaleToBeforeFile.yml`?

src/test/suite/fixtures/recorded/surroundingPair/parseTreeParity/takePairDouble.yml
```diff
@@ -5,7 +5,7 @@ command:
   action: clearAndSetSelection
   targets:
     - type: primitive
-      modifier: {type: surroundingPair, delimiter: any, delimiterInclusion: excludeInterior}
+      modifier: {type: surroundingPair, delimiter: any}
```
pokey, line 8: for these tests, please don't change the command to match the file name: it should always be the other way around.  For example, this test was recorded back when "pair" meant "bounds", but the name of the test was never updated

src/test/suite/fixtures/recorded/surroundingPair/parseTreeParity/takePairRound.yml
```diff
@@ -5,7 +5,7 @@ command:
   action: clearAndSetSelection
   targets:
     - type: primitive
-      modifier: {type: surroundingPair, delimiter: parentheses, delimiterInclusion: excludeInterior}
+      modifier: {type: surroundingPair, delimiter: parentheses}
```
pokey, line 8: as above, please revert this one

src/test/suite/fixtures/recorded/surroundingPair/textual/takePairDouble.yml
```diff
@@ -5,7 +5,7 @@ command:
   action: clearAndSetSelection
   targets:
     - type: primitive
-      modifier: {type: surroundingPair, delimiter: any, delimiterInclusion: excludeInterior}
+      modifier: {type: surroundingPair, delimiter: any}
```
pokey, line 8: as above

src/test/suite/fixtures/recorded/surroundingPair/textual/takePairRound.yml
```diff
@@ -5,7 +5,7 @@ command:
   action: clearAndSetSelection
   targets:
     - type: primitive
-      modifier: {type: surroundingPair, delimiter: parentheses, delimiterInclusion: excludeInterior}
+      modifier: {type: surroundingPair, delimiter: parentheses}
```
pokey, line 8: as above

src/test/suite/runSingleRecordedTest.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Add the file path ending of a recorded test and ONLY that one will run
+ */
+
+// example:
+// const filenameEnd = "textual/takePairRound.yml";
+
+const filenameEnd = "";
```
pokey, line 8: This setup is begging for trouble imo.  If someone accidentally leaves a filename here and we miss it, they will have deactivated our entire test suite.  I'd prefer having a special npm command for this one. Can we not just use [Mocha's `--grep`](https://mochajs.org/#-grep-regexp-g-regexp) feature? We could also try adding a npm arg and then passing it via env var or something

src/util/filterDuplicates.ts
```diff
@@ -0,0 +1,8 @@
+import { isEqual } from "lodash";
+
+export function filterDuplicates<T>(elements: T[]) {
```
pokey, line 3: why not just use [`uniq`](https://lodash.com/docs/4.17.15#uniq)?

pokey (commented):
Some more questions / comments.  Lmk if you are about to make any changes as I'm planning to pull this one locally and make some changes

src/test/suite/fixtures/recorded/surroundingPair/parseTreeParity/takePairDouble.yml
AndreasArvidsson: Something is wrong on these tests. If we are referencing bounds why is there only one selection?

src/processTargets/modifiers/PositionStage.ts
```diff
@@ -7,66 +7,36 @@ export default class implements ModifierStage {
...
-      trailingDelimiter,
-    } = target;
+    const { leadingDelimiter, trailingDelimiter } = target;
 
-    const common = {
-      position: this.modifier.position,
-      editor,
-      isReversed,
-    };
+    target.position = this.modifier.position;
```
pokey, line 13: I think targets should be immutable.  I would prefer to have a function `withPosition(position)` and let it return a new target with the right changes

src/processTargets/targets/RawSelectionTarget.ts
```diff
@@ -12,4 +13,10 @@ export default class RawSelectionTarget extends BaseTarget {
     super(parameters);
     this.delimiter = undefined;
   }
+
+  getEditNewLineContext(_isBefore: boolean): EditNewLineContext {
+    return {
+      delimiter: "",
```
pokey, line 19: Assuming this one fixes "drink just vest"?  Worth adding a test

src/test/suite/fixtures/recorded/hatTokenMap/bringHarpAndPointToEndOfThisAndEndOfWhaleTakeWhale.yml
```diff
@@ -1,7 +1,7 @@
 languageId: plaintext
 command:
   version: 1
-  spokenForm: bring harp and point to end of this and end of whale take whale
+  spokenForm: bring harp and point to end of this and end of whale
```
pokey, line 4: These tests are a bit strange.  The final "take whale" is there to indicate which token we're testing

src/test/suite/fixtures/recorded/hatTokenMap/bringPointAfterFirstCarWhaleTakeWhale.yml
```diff
@@ -1,7 +1,7 @@
 languageId: plaintext
 command:
   version: 1
-  spokenForm: bring point after first car whale take whale
+  spokenForm: bring point after first car whale
```
pokey, line 4: as above re "take whale"

src/processTargets/modifiers/PositionStage.ts
AndreasArvidsson: I thought about it but in that case we need the exact same code in every single target. Duplication of code is not something I'm fond of

src/test/suite/fixtures/recorded/hatTokenMap/bringHarpAndPointToEndOfThisAndEndOfWhaleTakeWhale.yml
AndreasArvidsson: I think you gonna have to explain that further.

src/processTargets/modifiers/PositionStage.ts
pokey: Why? Can't we handle at least part of it in the base class? And if not can't we use composition? Anyways maybe let's just figure it out when we pair today

src/test/suite/fixtures/recorded/hatTokenMap/bringHarpAndPointToEndOfThisAndEndOfWhaleTakeWhale.yml
pokey: https://www.cursorless.org/docs/contributing/test-case-recorder/#navigation-map-tests

src/processTargets/modifiers/PositionStage.ts
AndreasArvidsson: If you wanted to return a new instance of the same derived class that has to be done in each class. But we could do `target.setPosition(position)` then all those changes take place inside the target at least

src/core/commandRunner/CommandRunner.ts
```diff
@@ -83,17 +80,20 @@ export default class CommandRunner {
...
+        this.graph.debug.log(JSON.stringify(targetDescriptors, null, 3));
       }
 
+      const finalStages =
+        action.getFinalStages != null
+          ? action.getFinalStages(...actionArgs)
+          : [];
+
       const processedTargetsContext: ProcessedTargetsContext = {
+        finalStages,
```
pokey, line 96: this feels like an odd place for this.  might be worth revisiting tomorrow

AndreasArvidsson: It was just a quick way to get it into process primitive target while we figured out the better solution. Which should probably just pass it as a separate argument like we did before with action preference.

pokey:
One thought: we may want to consider using composition when constructing targets. For example, the position modifier could construct a new target called `PositionalTarget` that keeps a reference to the original target, then does what it needs to lazily. Then `BaseTarget` doesn't need to worry bout that stuff

This pattern could let us be even lazier: just wrap the original target and do what we need to on the fly when client accessed attributes

AndreasArvidsson:
Sounds like a good approach

src/processTargets/targets/WeakTarget.ts
```diff
@@ -21,6 +22,12 @@ export default class WeakTarget extends BaseTarget {
     return true;
   }
 
+  getEditNewContext(_isBefore: boolean): EditNewContext {
+    return {
+      delimiter: "\n",
```
pokey, line 27: shouldn't we just use a final stage of weak line in `EditNew` to convert a weak target to a line?

AndreasArvidsson: Sounds good

src/processTargets/modifiers/WeakContainingScopeStage.ts
```diff
@@ -0,0 +1,16 @@
+import { ContainingScopeModifier, Target } from "../../typings/target.types";
+import { ProcessedTargetsContext } from "../../typings/Types";
+import getModifierStage from "../getModifierStage";
+import { ModifierStage } from "../PipelineStages.types";
+
+export default class implements ModifierStage {
+  constructor(private nestedModifier: ContainingScopeModifier) {}
```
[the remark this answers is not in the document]
pokey: also tempted to pass in the stage rather than the modifier 🤔

Will-Sommers:
Heyo — starting to look through this. Since its so large, would y'all be up for a adding a few comments on notable pieces to review?

pokey:
> Heyo — starting to look through this. Since its so large, would y'all be up for a adding a few comments on notable pieces to review? 

It's a complete rewrite of the way we handle targets and modifiers, so I'd say `processTargets` is prob the place to look first

src/processTargets/processTargets.ts
```diff
@@ -0,0 +1,260 @@
...
+
+function processPrimitiveTarget(
+  context: ProcessedTargetsContext,
+  target: PrimitiveTargetDescriptor
+): Target[] {
+  const markStage = getMarkStage(target.mark);
+  let targets = markStage.run(context);
+
+  const modifierStages = [
+    // Reverse target modifies because they are in reverse order from the api. Slice is needed to create a copy or the modifiers will be in wrong order in the test recorder.
```
Will-Sommers, line 242:
```suggestion
    // Reverse target modifiers because they are in reverse order from the api. Slice is needed to create a copy or the modifiers will be in wrong order in the test recorder.
```

src/processTargets/processTargets.ts
```diff
@@ -0,0 +1,260 @@
...
+    ...target.modifiers.slice().reverse().map(getModifierStage),
+    ...context.finalStages,
+  ];
+
+  modifierStages.forEach((stage) => {
+    const stageTargets: Target[] = [];
+    for (const target of targets) {
+      stageTargets.push(...stage.run(context, target));
```
Will-Sommers, line 250: Although we're not shadowing `target` from the function's arguments, the usage or `target` and `targets` and then finally `target` within this class is a little bit unclear to me

AndreasArvidsson: Basically all targets from the previous stage goes through the current stage. The current stage can return multiple targets for the single input target. Two boundaries for one position for example. At the end of it stage the target list is updated with the result of this current stage.

src/processTargets/processTargets.ts
```diff
@@ -0,0 +1,296 @@
...
+  let currentTargets = markOutputTargets;
+
+  // Then we apply each stage in sequence, letting each stage see the targets
+  // one-by-one and concatenating the results before passing them on to the
+  // next stage.
+  modifierStages.forEach((stage) => {
+    currentTargets = currentTargets.flatMap((target) =>
+      stage.run(context, target)
+    );
+  });
```
pokey, line 288: @Will-Sommers ok I cleaned up the code and added some comments / docstrings.  Does that help?

src/processTargets/targets/BaseTarget.ts
```diff
@@ -0,0 +1,183 @@
...
+    return {
+      type: "delimiter",
+      delimiter: this.delimiter ?? "",
+    };
+  }
+
+  getRemovalRange(): Range {
+    switch (this.position) {
+      case "before":
+        return this.leadingDelimiterRange ?? this.contentRange;
```
pokey, line 148: How come we fall back to content range here?

src/processTargets/targets/BaseTarget.ts
```diff
@@ -0,0 +1,183 @@
...
+      ? leadingDelimiterRange
+      : undefined;
+  }
+
+  get trailingDelimiterRange() {
+    const { includeDelimitersInRemoval, trailingDelimiterRange } =
+      getTokenDelimiters(this.state.editor, this.state.contentRange);
+    return includeDelimitersInRemoval || this.position != null
+      ? trailingDelimiterRange
+      : undefined;
```
pokey, line 128: It seems misleading to return `undefined` when `includeDelimitersInRemoval` is `true` when there actually is a leading delimiter. I think it would be cleaner to instead have `getTokenDelimiters` just return `removalRange`.  I would make the change myself, but I'm concerned about how it will interact with `ContinuousRangeTarget`.

src/processTargets/targets/BaseTarget.ts
```diff
@@ -0,0 +1,183 @@
...
+    if (this.delimiter === "\n" && !isBefore) {
+      return { type: "command", command: "editor.action.insertLineAfter" };
+    }
+    return {
+      type: "delimiter",
+      delimiter: this.delimiter ?? "",
+    };
+  }
+
+  getRemovalRange(): Range {
```
pokey, line 145: I'd opt for a property `get removalRange()`, no?

src/processTargets/targets/BaseTarget.ts
```diff
@@ -0,0 +1,183 @@
...
+  get position() {
+    return this.state.position;
+  }
+  get editor() {
+    return this.state.editor;
+  }
+  get isReversed() {
+    return this.state.isReversed;
+  }
+  get contentRemovalRange() {
```
pokey, line 58: Hmm.  This property name is a bit confusing.  It's not immediately clear how it differs from `removalRange`.  It really feels like it should be private, as it's kind of an implementation detail.

I believe it's only used by `ContinuousRangeTarget`, no? Kind of makes me wonder whether all targets should actually be ranges, with today's targets just being the special case where the range consists of the same start and end scope. Although I believe some of our primitive targets already are ranges today, because eg a LineTarget is really a range of lines, right?  If so, then Targets could just implement a function called `makeRange(active, includeAnchor, includeActive)` which would just return another Target of the same type

Fwiw I'm happy for the target to downgrade itself when you form a range if the argument is not of the same type.  Eg if you say "take line air past token bat", it should downgrade to a WeakTarget.

pokey: now that everything is lazy, couldn't you just union the content ranges in the continuous range stage and return a new target?

pokey: fwiw when we fall back to weaktarget, I _think_ I would be ok with discarding the leading / trailing delimiters from the anchor / active and just using token delimiters provided by weaktarget if it simplifies things a bunch.  Eg "chuck line air past token bat" could just convert to a token range from start of line air through end of token bat

src/processTargets/targets/BaseTarget.ts
AndreasArvidsson: Is just because the removal action requires a target and in this case the content range will always be empty. This basically says that if we have a leading the limited range remove that or remove an empty range ie do nothing. We should probably make the method return undefined instead.

src/processTargets/targets/BaseTarget.ts
AndreasArvidsson: The method is specifically trailing range and should return undefined if non exists/should be used. 
This does indeed have ramifications for continuous range target and is why we have to keep those ranges separate.

src/processTargets/targets/BaseTarget.ts
pokey: Ah right. So it's a way of returning empty range. Maybe just add a comment for now

src/processTargets/targets/BaseTarget.ts
AndreasArvidsson: I really prefer proper methods for things that are not attributes but logic. I'm really unsure when to use what.

src/processTargets/targets/BaseTarget.ts
AndreasArvidsson: We still have syntax scope target that needs to transfer their delimiters to a new target and can't be done lazily today. 

This question is a bit too complex I think we need to have a discord about it

pokey: Yeah let's discuss over discord some time today

For syntax node, that could happen in its makeRange, because it's a lot less complex I think than what's happening here to account for lines

pokey: In case not clear from our discussion, I'd argue that eg for `SyntaxNodeTarget`, it should always have a `startNode`and `endNode`, and it's `makeContinuousRangeWith` function should just return a new `SyntaxNodeTarget` with start node from start and end node from end

pokey: Also, your unifyRemovalRanges could _potentially_ just use the new make range as well

pokey:
Also in case not clear from discussion, I think get insertionEdit (or whatever) could replace editNewContext. But if that's too painful can leave for follow-up

pokey:
Btw if user wants finer grained control over delimiters, they can say eg "chuck just air past trailing line bat"

src/processTargets/processTargets.ts
Will-Sommers: @pokey — definitely. Thanks!

src/test/suite/runSingleRecordedTest.ts
Will-Sommers: Any of these options would be super helpful from a contributor's perspective. The suite runs fast but not fast enough for a good feedback loop.

cursorless-talon/src/actions/call.py
```diff
@@ -6,5 +6,5 @@
 
 
 def run_call_action(target: dict):
-    targets = [target, IMPLICIT_TARGET]
+    targets = [target, IMPLICIT_TARGET.copy()]
```
pokey, line 9: should we just make a function that returns a new implicit target every time instead?  This copy stuff is a bit of a nightmare to have to remember

cursorless-talon/src/compound_targets.py
```diff
@@ -44,19 +44,19 @@ def cursorless_range(m) -> str:
         return primitive_targets[0]
 
     if len(primitive_targets) == 1:
-        start = BASE_TARGET.copy()
+        anchor = BASE_TARGET.copy()
```
pokey, line 47: same with base target; we should prob just have a function that constructs one

cursorless-talon/src/actions/call.py
AndreasArvidsson: Yes do that

cursorless-talon/src/compound_targets.py
AndreasArvidsson: Yes do that

src/processTargets/targets/TokenTarget.ts
```diff
@@ -10,6 +10,7 @@ import BaseTarget from "./BaseTarget";
 export default class TokenTarget extends BaseTarget {
   delimiterString = " ";
 
+  /** */
```
pokey, line 13: nice docstring 😄

AndreasArvidsson: :p

cursorless-talon/src/compound_targets.py
AndreasArvidsson: https://github.com/cursorless-dev/cursorless/issues/719

cursorless-talon/src/actions/call.py
AndreasArvidsson: https://github.com/cursorless-dev/cursorless/issues/719

docs/contributing/CONTRIBUTING.md
```diff
@@ -34,6 +34,10 @@ locally, you need to run the extension in debug mode. To do so you need to run
 the `workbench.action.debug.selectandstart` command in VSCode and then select either "Run
 Extension" or "Extension Tests".
 
+### Running a single test
+
+The entire test suite takes a little while to run (1-2 mins), so if you'd like to repeat a single test, you can edit [`runSingleRecordedTest`](../../src/test/suite/runSingleRecordedTest.ts) to refer to the desired test and use the "Run Single Extension Test" launch config instead of the usual "Extension Tests".
```
Will-Sommers, line 39: 🙌

cursorless-talon/src/csv_overrides.py
```diff
@@ -159,6 +159,14 @@ def update_dicts(
         key = obj["key"]
         if not is_removed(key):
             for k in key.split("|"):
+                if value == "pasteFromClipboard" and k.endswith(" to"):
+                    # FIXME: This is a hack to work around the fact that the
+                    # spoken form of the `pasteFromClipboard` action used to be
+                    # "paste to", but now the spoken form is just "paste" and
+                    # the two is part of the positional target. Users who had
```
pokey, line 166:
```suggestion
                    # the "to" is part of the positional target. Users who had
```

cursorless-talon/src/modifiers/delimiter_range.py
```diff
@@ -0,0 +1,23 @@
+from typing import Any
+
+from talon import Context, Module
+
+mod = Module()
+ctx = Context()
+
+
+delimiter_ranges = {
```
pokey, line 9: These should be overridable by csv, no?

cursorless-talon/src/modifiers/delimiter_range.py
```diff
@@ -0,0 +1,23 @@
...
+}
+
+mod.list("cursorless_delimiter_range", desc="Types of delimiter ranges")
+ctx.lists["self.cursorless_delimiter_range"] = delimiter_ranges.keys()
+
+
+@mod.capture(rule="{user.cursorless_delimiter_range}")
+def cursorless_delimiter_range(m) -> dict[str, Any]:
+    return {
+        "type": "delimiterRange",
```
pokey, line 21: Not in love with this identifier. "Range" is a bit confusing. Tbh I'd be tempted to just have two separate types `leadingDelimiter` and `trailingDelimiter`. Unless we can come up with a better name here

cursorless-talon/src/modifiers/delimiter_range.py
AndreasArvidsson: Since the positions are not in csv yet either I think we should do both or neither.

pokey: I'm a lot more confident about these. What do you expect to change here? These are fairly straightforward modifiers

cursorless-talon/src/modifiers/delimiter_range.py
AndreasArvidsson: Since they are handled by the same modifier stage I see no point in splitting them into two different types. I think delimiter range is clear enough. Is either the leading delimiter range or the trailing delimiter range that is the nomenclature we are using all over the place in the extension isn't it?

cursorless-talon/src/positional_target.py
```diff
@@ -0,0 +1,56 @@
...
+    "cursorless_positional_connective",
+    desc="The connective used to separate source and destination targets",
+)
+
+positional_connectives = {
+    **POSITION_BEFORE,
+    **POSITION_AFTER,
+}
+
+print(positional_connectives)
```
pokey, line 18:
```suggestion
```

cursorless-talon/src/positional_target.py
```diff
@@ -0,0 +1,56 @@
+from talon import Context, Module
+
+from .modifiers.position import POSITION_AFTER, POSITION_BEFORE
+
+mod = Module()
+ctx = Context()
+
+mod.list(
+    "cursorless_positional_connective",
+    desc="The connective used to separate source and destination targets",
```
pokey, line 10:
```suggestion
    desc='Connectives used to separate source and destination targets that indicate position, eg "before" or "after"',
```

cursorless-talon/src/modifiers/delimiter_range.py
AndreasArvidsson: Nothing really but I think all of them should go in a position.csv file

cursorless-talon/src/primitive_target.py
```diff
@@ -1,20 +1,22 @@
...
     "<user.cursorless_position>",  # before, end of
-    "<user.cursorless_selection_type>",  # token, line, file
     "<user.cursorless_head_tail>",  # head, tail
     "<user.cursorless_containing_scope>",  # funk, state, class
     "<user.cursorless_subtoken_scope>",  # first past second word
     "<user.cursorless_surrounding_pair>",  # matching/pair [curly, round]
     "<user.cursorless_to_raw_selection>",  # just
-    # "<user.cursorless_matching_paired_delimiter>",  # matching
+    "<user.cursorless_delimiter_inclusion>",  # inside, bounds
+    "<user.cursorless_delimiter_range>",  # leading, trailing
```
pokey, line 19: It is a bit funny that we use "delimiter" for both of these things, but we treat them completely differently 🤔.  Might be a bit late to try to tease these apart, as I believe some of this terminology is already baked into the csv.  Prob worth a quick chat today

cursorless-talon/src/actions/actions_makeshift.py
```diff
@@ -39,7 +39,7 @@ class MakeshiftAction:
         "editor.action.rename",
         restore_selection=True,
         await_command=False,
-        post_command_sleep_ms=150,
+        post_command_sleep_ms=200,
```
pokey, line 42: reminder to self to try this out locally

src/actions/Clear.ts
```diff
@@ -1,26 +1,20 @@
...
-  ];
-
   constructor(private graph: Graph) {
     this.run = this.run.bind(this);
   }
 
-  async run([targets]: [TypedSelection[]]): Promise<ActionReturnValue> {
+  async run([targets]: [Target[]]): Promise<ActionReturnValue> {
     const editor = ensureSingleEditor(targets);
+    const rawTargets = targets.map(toRawTarget);
```
pokey, line 15: We probably don't want raw targets, right?  We might want to remove that utility function as well so people aren't tempted to use raw targets for this kinda thing

cursorless-talon/src/primitive_target.py
AndreasArvidsson: We can definitely have a look at it

package.json
```diff
@@ -7,7 +7,7 @@
 		"color": "#00001A",
 		"theme": "dark"
 	},
-	"version": "0.25.0",
+	"version": "0.26.0-SNAPSHOT",
```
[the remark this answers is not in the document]
AndreasArvidsson: This seems to help which running the extension locally without being override from the marketplace. Just removing the snapshot when we are ready for release seems like a good way to handle this problem?

cursorless-talon/src/actions/actions_makeshift.py
AndreasArvidsson: 200 ms seems to work fine for me atleast

src/actions/Clear.ts
AndreasArvidsson: Why not? What matters here is that we don't have any removal delimiters. You can definitely switch it to a plain target if you prefer that.

docs/contributing/CONTRIBUTING.md
AndreasArvidsson: This was so easy to add and I don't know why we didn't do it months ago  :D

cursorless-talon/src/positional_target.py
```diff
@@ -0,0 +1,53 @@
...
+
+mod.list(
+    "cursorless_positional_connective",
+    desc='Connectives used to separate source and destination targets that indicate position, eg "before" or "after"',
+)
+
+ctx.lists["self.cursorless_positional_connective"] = {
+    "before": "before",
+    "after": "after",
+}
```
pokey, line 16: I moved these positional connectives here, because in my mind it's a bit of a hack that we allow them as modifiers; to me they are more properly considered to be connectives.  Open to tweaking if you think we should migrate them back to the other file; shouldn't be too big of a change

cursorless-talon/src/modifiers/position.py
```diff
@@ -1,23 +1,30 @@
+from typing import Any
```
pokey, line 1: Worth a look at this file; a bit cleaner now I think

AndreasArvidsson: If we are going to remove before and after I definitely think that we should add positions to csv file so user can customize them because start of an end of is to verbose. I also think that we should disable chuck after since if we're going to do this puristicly we're going to do it all the way.

cursorless-talon/src/positional_target.py
AndreasArvidsson: If we are going to remove before and after I definitely think that we should add positions to csv file so user can customize them because start of an end of is to verbose. I also think that we should disable chuck after since if we're going to do this puristicly we're going to do it all the way.

src/actions/CommandAction.ts
```diff
@@ -33,59 +30,49 @@ const defaultOptions: CommandOptions = {
...
-            editor.selections = updatedOriginalSelections;
-          }
-
-          return updatedTargetSelections.map((selection) => ({
-            editor,
-            selection,
-          }));
+        // Reset original selections
+        if (options.restoreSelection) {
+          setSelectionsWithoutFocusingEditor(editor, updatedOriginalSelections);
```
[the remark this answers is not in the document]
AndreasArvidsson: This code runs for each editor and we can't have focus on multiple editor that's why at the end we only restore selection to a single editor if needed.

pokey:
Just had a thought that we may want to reintroduce "chuck after" => "chuck trailing" in upgrade code. Otherwise users who haven't upgraded talon will see error message about using "trailing", but not yet have that command

AndreasArvidsson:
> Just had a thought that we may want to reintroduce "chuck after" => "chuck trailing" in upgrade code. Otherwise users who haven't upgraded talon will see error message about using "trailing", but not yet have that command

Readded

PR #676, opened 2022-05-14
Add condition scope type to C#

### LINKED ISSUE #674: C# does not yet support the 'condition' scope type
It may be as simple as using the pattern matcher for 'conditional_expression'

AndrewDant:
Closes #674
## Checklist

- [ X] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)


auscompgeek (commented):
Could you also add tests where the condition is:

- a simple value, e.g. `if (false) {}`
- includes logical not
- is a method call

AndrewDant:
@auscompgeek I've implemented the first two tests you asked for. Can you give me an example of which you want for a method call? I don't think the condition type matches method call arguments in existing implementations.

auscompgeek:
> I don't think the condition type matches method call arguments in existing implementations.

Huh, that sounds like a bug to me. Should fix that later if that's the case.

auscompgeek (approved):
LGTM

pokey:
> @auscompgeek I've implemented the first two tests you asked for. Can you give me an example of which you want for a method call? I don't think the condition type matches method call arguments in existing implementations.

I would do `if (foo(bar)) {}` => `if () {}`. Does that not work?

src/languages/csharp.ts
```diff
@@ -218,6 +220,10 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
   ifStatement: "if_statement",
   class: "class_declaration",
   className: "class_declaration[name]",
+  condition: cascadingMatcher(
+    conditionMatcher("*[condition]"),
+    patternMatcher("~binary_expression.binary_expression!")
```
pokey, line 225: This seems odd. Wouldn't this match eg the right hand side of `a = b + c`?

pokey (changes requested):

AndrewDant:
> I would do `if (foo(bar)) {}` => `if () {}`. Does that not work?

Yes that does work and I've added a test for it. I thought he meant a condition passed as an argument into a function.

src/languages/csharp.ts
auscompgeek: Ah, interesting. Could you add equivalent tests to the ones I requested above for while loops too?

AndrewDant:
@pokey as we discussed at the meetup today please update vscode-parse-tree to look at a commit on master for c# tree sitter instead of version 0.19.0, in order to fix the issue where the condition field is not recognized inside of while statements.

AndrewDant:
@pokey I think this is ready to go now. Let me know if you see any other issues.

src/languages/python.ts
```diff
@@ -101,6 +101,7 @@ const nodeMatchers: Partial<Record<ScopeType, NodeMatcherAlternative>> = {
     argumentMatcher("parameters", "argument_list"),
     matcher(patternFinder("call.generator_expression!"), childRangeSelector())
   ),
+  callee: "call.identifier!",
```
pokey, line 104: shouldn't this one be in #664?

src/typings/Types.ts
```diff
@@ -96,6 +96,7 @@ export type ScopeType =
   | "argumentOrParameter"
   | "anonymousFunction"
   | "attribute"
+  | "callee"
```
pokey, line 99: shouldn't this one be in #664 ?

src/test/suite/fixtures/recorded/languages/csharp/changeCondition2.yml
```diff
@@ -0,0 +1,23 @@
+languageId: csharp
```
pokey, line 1: would be good to also have a for-loop test where selection is in body of loop

AndrewDant: I've added two.

pokey (approved):
looks good!

PR #680, opened 2022-05-19
docusaurus.config.js: Remove trailing slash on url

### LINKED ISSUE #668: Figure out where extra `/` comes from when sharing link from chrome on mobile
# Steps
1. Go to cursorless.org in chrome on iOS
2. Click on custom actions in sidebar
3. Click the share button in the URL bar

# Expected behaviour
Shared URL is `https://www.cursorless.org/docs/user/customization/`

# Actual behaviour

URL is `https://www.cursorless.org//docs/user/customization/`. Note the `//` after the domain name. This seems to break things in Slack internal preview

auscompgeek:
Should fix #668. I haven't tested this yet.

## Checklist

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -61,7 +61,7 @@ function remarkPluginFixLinksToRepositoryArtifacts() {
 const config = {
   title: "Cursorless",
   tagline: "Structural voice coding at the speed of thought",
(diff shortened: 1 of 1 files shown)


pokey (approved):
Just tested locally and seems to fix the issue.  Thanks!

PR #684, opened 2022-05-19
Support delimiter based edit newline

### LINKED ISSUE #242: Allow "clone", "pour" etc to operate on other scopes

AndreasArvidsson:
In regards too `drink/pour`

Instead of relying on the built in vscode commands for inserting new lines we can now insert our own delimiters. 

`drink arg`, `pour funk` etc is now working. Some of the scope types don't have proper delimiters but as soon as we add that to the scope type definition they will automatically work with this action. 

1. Should we rename the action now that we can insert before arbitrary delimiters and not just lines?
2. The default implementation right now is to insert a new line delimiter. If we still want to use vscode command as default just change `getEditNewLineContext` in `BaseTarget.ts` to:
```
return {
  command: this.isBefore
    ? "editor.action.insertLineBefore"
    : "editor.action.insertLineAfter",
};
```

Fixes #242 

## Checklist

- [ ] Update existing tests

### DIFF
--- a/src/actions/EditNewLine.ts
+++ b/src/actions/EditNewLine.ts
@@ -11,7 +11,7 @@ import { createThatMark, ensureSingleEditor } from "../util/targetUtils";
 import { Action, ActionReturnValue } from "./actions.types";
 
 class EditNewLine implements Action {
-  constructor(private graph: Graph, private isAbove: boolean) {
+  constructor(private graph: Graph, private isBefore: boolean) {
     this.run = this.run.bind(this);
   }
 
@@ -20,7 +20,7 @@ class EditNewLine implements Action {
 
     const targetsWithContext = targets.map((target) => ({
       target,
-      context: target.getEditNewLineContext(this.isAbove),
+      context: target.getEditNewLineContext(this.isBefore),
     }));
     const commandTargets = targetsWithContext.filter(
       ({ context }) => !!(<any>context).command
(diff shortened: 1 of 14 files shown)


src/actions/EditNewLine.ts
```diff
@@ -1,66 +1,139 @@
...
-    if (this.isAbove) {
+      // Delimiter is one or more new lines. Handle as lines.
+      if (delimiter.includes("\n")) {
+        const lineNumber = this.isBefore
+          ? contentRange.start.line
+          : contentRange.end.line;
+        const line = editor.document.lineAt(lineNumber);
+        const characterIndex = line.isEmptyOrWhitespace
+          ? contentRange.start.character
+          : line.firstNonWhitespaceCharacterIndex;
```
pokey, line 64: This will always just be same indentation as previous line, right?  The built-in VSCode indentation code will automatically indent if, for example, you're on the guard of an `if` statement.  I don't think we can recover this behaviour without running the VSCode command or doing something like https://github.com/cursorless-dev/cursorless/issues/685

pokey (changes requested):
See below.  I think we could prob treat lines similar to notebook cells to fix the problem?

src/actions/EditNewLine.ts
AndreasArvidsson: Correct. Look in comment 2. We could just default to existing vscode command and only use delimiter for other than single line if you like.

pokey: I think I would just override the context in the Line class, the way you've done for notebook cells

PR #701, opened 2022-05-26
Exclude vscode sandbox

AndreasArvidsson:
## Checklist


.vscode/settings.json
```diff
@@ -1,16 +1,15 @@
 // Place your settings in this file to overwrite default and user settings.
 {
+	// Exclude the following files from explorer and search results
 	"files.exclude": {
-		"out": false // set this to true to hide the "out" folder with the compiled JS files
```
pokey, line 4: I would be inclined to leave this. I think we should only hide things from explorer if we have a really good reason to do so. They may want to browse the out folder to see the generated files

pokey (commented):
Looks good with one minor comment

pokey (approved):

PR #705, opened 2022-05-28
Allow for assertions against intentional non-match

Will-Sommers:
# What?

This PR adds the ability to capture errors and non-matches from issued commands in the integration tests.

# Why?

I wasn't able to do this as part of the TS-Scheme query PR and so added unit tests. I think it would be nice to have both. Non-matches help describe a complete spec of how expected behavior and unexpected behavior should be handled.

# How?

- Add a new VSCode command to capture error cases. This is "off" by default.
- Note if an error is thrown and pass it to the TestCaseRecorder
- Do not record `finalState` instead record `errorReturned` as a boolean
- If `errorReturned` is present on the fixture, ensure we throw 

# Todo
- Make a new test folder to assert on these errors
- Make `errorReturned` richer? I don't want to assert on an error message but I wonder if there is information that we can record to make the test cases more valuable.


package.json
```diff
@@ -63,6 +63,10 @@
 			{
 				"command": "cursorless.resumeRecording",
 				"title": "Cursorless: Resume test case recording"
+			},
+			{
+				"command": "cursorless.captureErrorTests",
```
pokey, line 68: Prob better to use existing test case recorder command and just add a field to its [config object](https://github.com/cursorless-dev/cursorless/blob/972cab15940ab20f3dcfb7205a10e01cc76d6edb/src/testUtil/TestCaseRecorder.ts#L19-L50).  See [the docs](https://www.cursorless.org/docs/contributing/test-case-recorder/#initial-setup) for how to use these fields from talon side (step 2)

src/core/commandRunner/CommandRunner.ts
```diff
@@ -137,6 +137,9 @@ export default class CommandRunner {
 
       return returnValue;
     } catch (e) {
+      if (this.graph.testCaseRecorder.captureErrors()) {
+        await this.graph.testCaseRecorder.postCommandHook(e);
+      }
       this.graph.testCaseRecorder.commandErrorHook();
```
pokey, line 143: couldn't you just do the if statement within the `commandErrorHook`?

src/testUtil/TestCase.ts
```diff
@@ -36,17 +36,19 @@ export type TestCaseFixture = {
   marksToCheck?: string[];
 
   initialState: TestCaseSnapshot;
-  finalState: TestCaseSnapshot;
+  finalState: TestCaseSnapshot | null;
```
pokey, line 39: might be worth a quick docstring saying this one is null in case of error

src/testUtil/TestCase.ts
```diff
@@ -36,17 +36,19 @@ export type TestCaseFixture = {
...
   returnValue: unknown;
   /** Inferred full targets added for context; not currently used in testing */
   fullTargets: Target[];
+  errorReturned: boolean;
```
pokey, line 43: I would capture name of error, and then check for that in test.  Can use `(err as Error).name`.  I think I'd have this field be `thrownError: ErrorInfo`, and to just have `name: string` field on the `ErrorInfo` type for now

src/testUtil/TestCase.ts
```diff
@@ -36,17 +36,19 @@ export type TestCaseFixture = {
...
 };
 
 export class TestCase {
   languageId: string;
   fullTargets: Target[];
   initialState: TestCaseSnapshot | null = null;
   finalState: TestCaseSnapshot | null = null;
+  errorState: Error | null = null;
```
pokey, line 51: I think I'd call this one `thrownError`

src/testUtil/TestCase.ts
```diff
@@ -131,9 +133,13 @@ export class TestCase {
     );
   }
 
-  toYaml() {
-    if (this.initialState == null || this.finalState == null) {
+  toYaml(captureErrors: boolean) {
+    if (
+      this.initialState == null ||
+      (this.finalState == null && !captureErrors)
```
pokey, line 139:
```suggestion
      (this.finalState == null && this.thrownError == null)
```

src/testUtil/TestCase.ts
```diff
@@ -131,9 +133,13 @@ export class TestCase {
...
+    ) {
       throw Error("Two snapshots must be taken before serializing");
+    } else {
```
pokey, line 142: what's the purpose of this branch?
```suggestion
```

src/testUtil/TestCase.ts
```diff
@@ -160,6 +167,11 @@ export class TestCase {
   }
 
   async recordFinalState(returnValue: unknown) {
+    if (returnValue instanceof Error) {
```
pokey, line 170: I'd just record this one in `commandErrorHook`, as mentioned above.  Then you don't need to abuse `returnValue`

src/testUtil/TestCaseRecorder.ts
```diff
@@ -243,15 +261,14 @@ export class TestCaseRecorder {
   }
 
   async postCommandHook(returnValue: any) {
-    if (this.testCase == null) {
+    if (this.testCase == null && !this.captureErrorTests) {
```
pokey, line 264: As implied above, I'd just let error handler do an `await this.finishTestCase();`, then this function can remain unchanged i believe

pokey (commented):
Great stuff!  Will be really nice to be able to record this type of test.  Left a bunch of minor comments

- [ ] Don't forget to add note [to docs](https://www.cursorless.org/docs/contributing/test-case-recorder/#initial-setup)

src/test/suite/fixtures/recorded/nonMatchErrors/takeList.yml
```diff
@@ -0,0 +1,19 @@
...
+  spokenForm: take list
+  action: setSelection
+  targets:
+    - type: primitive
+      modifier: {type: containingScope, scopeType: list, includeSiblings: false}
+initialState:
+  documentContents: |2
+      [1, 2, 3, [4, 5, 6]] 
+  selections:
+    - anchor: {line: 0, character: 0}
```
Will-Sommers, line 13: Test that we aren't expanding on whitespace to match the list.

src/test/suite/fixtures/recorded/nonMatchErrors/takeFunk.yml
```diff
@@ -0,0 +1,19 @@
+languageId: ruby
+command:
+  version: 1
+  spokenForm: take funk
```
Will-Sommers, line 4: ensure we aren't matching on the incorrect scope type.

src/testUtil/TestCase.ts
```diff
@@ -36,7 +40,10 @@ export type TestCaseFixture = {
   marksToCheck?: string[];
 
   initialState: TestCaseSnapshot;
-  finalState: TestCaseSnapshot;
+  /** The final state after a command is issued. If we are testing a non-match(error) case this is null. */
+  finalState: TestCaseSnapshot | null;
```
pokey, line 44: i'd be tempted to just use a `?`.  How come you've used `| null` here?

src/testUtil/TestCaseRecorder.ts
```diff
@@ -47,6 +47,12 @@ interface RecordTestCaseCommandArg {
    * Whether to flash a background for calibrating a video recording
    */
   showCalibrationDisplay?: boolean;
+
+  /**
+   * Whether we should record a test which yields an error
+   */
+
+  isErrorTest?: boolean;
```
pokey, line 55: I'd be tempted to call this one `recordErrors`, and update the docstring, because the user could start recording using this option, and then record some regular tests in addition to some error tests.  I could see that being useful for capturing where a scope type should and shouldn't match

Also, we usually don't include empty line between field and docstring

src/testUtil/TestCaseRecorder.ts
```diff
@@ -248,10 +257,9 @@ export class TestCaseRecorder {
       // command for a navigation map test
       return;
     }
+    await this.testCase!.recordFinalState(returnValue);
```
pokey, line 260: why do we need the `!` now?

docs/contributing/test-case-recorder.md
```diff
@@ -13,13 +13,19 @@ command run, and the final state, all in the form of a yaml document. See
...
 
     - https://github.com/pokey/pokey_talon/blob/9298c25dd6d28fd9fcf5ed39f305bc6b93e5f229/apps/vscode/vscode.talon#L468
     - https://github.com/pokey/pokey_talon/blob/49643bfa8f62cbec18b5ddad1658f5a28785eb01/apps/vscode/vscode.py#L203-L205
 
     It is quite unlikely you'll need this second step. Most tests don't check the navigation map.
 
+1.  If you'd like to be able to record tests which assert on non-matches, please add another command to your personal talon files. See the two files links above for context. Add the command below to your to your `vscode.py` and ensure that there is a matching Talon command.
+
+```
+  actions.user.vscode_with_plugin("cursorless.recordTestCase", {"isErrorTest": True})
```
pokey, line 26: don't forget to change `isErrorTest` if we change the name as suggested

pokey (commented):
Almost there!  Left another round of feedback

src/testUtil/TestCaseRecorder.ts
Will-Sommers: Got it, fixing. It would be cool to add this to our linters if possible. I guess I had a non-complete `if` check here prior and this was required.

pokey: Yeah good idea. Looks like [this](https://typescript-eslint.io/rules/no-unnecessary-type-assertion) is the rule? Odd thing is it appears that one should be on by default 🤔

src/testUtil/TestCase.ts
```diff
@@ -36,7 +40,10 @@ export type TestCaseFixture = {
   marksToCheck?: string[];
 
   initialState: TestCaseSnapshot;
-  finalState: TestCaseSnapshot;
+  /** The final state after a command is issued. Undefined if we are testing a non-match(error) case. */
+  finalState?: TestCaseSnapshot;
+  /** Used to assert if an error has been thrown. */
+  errorName?: ThrownError;
```
pokey, line 46: rename to `thrownError`

src/testUtil/TestCaseRecorder.ts
```diff
@@ -248,7 +257,6 @@ export class TestCaseRecorder {
       // command for a navigation map test
       return;
     }
-
```
pokey, line 251: re-add

pokey (approved):

pokey (approved):

PR #708, opened 2022-05-31
Implemented head tail modifier that works on multiple scope types

AndreasArvidsson:
`take head funk`
`clear tail pair`

## Checklist

### DIFF
--- a/src/processTargets/modifiers/HeadTailStage.ts
+++ b/src/processTargets/modifiers/HeadTailStage.ts
@@ -3,16 +3,35 @@ import { HeadModifier, TailModifier, Target } from "../../typings/target.types";
 import { ProcessedTargetsContext } from "../../typings/Types";
 import { ModifierStage } from "../PipelineStages.types";
 import TokenTarget from "../targets/TokenTarget";
+import WeakTarget from "../targets/WeakTarget";
+import { toLineTarget } from "./scopeTypeStages/LineStage";
 
 abstract class HeadTailStage implements ModifierStage {
-  abstract update(editor: TextEditor, range: Range): Range;
+  abstract update(
+    editor: TextEditor,
+    previousRange: Range,
+    nextRange: Range
+  ): Range;
 
   constructor(private isReversed: boolean) {}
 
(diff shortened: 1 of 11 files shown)


pokey:
Performed git surgery.  Original branch at https://github.com/cursorless-dev/cursorless/tree/bak/headTail; feel free to delete if this branch looks ok

pokey (commented):
Looks like merge caused test failure; I'll hold off reviewing until you get a chance to take a look

pokey (commented):
I'm not in love with having to modify all the targets for this one.  I think it would be pretty clean to have a capture `(head | tail) <modifier>?` that can accept any modifier, and then constructs a new modifier that runs its input modifier and then takes the head.  I think we could do that without having to touch any files

AndreasArvidsson:
The problem with that approach is that we want to be able to say things like `head that` or `tail air`. My point is that it's not just modifiers that can be proceeded by head or tail. We even have the implicit one `take head`
Another option is that instead of storing it in the targets themselves the pipeline could have that information in the context maybe?

pokey:
No the trailing modifier is optional, and defaults to "line". So "take head" will still be valid

AndreasArvidsson:
Okay that is what you mean with the question mark.

pokey:
Right sorry yeah couldn't remember the syntax without a talon file in front of me so just fell back to regex 😅

AndreasArvidsson:
No problem I understood that as soon as I grasped your intention :)
I will give it a try!

pokey:
The one thing to think about is what happens if they say something like "head every key". We may want to infer "head line every key"?

AndreasArvidsson:
Maybe. Maybe we should leave that one for now and try it out first?

AndreasArvidsson:
@pokey This implementation I think is more to your liking :)

AndreasArvidsson:
By the way now we are actually free with the ordering. If we want we could have `take funk head` instead of `take head funk`

pokey (approved):

PR #709, opened 2022-05-31
Added text based item scope

### LINKED ISSUE #294: `item` selects child instead of correct item in typescript
In the following code:

```typescript
[
  {
    foo: "bar",
  },
]
```

If we say "take item lace", it will select `foo: "bar"` instead of the containing map, which is the item we're actually after.

### LINKED ISSUE #357: Support generic comma-separated lists
### High-level

We'd like to support arbitrary list-like syntactic structures which consist of opening delimiters, closing delimiters, and separators.  Examples include the following:

- Argument lists
- List literals
- Dictionary literals
- Enclosed import lists
- Template parameters
- Probably lots of other things we haven't thought of, but that we can get for free

We'd like to use the same / similar code-path as surrounding pairs, so that it leverages parse tree if it has it, but works on text if it doesn't, and works generically without needing to manually specify acceptable tree-sitter types.

The idea would be that it would select anything between an opening / closing delimiter and a comma, as long as that comma isn't nested between delimiters not containing the input.  For example

```typescript
foo(bar(baz, bongo), hello)
```

If we said `"take item risk"`, it should skip over the first comma because it is nested

### Algorithm

**Note**: This algorithm could likely be simplified a lot if we decide to go with #484

- Add a new delimiter `side` called `"separator"`
- Potentially also need to have delimiters that create a level of nesting, so we ignore separators between them, but that shouldn't count as a container for a single item.  Good example would be `"`: should ignore commas between the quotes but not return its contents as an item, in contrast to parentheses, where if we found no commas between the parentheses, we'd just yield between them as a single-item list

#### In `generateUnmatchedDelimiters`
- we yield separators if and only if we're at top-level, ie `sum(Object.values(delimiterBalances)) === 0`.  Otherwise we just ignore them

#### In `findDelimiterPairContainingSelection`
- After rightward pass, we look to see whether the yielded delimiter is a separator or a closing delimiter
- If right delimiter is a closing delimiter, then:
  - Sweep leftwards looking for either the matching left delimiter or any separator.  We need the `generateUnmatchedDelimiters` function to keep track of all delimiter pairs even though it's only looking for the matching left delimiter.  This way we properly ignore nested separators.  Could do this either by having a `trackedDelimiters` arg that contains all the delimiters, or by just looping in `findDelimiterPairContainingSelection` until we get a matching delimiter.  
  - If we hit either the opening delimiter we want or a separator, then we check if it is left of selection
    - If left of selection, return it
    - If not,
      - if it is the opening delimiter, we go back to the outer loop
      - if it is a separator, we continue left
- If right delimiter is a separator, then:
  - Sweep leftwards looking for any delimiter or separator,
  - If yielded delimiter is left of selection, return it
  - If not, check whether the yielded delimiter is a separator or an opening delimiter
    - If separator, continue leftward
    - If opening delimiter,
      - sweep rightward looking only for the matching closing delimiter, ignoring separators or other delimiters.  I think can just set `acceptableDelimiters` for that one.  Though depending how much we care about mismatched delimiters, it might be safer to leave `acceptableDelimiters` and just loop checking the yielded delimiter
      - once we hit the matching closing delimiter, we just go back to the outer loop

#### In `findDelimiterPairAdjacentToSelection`

- If it is an opening delimiter, we look to the right for its closing delimiter or for a separator, making sure to use `trackedDelimiters` to ensure that we still properly ignore nested separators
- If it is a closing delimiter, we look to the left for its opening delimiter or for a separator
- If it is a separator, we look to the left for matching separator or any opening delimiter

#### Returning selection

- We consider any whitespace to the right of the left delimiter, or to the left of the right delimiter, to be part of the delimiter
- We don't support interiorOnly / excludeInterior
- On `selectionContext`, set trailing / leading delimiter and `inDelimitedList` in the same way we do for args / items today

#### Unenclosed lists

Examples:
  - CSVs
  - unenclosed import lists, like in Python
  - `|`-separated type disjunctions in Typescript

We might be able to inject dummy zero-length opening / closing delimiter pair by leveraging a bit of language-specific context here:

- For the Python unenclosed import lists, inject a dummy delimiter after `import` and before end of statement
- For Typescript `|`, any node where child is `union_type` but parent is not, inject dummy delimiters before and after `union_type`
- For the CSV example, inject dummy delimiters at start and end of line

Alternately, could just limit scope of search instead of injecting dummy delimiters, but then we'd need to indicate to lower level that it's ok to yield even if it didn't find one or both sides, which feels weird

### LINKED ISSUE #441: Typescript: `item` doesn't work for const imports
```typescript
import { A, B, C } from 'foo'
```

In the above example, you would expect `chuck item bat` to delete the const import `B`, but you get the error `Couldn't find containing collectionItem`

### LINKED ISSUE #566: `"chuck every item"` leaves a comma
In the following python code:

```python
{
    "hello": "there",
    "testing": "whatever",
}
```

Saying `"chuck every item"` results in the following:

```python
{
    ,
}
```

Note: this behaviour only occurs when there is a trailing comma

AndreasArvidsson:
Fixes #357
Fixes #566
Fixes #441
Fixes #294

If we replace the language based implementation with the text based one like we did for string we can close a lot of language specific issues with item

## Checklist


pokey (commented):
As discussed, let's aim to

- [ ] Support lists with a single item (ie no comma appears inside the parens)
- [ ] Reuse the existing `collectionItem` modifier

Ideally we'd reuse existing code from [`generateUnmatchedDelimiters`](https://github.com/cursorless-dev/cursorless/blob/main/src/processTargets/modifiers/surroundingPair/generateUnmatchedDelimiters.ts) as well, but prob not a showstopper given this file is well contained, not too long, and will probably be reworked after #484

pokey:
Performed git surgery.  Original branch at https://github.com/cursorless-dev/cursorless/tree/bak/textTheBasedItemScope; feel free to delete if this branch looks ok

pokey (commented):
Looks like updated code causes type checker to be upset; I'll hold off reviewing till you get a chance to take a look

AndreasArvidsson:
@pokey The tests are now clearing locally

pokey:
plan:

- [ ] Remove "item" scope type from tree-sitter defs for languages that have comma-separated items
- [ ] Leave "arg" as is in those languages
- [ ] Duplicate "arg" tests for those languages with plaintext as lang and "item" as scope
- [ ] Leave languages alone that don't have comma-separated lists

AndreasArvidsson:
@pokey All languages that can are now relying on the text based implementation and I have added a bunch of new tests

src/processTargets/modifiers/scopeTypeStages/ItemStage.ts
```diff
@@ -0,0 +1,361 @@
...
+      ).run(context, target);
+    } catch (_error) {}
+
+    if (this.modifier.type === "everyScope") {
+      return this.getEveryTarget(context, target);
+    }
+    return [this.getSingleTarget(context, target)];
+  }
+
+  private getEveryTarget(context: ProcessedTargetsContext, target: Target) {
```
pokey, line 36: Only expand to iteration scope if target is weak

src/processTargets/modifiers/scopeTypeStages/ItemStage.ts
```diff
@@ -0,0 +1,361 @@
...
+  SimpleContainingScopeModifier,
+} from "./ContainingSyntaxScopeStage";
+import { fitRangeToLineContent } from "./LineStage";
+
+export default class ItemStage implements ModifierStage {
+  constructor(private modifier: ContainingScopeModifier | EveryScopeModifier) {}
+
+  run(context: ProcessedTargetsContext, target: Target): Target[] {
+    try {
+      return new ContainingSyntaxScopeStage(
```
pokey, line 25: add comment

src/processTargets/modifiers/scopeTypeStages/ItemStage.ts
```diff
@@ -0,0 +1,361 @@
+import { Range, TextEditor } from "vscode";
```
pokey, line 1: Please move into a directory so that you can factor out all the different functions

src/processTargets/modifiers/scopeTypeStages/ItemStage.ts
```diff
@@ -0,0 +1,361 @@
...
+    itemInfo.leadingDelimiterRange != null
+      ? getText(itemInfo.leadingDelimiterRange)
+      : defaultDelimiterInsertion,
+  ];
+  // Use longest delimiter text for insertion
+  delimiters.sort((a, b) => b.length - a.length);
+  return delimiters[0];
+}
+
+function getItemInfos(context: ProcessedTargetsContext, target: Target) {
```
pokey, line 97: Add doc string and or change name

pokey:
- [ ] File issue for trailing comma with "chuck all arg", esp Python

pokey:
Support "item" within strings

AndreasArvidsson:
I have now added python specific implementation for item in import list

src/processTargets/modifiers/ItemStage/tokenizeRange.ts
```diff
@@ -0,0 +1,129 @@
...
+      { type: "boundary", range: collectionBoundary[0] },
+      ...tokens,
+      { type: "boundary", range: collectionBoundary[1] },
+    ];
+  }
+
+  return tokens;
+}
+
+export function joinLexemesBySkippingMatchingPairs(lexemes: string[]) {
```
pokey, line 59: Docstring please, including a simple example

src/processTargets/modifiers/ItemStage/tokenizeRange.ts
```diff
@@ -0,0 +1,129 @@
+import { Range, TextEditor } from "vscode";
+
+export function tokenizeRange(
```
pokey, line 3: Docstring and return type please

src/processTargets/modifiers/ItemStage/tokenizeRange.ts
```diff
@@ -0,0 +1,129 @@
...
+    // Whitespace found. Just skip
+    if (lexeme.trim().length === 0) {
+      offset += lexeme.length;
+      return;
+    }
+
+    // Separator delimiter found.
+    if (lexeme === delimiter) {
+      tokens.push({
+        type: "delimiter",
```
pokey, line 25: Let's call this `"separator"`.  The term "delimiter" is too vague to be useful here, as you could easily interpret it to include boundaries

src/processTargets/modifiers/ItemStage/ItemStage.ts
```diff
@@ -0,0 +1,169 @@
...
+        return new Range(token.range.end, tokens[i + 1].range.end);
+      }
+      return undefined;
+    })();
+    // Leading boundary is excluded and leading delimiter is included
+    const leadingMatchStart =
+      tokens[i - 1]?.type === "boundary"
+        ? tokens[i - 1].range.end
+        : tokens[i - 1]?.type === "delimiter"
+        ? tokens[i - 1].range.start
```
pokey, line 145: I believe this should be
```suggestion
        ? tokens[i - 1].range.end
```

The fact that this ends up going left when cursor is directly before `,` happens in a fairly opaque way; both this item and the proceeding one will match the domain, and it's just the fact that the previous item ends up having its match earlier in the list of matches that causes it to get returned

src/processTargets/modifiers/ItemStage/ItemStage.ts
```diff
@@ -0,0 +1,169 @@
...
+      throw new NoContainingScopeError(this.modifier.scopeType.type);
+    }
+
+    itemInfoWithIntersections.sort(
+      (a, b) =>
+        rangeLength(target.editor, b.intersection!) -
+        rangeLength(target.editor, a.intersection!)
+    );
+
+    return this.itemInfoToTarget(target, itemInfoWithIntersections[0].itemInfo);
```
pokey, line 70: Should just make a range from first to last item.  See test I added.  That behaviour is consistent with "line", and desired behaviour in https://github.com/cursorless-dev/cursorless/issues/484

src/processTargets/modifiers/surroundingPair/index.ts
```diff
@@ -33,11 +34,40 @@ export function processSurroundingPair(
...
+ * @returns The new selection expanded to the containing surrounding pair or
+ * `null` if none was found
+ */
+export function processSurroundingPairForDelimiters(
+  context: ProcessedTargetsContext,
+  editor: TextEditor,
+  range: Range,
+  scopeType: SurroundingPairScopeType,
+  delimiters: SimpleSurroundingPairName[]
+): SurroundingPairInfo | null {
```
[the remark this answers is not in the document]
AndreasArvidsson: I'm also not that happy with it. But it is the same signature as [findSurroundingPairTextBased](https://github.com/cursorless-dev/cursorless/blob/022efb15a6b94ecbef32f7ecbcc6f2b188ac03ef/src/processTargets/modifiers/surroundingPair/findSurroundingPairTextBased.ts#L68-L180) so the problem is deeper. May be just make it so that delimiter can optionally be a list?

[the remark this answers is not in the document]
pokey: What's wrong with using a complex scope type, as per link above?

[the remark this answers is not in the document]
pokey: Fair point re `findSurroundingPairTextBased`, but I don't think that means the problem is deeper.  We can fix it here using a new complex surrounding pair type  so at least we localise that problem to the impl of surrounding pair

src/processTargets/modifiers/ItemStage/getIterationScope.ts
```diff
@@ -0,0 +1,93 @@
...
+function getParentSurroundingPair(
+  context: ProcessedTargetsContext,
+  editor: TextEditor,
+  pairInfo: SurroundingPairInfo
+) {
+  // Step out of this pair and see if we have a parent
+  const position = editor.document.positionAt(
+    editor.document.offsetAt(pairInfo.contentRange.start) - 1
+  );
+  if (position.isEqual(pairInfo.contentRange.start)) {
```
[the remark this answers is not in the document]
AndreasArvidsson: Correct. We can do that

src/processTargets/modifiers/surroundingPair/index.ts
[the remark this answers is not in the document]
AndreasArvidsson: I'm fine with that. Just have to decide which solution is the best one.

src/test/suite/fixtures/recorded/itemTextual/chuckItem.yml
```diff
@@ -0,0 +1,32 @@
...
+        aaa,
+    ]
+  selections:
+    - anchor: {line: 1, character: 4}
+      active: {line: 1, character: 4}
+  marks: {}
+finalState:
+  documentContents: |-
+    [
+
```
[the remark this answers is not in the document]
AndreasArvidsson: I had the same thought. But as you say it's not actually connected to textual item.

src/processTargets/modifiers/surroundingPair/index.ts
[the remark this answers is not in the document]
pokey: Do you have another solution in mind?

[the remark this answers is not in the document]
pokey: Hmm not a huge fan of variadic types like that, plus having it formalised puts its definition in one place, rather than baked into this code

src/processTargets/modifiers/ItemStage/ItemStage.ts
AndreasArvidsson: No it is correct afaict. The separator is included in the leading range and excluded in the trailing range. I'm actually sorting them on intersection range length to always get the correct one.  
itemInfoWithIntersections.sort(
      (a, b) =>
        rangeLength(target.editor, b.intersection!) -
        rangeLength(target.editor, a.intersection!)
    );

src/processTargets/modifiers/ItemStage/ItemStage.ts
AndreasArvidsson: To day I don't think we want `take item comma` to select both items. This would break with the behavior of arguments.

src/util/nodeSelectors.ts
```diff
@@ -385,3 +379,22 @@ export function delimitedSelector(
     };
   };
 }
+
+export function getInsertionDelimiter(
```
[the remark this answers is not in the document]
AndreasArvidsson: Exactly

src/processTargets/modifiers/surroundingPair/index.ts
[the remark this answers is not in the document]
AndreasArvidsson: I agree that a least always is a better typing. That we are hesitant to do the better solution because of migrations is really indicative in how laborious our migrations are. Ideally it would just be writing a simple migration and be done with it.

Could you elaborate on the complex surrounding pair type and how that would look?

src/processTargets/modifiers/ItemStage/ItemStage.ts
AndreasArvidsson: Okay then let's do that

src/languages/python.ts
```diff
@@ -49,13 +53,28 @@ export const getTypeNode = (node: SyntaxNode) =>
...
+  const finder = argumentNodeFinder("import_from_statement");
+  return (node: SyntaxNode, selection?: Selection) => {
+    const childNode = finder(node, selection);
+    if (
+      childNode?.type !== "dotted_name" ||
+      childNode?.id === childNode?.parent?.namedChildren[0].id
+    ) {
+      return null;
+    }
+    return childNode;
```
pokey, line 66: nit:

```suggestion
    if (
      childNode?.type === "dotted_name" &&
      childNode.equals(childNode.parent.firstNamedChild)
    ) {
      return childNode;
    }

    return null;
```

src/processTargets/modifiers/ItemStage/tokenizeRange.ts
```diff
@@ -0,0 +1,162 @@
+import { Range, TextEditor } from "vscode";
+
+/**
+ * Takes the range for a collection and returns a list of tokens within that collection
```
pokey, line 4: Please define  what a "token" is here, as we use the term "token" all over the place and this is not what we mean.  Tbh I'm tempted to not use "token" at all as a term here given that it violates our naming conventions, but at the very least it would be good to have a comment indicating how you're using the term here

An example would prob be good; see what I did on lines 19-20 below

src/processTargets/modifiers/ItemStage/tokenizeRange.ts
```diff
@@ -0,0 +1,162 @@
...
+    ];
+  }
+
+  return tokens;
+}
+
+/**
+ * Takes a list of lexemes and joins them by skipping matching pairs (), {}, etc
+ * @param lexemes List of lexemes to operate on
+ * @returns List of lexemes with equal or less length then {@link lexemes}
```
pokey, line 78:
```suggestion
 * @returns List of merged lexemes.  Note that its length will be less than or equal to {@link lexemes}
```

src/processTargets/modifiers/ItemStage/tokenizeRange.ts
```diff
@@ -0,0 +1,162 @@
...
+      ...tokens,
+      { type: "boundary", range: boundary[1] },
+    ];
+  }
+
+  return tokens;
+}
+
+/**
+ * Takes a list of lexemes and joins them by skipping matching pairs (), {}, etc
```
pokey, line 76:
```suggestion
 * Takes a list of lexemes and joins them into a list of alternating items and separators, skipping matching pairs (), {}, etc
```

pokey (commented):
Looks good!  Down to the last few minor comments and then let's ship it

AndreasArvidsson:
@pokey Your latest change requests are now implemented

pokey (approved):
Ok I made a couple of final tweaks in https://github.com/cursorless-dev/cursorless/pull/709/commits/0791c3e8de28bf57865347daf3f16439ce866d5f.  If they look good to you merge away!

PR #715, opened 2022-06-01
Fit interior range to content

### LINKED ISSUE #713: Update inside content range to only contain actual content and not whitespace
* content range should shrink down to fit actual content
* the full interior should be removal range
* on empty interior content range should be the same as removal range

AndreasArvidsson:
Fixes #713 

## Checklist

### DIFF
--- a/src/processTargets/targets/InteriorTarget.ts
+++ b/src/processTargets/targets/InteriorTarget.ts
@@ -1,0 +1,55 @@
+import { Range } from "vscode";
+import { Target, TargetType } from "../../typings/target.types";
+import { shrinkRangeToFitContent } from "../../util/selectionUtils";
+import BaseTarget, {
+  CloneWithParameters,
+  CommonTargetParameters,
+} from "./BaseTarget";
+import { createContinuousRangeWeakTarget } from "./WeakTarget";
+
+export default class InteriorTarget extends BaseTarget {
+  constructor(parameters: CommonTargetParameters) {
+    super(parameters);
+  }
+
+  get type(): TargetType {
+    return "weak";
+  }
+  get delimiter() {
+    return "";
+  }
+
+  get contentRange() {
+    return shrinkRangeToFitContent(this.editor, this.state.contentRange);
+  }
+
+  getRemovalRange(): Range {
+    return this.state.contentRange;
+  }
+
+  cloneWith(parameters: CloneWithParameters) {
+    return new InteriorTarget({
+      ...this.getCloneParameters(),
+      ...parameters,
+    });
+  }
+
+  createContinuousRangeTarget(
+    isReversed: boolean,
+    endTarget: Target,
+    includeStart: boolean,
+    includeEnd: boolean
+  ): Target {
+    return createContinuousRangeWeakTarget(
+      isReversed,
+      this,
+      endTarget,
+      includeStart,
+      includeEnd
+    );
+  }
+
+  protected getCloneParameters() {
+    return this.state;
+  }
+}
--- a/src/util/selectionUtils.ts
+++ b/src/util/selectionUtils.ts
@@ -1,4 +1,5 @@
-import { Position, Range, Selection } from "vscode";
+import { Position, Range, Selection, TextEditor } from "vscode";
+import { fitRangeToLineContent } from "../processTargets/modifiers/scopeTypeStages/LineStage";
 import { SelectionWithEditor } from "../typings/Types";
 
 export function isForward(selection: Selection) {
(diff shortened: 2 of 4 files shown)


pokey:
Performed git surgery.  Original branch at https://github.com/cursorless-dev/cursorless/tree/bak/fitInteriorToContent; feel free to delete if this branch looks ok

src/processTargets/targets/InteriorTarget.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export default class InteriorTarget extends BaseTarget {
+  insertionDelimiter = "";
+
+  constructor(parameters: CommonTargetParameters) {
+    super(parameters);
+  }
+
+  get contentRange() {
+    return shrinkRangeToFitContent(this.editor, this.state.contentRange);
```
pokey, line 12: huh.  It's a bit confusing that the `contentRange` is actually the `removalRange` for this target 🤔.  Not sure I have a better solution, but it's a bit surprising.  Maybe `InteriorTarget` is the exception to `CommonTargetParameters` containing `contentRange`?

Let's Discord this one when you're back

pokey (changes requested):
Nice direction; left some comments

src/util/selectionUtils.ts
```diff
@@ -42,3 +42,58 @@ export function selectionFromRange(isReversed: boolean, range: Range) {
   const { start, end } = range;
   return isReversed ? new Selection(end, start) : new Selection(start, end);
 }
+
+export function shrinkRangeToFitContent(editor: TextEditor, range: Range) {
+  const { lineAt } = editor.document;
+
+  const start = (() => {
```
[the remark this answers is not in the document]
AndreasArvidsson: That's definitely one way of doing it. I think both will give the same results but the current solution is probably more performance effective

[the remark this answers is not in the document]
pokey: Maybe?  We don't know how `lineAt` and `line.text` work.  I would be inclined to avoid premature optimisation if we can shed 40 lines of code

src/processTargets/targets/InteriorTarget.ts
pokey: I'd be tempted to have something like `definingRange` as the common parameter, and then `contentRange` by default returns it, but things can override it

AndreasArvidsson: We can do that

src/util/selectionUtils.ts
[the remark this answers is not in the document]
AndreasArvidsson: Updated. Is the current solution better?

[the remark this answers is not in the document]
pokey: Much cleaner!

src/util/selectionUtils.ts
```diff
@@ -42,3 +42,15 @@ export function selectionFromRange(isReversed: boolean, range: Range) {
   const { start, end } = range;
   return isReversed ? new Selection(end, start) : new Selection(start, end);
 }
+
+export function shrinkRangeToFitContent(editor: TextEditor, range: Range) {
```
pokey, line 46: Docstring please 😊

src/processTargets/targets/InteriorTarget.ts
AndreasArvidsson: You want me to go ahead and do that change in this pr?

pokey: Might be worth a quick convo

AndreasArvidsson: Sounds good

pokey (commented):
- [ ] Stop using `CommonTargetParameters` for parameters to `InteriorTarget`; instead use `fullInteriorRange` and then construct `contentRange` before passing to `BaseTarget`

src/test/suite/fixtures/recorded/selectionTypes/clearHeadCoreCurly.yml
```diff
@@ -23,12 +23,13 @@ initialState:
   marks: {}
 finalState:
   documentContents: |-
```
pokey, line 25: do we have a test for "chuck head core curly"?  If not, worth adding I think

src/util/selectionUtils.ts
```diff
@@ -42,3 +42,22 @@ export function selectionFromRange(isReversed: boolean, range: Range) {
   const { start, end } = range;
   return isReversed ? new Selection(end, start) : new Selection(start, end);
 }
+
+/**
+ * Shrink given range to fit text content.
```
pokey, line 47:
```suggestion
 * Return a copy of {@link range} excluding any leading or trailing whitespace.
```

src/util/selectionUtils.ts
```diff
@@ -42,3 +42,22 @@ export function selectionFromRange(isReversed: boolean, range: Range) {
...
+ * If the given range has no text content the given range will be returned unchanged.
```
pokey, line 48: What does "no text content" mean?  Does that mean "if the range contains only whitespace"?

pokey (commented):
Basically there!  In addition to a few minor comments inline:

- [ ] Would be good to add a "change" test where the core is empty
- [ ] Would be good to add a "change" test where the core includes only whitespace

src/util/selectionUtils.ts
AndreasArvidsson: That or empty range

src/test/suite/fixtures/recorded/selectionTypes/clearHeadCoreCurly.yml
AndreasArvidsson: It will remove the same range since we are only looking at the content ranges in head/tail stage.

pokey: Hmm that is a bit unfortunate. I'm not entirely sure what is the right thing to do in that situation. I'm not sure there's anything else we can do reliably, but I was kind of expecting it to include the white space

AndreasArvidsson: In that case we probably need to pass removal and content range to the head/tail stage. Or pass the leading and trailing delimiters or something like that.

pokey: Yeah I'm just not sure there's a reliable way to do it without all targets needing to know something about head and tail. Because targets need to be able to be smart about constructing removal range, and if head is just taking it and slicing it in half, that could violate the logic, eg we could possibly end up merging tokens

I don't think there's a simple way to do better than what you have here. Unless you have an idea?

AndreasArvidsson: Unfortunately I don't have any good answers right now

pokey: No worries I'm happy to ship it and see if we find ourselves wanting this. Head tail does return a token target so at least tries to do some cleanup. Let's see if that's good enough

PR #718, opened 2022-06-01
Add the ability to specify a language specific tokenizer, starting with CSS/SCSS

Will-Sommers:
## What?
This adds a custom tokenizer regex for CSS/SCSS to allow for tokenization of `kebab-case-properties` as single tokens.

## Why?
In SCSS/CSS, this is a convention and properties are treated as single keywords.

## How?
Tweak the `tokenizer.ts` API to allow for custom definitions to be specified by `languageId`. I've added tests as well and a pattern for other language specific tokenizations, should we need them.  

## Follow ups?
- I'm considering splitting up the tokenizer file for clarity, pulling out the default definitions. Thoughts?

## Checklist

- Add other tests, add integration tests?


src/test/suite/tokenizer.test.ts
```diff
@@ -60,13 +67,45 @@ const tests: TestCase[] = [
   ["aåäöb", ["aåäöb"]],
 ];
 
+const languageTokenizerTests: LanguageTokenizerTests = {
+  css: {
+    tests: [
+      ["min-height", ["min-height"]],
+      ["-webkit-font-smoothing", ["-webkit-font-smoothing"]],
+    ],
```
auscompgeek, line 75: Can we add a media query condition here too?

src/core/tokenizer.ts
```diff
@@ -39,29 +42,79 @@ const FIXED_TOKENS = [
...
+
+  // Order matters here.
+  const regex = [
+    fixedTokensRegex,
+    tokenizerDefinition.numbersRegex,
+    tokenizerDefinition.identifiersRegex,
+    repeatableSymbolsRegex,
+    tokenizerDefinition.singleSymbolsRegex,
+  ].join("|");
+  console.log("lang:", languageId, regex);
```
pokey, line 88: Assuming we don't want to ship this?  If you did intend to keep it, it should at least be `console.debug`
```suggestion
```

src/core/tokenizer.ts
```diff
@@ -39,29 +42,79 @@ const FIXED_TOKENS = [
...
+function generateTokenMatcher(languageId?: SupportedLanguageId): RegExp {
+  let settings = {};
+  if (languageId && languageSettings[languageId]) {
+    settings = languageSettings[languageId] ?? {};
+  }
+
+  const tokenizerDefinition = Object.assign(
+    defaultLanguageTokenizerSetting,
+    settings
+  );
```
pokey, line 69: This mutates `defaultLanguageTokenizerSetting`.  Please use `...` spread instead

src/core/tokenizer.ts
```diff
@@ -39,29 +42,79 @@ const FIXED_TOKENS = [
...
+  numbersRegex: NUMBERS_REGEX,
+  singleSymbolsRegex: SINGLE_SYMBOLS_REGEX,
+};
+
+export const TOKEN_MATCHER = generateTokenMatcher();
+
+function generateTokenMatcher(languageId?: SupportedLanguageId): RegExp {
+  let settings = {};
+  if (languageId && languageSettings[languageId]) {
+    settings = languageSettings[languageId] ?? {};
```
pokey, line 63: why do you need `??` if you're already checking it's non-null in guard?

src/core/tokenizer.ts
```diff
@@ -39,29 +42,79 @@ const FIXED_TOKENS = [
...
+
+const languageSettings: Partial<
+  Record<SupportedLanguageId, Partial<DefaultLanguageTokenizerSettings>>
+> = {
+  css,
+};
+
+const tokenMatchersForLanguage: Partial<Record<SupportedLanguageId, RegExp>> = {
+  css: generateTokenMatcher("css"),
+  scss: generateTokenMatcher("css"),
```
pokey, line 100: why not just pass the `css` object directly into `generateTokenMatcher`?  A bit confusing to have this object as well as `languageSettings`.  I'd also be tempted to not require repeating `generateTokenMatcher` every time; could just [map the object values](https://lodash.com/docs/4.17.15#mapValues) through `generateTokenMatcher`

src/core/tokenizer.ts
```diff
@@ -39,29 +42,79 @@ const FIXED_TOKENS = [
...
-].join("|");
+const defaultLanguageTokenizerSetting: DefaultLanguageTokenizerSettings = {
+  fixedTokens: FIXED_TOKENS,
+  repeatableSymbols: REPEATABLE_SYMBOLS,
+  identifiersRegex: IDENTIFIERS_REGEX,
+  numbersRegex: NUMBERS_REGEX,
+  singleSymbolsRegex: SINGLE_SYMBOLS_REGEX,
+};
+
+export const TOKEN_MATCHER = generateTokenMatcher();
```
pokey, line 58: I'd rename this one to `defaultTokenMatcher`, and not export it

src/core/tokenizer.ts
```diff
@@ -39,29 +42,79 @@ const FIXED_TOKENS = [
...
+  scss: generateTokenMatcher("css"),
+};
+
+export function getTokenMatcher(languageId: string): RegExp {
+  const languageMatcher =
+    tokenMatchersForLanguage[languageId as SupportedLanguageId];
+  if (languageMatcher) {
+    return languageMatcher;
+  }
+  return TOKEN_MATCHER;
```
pokey, line 109: can replace all this with a null-ish coalescence `??`

src/test/suite/fixtures/recorded/languages/scss/takeFirstWordCap.yml
```diff
@@ -0,0 +1,40 @@
+languageId: css
```
pokey, line 1: not worth re-recording, but please use "clear" / "change" instead of "take"

src/test/suite/tokenizer.test.ts
```diff
@@ -60,13 +67,47 @@ const tests: TestCase[] = [
...
+  Object.entries(languageTokenizerTests).forEach(
+    ([language, { tests, exclusionPredicate = () => true }]) => {
+      tests.forEach(([input, expectedOutput]) => {
+        test(`${language} custom tokenizer, input: ${input}`, () => {
+          const output = tokenize(input, language, (match) => match[0]);
+          assert.deepStrictEqual(output, expectedOutput);
+        });
+      });
+
+      tests.forEach(([input, expectedOutput]) => {
```
pokey, line 100: I believe you're intending to use the `tests` variable from the global scope here, but it is being shadowed by the `tests` parameter

src/typings/Types.ts
```diff
@@ -181,6 +181,16 @@ export interface Edit {
   isReplace?: boolean;
 }
 
+/** Represents a custom tokenizer for a language */
+
+export interface DefaultLanguageTokenizer {
```
pokey, line 186: Shouldn't this be named something like
```suggestion
export interface LanguageTokenizerComponents {
```

src/core/languageTokenizers/css.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Add custom identifierRegex to ensure that kebab case identifiers, properties
+ * and class-names are tokenized as a single token.
+ */
+export default {
+  identifiersRegex: "[\\p{L}_|\\-0-9]+",
```
pokey, line 6: what's the `|` doing here?

pokey (changes requested):
Looks good!  Left a bunch of minor comments.  In addition, it would be good to have the following tests:

- [ ] "take every token" on a line in a `css` file to ensure it's using the `css` tokenizer
- [ ] [Hat-map update tests](https://www.cursorless.org/docs/contributing/test-case-recorder/#testing-the-hat-map).  Use the following initial document for each test:
    ```
    foo-bar baz-bongo
    ```
    Then test each of the following:
    - [ ] "move last word zip after fine take fine"
    - [ ] "move last word zip after fine take zip"
    - [ ] "move first word zip before fine take fine"
    - [ ] "move first word zip before fine take zip"

src/core/tokenizer.ts
Will-Sommers: I think this might have something to do with `Partial`. I'm getting a type error here that `settings` could be `undefined` which is why I added null-ish coalesce.

Will-Sommers:
Recording a few more tests as well

src/core/tokenizer.ts
pokey: It's prob because you access the array twice: once in the guard and once here

Will-Sommers: ty

Will-Sommers:
Heyo @pokey — I think I recorded the navigation tests correctly. Let me know.

src/core/tokenizer.ts
```diff
@@ -39,29 +44,67 @@ const FIXED_TOKENS = [
...
+    fixedTokensRegex,
+    tokenizerDefinition.numbersRegex,
+    tokenizerDefinition.identifiersRegex,
+    repeatableSymbolsRegex,
+    tokenizerDefinition.singleSymbolsRegex,
+  ].join("|");
+  return new RegExp(regex, "gu");
+}
+
+const tokenMatchersForLanguage: Partial<Record<SupportedLanguageId, RegExp>> =
```
pokey, line 90: nit: I'd prob store `{ css, scss }` in a variable first, so that the dev doesn't need to look at this complex statement to add a new one

src/test/suite/fixtures/recorded/languages/scss/moveLastWordZipAfterFineTakeZip.yml
```diff
@@ -0,0 +1,52 @@
...
+      active: {line: 0, character: 17}
+  marks:
+    default.z:
+      start: {line: 0, character: 8}
+      end: {line: 0, character: 17}
+    default.f:
+      start: {line: 0, character: 0}
+      end: {line: 0, character: 7}
+finalState:
+  documentContents: foo-bar bongo baz
```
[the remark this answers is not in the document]
Will-Sommers: @pokey — it appears that `take last word` and `take first word` might be broken for:

`kebab-case`
`CONSTANT_CASE`

Just tried it in plaintext on main.

src/test/suite/fixtures/recorded/languages/scss/takeEveryToken.yml
```diff
@@ -0,0 +1,38 @@
+languageId: scss
```
pokey, line 1: nice 😊

src/test/suite/tokenizer.test.ts
```diff
@@ -1,8 +1,15 @@
...
 import { flatten, range } from "lodash";
+import { SupportedLanguageId } from "../../languages/constants";
 
 type TestCase = [string, string[]];
+type LanguageTokenizerTests = Partial<
+  Record<
+    SupportedLanguageId,
+    { tests: TestCase[]; exclusionPredicate?: (input: string) => boolean }
+  >
+>;
```
pokey, line 12: I think I'd have this type be the smaller one, and then add a docstring on `exclusionPredicate`, because it's not obvious what that does 😅
```suggestion
/** Language-specific tokenizer test configuration object */
interface LanguageTokenizerTests {
   /** Language-specific test cases to run in addition to the global tests for this language */
   additionalTests: TestCase[];

   /**
    * By default we run all global tests in the given language, in addition to the specific tests.  We exclude global test inputs that match this predicate.
    */
   exclusionPredicate?: (input: string) => boolean;
}
```

src/test/suite/tokenizer.test.ts
```diff
@@ -60,13 +67,53 @@ const tests: TestCase[] = [
...
+    ["min-height", ["min-height"]],
+    ["-webkit-font-smoothing", ["-webkit-font-smoothing"]],
+    ["(min-width: 400px)", ["(", "min-width", ":", "400px", ")"]],
+    ["prefers-reduced-motion", ["prefers-reduced-motion"]],
+  ],
+  // Leave kebab and dashes to css language specific tests.
+  exclusionPredicate: (input: string) => !!input.match("-"),
+};
+
+const languageTokenizerTests: LanguageTokenizerTests = {
```
pokey, line 81: Then here you just use 
```suggestion
const languageTokenizerTests: Partial<Record<SupportedLanguageId, LanguageTokenizerTests>> = {
```

src/test/suite/tokenizer.test.ts
```diff
@@ -60,13 +67,53 @@ const tests: TestCase[] = [
...
+  css: cssDialectTokenizerTests,
+  scss: cssDialectTokenizerTests,
+};
+
 suite("tokenizer", () => {
   tests.forEach(([input, expectedOutput]) => {
-    test(input, () => {
-      const output = tokenize(input, (match) => match[0]);
+    test(`tokenizer test, input: ${input}`, () => {
```
pokey, line 88:
```suggestion
    test(`tokenizer test, input: "${input}"`, () => {
```

src/test/suite/tokenizer.test.ts
```diff
@@ -60,13 +67,53 @@ const tests: TestCase[] = [
...
+      const output = tokenize(input, "anyLang", (match) => match[0]);
```
pokey, line 89: Aha this is why you need the typecast in `tokenize`.  Hmm.  I don't think that our supported languages for parser and supported languages for tokenizer need to be the same, so I wonder if that's even the right type to use?  Not quite sure what the right answer is here 🤔

Maybe we should have a type `LanguageId` that is a superset of `SupportedLanguageId`?  And then we could include `plaintext`, and use that here?

cc/ @auscompgeek @AndreasArvidsson

src/test/suite/tokenizer.test.ts
```diff
@@ -60,13 +67,53 @@ const tests: TestCase[] = [
...
     });
   });
+
+  Object.entries(languageTokenizerTests).forEach(
+    ([
+      language,
+      { tests: languageSpecificTests, exclusionPredicate = () => true },
+    ]) => {
+      languageSpecificTests.forEach(([input, expectedOutput]) => {
+        test(`${language} custom tokenizer, input: ${input}`, () => {
```
pokey, line 100:
```suggestion
        test(`${language} custom tokenizer, input: "${input}"`, () => {
```

src/test/suite/tokenizer.test.ts
```diff
@@ -1,8 +1,15 @@
...
 type TestCase = [string, string[]];
+type LanguageTokenizerTests = Partial<
+  Record<
+    SupportedLanguageId,
+    { tests: TestCase[]; exclusionPredicate?: (input: string) => boolean }
+  >
+>;
 const singleSymbolTests: TestCase[] = getAsciiSymbols().map((s) => [s, [s]]);
 
 const tests: TestCase[] = [
```
pokey, line 15: Maybe let's rename this one so the shadowing thing doesn't happen again
```suggestion
/** Tokenizer tests that will be run on the default tokenizer, as well as on all language-specific tokenizers */
const globalTests: TestCase[] = [
```

src/test/suite/tokenizer.test.ts
```diff
@@ -60,13 +67,53 @@ const tests: TestCase[] = [
...
+          const output = tokenize(input, language, (match) => match[0]);
+          assert.deepStrictEqual(output, expectedOutput);
+        });
+      });
+
+      tests.forEach(([input, expectedOutput]) => {
+        if (exclusionPredicate(input)) {
+          return;
+        }
+        test(`${language} custom tokenizer, input: ${input}`, () => {
```
pokey, line 110:
```suggestion
        test(`${language} custom tokenizer, input: "${input}"`, () => {
```

pokey (changes requested):
Ok getting close; left a bunch of minor comments

src/core/HatAllocator.ts
```diff
@@ -37,6 +37,10 @@ export class HatAllocator {
         this.toggleDecorations
       ),
 
+      // An event that fires when a text document opens
+      vscode.workspace.onDidOpenTextDocument(this.addDecorationsDebounced),
+      // An event that fires when a text document closes
+      vscode.workspace.onDidCloseTextDocument(this.addDecorationsDebounced),
```
pokey, line 43: Before this change, we didn't retokenize a document when we changed the language to css.  VSCode treats a language change as closing and re-opening the editor, so we need to listen to these events to catch that

src/test/suite/fixtures/recorded/languages/scss/moveLastWordZipAfterFineTakeZip.yml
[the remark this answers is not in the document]
pokey: This seems to work now after a rebase 🤷‍♂️

pokey (approved):
LGTM!

src/core/HatAllocator.ts
auscompgeek: Can we just listen to editor open? Adding decorations to an editor that has "closed" sounds like a weird thing to do :sweat_smile:

pokey: It doesn't add decorations to the closed editor, but I think probably closing an editor could affect hats in other editors 🤷‍♂️.  It's debounced as well so shouldn't make much difference

PR #722, opened 2022-06-02
Add unit sub-selection for values in SCSS/CSS

Will-Sommers:
## Checklist


Will-Sommers:
@pokey — this one is ready to go again.

docs/user/README.md
```diff
@@ -154,6 +154,7 @@ For programming languages where Cursorless has rich parse tree support, we suppo
 | `"string"`     | string                                              |
 | `"tags"`       | xml both tags                                       |
 | `"type"`       | a type annotation or declaration                    |
+| `"unit"`       | a unit, eg `px` in `100px`                          |
```
pokey, line 157: Do we want to support this one in other languages via a regex? Might be useful in talon files for example

pokey (commented):
Looks good! Nice job adding tests that start from multiple places. See inline comment

docs/user/README.md
auscompgeek: Also CSS-in-JS :)

Will-Sommers: Working on handling selection in these final cases. This is relatively straightforward — we run `matchAll` and then check where the selection is and take the one closest. 

`const padding = "10px 50px 30px 0px";`

Working through this case now `"take unit 0"`. We shouldn't find a match there. 
`const padding = "10px 50px 30px 0";`

auscompgeek:
:wave: Hey @Will-Sommers, how's this PR going? I'd personally love to see this land; let us know if someone else will need to pick this up.

Will-Sommers:
Heyo @auscompgeek feel free to take it over! 

This weekend I should be able to push up my code related to the regex version of this. I need to scroll back in my convos with @pokey to see where we fell on `take every unit` within a string.

docs/user/README.md
pokey: Let's stick syntax tree for now; this way we can do cool things like select all the units at once if they're in the property name

src/core/commandVersionUpgrades/upgradeV2ToV3/targetDescriptorV2.types.ts
```diff
@@ -68,7 +68,7 @@ export type SurroundingPairName =
   | SimpleSurroundingPairName
   | ComplexSurroundingPairName;
 
-export type SimpleScopeTypeType =
+export type SimpleScopeTypeTypeV2 =
```
pokey, line 71: @AndreasArvidsson I believe this was an oversight in #982, and caused a type error when a new scope type was added

src/test/suite/fixtures/recorded/languages/css/clearUnit5.yml
```diff
@@ -0,0 +1,37 @@
...
+    }
+      
+  selections:
+    - anchor: {line: 1, character: 4}
+      active: {line: 1, character: 4}
+  marks: {}
+finalState:
+  documentContents: |-
+    div {
+        padding: 25 50 75 100;
```
pokey, line 25: I think this is nice.  wdyt @auscompgeek @AndreasArvidsson

pokey (approved):
I made some tweaks.  I'm happy but will wait for 👀 from @AndreasArvidsson or @auscompgeek as I'm probably a bit too close to this one now

pokey (approved):

src/test/suite/fixtures/recorded/languages/css/clearUnit5.yml
AndreasArvidsson: Do I have this correct that it will operate on all units in the value if you target the key? That is fine.

pokey: yep!

AndreasArvidsson:
Looks good to me. If there are any details we can discuss them in the meeting.

pokey:
> Looks good to me. If there are any details we can discuss them in the meeting.

well there's the ever-present parser vs regex discussion 😅 https://github.com/cursorless-dev/cursorless/pull/722#discussion_r893104146.  In this case I think parser wins due to the multiple unit from key thing, but curious what you think

AndreasArvidsson:
> > Looks good to me. If there are any details we can discuss them in the meeting.
> 
> well there's the ever-present parser vs regex discussion 😅 [#722 (comment)](https://github.com/cursorless-dev/cursorless/pull/722#discussion_r893104146). In this case I think parser wins due to the multiple unit from key thing, but curious what you think

I think that the syntax tree is a better match here. Personally I don't really have the need to support units in any other language anyhow. Not one that I can think of on top of my head.

src/test/suite/fixtures/recorded/languages/css/clearUnit5.yml
auscompgeek: Seems fine, not sure what else it could do in that case anyway.

auscompgeek:
> Personally I don't really have the need to support units in any other language anyhow. Not one that I can think of on top of my head.

Talon script within `sleep()`?

AndreasArvidsson:
> > Personally I don't really have the need to support units in any other language anyhow. Not one that I can think of on top of my head.
> 
> Talon script within `sleep()`?

The amount of times I want to change just the unit in a Talon script sleep is more or less zero. Turning `500ms` into `1s` sure but in that case I would not target just the unit.

src/core/commandVersionUpgrades/upgradeV1ToV2/upgradeV1ToV2.ts
```diff
@@ -41,7 +41,7 @@ function upgradeModifier(modifier: ModifierV0V1): ModifierV2[] {
         {
           type: includeSiblings ? "everyScope" : "containingScope",
           scopeType: {
-            type: scopeType as SimpleScopeTypeType,
+            type: scopeType as SimpleScopeTypeTypeV2,
```
auscompgeek, line 44: Can we get rid of this type cast somehow?

pokey: Possibly, but doesn't seem like a huge deal.  We should probably figure out a better migration setup as part of #983, and we should consider not pinning scope type list during migrations at that point.  But for now this doesn't strike me as a big loophole, but happy to give it a bit more thought if you think it's a risk

Issue #598, opened 2022-03-09
Tests failing locally


Will-Sommers:
While working on https://github.com/cursorless-dev/cursorless-vscode/pull/597 I had erroneously reported failing tests locally. The suite passed on the PR build action but locally I am having failing tests. I also checked on the main branch and also have similar failures. The test run hasn't finished yet so I'm not sure if they are deterministically failing but I'll update the issue with more information. Here's an example locally failing test from the PR:

```
  22) recorded test cases
       /Users/.../recorded/languages/typescript/ifElseWrapThis:

      AssertionError [ERR_ASSERTION]: Unexpected final state
      + expected - actual

       {
      -  "documentContents": "if () {\n  const foo = \"hello\";\n} else {\n  \n}"
      +  "documentContents": "if () {\n    const foo = \"hello\";\n} else {\n    \n}"
         "selections": [
           {
             "active": {
               "character": 4

```

Also, noted in Slack. I currently have to quit VSCode to run the suite. Probably a separate issue but worth noting.

pokey:
@wenkokke was having this issue too. Do you remember if we got to the bottom of it?

wenkokke:
We never managed to fix it, I simply ignored it and you didn't have this issue locally.

pokey:
> We never managed to fix it, I simply ignored it and you didn't have this issue locally.

Right ok that was my recollection but was hoping my recollection was wrong 😅

I believe this one has to do with VSCode space insertion.  When you define a snippet in VSCode, it reinterprets tab characters in the snippet body as spaces according to your configuration for the given file / language.  So I believe it has something to do with your indentation settings.  We somehow need to get VSCode to ignore your custom indentation settings I believe

Will-Sommers:
Hrm, got it. I'll think about this a bit. I wonder if it would make sense to give local developers the ability to run tests akin to the GH actions, in a local container.

pokey:
oh that's a cool idea.  I wonder if that would fix the insiders' thing too

Will-Sommers:
This looks pretty cool, although new, and might obviate the need for a separate and possibly diverging config. https://github.com/nektos/act

pokey:
Funny I just ran across that today for the first time

Interesting idea, and might simplify things, tho seems a bit heavy handed, as I believe it installs VSCode in the container every time. But prob not a showstopper; might not be too slow and could potentially be cached

[...]

Issue #721, opened 2022-06-01
Add test for "pour" when target is in opposite split


pokey:
Ensure that the other editor gets focus.  Could potentially re-use some stuff from [`crossCellsSetSelection`](https://github.com/cursorless-dev/cursorless/blob/main/src/test/suite/crossCellsSetSelection.test.ts)

PR #731, opened 2022-06-03
Force tab size on test run

Will-Sommers:
- This fixes test failures related to different local configurations
- Tests expect tabSize to be 4

Fixes: https://github.com/cursorless-dev/cursorless/issues/598

## Checklist

- **N/A** - I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)


AndreasArvidsson:
This will only force tab size to be for when re-running the tests. We need to add the same to the test recorder otherwise every test will be recorded with the user tab setting, but run with a fixed setting.

pokey:
> This will only force tab size to be for when re-running the tests. I think we need to add the same to the test recorder otherwise every test will be recorded with the user tab setting, but run with a fixed setting.

I'll second what @AndreasArvidsson says here.  I think if we detect that the extension is running in debug mode, we should make the same `tabSize` change

Will-Sommers:
Not important, but I'm still a little curious about why this didn't break for me, with other tests that I've recorded. 

@pokey — I'll likely add the setting here as well rather than in the actual extension.

https://github.com/cursorless-dev/cursorless/blob/main/src/test/suite/recorded.test.ts#L74-L78

pokey:
> Not important, but I'm still a little curious about why this didn't break for me, with other tests that I've recorded.

did you record tests that include snippet wrapping?  I think those are the ones that will break

> 
> @pokey — I'll likely add the setting here as well rather than in the actual extension.
> 
> https://github.com/cursorless-dev/cursorless/blob/main/src/test/suite/recorded.test.ts#L74-L78

Sorry not sure I follow—that won't impact recording right?

Will-Sommers:
> did you record tests that include snippet wrapping? I think those are the ones that will break

Nope, you're right. One other indent and fold test broke as well though

> Sorry not sure I follow—that won't impact recording right?

Multi-tasking between emails, sorry. You're right.

src/core/commandRunner/CommandRunner.ts
```diff
@@ -60,6 +61,10 @@ export default class CommandRunner {
       if (this.graph.debug.active) {
         this.graph.debug.log(`command:`);
         this.graph.debug.log(JSON.stringify(command, null, 3));
+        if (vscode.window.activeTextEditor) {
+          vscode.window.activeTextEditor.options.tabSize =
+            DEFAULT_TAB_SIZE_FOR_TESTS;
+        }
```
pokey, line 67: hmm maybe we should only do this if they're recording a test?  Otherwise we'll clobber their tab size if they have cursorless debug mode on

pokey (commented):
looks good with minor tweak

src/core/commandRunner/CommandRunner.ts
Will-Sommers: Heyo @pokey — I moved this entire step into the `TestCaseRecorder` since we have state there about whether we should or should not set the tabs. That also bring the code out of the `CommandRunner` loop. We clean up at the end and reset to the original tab setting, if a user stops recording their test.

src/core/constants.ts
```diff
@@ -35,3 +35,5 @@ export interface HatStyle {
   color: HatColor;
   shape: HatShape;
 }
+
+export const DEFAULT_TAB_SIZE_FOR_TESTS = 4 as const;
```
pokey, line 39: why the `as const` here?

pokey (commented):
I think we should do this one in the `preCommandHook` / `errorHook` / `postCommandHook` so that

1. The code should be simpler, I think, because we don't need to set / unset it in so many places
2. We can properly handle splits in the future.  Your current code will break with multiple splits

I would a keep list of visible editors and their original tab stops in `preCommandHook`, and reset them on those editors in `postCommandHook` / `errorHook`

Does that make sense?

Alternately, you can just do `activeTextEditor`, if the visible editors thing looks painful.  I am considering adding split testing support in https://github.com/cursorless-dev/cursorless/issues/721; debating whether it's worth supporting recording tests this way.  So if it looks painful to do all visible editors, active is prob fine, but I still think it's simpler to do it in `pre` / `postCommandHook`.  

But I don't feel _too_ strongly if you think it's going to be better this way 😊

src/testUtil/TestCaseRecorder.ts
```diff
@@ -230,10 +233,27 @@ export class TestCaseRecorder {
   }
 
   stop() {
+    this.toggleUserTabSetting(false);
+
     this.active = false;
     this.paused = false;
   }
 
+  private toggleUserTabSetting(shouldSetTabs: boolean) {
```
auscompgeek, line 242: This seems like a weird thing to have a toggle method for, especially since basically the entire method is in the if statement. This probably should just be two different methods.

src/test/suite/recorded.test.ts
```diff
@@ -75,6 +76,8 @@ async function runTest(file: string) {
     fixture.initialState.documentContents,
     fixture.languageId
   );
+  // Override any user defaults, all tests are recorded as tabSize = 4
+  editor.options.tabSize = DEFAULT_TAB_SIZE_FOR_TESTS;
```
auscompgeek, line 80: I think we also want to force soft tabs too. I have `"editor.insertSpaces": false` and all my snapshots changed 😅

pokey (approved):
Looks good to me!  Needed to make some tweaks because the restore code was broken

Issue #450, opened 2021-06-22
Add a command for toggling decorations


brxck:
It would be nice to have something hooked up to `cursorless.toggleDecorations` by default.

_Edit_ (@pokey): We've decided to just add instructions for how to set up your own voice command for this to the wiki under FAQ and recipes

pokey:
I'd be tempted to just have something in the wiki describing how to do this.  WDYT cc/ @AndreasArvidsson

AndreasArvidsson:
Sounds good to me

PR #733, opened 2022-06-04
Add docs to cover toggling hats

Will-Sommers:
Answers a question I had on day 1. 

Closes: https://github.com/cursorless-dev/cursorless/issues/450

## Checklist

- [x] **N/A** I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)

### DIFF
--- a/docs/user/customization.md
+++ b/docs/user/customization.md
@@ -60,6 +60,15 @@ push down, editor.action.moveLinesDownAction
 
 Now when you say eg "push down air and bat", cursorless will first select the two tokens with a gray hat over the `a` and `b`, then issue the VSCode command `editor.action.moveLinesDownAction`, and then restore your original selection.
 
+## Toggling "hat" display
+
+Most users, often after less than a week of using Cursorless, do not find the hat display over tokens to be obtrustive. However, should you need to remove the display of hats within your editor, there is a command registered for you.
+
+- Bring up the command pallet(`Shift + CMD + P`).
+- Enter `cursorless.toggleDecorations` or type `Cursorless Toggle` to see the command in the dropdown.
+
+Use the same steps above to reenable hat positioning. During the time when hats are not positioned, you will not be able to address any marks, eg. `take air`. However, Cursorless will still recognize scopes, eg. `take funk`.
+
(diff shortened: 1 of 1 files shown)


docs/user/customization.md
```diff
@@ -60,6 +60,15 @@ push down, editor.action.moveLinesDownAction
 
 Now when you say eg "push down air and bat", cursorless will first select the two tokens with a gray hat over the `a` and `b`, then issue the VSCode command `editor.action.moveLinesDownAction`, and then restore your original selection.
 
+## Toggling "hat" display
+
+Most users, often after less than a week of using Cursorless, do not find the hat display over tokens to be obtrustive. However, should you need to remove the display of hats within your editor, there is a command registered for you.
```
pokey, line 65:
```suggestion
It is possible to show / hide the hats with a command.  Keep in mind that many users, often after less than a week using Cursorless, find that their brain starts to tune out the hat display. Some start to miss them when they're gone 🥲

However, should you like to show / hide the hats within your editor, you can use the following approach.
```

docs/user/customization.md
```diff
@@ -60,6 +60,15 @@ push down, editor.action.moveLinesDownAction
...
+
+- Bring up the command pallet(`Shift + CMD + P`).
+- Enter `cursorless.toggleDecorations` or type `Cursorless Toggle` to see the command in the dropdown.
+
+Use the same steps above to reenable hat positioning. During the time when hats are not positioned, you will not be able to address any marks, eg. `take air`. However, Cursorless will still recognize scopes, eg. `take funk`.
```
pokey, line 70:
```suggestion
You should also be able to find the above command by saying `"please toggle"`.  The same command can be repeated to turn the hats back on.  

While the hats are hidden, you will not be able to address any marks, eg `"take air"`. However, Cursorless will still recognize scopes, eg `"take funk"`.

If you'd like to map a voice command to toggle the hats, have a look at https://youtu.be/oWUJyDgz63k
```

pokey (commented):
Thanks!  I left some tweaks; lmk if they seem ok to you

PR #734, opened 2022-06-05
typescript fix assignment expression name scope

### LINKED ISSUE #716: `"name"` scope type in typescript should support anything on the left and side
For example:

```typescript
this.leadingDelimiterRange_ = parameters.leadingDelimiterRange;
```

We should get `this.leadingDelimiterRange_`.

We should probably exclude `const` and `let` keywords though

AndrewDant:
Closes #716
## Checklist

- [ x] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)


AndreasArvidsson (approved):

pokey (approved):
Looks great thank you! Will be great to have, plus consistency with Python is nice

Issue #193, opened 2021-08-06
Implement next-gen inference engine


pokey:
capturing this mostly as some notes for posterity

![Image from iOS](https://user-images.githubusercontent.com/755842/128505717-16862274-1a58-47fb-b06b-358c1f094b7f.jpg)
![cursorless engine 2 0](https://user-images.githubusercontent.com/755842/128505475-5ac7fe28-c615-40f6-99af-48c448922ada.jpg)

PR #735, opened 2022-06-05
Fixed regression in vertical range target

AndreasArvidsson:
```
short
something longer
something even longer
```

"chuck line R slice M"

All three lines are completely removed because the new targets are still lines. This is a regression in how vertical range target is supposed to work. The solution is to always default to a plain target except in the case of a position target.

## Checklist

### DIFF
--- a/src/processTargets/processTargets.ts
+++ b/src/processTargets/processTargets.ts
@@ -10,6 +10,8 @@ import { ProcessedTargetsContext } from "../typings/Types";
 import { ensureSingleEditor } from "../util/targetUtils";
 import getMarkStage from "./getMarkStage";
 import getModifierStage from "./getModifierStage";
+import PlainTarget from "./targets/PlainTarget";
+import PositionTarget from "./targets/PositionTarget";
 
 /**
  * Converts the abstract target descriptions provided by the user to a concrete
@@ -144,7 +146,17 @@ function processVerticalRangeTarget(
       anchorTarget.contentRange.end.character
     );
 
-    results.push(anchorTarget.withContentRange(contentRange));
+    if (anchorTarget instanceof PositionTarget) {
+      results.push(anchorTarget.withContentRange(contentRange));
+    } else {
+      results.push(
+        new PlainTarget({
+          editor: anchorTarget.editor,
+          isReversed: anchorTarget.isReversed,
+          contentRange,
+        })
+      );
+    }
 
     if (i === activeLine) {
       return results;
--- a/src/test/suite/fixtures/recorded/compoundTargets/chuckLineRiskSliceMade.yml
+++ b/src/test/suite/fixtures/recorded/compoundTargets/chuckLineRiskSliceMade.yml
@@ -1,0 +1,51 @@
+languageId: plaintext
+command:
+  spokenForm: chuck line risk slice made
+  version: 2
+  targets:
+    - type: range
+      anchor:
+        type: primitive
+        mark: {type: decoratedSymbol, symbolColor: default, character: r}
+        modifiers:
+          - type: containingScope
+            scopeType: {type: line}
+      active:
+        type: primitive
+        mark: {type: decoratedSymbol, symbolColor: default, character: m}
+      excludeAnchor: false
+      excludeActive: false
+      rangeType: vertical
+  usePrePhraseSnapshot: true
+  action: {name: remove}
+initialState:
+  documentContents: |-
+    short
+    something longer
+    something even longer
+  selections:
+    - anchor: {line: 0, character: 0}
+      active: {line: 0, character: 0}
+  marks:
+    default.r:
+      start: {line: 0, character: 0}
+      end: {line: 0, character: 5}
+    default.m:
+      start: {line: 2, character: 0}
+      end: {line: 2, character: 9}
+finalState:
+  documentContents: |-
+
+    hing longer
+    hing even longer
+  selections:
+    - anchor: {line: 0, character: 0}
+      active: {line: 0, character: 0}
+  thatMark:
+    - anchor: {line: 0, character: 0}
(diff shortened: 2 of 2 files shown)


pokey:
Hmm. What you describe is the behaviour I would expect. The behaviour you want is the behaviour I'd want for "chuck risk slice made". If you want that behaviour, why say "line"?

pokey:
Fwiw in next generation inference engine, I'd argue that "line" should be applied to output of slice range, so you'll get this behaviour, plus things like "change line risk slice made" will work as well

AndreasArvidsson:
> Hmm. What you describe is the behaviour I would expect. The behaviour you want is the behaviour I'd want for "chuck risk slice made". If you want that behaviour, why say "line"?

Because you want to create a selection the width of the entire line. Imagine you have multiple tokens on the line and you want to remove them all in a vertical column.

AndreasArvidsson:
> Fwiw in next generation inference engine, I'd argue that "line" should be applied to output of slice range, so you'll get this behaviour, plus things like "change line risk slice made" will work as well

Okay that might be useful, but it is an change on how it works on main today that we have not planned as far as I know.

pokey:
Btw one argument for what you're proposing here is that soon we will have 
"chuck every line risk past made", which accomplishes what I'm proposing

AndreasArvidsson:
> Btw one argument for what you're proposing here is that soon we will have "chuck every line risk past made", which accomplishes what I'm proposing

That one is actually interesting, but when it comes to removal if you just want to remove all lines stand here's no difference in doing "chuck line risk past made"
"take every line risk past made" is actually really useful

pokey:
Although we'll need next-gen slice semantics to support "string air slice bat", which I often find myself wanting in markdown tables of cursorless commands, so next-gen "every" doesn't hit every use case for next-gen "slice"

I think it's quite rare that you'd actually want the rigid vertical slice behaviour from this PR. I'm hesitant to go out of our way to support / test behaviour that we're planning to change soon, and which users have expressed surprise at

But tbh I don't feel too strongly on this one, as I don't think many users will hit it one way or the other. If you feel strongly I'm fine to merge this one in

src/processTargets/processTargets.ts
```diff
@@ -144,7 +146,17 @@ function processVerticalRangeTarget(
       anchorTarget.contentRange.end.character
     );
 
-    results.push(anchorTarget.withContentRange(contentRange));
+    if (anchorTarget instanceof PositionTarget) {
+      results.push(anchorTarget.withContentRange(contentRange));
+    } else {
+      results.push(
+        new PlainTarget({
```
pokey, line 153: Do we want to use a `WeakTarget` instead? Might be nice to clean up white space and do weak expansion, given these are more like a bunch of marks in my mind

AndreasArvidsson:
> Although we'll need next-gen slice semantics to support "string air slice bat", which I often find myself wanting in markdown tables of cursorless commands, so next-gen "every" doesn't hit every use case for next-gen "slice"
> 
> I think it's quite rare that you'd actually want the rigid vertical slice behaviour from this PR. I'm hesitant to go out of our way to support / test behaviour that we're planning to change soon, and which users have expressed surprise at
> 
> But tbh I don't feel too strongly on this one, as I don't think many users will hit it one way or the other. If you feel strongly I'm fine to merge this one in

I want that too. I'm not arguing against that next gen implementation. All I'm saying right now is that the changes we made to compositional grammar made the behavior of content and removal ranges for slice targets diverge which must be viewed as a regression. Once we actually can support modifies after sliced range this will work as intended for both selection and removal. 

Users has only expressed surprise of the content range of sliced targets which this pr does nothing to change. All I want is to have a consistency between selection and removal behavior. 

I'm going to go ahead and merge this and then we can change it when we can do it properly.

pokey:
Go for it, but see my comment re weak vs plain

src/processTargets/processTargets.ts
AndreasArvidsson: You do make a point but at the same time I do feel that the behavior of sliced targets right now is a fixed range that expands to a column and if I tell it to slice a specific column I don't want anything unexpected to change that range. This is a real edge case but I believe that plain target is more correct.

pokey: Ok we have pretty different expectations of slice semantics. Let's leave as plain for now but aim to discuss

Will you miss your rigid block slice semantics if we move to post-slice modifier application? If so then maybe we need two different kinds of slice?

AndreasArvidsson: Agreed we definitely need to iron this one out 

I guess in that case I can just use just/raw to get the same results. But that is definitely something we need to think about.

pokey: I'm not sure "just" helps you here. What exactly were you planning to say?

Fwiw we could just add a new range modifier called "strict slice"

AndreasArvidsson: "take line (just air slice bat)"
I have no idea how we should do the grammar so that line should run after but just should run before. hmm

Because I don't think we should make it so all modifiers ran after the slice range.

pokey: Ok let's plan to discuss

AndreasArvidsson: Definitely

AndreasArvidsson: We could change the grammar a bit
"take line slice air past bat"
"take slice line air past bat"

pokey: yeah worth thinking bout.  Fwiw I updated https://github.com/cursorless-dev/cursorless/issues/193 in prep for this discussion.  Lmk if that formalism makes sense

AndreasArvidsson: Works for me

PR #737, opened 2022-06-05
implement 'short paint' scope type

### LINKED ISSUE #587: Add "short paint" modifier
Defined as intersection of "paint" and "string", so for example:

```typescript
foo("bar-baz bongo")
```

Saying "take small paint zip" would select `bar-baz`.  This modifier will be really useful for tailwind.css

Once we have #69, we can implement this one using an `intersection` pipeline constructor, which will take `n` pipelines as its arguments, and construct a new pipeline stage that applies each of its pipeline arguments to its input, and outputs the intersection of the outputs of its pipelines

AndrewDant:
Closes #587

## Checklist


cursorless-talon/src/modifiers/containing_scope.py
```diff
@@ -69,6 +69,7 @@ def cursorless_containing_scope(m) -> dict[str, dict[str, Any]]:
     "file": "document",
     "line": "line",
     "paint": "nonWhitespaceSequence",
+    "small paint": "nonwhiteSpaceOrDelimiterSequence",
```
Will-Sommers, line 72: Heyo @AndrewDant — I'm wondering if we should consider moving this definition to [selection_type.py](https://github.com/cursorless-dev/cursorless/blob/main/cursorless-talon/src/modifiers/selection_type.py#L9) rather than defining a separate selection type. It seems to me like `small` is a modifier on `paint` and we implement an optional preceding parameter in the `selection_type` file. 

We could then send a different sub type of `SelectionType` into the TS side and map the regex that way. 

This is my preference but I'm also not up to speed on the fully compositional modifiers changes. 

I'm definitely down to talk about it, just let me know. Also interested to hear others' opinions.

src/processTargets/processSelectionType.ts
```diff
@@ -44,6 +44,13 @@ export default function (
         selection,
         selectionContext
       );
+    case "nonwhiteSpaceOrDelimiterSequence":
```
Will-Sommers, line 47: Since all three of these call the same function with a different regex, I wonder if the case should call through on 2 of the 3 and then then in the third, we look up the regex from a hash. Alternative would be to have the modifier as a class that has a default regex stored on it, and use that.

Will-Sommers:
Looking good!

cursorless-talon/src/modifiers/containing_scope.py
pokey: I think I'd prefer it as is. I'll prob remap it to a single word, which I couldn't do if it were a modifier on "paint"

src/processTargets/processSelectionType.ts
pokey: This is setup is going to change a lot when big PR merges in, so during conflict resolution it will in fact end up close to what you're proposing

pokey (commented):
Nice! Would be good to add a couple tests that exercise the "paint" aspect—all your tests just focus on the "small" part. So I'd have the following tests, each time testing both "clear" and "chuck":

- [ ] `"hello-world"`
- [ ] `"hello-world testing"`
- [ ] `"some hello-world test"`

pokey:
Also, re excluding other delimiters, it's a good question. I prob wouldn't want to exclude square brackets, because I'd prob want to be able to grab `foo[0]`. Maybe let's keep it to quotes for now?

AndrewDant:
> Also, re excluding other delimiters, it's a good question. I prob wouldn't want to exclude square brackets, because I'd prob want to be able to grab `foo[0]`. Maybe let's keep it to quotes for now?

I'm content with that.

cursorless-talon/src/modifiers/containing_scope.py
Will-Sommers: 💯  got it — I wanted to get some thoughts down to understand where we might differ.

pokey:
Should we switch this one to "short paint"? That will be more consistent with "long head" when we add that

cursorless-talon/src/modifiers/containing_scope.py
```diff
@@ -52,6 +52,7 @@
     "file": "document",
     "line": "line",
     "paint": "nonWhitespaceSequence",
+    "small paint": "nonWhitespaceOrQuoteSequence",
```
AndreasArvidsson, line 55: "nonWhitespaceOrQuoteSequence" is not an optimal name. "nonWhitespaceOrPairSequence" maybe?
@pokey Any ideas?

src/processTargets/getModifierStage.ts
```diff
@@ -29,6 +29,9 @@ import {
   UrlStage,
 } from "./modifiers/scopeTypeStages/RegexStage";
 import TokenStage from "./modifiers/scopeTypeStages/TokenStage";
+import SmallPaintStage, {
```
AndreasArvidsson, line 32: The stage should preferably be called the same as the modifier

src/processTargets/modifiers/SmallPaintStage.ts
```diff
@@ -0,0 +1,66 @@
...
+    });
+
+    const surroundingPairStage = new InteriorOnlyStage({
+      type: "interiorOnly",
+    });
+
+    const paintTargets = paintStage.run(context, target);
+
+    try {
+      const pairTargets = surroundingPairStage.run(context, target);
```
AndreasArvidsson, line 41: Instead of using the stage we could probably just call the function directly
https://github.com/cursorless-dev/cursorless/blob/256d656b2e84ebc6d29ab9ee5515f3eee396fb2d/src/processTargets/modifiers/SurroundingPairStage.ts#L37

Just check if `pairInfo` is not null and then do an intersection with `pairInfo.interiorRange`

src/processTargets/modifiers/SmallPaintStage.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Intersection of NonWhitespaceSequenceStage and InteriorOnlyStage
+ * Expand the target until reaching a white space or surrounding pair.
+ * If there is no surrounding pair defaults to the non white space sequence
+ */
+export default class SmallPaintStage implements ModifierStage {
+  constructor(private modifier: NonWhitespaceOrQuoteSequenceModifier) {}
+
+  run(context: ProcessedTargetsContext, target: Target): Target[] {
+    const paintStage = new NonWhitespaceSequenceStage({
+      type: "containingScope",
```
AndreasArvidsson, line 30: We don't just want containing scope. We should pass in the same type as used for the small paint modifier. ie if the user says "take every small paint" we want the call to `NonWhitespaceSequenceStage` to be every as well

cursorless-talon/src/modifiers/containing_scope.py
AndrewDant: Right, not accurate any longer. Maybe "nonWhitespaceSurroundingPairIntersect"? Pretty long but conceptually accurate.

pokey, line 55: As mentioned in Slack, I'd be tempted to go with

```suggestion
    "short paint": "nonWhitespaceOrQuoteSequence",
```

To continue the [`"short"` / `"long"`](http://redstartsystems.com/human-machine-grammar-the-rules#:~:text=the%20No%20button.-,Rule%2013,-Use%20phrase%20modes) terminology

src/core/commandVersionUpgrades/upgradeV1ToV2/commandV1.types.ts
```diff
@@ -84,6 +84,7 @@ type SelectionType =
   | "paragraph"
   | "document"
   | "nonWhitespaceSequence"
+  | "nonWhitespaceOrQuoteSequence"
```
pokey, line 87: These files should never change.  Or is that necessary for some old tests you have?  I guess that's ok? cc/ @AndreasArvidsson

cursorless-talon/src/modifiers/containing_scope.py
pokey: We generally prefer less verbose names for the identifier, to give us more flexibility to tweak behaviour in the future.  Also wrecks layout in the cheatsheet if they're too long 😅

Maybe `boundedNonWhitespaceSequence`?

src/processTargets/modifiers/SmallPaintStage.ts
pokey: Hmm not sure.  I'd be tempted to handle "every" specially for this modifier.  When target is empty, I'd expand to nearest containing "inside", and then run "every paint" on that.  Or maybe that's what you're suggesting?

Tbh I'm not really sure what to do with "every" when target is not empty 🤔

src/processTargets/modifiers/SmallPaintStage.ts
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Intersection of NonWhitespaceSequenceStage and InteriorOnlyStage
+ * Expand the target until reaching a white space or surrounding pair.
+ * If there is no surrounding pair defaults to the non white space sequence
+ */
+export default class SmallPaintStage implements ModifierStage {
+  constructor(private modifier: NonWhitespaceOrQuoteSequenceModifier) {}
+
+  run(context: ProcessedTargetsContext, target: Target): Target[] {
+    const paintStage = new NonWhitespaceSequenceStage({
```
pokey, line 29: I have mixed feelings about instantiating the stage directly.  Wouldn't it be better not to tightly couple ourselves to the class that implements the stage?  We have `getModifierStage`, so I'm not sure why we don't want to use it? cc/ @AndreasArvidsson

src/processTargets/modifiers/SmallPaintStage.ts
pokey: As above, I'd be inclined not to tightly couple where possible, so I'd lean towards using a stage.  But maybe this simplifies things somehow?

src/processTargets/modifiers/SmallPaintStage.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    const surroundingPairStage = new InteriorOnlyStage({
+      type: "interiorOnly",
+    });
+
+    const paintTargets = paintStage.run(context, target);
+
+    try {
+      const pairTargets = surroundingPairStage.run(context, target);
+      return zip(paintTargets, pairTargets).map(([paintTarget, pairTarget]) => {
```
pokey, line 42: why are we zipping here?

src/test/suite/fixtures/recorded/selectionTypes/changeSmallPaint.yml
```diff
@@ -0,0 +1,22 @@
+languageId: plaintext
+command:
+  version: 1
+  spokenForm: change small paint
+  action: clearAndSetSelection
+  targets:
+    - {type: primitive, selectionType: nonwhiteSpaceOrDelimiterSequence}
+initialState:
+  documentContents: "' another '"
```
pokey, line 9: You have a bunch of tests like this, where it's targeting one token.  It's worth having one test like this, but most tests should probably use something with a dash in it or something, eg `foo-bar`, to test the "paint" part of it, rather than focusing only on the "small" part 😊

pokey (commented):
Interesting direction.  I still don't totally understand how "every" should work, but I left a bunch of minor comments.

Fwiw, the advantage of the original approach is that "every" becomes trivial, though I think in that case I'd still want "every" to expand to nearest containing string if input target is empty

src/core/commandVersionUpgrades/upgradeV1ToV2/commandV1.types.ts
AndreasArvidsson: Agreed we should not change this file.

src/processTargets/modifiers/SmallPaintStage.ts
AndreasArvidsson: Just less abstraction. If that is good or not is something has probably varies from person to person.

src/processTargets/modifiers/SmallPaintStage.ts
AndreasArvidsson: Because the stages return an array but that will change for surrounding pair

AndreasArvidsson:
@pokey Every will be solved by just passing every to the paint modifier. Then running all the target through an intersection. Basically you will get every paint within the matching pair.

src/processTargets/modifiers/SmallPaintStage.ts
AndreasArvidsson: Every will be solved by just passing every to the paint modifier. Then running all the target through an intersection. Basically you will get every paint within the matching pair. If we have no surrounding pair paint and short paint becomes equivalent.

cursorless-talon/src/modifiers/containing_scope.py
AndrewDant: I personally quite like that suggestion.

src/core/commandVersionUpgrades/upgradeV1ToV2/commandV1.types.ts
AndrewDant: My bad, I was looking for anywhere that nonWhitespaceSequence appeared.

src/processTargets/modifiers/SmallPaintStage.ts
AndrewDant: I've updated this first instance to use getModifierStage, and the second one to call processSurroundingPair directly.

src/processTargets/modifiers/SmallPaintStage.ts
AndrewDant: Is the solution that I've gone with and I think the way works is pretty intuitive.

src/processTargets/modifiers/SmallPaintStage.ts
AndrewDant: I've gone with this suggestion because it does make the code simpler to write and understand.

AndreasArvidsson:
To fix the bug with paint just replace `range.end.isBeforeOrEqual(end)` with `range.start.isBeforeOrEqual(end)` at `RegexStage.ts:L41`

pokey:
Also a couple more tests to add:

- [ ] "change every small paint pair paren"
- [ ] "change every small paint inside paren"

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
```diff
@@ -38,7 +38,10 @@ class RegexStage implements ModifierStage {
     for (let i = start.line; i <= end.line; ++i) {
       this.getMatchesForLine(editor, i).forEach((range) => {
         // Regex match and selection intersects
-        if (range.end.isAfterOrEqual(start) && range.end.isBeforeOrEqual(end)) {
+        if (
+          range.end.isAfterOrEqual(start) &&
+          range.start.isBeforeOrEqual(end)
```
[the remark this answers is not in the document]
AndrewDant: I don't mind recording them but can you give some examples of what would test this? I don't really remember what the failing case was since I only saw it in that one discord call.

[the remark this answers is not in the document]
pokey: cc/ @AndreasArvidsson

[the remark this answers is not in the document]
AndreasArvidsson: `aa|aa bb|bb`
`"clear every paint"`

src/test/suite/fixtures/recorded/selectionTypes/changeEveryPaintPairParen2.yml
```diff
@@ -0,0 +1,36 @@
...
+  documentContents: foo(aa|aa-().c_d bb|bb-().'') { return 1; }
+  selections:
+    - anchor: {line: 0, character: 1}
+      active: {line: 0, character: 1}
+  marks:
+    default.(:
+      start: {line: 0, character: 3}
+      end: {line: 0, character: 4}
+finalState:
+  documentContents: "  { return 1; }"
```
AndrewDant, line 25: It may be unintuitive for some users that the first paint extends outside the selected range. Not sure if that's something we'd want to change though since it is one contiguous nonwhitespace sequence.

AndrewDant:
I've added more tests, we might be ready to go now but let me know if I missed anything.

src/test/suite/fixtures/recorded/selectionTypes/changeEveryPaintPairParen2.yml
AndreasArvidsson: This looks like the correct implementation with regards to what we talked about

pokey (approved):
Looks good! Ship it 🙌

Issue #740, opened 2022-06-06
Indicate that VSCode must be open for Cursorless help / reference commands


hakanErgin:
It is not clear in the documentation that these comments work only on vscode:
`cursorless help`, `cursorless instructions`

AndreasArvidsson:
This is now fixed

PR #748, opened 2022-06-07
Update installation to re-state 'restart talon' instructions

Will-Sommers:
Add extra "restart talon" statements.


pokey (approved):

pokey:
This PR is nice, but I don't think it has anything to do with #740 😅.  I've unlinked that issue

Will-Sommers:
Thanks! What I get for opening a PR at 9:41!

Issue #449, opened 2021-07-02
Add regression tests for voice commands


pokey:
We'd like to ensure that voice commands don't break.  Probably the most comprehensive way to do this would be to convert all of our cursorless VSCode [recorded tests](https://github.com/pokey/cursorless-vscode/tree/main/src/test/suite/fixtures/recorded) into voice command tests.  We'd basically take the `spokenForm` field of the test, call `mimic` on it, and ensure that its output matches the `command` in the test.  We'd need to do some bulk processing to fix the spoken forms that didn't come from knausj

This effort would require https://github.com/talonvoice/talon/issues/375 in order to run in CI, but will start by getting some tests running locally to function as a starting point for a conversation around our needs for a talon CI runner

See also [knausj test setup](https://github.com/knausj85/knausj_talon/blob/main/tests), esp:
- https://github.com/knausj85/knausj_talon/blob/main/tests/stubs/talon/__init__.py
- https://github.com/knausj85/knausj_talon/blob/main/tests/test_code_modified_function.py

Also maybe worth having a look at contract / hypothesis testing?
- https://github.com/mristin/icontract-hypothesis
- https://hypothesis.readthedocs.io/en/latest/index.html
- https://crosshair.readthedocs.io/en/latest/contracts.html

## Approach

- [x] Add Cursorless extension VSCode command `cursorless.generate_spoken_form_tests` that iterates recorded test cases and populates a tmp directory with spoken form tests, returning the path to that directory.  Might look at https://github.com/cursorless-dev/cursorless/blob/037389f87b16842e32511dbcffeca0d6e95969d2/packages/cursorless-engine/src/core/Cheatsheet.ts#L58 for inspiration
- [x] Add Talon tag `user.cursorless_default_vocabulary` that switches all spoken forms to their normal forms (eg stock knausj alphabet, no modifier overrides)
- [x] Add Talon action that does the following:
  1. Calls the `cursorless.generate_spoken_form_tests` VSCode action
  2. Enables `user.cursorless_default_vocabulary`
  3. Enables a tag that overrides the cursorless command actions from https://github.com/cursorless-dev/cursorless/blob/main/cursorless-talon/src/command.py to just assert that the payload about to be sent matches expectation, and then doesn't send the action
  4. Iterates through the generated tests and runs them using `mimic` on the spoken form, setting up the payload checker with the payload.  Possibly we should run 3) once per test case
  5. Disable both tags above
[...]

PR #749, opened 2022-06-07
Fix `"bring line before bat"`

### LINKED ISSUE #747: Fix `"bring line before bat"`
It should be interpreted as a two-target "bring", that brings the current line before line bat.  However, Talon interprets it as a one-arg bring with a "line before" modifier chain.

Maybe the easiest solution is to force positional modifiers to be the first in a modifier chain?

See also #464

pokey:
Fixes #747

## Checklist

- [x] Add tests (in the future they will turn into Talon tests via #449)
- [x] Do some local testing
- [x] Test DFA on public


pokey:
Unfortunately, this does affect DFA on public 😕

Avg on `main` with a Python file: 622ms
Avg on this branch with a Python file: 850ms
Avg on `main` with a Talon file: 627ms
Avg on this branch with a Talon file: 765ms

Btw not sure how our DFA crept up to ~620ms; I believe it used to be at or below 500ms 🤷.  But 200ms bump is pretty bad so I think we're going to have to add this one to the list of PRs waiting on 0.3 😕

AndreasArvidsson:
@pokey The dfa improvements are now shipped on all platforms afaik

AndreasArvidsson (approved):

PR #751, opened 2022-06-07
Disabled escaped characters in tokenizer

### LINKED ISSUE #726: Removed escaped characters from the tokenizer
Today `\t` and `\r` creates problem with file paths in windows
`...\recorded\typescript...`

pokey:
Fixes #726

Originally opened as #727, but that suffered from a botched git surgery from which it never recovered 💀

## Checklist


pokey:
Sorry I had to re-open; lmk if I botched anything.  Also would be good to add a test including a Windows path so this doesn't break in the future

pokey (commented):
Looks fine to me, but

- [ ] would be good to add a test including a Windows path so this doesn't break in the future

AndreasArvidsson (approved):

pokey:
Looks good to me feel free to merge

Issue #694, opened 2022-05-26
Migrate v1 test cases to v2


pokey:
We'd like to upgrade all test cases to the new v2 api introduced in #672.  We would like to duplicate a few tests that use v1, though, to ensure that the on-the-fly command upgrader works.  We can proceed as follows:

- [ ] Look through the v1 to v2 upgrade code and figure out what needs to be tested
- [ ] Add a field to tests that cause the [upgrader](https://www.cursorless.org/docs/contributing/test-case-recorder/#upgrading-fixtures) to ignore them
- [ ] Select tests to test the paths through the upgrade code and duplicate them to a backwards compatibility test folder, adding the above field
- [ ] Run the upgrader
- [ ] Rename "chuck after" / "chuck after" tests and change spoken form

pokey:
done in https://github.com/cursorless-dev/cursorless/pull/2433

PR #764, opened 2022-06-11
Fix bug in testcase upgrade script

pokey:
## Checklist

### DIFF
--- a/src/core/commandRunner/CommandRunner.ts
+++ b/src/core/commandRunner/CommandRunner.ts
@@ -1,6 +1,6 @@
 import * as vscode from "vscode";
 import { ActionType } from "../../actions/actions.types";
-import { ActionableError } from "../../errors";
+import { OutdatedExtensionError } from "../../errors";
 import processTargets from "../../processTargets";
 import { Graph, ProcessedTargetsContext } from "../../typings/Types";
 import { isString } from "../../util/type";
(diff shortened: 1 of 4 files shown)


src/core/commandRunner/CommandRunner.ts
```diff
@@ -155,6 +155,21 @@ export default class CommandRunner {
...
+    const item = await vscode.window.showErrorMessage(
+      err.message,
+      "Check for updates"
+    );
+
+    if (item == null) {
+      return;
+    }
+
+    await vscode.commands.executeCommand(
```
Will-Sommers, line 168: This is basically an auto-upgrade? Do we require a restart?

Will-Sommers (approved):
lgtm

pokey:
@Will-Sommers this one fixes the vscode import issue in the test case upgrade script.  Fwiw the test case upgrade script still can't be run until we tackle https://github.com/cursorless-dev/cursorless/issues/694.  But just to be clear, the test case upgrade script is different from the snapshot updater.  The former is a script that upgrades the command versions in recorded tests to the latest; the latter is a launch config that you run instead of Run Tests, that will update the post-command snapshot instead of failing if it differs from expected

src/core/commandRunner/CommandRunner.ts
pokey: I don't think so, I believe it just checks for upgrades, and the user can decide what to do.  Cursorless shouldn't require a restart on upgrade

PR #766, opened 2022-06-12
Breakpoint <target> puts a breakpoint on the current line instead of the target line

### LINKED ISSUE #761: Breakpoint <target> puts a breakpoint on the current line instead of the target line
If I say something like 'breakpoint air' in vscode cursorless will put a breakpoint on the current line and then talon will type the letter a.

pokey:
Fixes #761

## Checklist


auscompgeek (approved):
LGTM. Is there a plan to inform existing users that their setup might need fixing?

PR #768, opened 2022-06-13
Fix Notebook CI failures; pin vscode version

### LINKED ISSUE #767: Fix CI errors related to notebook cells
We currently have breakages in CI, eg https://github.com/cursorless-dev/cursorless/runs/6844372300?check_suite_focus=true.  They are likely due to the new VSCode version changing the notebook API, which is probably breaking the following
https://github.com/cursorless-dev/cursorless/blob/004ce1405162efda251826f6ae01c1fc1bf697db/src/util/notebook.ts#L10

pokey:
Fixes #767
Replaces #497

Adds backwards compatibility checks for old VSCode versions

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,13 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
+        vscode_version: [stable]
+        include:
+          - os: ubuntu-latest
+            vscode_version: 1.58.1
     runs-on: ${{ matrix.os }}
+    env:
+      VSCODE_VERSION: ${{ matrix.vscode_version }}
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
--- a/package.json
+++ b/package.json
@@ -15,10 +15,11 @@
 		"url": "https://github.com/cursorless-dev/cursorless.git"
 	},
 	"engines": {
-		"vscode": "^1.61.0"
+		"vscode": "^1.58.0"
 	},
 	"extensionKind": [
-		"workspace"
+		"workspace",
+		"ui"
 	],
 	"categories": [
 		"Other"
@@ -519,7 +520,7 @@
 		"@types/node": "^16.11.3",
 		"@types/semver": "^7.3.9",
 		"@types/sinon": "^10.0.2",
-		"@types/vscode": "^1.61.0",
+		"@types/vscode": "~1.58.0",
 		"@typescript-eslint/eslint-plugin": "^5.20.0",
 		"@typescript-eslint/parser": "^5.20.0",
 		"esbuild": "^0.11.12",
--- a/src/test/runTest.ts
+++ b/src/test/runTest.ts
@@ -7,6 +7,7 @@ import {
   downloadAndUnzipVSCode,
 } from "vscode-test";
 import { extensionDependencies } from "./extensionDependencies";
+import { env } from "process";
 
 async function main() {
   try {
@@ -18,7 +19,8 @@ async function main() {
     // Passed to --extensionTestsPath
     const extensionTestsPath = path.resolve(__dirname, "./suite/index");
 
-    const vscodeExecutablePath = await downloadAndUnzipVSCode();
+    const vscodeVersion = env.VSCODE_VERSION ?? "stable";
+    const vscodeExecutablePath = await downloadAndUnzipVSCode(vscodeVersion);
     const cliPath =
       resolveCliPathFromVSCodeExecutablePath(vscodeExecutablePath);
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -152,15 +152,15 @@
   dependencies:
     "@types/sinonjs__fake-timers" "*"
 
+"@types/vscode@~1.58.0":
+  version "1.58.1"
+  resolved "https://registry.yarnpkg.com/@types/vscode/-/vscode-1.58.1.tgz#7deae08792adc73fa57383244a0c79d3530df4f7"
+  integrity sha512-sa76rDXiSif09he8KoaWWUQxsuBr2+uND0xn1GUbEODkuEjp2p7Rqd3t5qlvklfmAedLFdL7MdnsPa57uzwcOw==
+
 "@types/sinonjs__fake-timers@*":
   version "8.1.2"
   resolved "https://registry.yarnpkg.com/@types/sinonjs__fake-timers/-/sinonjs__fake-timers-8.1.2.tgz#bf2e02a3dbd4aecaf95942ecd99b7402e03fad5e"
   integrity sha512-9GcLXF0/v3t80caGs5p2rRfkB+a8VBGLJZVih6CNFkx8IZ994wiKKLSRs9nuFwk1HevWs/1mnUmkApGrSGsShA==
-
-"@types/vscode@^1.61.0":
-  version "1.66.0"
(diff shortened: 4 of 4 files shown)


third-party-licenses.csv
```diff
@@ -1,19 +1,11 @@
...
 "ansicolor@1.1.100","Unlicense","https://github.com/xpl/ansicolor","https://github.com/xpl/ansicolor/raw/master/LICENSE","metals"
 "brace-expansion@2.0.1","MIT","https://github.com/juliangruber/brace-expansion","https://github.com/juliangruber/brace-expansion/raw/master/LICENSE","vscode-fileutils"
-"clsx@1.1.1","MIT","https://github.com/lukeed/clsx","https://github.com/lukeed/clsx/raw/master/license","website"
 "immutability-helper@3.1.1","MIT","https://github.com/kolodny/immutability-helper","https://github.com/kolodny/immutability-helper/raw/master/LICENSE","cursorless"
 "jsonc-parser@2.3.1","MIT","https://github.com/microsoft/node-jsonc-parser","https://github.com/microsoft/node-jsonc-parser/raw/master/LICENSE.md","parse-tree"
-"mdast-util-find-and-replace@2.1.0","MIT","https://github.com/syntax-tree/mdast-util-find-and-replace","https://github.com/syntax-tree/mdast-util-find-and-replace/raw/master/license","website"
+"lodash@4.17.21","MIT","https://github.com/lodash/lodash","https://github.com/lodash/lodash/raw/master/LICENSE","cursorless"
```
auscompgeek, line 7: Interesting that this list has changed, despite the affected packages not changing in package.json or yarn.lock.

.github/workflows/test.yml
```diff
@@ -15,7 +15,13 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
+        vscode_version: [stable]
+        include:
+          - os: ubuntu-latest
+            vscode_version: 1.66.0
```
pokey, line 21: Note that we now test on VSCode version `1.66.0`.  I figured two versions back was good enough.  We can bump this one in the future if we want

package.json
```diff
@@ -18,7 +18,8 @@
 		"vscode": "^1.61.0"
 	},
 	"extensionKind": [
-		"workspace"
+		"workspace",
+		"ui"
```
pokey, line 22: Added this based on user requests.  Should help with running in various VSCode contexts, even if command-server might not work there

package.json
```diff
@@ -519,7 +520,7 @@
 		"@types/node": "^16.11.3",
 		"@types/semver": "^7.3.9",
 		"@types/sinon": "^10.0.2",
-		"@types/vscode": "^1.61.0",
+		"@types/vscode": "~1.61.0",
```
pokey, line 523: Pinned for now so that we don't break backwards compatibility

.github/workflows/test.yml
```diff
@@ -15,7 +15,13 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
+        vscode_version: [stable]
+        include:
+          - os: ubuntu-latest
+            vscode_version: legacy
```
pokey, line 21: Note that we now test on older VSCode versions as well

src/test/runTest.ts
```diff
@@ -18,7 +19,8 @@ async function main() {
     // Passed to --extensionTestsPath
     const extensionTestsPath = path.resolve(__dirname, "./suite/index");
 
-    const vscodeExecutablePath = await downloadAndUnzipVSCode();
+    const vscodeVersion = env.VSCODE_VERSION === "legacy" ? "1.66.0" : "stable";
```
pokey, line 22: I figured two versions back was good enough.  We can bump this one in the future if we want

src/test/runTest.ts
```diff
@@ -18,7 +19,11 @@ async function main() {
     // Passed to --extensionTestsPath
     const extensionTestsPath = path.resolve(__dirname, "./suite/index");
 
-    const vscodeExecutablePath = await downloadAndUnzipVSCode();
+    // NB: We include the exact version here instead of in `test.yml` so that
+    // we don't have to update the branch protection rules every time we bump
+    // the legacy VSCode version.
```
auscompgeek, line 24: Hmm, you could set the job name using the matrix, but I guess this works as well.

Alternatively we could have a job that requires test, but seems excessive for what we need.

src/util/notebook.ts
```diff
@@ -8,21 +16,39 @@ import { NotebookDocument, TextDocument } from "vscode";
  * given cell
  */
 export function getNotebookFromCellDocument(document: TextDocument) {
-  return (document as any).notebook as NotebookDocument | undefined;
+  if (lt(version, "1.68.0")) {
+    return getNotebookFromCellDocumentLegacy(document);
+  }
+
+  // FIXME: All these type casts are necessary because we've pinned VSCode
+  // version type defs.  Can remove them once we are using more recent type defs
```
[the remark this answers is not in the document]
Will-Sommers: I'm not fully understanding why we can't bump the VSCode type defs, though. Can you walk through it?

src/util/setSelectionsAndFocusEditor.ts
```diff
@@ -46,48 +55,47 @@ export async function focusEditor(editor: TextEditor) {
   } else {
     // If the view column is null we see if it's a notebook and try to see if we
     // can just move around in the notebook to focus the correct editor
-    const activeTextEditor = window.activeTextEditor;
 
-    if (activeTextEditor == null) {
-      return;
+    if (lt(version, "1.68.0")) {
```
Will-Sommers, line 59: Might be overkill, but it might make sense to store this in a constants file. We should likely check `lt` and also ensure `gt` from some baseline.

src/util/setSelectionsAndFocusEditor.ts
```diff
@@ -46,48 +55,47 @@ export async function focusEditor(editor: TextEditor) {
...
-      );
-    }
+  await commands.executeCommand(
+    columnFocusCommands[
+      desiredNotebookEditor.viewColumn as keyof typeof columnFocusCommands
+    ]
+  );
 
-    const cellOffset = editorIndex - activeEditorIndex;
+  const activeTextEditor = window.activeTextEditor;
```
Will-Sommers, line 82: It seems like we do this in many cases, might be interesting to think if there are ways we can add an AOP type filter for various actions.

Will-Sommers (dismissed):

src/util/setSelectionsAndFocusEditor.ts
pokey: Sorry not sure I'm following. Fwiw looks like I don't use the active text editor below, so I can remove this null check. It was a holdover from a previous approach. But maybe that's not what you're referring to?

src/util/notebook.ts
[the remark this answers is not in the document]
pokey: There is a check in the `vsce package` command that checks whether the [minimum VSCode engine version](https://github.com/cursorless-dev/cursorless/blob/5f75975f577cdb43e68a6b0a9cfafb1f8560dd01/package.json#L18) is less than the [VSCode type defs version](https://github.com/cursorless-dev/cursorless/blob/5f75975f577cdb43e68a6b0a9cfafb1f8560dd01/package.json#L523).  So if we use the most recent type defs, we will have to bump our VSCode minimum engine, which would force all users to upgrade to the very latest VSCode, which seems a bit aggressive

Maybe there's a way to disable that check?

src/test/runTest.ts
pokey: I'm going to ship this one as is to get CI working again, but feel free to add comments here if you have a better solution

Issue #559, opened 2022-02-13
Switch back to our own copy-paste implementation


pokey:
I think the best way to solve this is
* Revert to our own implementation of copy/cut/paste
* Try to leverage as much from bring/move as we can
* Use the same implementation that vscode does with an internal clipboard per cursor copied
* Support clipboard history. 
* Contribute our own ctrl/cmd+c/x/v keyboard bindings with a when clause that the editor is focused

@pokey Sounds good?

_Originally posted by @AndreasArvidsson in https://github.com/cursorless-dev/cursorless-vscode/issues/128#issuecomment-1024389906_

PR #771, opened 2022-06-15
Support delimiter insertion for "paste after"

### LINKED ISSUE #128: Add "after" delimiter-adding support to "paste"
See https://github.com/pokey/cursorless-vscode/blob/master/src/actions/BringMoveSwap.ts#L111-L121

pokey:
Fixes #128
The main drawback here is that "paste after" inserts an extra undo stop.  I think that's unavoidable until we have #559 

## Checklist


src/actions/actions.types.ts
```diff
@@ -8,6 +8,7 @@ export type ActionType =
   | "copyToClipboard"
   | "cutToClipboard"
   | "deselect"
+  | "editNew"
```
pokey, line 11: Added this so that paste and insert snippet can use it; could be useful to map a spoken form at some point 🤷.  Kinda like a fancy "change" that will insert delimiters if you give it a positional modifier

src/test/suite/fixtures/recorded/actions/pasteAfterLineSpunAndAfterBlockLookAndBeforeLineSpun.yml
```diff
@@ -0,0 +1,82 @@
+languageId: plaintext
+command:
+  spokenForm: paste after line spun and after block look and before line spun
```
pokey, line 3: Shockingly, Cursorless handled this one with no problem, inserting the right delimiters and leaving cursors in the right spot 😯

src/testUtil/TestCase.ts
```diff
@@ -139,8 +140,20 @@ export class TestCase {
...
+      ? ["pasteFromClipboard"]
+      : ["copyToClipboard", "cutToClipboard"];
+
+    const visibleRangeActions: ActionType[] = [
+      "foldRegion",
+      "unfoldRegion",
+      "scrollToBottom",
+      "scrollToCenter",
+      "scrollToTop",
+    ];
```
pokey, line 153: These had rotted when we upgraded action names.  I added explicit types to prevent that happening in the future

src/actions/actions.types.ts
AndreasArvidsson: Then maybe we should just merge this with change and allowed that option?

pokey: That was my suggestion at meet-up, but the issue is that highlighting becomes a bit complex, as the targets will need to own a bit more. I'm not opposed to it, but maybe out of scope for this PR? We'll need to think a bit about giving targets more autonomy wrt highlighting

AndreasArvidsson: I think we can do that in a follow up pr

AndreasArvidsson: I mean we already have removal highlight range in the targets so we could just add content highlight range if we want.

pokey: I think there is a fairly clean solution here that would also simplify the highlight graph component, but it's worth pairing on and I think out of scope

If you prefer for this PR I can just have the paste action keep its own `EditNew` object; that was my original solution

AndreasArvidsson (approved):

Issue #579, opened 2022-02-24
Add keyword for alternatives in a match construct


wenkokke:
## _edit @pokey_:

As discussed in https://github.com/cursorless-dev/cursorless/issues/579#issuecomment-1161627378, we'd like to proceed as follows:

- [x] #1172
- [ ] #1173
- [x] #1174
- [ ] #1175
- [x] #1176

## Original
There currently is no keyword for selecting alternatives in a match construct. Examples of match constructs are:
- Python 
   ```python
   match x:
   case Alt1: exp1
   …
   case AltN: expN
   ```
- Haskell
   ```haskell
   f = case x of
     Alt1 -> exp1
     …
     AltN -> expN
   ```
   or
   ```haskell
   f Alt1 = exp1
   …
   f AltN = expN
   ```

We probably want three keywords:
- `matched` to select the value matched on, i.e., `x`;
- `<ordinal> pattern` to select the *i*’th pattern, i.e., `Alt1` to `AltN` above; and
- `<ordinal> case` to select the *i*’th case, i.e., `exp1` to `expN` above.

Unmodified `pattern` and `case` should probably to the scopes relative to the named mark or cursor, i.e., `pattern` refers to `Alt1` if the mark or cursor is in either `Alt1` or `exp1`, and similarly for `case`.

pokey:
Good stuff thanks for filing.  A couple thoughts:

- I'd be tempted to fold "pattern" into "value" and "matched" into "arg", but don't feel super strongly
- I wouldn't worry about ordinals; those should come for free with #51 as long as you implement them such that  "every" works

wenkokke:
Patterns and values are rather different and merging them might let to very confusing selections...

Are you suggesting that "arg" matches the matched value when the cursor is on "match"? There could be arbitrarily nested function calls in x, so having a keyword that jumps directly to the top of the match construct might be extremely useful?

pokey:
Let's talk this one out when we discuss the haskell stuff.  I like to merge when I can to keep vocabulary down but when it doesn't make sense I'm always happy to add another scope type

purpleP:
I think it makes more sense to discuss this while we’re working on russet support as rust pattern matching capabilities and syntax is more complex than Haskell.

pokey:
## General philosophy

> That’s probably a technical limitation, but if it isn’t I would actually allowed each language to define it’s own scopes instead of trying to define a set of common ones and having languages where not all of them are defined

[...]

PR #775, opened 2022-06-17
Feature/rust support

purpleP:
## What

Adds support for the Rust programming language

## Checklist

- [x] Support pattern matching
- [x] Support tuples
- [x] Support RHS `value`
- [x] Support attributes


pokey:
Don't forget to use [PR template](https://github.com/cursorless-dev/cursorless/blob/main/.github/PULL_REQUEST_TEMPLATE/new_programming_language.md?plain=1) 😊

src/languages/rust.ts
```diff
@@ -0,0 +1,106 @@
...
+  leadingMatcher,
+  trailingMatcher,
+  cascadingMatcher,
+} from "../util/nodeMatchers";
+import { NodeMatcherAlternative } from "../typings/Types";
+import { SimpleScopeTypeType } from "../typings/targetDescriptor.types";
+
+// Generated by the following command:
+// `curl https://raw.githubusercontent.com/tree-sitter/tree-sitter-rust/master/src/node-types.json | jq '[.[] | select(.type == "_declaration_statement" or .type == "_expression") | .subtypes[].type]'`
+const STATEMENT_TYPES = [
```
pokey, line 13: https://github.com/cursorless-dev/cursorless/pull/546/files#r811864233

src/test/suite/fixtures/recorded/languages/rust/changeState.yml
```diff
@@ -0,0 +1,28 @@
...
+  version: 1
+  spokenForm: change state
+  action: clearAndSetSelection
+  targets:
+    - type: primitive
+      modifier: {type: containingScope, scopeType: statement, includeSiblings: false}
+initialState:
+  documentContents: |-
+    let expensive_closure = |num| {
+        println!("calculating slowly...");
```
pokey, line 12: https://github.com/cursorless-dev/cursorless/pull/546/files#r830060769

src/languages/rust.ts
```diff
@@ -0,0 +1,106 @@
...
+  functionCall: ["call_expression", "macro_invocation"],
+  comment: ["line_comment", "block_comment"],
+  list: "array_expression",
+  namedFunction: "function_item",
+  type: leadingMatcher(
+    [
+      "let_declaration[type]",
+      "parameter[type]",
+      "generic_type",
+      "struct_item.type_identifier!",
```
pokey, line 91: this should be the whole struct, not just the identifier; +test 😊

src/languages/rust.ts
```diff
@@ -0,0 +1,106 @@
...
+  functionName: ["function_item[name]"],
+  anonymousFunction: "closure_expression",
+  argumentOrParameter: cascadingMatcher(
+    argumentMatcher("arguments"),
+    trailingMatcher(["parameter"], [","]),
+    trailingMatcher(["self_parameter"], [","])
+  ),
+  name: ["let_declaration.identifier!", "parameter.identifier!"],
+  class: ["struct_item", "struct_expression"],
+};
```
pokey, line 104: https://github.com/cursorless-dev/cursorless/pull/546/files#r811870350

src/languages/rust.ts
```diff
@@ -0,0 +1,106 @@
...
+  "expression_statement",
+];
+
+const nodeMatchers: Partial<
+  Record<SimpleScopeTypeType, NodeMatcherAlternative>
+> = {
+  statement: STATEMENT_TYPES,
+  string: ["raw_string_literal", "string_literal"],
+  ifStatement: "if_expression",
+  functionCall: ["call_expression", "macro_invocation"],
```
pokey, line 82: https://github.com/cursorless-dev/cursorless/pull/546/files#r811869556

pokey:
> @pokey I removed ability to use `arg` modifier on type arguments as one can use `type` modifier on them already.

"type" should not match type args.  It should only match the top-level type.  So "arg" is still useful

> What is still missing in this PR?

It's quite close, but I don't think most of my feedback from https://github.com/cursorless-dev/cursorless/pull/546#pullrequestreview-889681412 has been addressed, including the comments in the body of that review as well as the inline comments.  I added comments on this PR linking to comments on the original PR in case helpful, but see also description in that review

> And what should we do with `match` and pattern matching in general?

I believe that issue is not yet resolved in Cursorless; feel free to chime in on https://github.com/cursorless-dev/cursorless/issues/579, but I don't think it needs to be in scope for this PR.  But maybe I've missed your meaning

src/languages/rust.ts
purpleP: Not sure I follow you there. There is no if statement in Rust, only if expression. You just don’t have to use the returned value, that’s all. I will probably need some examples of what you want here. Maybe like the ones I did with hand-drawn scopes.

purpleP:
> "type" should not match type args. It should only match the top-level type. So "arg" is still useful

Why? It is a type… And what’s even top-level type mean? Do you mean type definition? But then it should match type annotations also… I’m a bit confused here.

I’ve added a test for associated function call (what you called a static method). Is it okay?

There is a way to instantiate a type in Rust like this.
4482DEA4-AE33-4651-800E-5A67CC86F23C
What do you want a user to be able to do here?

I think I’ve managed to create a query to match array items. But I don’t know how to express the same in cursorless.
D124D615-BE23-4B65-BE74-73339BBE8EE9

src/languages/rust.ts
auscompgeek: What do we do for other expression-oriented languages? I recall we have Clojure support?

pokey:
> Why? It is a type… And what’s even top-level type mean? Do you mean type definition? But then it shouldn’t  match type annotations also… I’m a bit confused here.

"type" is short for "type annotation" or "type declaration". A type arg is neither. So in that specific case by top-level type I meant the type annotation

> I’ve added a test for associated function call (what you called a static method). Is it okay?

Yep looks good! Minor comment on that test tho: for future tests please use "change" instead of "chuck" unless you're specifically testing removal behaviour (eg cleaning up trailing delimiters)

> 4482DEA4-AE33-4651-800E-5A67CC86F23C

This one should be "call"

> 5789D837-F262-4698-9DDD-E5954204DA2B

This one should be "type" (because it is a type declaration)

> A0212520-3D7C-444A-B40C-274B08291B44

This one should be "call"

> Also I think I’ve managed to make `item` work for arrays. Could you take a look?

Yep looks good! Please add a test or two for "change item" as well though

> ED94AEF5-5FD4-4A45-B146-F57BB5E44AB0

Yes this should be "type"

pokey:
Also prob worth adding tests for the modifier cases discussed in https://github.com/cursorless-dev/cursorless/pull/546#issuecomment-1158864553

src/languages/rust.ts
pokey: A statement is supposed to be a standalone thing.  Here are some examples:

## Statements

image
image
image
image

## Not statements

image
image
image
image

purpleP:
> "type" is short for "type annotation" or "type declaration". A type arg is neither. So in that specific case by top-level type I meant the type annotation

Do you want me to try to complicate a matcher to exclude types that are type arguments? Because as I’ve said from the point of tree-sitter parser these all just types, so it would require additional effort.

pokey:
> Do you want me to try to complicate a matcher to exclude types that are type arguments? Because as I’ve said from the point of tree-sitter parser these all just types, so it would require additional effort.

Yes, please do. That is how it works in all other languages; I'm not sure why we would want Rust to behave differently here?

pokey:
> I don’t understand why cursorless can’t recognise `state` here although there is special `expression_statement ` in the parser. 49ABA7DF-57B5-49CF-B13C-C3E2C4C32CE4 1E0076F3-CF48-4A31-9147-53B67452B770

I think that was due to an old version of tree-sitter-rust.  I just published a new version of the [parse tree extension](https://marketplace.visualstudio.com/items?itemName=pokey.parse-tree) which should fix that.  I also took the liberty of implementing "state" in https://github.com/purpleP/cursorless-vscode/pull/1; feel free to merge it into your branch if it looks ok

One thing I didn't know quite what to do with was the optional final expression at the end of a function.  Eg

image

Should this be "state"?  Should it be "value"?  Both?  Might be cool if it could be referred to as "value" from anywhere within the function

Cc/ @auscompgeek 

> I don’t understand how to make cursorless distinguish between `impo_item` and `impl_item[type]` here.
> 
> A34747CE-3C80-49FA-AC30-238179D8804A
> 
> 105CCC8B-91E5-427C-B1EE-24200B9B32D1

I haven't tested locally, but I'm guessing that will be fixed with the new parse tree extension mentioned above; try upgrading that extension and see if things start working better

> Also what should I do with tuples? Allow `item` to be used to select it’s parts?

Makes sense to me!

auscompgeek:
I'm honestly not sure what scope I would call the return value of a function. We have an expression scope right? At the very least that should work, and should hopefully work consistently between languages.

pokey:
We use "value" in other languages. Eg in typescript:

```typescript
function foo() {
   return "hello";
}
```

You could select `"hello"` from anywhere in the return statement by saying `"take value"`. So for consistency I think I'd argue that we should support that as well, both for explicit return statements and for the optional expression at the end of a function. The only issue is that to me, as a non-Rust programmer, it looks like a statement. So I would be tempted to also support "state" for that, as I don't see much harm in doing so. 

The additional question is whether the user should be able to say "value" from anywhere in the block to refer to it.  If it is the block of a function, it might be nice to support saying it from anywhere in the function, not just the block

purpleP:
> I also took the liberty of implementing "state" in [purpleP#1](https://github.com/purpleP/cursorless-vscode/pull/1); feel free to merge it into your branch if it looks ok

Thanks I took a look on how it was implemented in ruby, but I found that rust tree-sitter parser has special type for these if expressions where returned value isn’t used. So I don’t see the need in these `ancestorChain` thing. But maybe I’m missing something.

pokey:
> Thanks, I took a look on how it was implemented in ruby, but I found that rust tree-sitter parser has special type for these if expressions where returned value isn’t used. So I don’t see the need for these `ancestorChain` thing. But maybe I’m missing something.

Yes, that's true for expression statements, but not for the other kinds of statements in the list, eg macros

src/test/suite/fixtures/recorded/languages/rust/chuckFunkNameFine.yml
```diff
@@ -0,0 +1,33 @@
+languageId: rust
```
pokey, line 1: Nice.  Worth also testing that you can just use "name" for this one as well

Also, not worth re-recording, but we prefer "change" to "chuck" if you're not explicitly trying to test removal semantics, as otherwise you're implicitly testing our whitespace removal code

src/test/suite/fixtures/recorded/languages/rust/chuckTypeLook.yml
```diff
@@ -0,0 +1,45 @@
+languageId: rust
```
pokey, line 1: 👍

src/test/suite/fixtures/recorded/languages/rust/chuckTypeOdd.yml
```diff
@@ -0,0 +1,36 @@
+languageId: rust
```
pokey, line 1: 👍

src/test/suite/fixtures/recorded/languages/rust/chuckValueNear.yml
```diff
@@ -0,0 +1,32 @@
...
+  selections:
+    - anchor: {line: 0, character: 24}
+      active: {line: 0, character: 24}
+  marks:
+    default.n:
+      start: {line: 0, character: 9}
+      end: {line: 0, character: 13}
+finalState:
+  documentContents: |
+    let x = ;
```
pokey, line 25: Is it valid to have eg `let x: u32` without an ` = ` (ie declaration without initialization)?  If so, I'd be tempted to remove the ` = ` when you "chuck value" here.  Would leave it for `"change"`

src/test/suite/fixtures/recorded/languages/rust/clearState4.yml
```diff
@@ -0,0 +1,32 @@
...
+        foo("bar");
+    }
+  selections:
+    - anchor: {line: 1, character: 8}
+      active: {line: 1, character: 8}
+  marks: {}
+finalState:
+  documentContents: |-
+    fn foo() {
+
```
pokey, line 24: see how `;` is removed here, but not in your test above.  Strange

src/languages/rust.ts
```diff
@@ -0,0 +1,75 @@
...
+    ),
+    leadingMatcher(["function_item[return_type]"], ["->"]),
+    patternMatcher("struct_item", "trait_item", "impl_item", "impl_item[type]")
+  ),
+  functionName: ["function_item[name]"],
+  anonymousFunction: "closure_expression",
+  argumentOrParameter: cascadingMatcher(
+    argumentMatcher("arguments"),
+    trailingMatcher(["parameter"], [","]),
+    trailingMatcher(["self_parameter"], [","])
```
pokey, line 68: I don't think this will work in all cases.  I believe it will break if you say "chuck arg" on the final parameter if there is no trailing comma.  In that case, it is not a trailing `,` that you want to delete, but leading.  That's why we always use `argumentMatcher` for comma-separated lists.  Admittedly that name could be clearer, to indicate it's for all comma-separated lists 😅.  Maybe you'll need to tweak `argumentMatcher` to support specifying child type rather than parent type?

src/languages/rust.ts
```diff
@@ -0,0 +1,75 @@
...
+  ),
+  name: ["let_declaration.identifier!", "parameter.identifier!"],
+  class: ["struct_item", "struct_expression"],
+  value: ["let_declaration[value]", "block[-1]"],
+};
```
pokey, line 73: Looks like you're missing `className`

src/languages/rust.ts
```diff
@@ -0,0 +1,75 @@
...
+    patternMatcher("struct_item", "trait_item", "impl_item", "impl_item[type]")
+  ),
+  functionName: ["function_item[name]"],
+  anonymousFunction: "closure_expression",
+  argumentOrParameter: cascadingMatcher(
+    argumentMatcher("arguments"),
+    trailingMatcher(["parameter"], [","]),
+    trailingMatcher(["self_parameter"], [","])
+  ),
+  name: ["let_declaration.identifier!", "parameter.identifier!"],
```
pokey, line 70: This should also include class names and function names

src/languages/rust.ts
```diff
@@ -0,0 +1,128 @@
...
+  list: ["array_expression", "tuple_expression"],
+  collectionItem: argumentMatcher("array_expression", "tuple_expression"),
+  namedFunction: "function_item",
+  type: cascadingMatcher(
+    leadingMatcher(
+      ["let_declaration[type]", "parameter[type]", "field_declaration[type]"],
+      [":"]
+    ),
+    leadingMatcher(["function_item[return_type]"], ["->"]),
+    patternMatcher("struct_item", "trait_item", "impl_item", "impl_item[type]")
```
pokey, line 115: Hmm looks like the order got switched back. But either way it just occurred to me that this won't actually work with either order. The problem is that with `impl_item[type]`, the node matcher is still just looking for an `impl_item`, and then returning `[type]` if that field exists. That means that this matcher will return the type no matter where you are in the `impl_item`

@AndreasArvidsson I'm assuming there's no way to do this using our pattern matchers?

Your best bet is prob to write a custom finder. Shouldn't be too bad. Basically just check if parent is `impl_item` and if the node is its `[type]` field

AndreasArvidsson: You are correct

purpleP: How do I check if the node is inside his `type` field? By checking ranges?

pokey: Write a finder where given `node`, check if `node.parent.type === `"impl_item"` and then use [`getChildNodesForFieldName`](https://github.com/cursorless-dev/cursorless/blob/81d94e2b3926499c0a85299c023fde8da66c28d3/src/util/treeSitterUtils.ts#L18): `getChildNodesForFieldName(node.parent, "type") === node`

src/languages/rust.ts
```diff
@@ -0,0 +1,128 @@
...
+  functionName: ["function_item[name]"],
+  anonymousFunction: "closure_expression",
+  argumentOrParameter: argumentMatcher("arguments", "parameters"),
+  name: ["let_declaration.identifier!", "parameter.identifier!"],
+  class: ["struct_item", "struct_expression"],
+  value: cascadingMatcher(
+    patternMatcher("let_declaration[value]"),
+    matcher(returnValueFinder)
+  ),
+};
```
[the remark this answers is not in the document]
pokey: I'm guessing https://github.com/cursorless-dev/cursorless/pull/546#issuecomment-1158864553?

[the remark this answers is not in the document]
auscompgeek: `attribute` should match attributes, see https://doc.rust-lang.org/book/ch19-06-macros.html#procedural-macros-for-generating-code-from-attributes

src/languages/rust.ts
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * Returns "impl_item[type]" node higher in the chain
+ * @param node The node which we will start our search from
+ * @returns node or null
+ */
+function traitTypeMatcher(node: SyntaxNode) {
+  var parentNode = node.parent;
+  // Try find parentNode which type is "impl_item"
+  while (true) {
```
pokey, line 56: you don't need to loop here.  All of the marchers are called repeatedly as part of an algorithm that walks up the ancestors chain.  Just check that the node looks like you want it to, and return it if so

src/languages/rust.ts
```diff
@@ -0,0 +1,175 @@
...
+ * Scope types allowed to be parents of a statement
+ */
+const STATEMENT_PARENT_TYPES = ["source_file", "block", "declaration_list"];
+
+/**
+ * Returns "impl_item[type]" node higher in the chain
+ * @param node The node which we will start our search from
+ * @returns node or null
+ */
+function traitTypeMatcher(node: SyntaxNode) {
```
pokey, line 53: This name isn't really accurate, as you're matching impl item.  But you probably also want to match trait, right?  Maybe just ignore the parent node type entirely and check if this node is at the `type` field of its parent?  Then just call this function `typeFieldFinder` or something.  

Could prob just have this be a generic matcher, where it takes a field name, and returns a matcher that looks for a node that appears as that field

src/languages/rust.ts
```diff
@@ -0,0 +1,175 @@
...
+  anonymousFunction: "closure_expression",
+  argumentOrParameter: argumentMatcher("arguments", "parameters"),
+  name: [
+    "let_declaration.identifier!",
+    "parameter.identifier!",
+    "function_item[name]",
+    "struct_item[name]",
+    "enum_item[name]",
+    "trait_item[name]",
+    "const_item[name]",
```
pokey, line 164: you could maybe just replace all of these by `*[name]`.  Are you trying to exclude something in particular?

pokey (commented):
Ok almost there.  I think just attributes, and a couple comments below

src/languages/rust.ts
purpleP: Not really, but I don’t want to match something unintentionally now or in the future.

src/languages/rust.ts
purpleP: > This name isn't really accurate, as you're matching impl item

I agree, I’ll think about it some more.

> But you probably also want to match trait, right?

I could do that, but I don’t have to. A simple matcher for “impl_item” does that already.

> Maybe just ignore the parent node type entirely and check if this node is at the type field of its parent? Then just call this function typeFieldFinder or something

We can certainly make it more generic. But I would argue that there should be two arguments to this function the parent type in a chain and its child’s type. It should match if we are anywhere inside specific subtree in parse tree.

src/languages/rust.ts
pokey: We do this in lots of places; remember that the cost of matching something that you don't want is low: you just end up stopping a little bit lower in the parse tree than you hoped.  And it's quite unlikely that if it's called "name", it's not what you want.  I'd actually argue that you're future-proofing yourself by doing so

src/languages/rust.ts
pokey: Ah right I guess you don't need it for `trait_item`.  I guess fine to just rename the function to `implItemTypeFinder` and leave the more generic thing for now

src/languages/rust.ts
purpleP: I don’t think It’ll work. I want to match if I’m at the correct subtree. Please check.

src/languages/rust.ts
```diff
@@ -0,0 +1,75 @@
...
+  functionName: ["function_item[name]"],
+  anonymousFunction: "closure_expression",
+  argumentOrParameter: cascadingMatcher(
+    argumentMatcher("arguments"),
+    trailingMatcher(["parameter"], [","]),
+    trailingMatcher(["self_parameter"], [","])
+  ),
+  name: ["let_declaration.identifier!", "parameter.identifier!"],
+  class: ["struct_item", "struct_expression"],
+  value: ["let_declaration[value]", "block[-1]"],
```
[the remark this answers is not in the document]
purpleP: It doesn’t exist in rust

src/languages/rust.ts
pokey: Your function is called as part of https://github.com/cursorless-dev/cursorless/blob/main/src/processTargets/modifiers/scopeTypeStages/ContainingSyntaxScopeStage.ts#L96-L110

src/test/suite/fixtures/recorded/languages/rust/chuckValueNear.yml
purpleP: You are right, but how do I achieve that?

src/languages/rust.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Scope types allowed to be parents of a statement
+ */
+const STATEMENT_PARENT_TYPES = ["source_file", "block", "declaration_list"];
+
+const nodeMatchers: Partial<
+  Record<SimpleScopeTypeType, NodeMatcherAlternative>
+> = {
+  statement: STATEMENT_TYPES,
```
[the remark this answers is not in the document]
pokey: Btw worth adding a test for 

```rust
let x = vec![1,2,3];
```

with cursor on `vec`, saying "take state", to make sure we don't get trapped in macros

src/test/suite/fixtures/recorded/languages/rust/changeState2.yml
```diff
@@ -0,0 +1,36 @@
...
+        num
+    };
+  selections:
+    - anchor: {line: 1, character: 15}
+      active: {line: 1, character: 15}
+  marks: {}
+finalState:
+  documentContents: |
+    let expensive_closure = |num| {
+        ;
```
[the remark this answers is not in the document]
pokey: That may well be, but the parser views it in a fairly strange way, so I had assumed it was considered non-idiomatic:

```
source_file [0, 0] - [3, 0]
  if_expression [0, 0] - [2, 1]
    condition: boolean_literal [0, 3] - [0, 7]
    consequence: block [0, 8] - [2, 1]
      macro_invocation [1, 4] - [1, 21]
        macro: identifier [1, 4] - [1, 11]
        token_tree [1, 12] - [1, 21]
          string_literal [1, 13] - [1, 20]
  empty_statement [2, 1] - [2, 2]
```

src/languages/rust.ts
pokey: The code looks unchanged to me, so I'm not sure what to check?

src/test/suite/fixtures/recorded/languages/rust/chuckArgDrum.yml
```diff
@@ -0,0 +1,34 @@
...
+  selections:
+    - anchor: {line: 2, character: 0}
+      active: {line: 2, character: 0}
+  marks:
+    default.d:
+      start: {line: 0, character: 9}
+      end: {line: 0, character: 14}
+finalState:
+  documentContents: |
+    #[derive()]
```
auscompgeek, line 26: Nice. Can you also add some tests with `cfg` and `cfg_attr` attributes? For example:
```rust
#[cfg(test)]
#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
struct Test;
```
I'd expect to be able to refer to `name`/`value` in `feature = "serde"`.

src/languages/rust.ts
```diff
@@ -0,0 +1,172 @@
...
+    matcher(implItemTypeFinder),
+    patternMatcher("struct_item", "trait_item", "impl_item")
+  ),
+  functionName: ["function_item[name]"],
+  anonymousFunction: "closure_expression",
+  argumentOrParameter: argumentMatcher(
+    "arguments",
+    "parameters",
+    "meta_arguments"
+  ),
```
auscompgeek, line 158: Can we test targeting lifetime and type parameters?

src/languages/rust.ts
pokey: Discord?

src/test/suite/fixtures/recorded/languages/rust/changeState2.yml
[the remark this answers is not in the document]
purpleP: I’m not sure I get what you want here. If you want semicolon to be removed when we change statement then I’m not sure how to achieve that. Do I need to wrap ancestorChain matcher into trailing matcher with semicolon or something?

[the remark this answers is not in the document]
pokey: ah wait a minute; that semicolon is actually extra isn't it?  because `println!` is a macro, and macros don't actually require a semicolon.  This is fine

pokey:
@auscompgeek I think this is pretty much good to go but wanted to give you a last chance to look before I merge it

src/test/suite/fixtures/recorded/languages/rust/chuckArgTrap.yml
```diff
@@ -0,0 +1,32 @@
...
+  selections:
+    - anchor: {line: 1, character: 0}
+      active: {line: 1, character: 0}
+  marks:
+    default.t:
+      start: {line: 0, character: 21}
+      end: {line: 0, character: 22}
+finalState:
+  documentContents: |
+    struct S<'a, 'b: 'a, >;
```
pokey, line 25: This one doesn't look right to me.  Shouldn't this one have removed the trailing comma?

pokey:
Made a couple quick tweaks so make sure to pull

auscompgeek:
I'll test this out this evening my time.

src/test/suite/fixtures/recorded/languages/rust/chuckArgTrap.yml
purpleP: Yes, that’s strange.

pokey: That's the `: 'a` thing? Agreed

purpleP: Yes. It’s rarely used so I forgot the official name.

purpleP: Okay, It’s actually just a trait bound for the tree sitter.

pokey: Cool so "type" / "name" just works out of the box then?

Any idea why comma is sticking around?

src/test/suite/fixtures/recorded/languages/rust/changeTypeUrge.yml
```diff
@@ -0,0 +1,32 @@
...
+  selections:
+    - anchor: {line: 1, character: 0}
+      active: {line: 1, character: 0}
+  marks:
+    default.u:
+      start: {line: 0, character: 10}
+      end: {line: 0, character: 12}
+finalState:
+  documentContents: |
+    const X: [; 2] = [0, 0];
```
pokey, line 25: Huh I think I'd expect the whole `[u8; 2]` to be `"type"` here? I don't feel super strongly but that would be my leaning. Cc/ @auscompgeek @AndreasArvidsson

AndreasArvidsson: Agreed. I would expect the entire `[u8, 2]` to be the type

auscompgeek: Hmm, I think this behaviour is reasonable, not sure how I'd select the type the array contains otherwise if it's more complex, e.g. `[HashMap<u32, u64>; 4]`. The array type can be reasonably selected in other ways.

purpleP: > Hmm, I think this behaviour is reasonable, not sure how I'd select the type the array contains otherwise if it's more complex, e.g. `[HashMap<u32, u64>; 4]`. The array type can be reasonably selected in other ways.

My reasoning exactly. The whole type can be selected with `take type plex` etc.

src/util/nodeSelectors.ts
```diff
@@ -124,6 +133,8 @@ export function argumentSelectionExtractor(): SelectionExtractor {
       node.type === ")" ||
       node.type === "[" ||
       node.type === "]" ||
+      node.type === ">" ||
+      node.type === "<" ||
```
pokey, line 137: As above, I wonder if less-than or greater-than will be an issue here

pokey (commented):
Couple more thoughts but otherwise good to go from my perspective

pokey:
> I'll test this out this evening my time.

@auscompgeek just ping here when you've had a chance to take it for a spin and then I think we're in good shape

src/test/suite/fixtures/recorded/languages/rust/changeTypeUrge.yml
AndreasArvidsson: Not at all

src/util/nodeFinders.ts
```diff
@@ -108,8 +108,8 @@ export const typedNodeFinder = (...typeNames: string[]): NodeFinder => {
 const toPosition = (point: Point) => new Position(point.row, point.column);
 
 export const argumentNodeFinder = (...parentTypes: string[]): NodeFinder => {
-  const left = ["(", "{", "["];
-  const right = [")", "}", "]"];
+  const left = ["(", "{", "[", "<"];
+  const right = [")", "}", "]", ">"];
```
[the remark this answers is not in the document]
AndreasArvidsson: They are looking for these delimiters to the left and to the right of an already found argument. As long as the language don't support arithmetic operators in the argument list we should be fine.

src/test/suite/fixtures/recorded/languages/rust/changeTypeUrge.yml
pokey: Cool so let's just leave this one as is for now.  We can add a special scope type for `length` later if we feel we need it 🤷

pokey (approved):

pokey:
@auscompgeek I just shipped it, but when you get a minute just go ahead and update your extension to test it on mainline, and we can open a follow-up for anything that doesn't feel right.  I'll make an announcement once you're happy it works as expected

src/util/nodeSelectors.ts
purpleP: Good question…

Issue #714, opened 2022-06-01
Add `no-unnecessary-type-assertion` to typescript lint rules


pokey:
Looks like [this](https://typescript-eslint.io/rules/no-unnecessary-type-assertion) is the rule? Odd thing is it appears that one should be on by default 🤔

__Originally posted by @pokey in https://github.com/cursorless-dev/cursorless/pull/705#discussion_r886347414__

auscompgeek:
> [`explicit-function-return-type`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/explicit-function-return-type.md)

Should we use [`explicit-module-boundary-types`](https://typescript-eslint.io/rules/explicit-module-boundary-types/) rather than `explicit-function-return-type`?

auscompgeek:
Unassigning myself because I realistically won't have time to enable more lint rules.

It'll be good to consider the lint rules that use type information, but we should consider how enabling type checking in linting affects the linter performance.

PR #776, opened 2022-06-19
Add ESLint recommended rules

auscompgeek:
Part of #714

Adds the following ESLint rules:
- `eslint:recommended`
- `typescript-eslint/recommended`
- [`@typescript-eslint/consistent-type-assertions`](https://typescript-eslint.io/rules/consistent-type-assertions/)

Ran `yarn lint --fix` then fixed the unautofixable lints.

Todo:
- [ ] Look into [explicit-function-return-type](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/explicit-function-return-type.md) (maybe for later)


src/test/util/asyncSafety.ts
```diff
@@ -12,20 +12,19 @@ import { AsyncFunc, Context, Done } from "mocha";
  * @param fn The test function to run
  * @returns A safely wrapped test function
  */
-export default function asyncSafety(fn: AsyncFunc) {
+export default function asyncSafety(fn: () => Promise<void>) {
```
auscompgeek, line 15: Changed this to remove the @ts-ignore

src/core/commandVersionUpgrades/upgradeV1ToV2/upgradeV1ToV2.ts
```diff
@@ -131,10 +131,12 @@ function upgradePrimitiveTarget(
         if (modifier?.type === "subpiece") {
           break;
         }
+      // fallthrough
       case "line":
         if (mark?.type === "lineNumber") {
           break;
         }
+      // fallthrough
```
auscompgeek, line 139: I presume these fallthroughs are deliberate?

src/extension.ts
```diff
@@ -25,7 +25,6 @@ export async function activate(context: vscode.ExtensionContext) {
     commandServerApi: () => commandServerApi,
     getNodeAtLocation: () => getNodeAtLocation,
   } as FactoryMap<Graph>);
-  graph.debug.init();
```
auscompgeek, line 28: `Debug#init` was empty, so removed it altogether, since it's not implementing anything.

src/extension.ts
```diff
@@ -94,4 +93,6 @@ export async function activate(context: vscode.ExtensionContext) {
 }
 
 // this method is called when your extension is deactivated
-export function deactivate() {}
+export function deactivate() {
+  // do nothing
```
auscompgeek, line 97: Squashes `no-empty-function`.

.eslintrc.json
```diff
@@ -25,6 +32,7 @@
 				"null": "never"
 			}
 		],
+		"no-constant-condition": ["error", { "checkLoops": false }],
```
auscompgeek, line 41: There's one `while (true)` loop left in `findDelimiterPairContainingSelection`. Figured those are probably more useful than the lint rule, and IMO `for (;;)` is ugly.

src/processTargets/marks/CursorStage.ts
```diff
@@ -1,13 +1,10 @@
 import { Target } from "../../typings/target.types";
-import { CursorMark } from "../../typings/targetDescriptor.types";
 import { ProcessedTargetsContext } from "../../typings/Types";
 import { isReversed } from "../../util/selectionUtils";
 import { MarkStage } from "../PipelineStages.types";
 import WeakTarget from "../targets/WeakTarget";
 
 export default class CursorStage implements MarkStage {
-  constructor(_modifier: CursorMark) {}
```
auscompgeek, line 9: `MarkStage` doesn't require any particular constructor signature, so removing this to squash `no-empty-function`.

auscompgeek:
Might leave explicit-function-return-type for later, that one might get a bit gnarly.

src/core/commandVersionUpgrades/upgradeV1ToV2/upgradeV1ToV2.ts
pokey: I believe so. Slightly confusing way of doing this logic imo but prob ok as it's just in migration

Cc/ @AndreasArvidsson

src/extension.ts
pokey: The way the graph works, accessing an attribute will cause the component to be instantiated, so it's possible this line was actually doing something

src/processTargets/getMarkStage.ts
```diff
@@ -10,7 +10,7 @@ import { MarkStage } from "./PipelineStages.types";
 export default (mark: Mark): MarkStage => {
   switch (mark.type) {
     case "cursor":
-      return new CursorStage(mark);
+      return new CursorStage();
```
pokey, line 13: Hmm I'm hesitant to move away from the uniformity here, as we may turn this case statement into a map

Slightly mixed feelings about the empty function thing, as it can be useful. I guess forcing the comment saying it's empty on purpose is maybe not a bad thing either way?

Cc/ @AndreasArvidsson

pokey (commented):
Left some initial comments

src/processTargets/getMarkStage.ts
AndreasArvidsson: I agree and would prefer to keep the uniformity.

src/testUtil/serialize.ts
```diff
@@ -1,41 +1,39 @@
-// @ts-nocheck
 // From https://github.com/nodeca/js-yaml/issues/586#issuecomment-814310104
```
pokey, line 1: Epic!  Presumably you tried running this?  I don't believe it is actually tested.  Best test is prob to run the "Update snapshots" debug launch config in VSCode

src/util/typeUtils.ts
```diff
@@ -5,7 +5,7 @@
  * @param b The second object
  * @returns True if `a` and `b` have the same constructor
  */
-export function isSameType(a: unknown, b: unknown): boolean {
+export function isSameType<T>(a: T, b: unknown): b is T {
```
pokey, line 8: oooh fancy 🎩😊

src/test/util/asyncSafety.ts
pokey: Is there any consequence to only supporting functions that don't return anything?

src/processTargets/marks/CursorStage.ts
pokey: See above re uniformity

src/processTargets/processTargets.ts
```diff
@@ -138,7 +138,7 @@ function processVerticalRangeTarget(
   const activeLine = activePosition.line - (excludeActive ? delta : 0);
 
   const results: Target[] = [];
-  for (let i = anchorLine; true; i += delta) {
+  for (let i = anchorLine; ; i += delta) {
```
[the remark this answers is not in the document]
auscompgeek: Yeah. I "fixed" this one before I found the `while (true)` I couldn't refactor though.

src/test/util/asyncSafety.ts
auscompgeek: We're only using this in one place AFAICT, namely the recorded test runner. If we start using it elsewhere where the promise does resolve to something, that's probably a bug since the value would be unused.

src/testUtil/serialize.ts
auscompgeek: Huh, it seems like I may have accidentally made `documentContents` serialise as an inline string. I'll look into that.

auscompgeek: Oh wait, that might just be because I have hard tabs by default, so the YAML serialiser wants to serialise a tab character as `"\t"`:
```yaml
documentContents: "try {\n\tif (true) {\n\t    const foo = \"hello\";\n\t}\n} catch (err) {\n\t\n}\n\ntry {\n\tconst bar = \"hello\";\n} catch (err) {\n\t\n}"
```

pokey: Yeah probably.  Might be worth just changing your local setting temporarily so we don't have to wait on #731?

pokey: Ok well it has nothing to do with what you've done in this PR, so I'd consider this a passing test from that perspective.

As for why it's doing that, idk that's weird.  Are tests passing for you locally?

auscompgeek: Oddly enough they are passing locally after blowing away those changes. 🤔

pokey: Strange.  Possible that test fails nondeterministically.  We retry all tests 5 times, but when you do a snapshot update it will just use the first result so any failure will just become a snapshot update and not re-run

src/extension.ts
pokey: I believe this is the only remaining issue on this PR

auscompgeek: Reverted this change. I'm not convinced this line is doing much in practice, but we can look into it later.

pokey (approved):
Looks good thanks!

PR #777, opened 2022-06-19
Fix positional inference

pokey:
Separates inference so that we handle positional modifiers and non-positional modifiers separately, the way we did before #672.  This way with the command `"bring air after line bat and block cap"`, we infer `"bring air after line bat and *after* block cap"`, as desired, just as we used to do before #672

## Checklist

### DIFF
--- a/src/test/suite/fixtures/recorded/hatTokenMap/bringPointAndHarpToEndOfSecondCarWhaleAndEndOfJustWhaleTakeWhale.yml
+++ b/src/test/suite/fixtures/recorded/hatTokenMap/bringPointAndHarpToEndOfSecondCarWhaleAndEndOfJustWhaleTakeWhale.yml
@@ -1,6 +1,6 @@
 languageId: plaintext
 command:
-  spokenForm: bring point and point to end of second car whale and end of whale take whale
+  spokenForm: bring point and harp to end of second car whale and end of just whale take whale
   version: 2
   targets:
     - type: list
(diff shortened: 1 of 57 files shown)


src/test/suite/fixtures/recorded/hatTokenMap/bringPointAndHarpToEndOfSecondCarWhaleAndEndOfJustWhaleTakeWhale.yml
```diff
@@ -13,15 +12,22 @@ command:
...
+              active: 1
+              excludeAnchor: false
+              excludeActive: false
         - type: primitive
-          position: after
-          insideOutsideType: inside
           mark: {type: decoratedSymbol, symbolColor: default, character: w}
+          modifiers:
+            - {type: position, position: end}
+            - {type: toRawSelection}
```
pokey, line 28: Note that we add "just" here to break inference chain, because we actually had a bug where we weren't doing inference for subtoken modifiers that had a position, even though we should have been

src/test/suite/fixtures/recorded/inference/bringHarpAfterLineTrapAndBlockSpun.yml
```diff
@@ -0,0 +1,62 @@
+languageId: plaintext
```
pokey, line 1: This test is the punchline: having "block" as a modifier on the second destination target still allows it to infer "after" from the first destination target

AndreasArvidsson (approved):

PR #782, opened 2022-06-23
Fix `funk` and `funk name` scope for Ruby Class methods

### LINKED ISSUE #760: Scope `funk` doesn't match on Class methods - ruby
I was hoping to try and fix this myself as a way to get familiar with how cursorless + tree sitter works but it might take me a bit to get up to speed. Here is the issue:

Given this code:

```ruby
class Test
  class << self
    # `take funk` doesn't work
    def baz
    end
  end

  # `take funk` doesn't work
  def self.foo
    new.process
  end

  # `take funk` doesn't work
  def Test.bar
    new.process
  end

  # `take funk` works :D
  def process
    puts "process"
  end
end
```

Current use cases/scopes not covered:
- [x] `"change funk"` on a function defined `self.`
- [x] `"change funk"` on a function defined `ClassName.`
- [x] `"change funk"` on a function defined within `class << self .... end` block

richardmcmillen:
Tree sitter sees these as `singleton_methods` and this was missing from
the `namedFunction` and `functionName` definitions.

## Checklist

---

Fixes #760


src/languages/ruby.ts
```diff
@@ -167,8 +167,8 @@ const nodeMatchers: Partial<
   ifStatement: "if",
   functionCall: "call",
   comment: "comment",
-  namedFunction: "method",
-  functionName: "method[name]",
+  namedFunction: ["method", "singleton_method"],
+  functionName: ["method[name]", "singleton_method[name]"],
```
pokey, line 171: Please also add to `name` (line 187 below), and add a test for that. Just one test for that one is fine

pokey (commented):
Looks good! See inline comment, and also

- [ ] Worth adding a test for "arg" within an arg list for one of these functions

I'll also give @Will-Sommers a chance to have a look

richardmcmillen:
>Worth adding a test for "arg" within an arg list for one of these functions

I added a new test case which I think covers this if I understand correctly, and make the other change as well.

src/languages/ruby.ts
Will-Sommers: Sorry I missed this suggestion in the issue @richardmcmillen!

Will-Sommers (approved):
lgtm, thank you!

pokey (approved):
Yep looks good to me too thanks!

PR #786, opened 2022-06-26
Remove unsupported command from documentation.

wolfmanstout:
This tripped me up when I was first getting started, and is apparently still in the documentation and unsupported.

## Checklist

### DIFF
--- a/docs/user/README.md
+++ b/docs/user/README.md
@@ -104,10 +104,9 @@ minimize syllables.
 
 ##### `"this"`
 
-The word `"this"` can be used as a mark to refer to the current cursor(s) or selection(s) as a target. Note that when combined with a modifier, the `"this"` mark can be omitted, and it will be implied.
+The word `"this"` can be used as a mark to refer to the current cursor(s) or selection(s) as a target. Note that when combined with a modifier, the `"this"` mark must be omitted, and it will be implied.
 
 - `chuck this`
-- `take this funk`
 - `pre funk`
 - `chuck line`
 


docs/user/README.md
```diff
@@ -104,10 +104,9 @@ minimize syllables.
 
 ##### `"this"`
 
-The word `"this"` can be used as a mark to refer to the current cursor(s) or selection(s) as a target. Note that when combined with a modifier, the `"this"` mark can be omitted, and it will be implied.
+The word `"this"` can be used as a mark to refer to the current cursor(s) or selection(s) as a target. Note that when combined with a modifier, the `"this"` mark must be omitted, and it will be implied.
```
pokey, line 107:
```suggestion
The word `"this"` can be used as a mark to refer to the current cursor(s) or selection(s) as a target. Note that when combined with a modifier, the `"this"` mark can be omitted, and it will be implied.
```

docs/user/README.md
```diff
@@ -104,10 +104,9 @@ minimize syllables.
...
 
 - `chuck this`
-- `take this funk`
```
pokey, line 110: This should actually be:

- `take funk this`

pokey (changes requested):
ooh ha nice catch.  Not quite right, though.  You can in fact omit the mark, and it will be `this`, it's just that the mark appears after all modifiers.  The problem was the word order in that command

docs/user/README.md
wolfmanstout: Ahhhh!! Cool.

wolfmanstout:
Okay fixed up. Mental model also fixed :-)

pokey (approved):
Looks good thanks!

PR #791, opened 2022-06-27
Use removal highlight range instead of removal range

### LINKED ISSUE #787: Incorrect highlighting for move line
The source line also highlights the line afterwards when it should only highlight the line itself

AndreasArvidsson:
fixes #787

## Checklist

### DIFF
--- a/src/test/suite/fixtures/recorded/decorations/moveLineAir.yml
+++ b/src/test/suite/fixtures/recorded/decorations/moveLineAir.yml
@@ -1,0 +1,53 @@
+languageId: plaintext
+command:
+  spokenForm: move line air
+  version: 2
+  targets:
+    - type: primitive
+      mark: {type: decoratedSymbol, symbolColor: default, character: a}
+      modifiers:
+        - type: containingScope
+          scopeType: {type: line}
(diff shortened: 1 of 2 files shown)


src/test/suite/fixtures/recorded/decorations/moveLineAir.yml
```diff
@@ -0,0 +1,53 @@
...
+      active: {line: 0, character: 0}
+decorations:
+  - name: pendingDeleteBackground
+    type: line
+    start: {line: 0, character: 0}
+    end: {line: 0, character: 5}
+  - name: pendingModification0Background
+    type: token
+    start: {line: 3, character: 0}
+    end: {line: 3, character: 0}
```
pokey, line 48: Any idea where this is coming from?

AndreasArvidsson: Yes that is my current selection. ie the destination we are replacing.

pokey: ohh interesting got it

pokey (approved):

PR #792, opened 2022-06-27
Fix new line bug with pour arg

### LINKED ISSUE #788: Fix `"pour arg"` with multiline argument lists
```typescript
async run(
    [targets]: [Target[]],
    snippetName: string,
    substitutions: Record<string, string>
  ): Promise<ActionReturnValue> {

  }
```

If you say `"pour argue made"` with `"made"` referring to `snippetName`, you end up with two commas at the end of that line and the cursor on the line below

- [ ] "bring arg air after bat" is broken similarly with multiline arg lists

AndreasArvidsson:
Fixes #788 

## Checklist

- [x} I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)

### DIFF
--- a/src/processTargets/targets/PositionTarget.ts
+++ b/src/processTargets/targets/PositionTarget.ts
@@ -49,7 +49,8 @@ export default class PositionTarget extends BaseTarget {
     useLinePadding: boolean
   ): EditWithRangeUpdater {
     const delimiter = this.insertionDelimiter;
-    const isLine = delimiter.includes("\n");
+    // It's only considered a line if the delimiter is only new line symbols
+    const isLine = /^(\n)+$/.test(delimiter);
     const isBefore = this.position === "before";
 
     const range = getEditRange(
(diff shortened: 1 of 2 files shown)


pokey (commented):
Nice!  Would also be good to add a few more tests:

- [x] `"bring arg air after bat"` (see checkbox in #788)
- [x] `"pour arg"` on final arg in multiline arg list without trailing comma
- [ ] `"pour arg"` on final arg in multiline arg list with trailing comma
- [ ] `"drink arg"` on final arg in multiline arg list without trailing comma
- [ ] `"drink arg"` on final arg in multiline arg list with trailing comma
- [ ] `"drink arg"` on first arg in multiline arg list
- [ ] `"drink arg"` on middle arg in multiline arg list
- [ ] A couple multi-line tests for "item" (esp because #709 on the horizon)

pokey (commented):
Great!  Sorry but I thought of a couple more quick tests 😅:

- [ ] `"bring arg air after bat"` when bat already has a trailing comma (eg in middle of list)
- [ ] `"bring arg air before bat"`

AndreasArvidsson:
No problem. Tests added.

pokey (approved):
Looks good!

PR #794, opened 2022-06-27
Enable java return value as scope type value

AndreasArvidsson:
## Checklist


pokey (commented):
Looks good, but please also add test where mark is in `return` keyword

AndreasArvidsson:
@pokey Done!

pokey (approved):
Looks good!

PR #810, opened 2022-06-30
Reuse editor for recorded tests

AndreasArvidsson:
I took a few performance measurements. All data is the delta time for a single pass run on my seven year old dual core elite book using ubuntu.

Test subset with all recorded tests (1705 tests)

original: 6m
reuse editor: 124s
removed everything after cursorless command: 126s
removed cursorless command: 76s
removed everything after reuseEditor: 47s
removed reuseEditor: 7s
removed rest of test case: 4s

I also did a few measurements per function call:

openNewEditor: 188s
reuseEditor: 44.083s
executeCommand: 59.752s
addDecorations: 6549ms
getReadableMap: 2411ms
getCursorlessApi: 771ms

## Checklist


pokey (approved):
🙌

PR #813, opened 2022-07-02
Sort VSCode settings to put common settings first

### LINKED ISSUE #389: Order settings
Would be nice if all the simple settings like decoration time and show on start were above all the different settings for the hats

https://code.visualstudio.com/updates/v1_63#_settings-editor-improvements

paul-schaaf:
closes #389 

I've taken the liberty to also group the hat settings. let me know if I should revert it!

For some reason one of the hat settings still shows up first. I've looked into the vscode docs but wasn't able to find anything on it.I am guessing it is a bug in vscode. But maybe I also made a mistake somewhere.


README.md
```diff
@@ -42,13 +42,13 @@ See [installation](https://www.cursorless.org/docs/user/installation/) for insta
...
+- `cursorless.hatSettings.hatVerticalOffset`: How much to vertically shift the hats as a percentage of font size; positive is up
+- `cursorless.hatSettings.hatEnablement.colors`: Whether to enable particular hat colors.
+- `cursorless.hatSettings.hatEnablement.shapes`: Whether to enable particular hat shapes.
+- `cursorless.hatSettings.hatPenalties.colors`: How much to penalize each hat color. You will probably want to set this one to the number of syllables in the given style. Cursorless will then sort every style combination by number of syllables to refer to it.
+- `cursorless.hatSettings.hatPenalties.shapes`: How much to penalize each hat shape. You will probably want to set this one to the number of syllables in the given style. Cursorless will then sort every style combination by number of syllables to refer to it.
+- `cursorless.hatSettings.maximumHatStylePenalty`: The maximum allowed penalty for a hat style. Any hat style whose penalty is greater than this amount will not be used. A hat style penalty is defined to be the shape penalty plus the colour penalty. Setting this value less than or equal to zero is treated as no maximum.
```
auscompgeek, line 51: If these names change then we need some migration code for those that have modified these settings. I don't see any.

auscompgeek (changes requested):

package.json
```diff
@@ -335,14 +345,16 @@
 					},
 					"additionalProperties": false
 				},
-				"cursorless.maximumHatStylePenalty": {
+				"cursorless.hatSettings.hatPenalties.maximumHatStylePenalty": {
```
pokey, line 348: This doesn't really have any impact on hat penalties, so I don't think I'd put it in that group
```suggestion
				"cursorless.hatSettings.maximumHatStylePenalty": {
```

pokey (changes requested):
Thanks for taking a crack at this one!  Good direction, but as @auscompgeek mentions, this PR will break everyone's Cursorless who has used any hat settings 😅.  You'll need to modify the code that reads the settings to be able to handle the old as well as the new paths

src/core/Decorations.ts
```diff
@@ -94,15 +94,15 @@ export default class Decorations {
...
 
     const userVerticalOffset = vscode.workspace
-      .getConfiguration("cursorless")
+      .getConfiguration("cursorless.hatSettings")
       .get<number>(`hatVerticalOffset`)!;
 
     const userIndividualAdjustments = vscode.workspace
-      .getConfiguration("cursorless")
+      .getConfiguration("cursorless.hatSettings")
       .get<IndividualHatAdjustmentMap>("individualHatAdjustments")!;
```
pokey, line 106: You can prob add a new function like the following:

```typescript
function getHatSetting<T>(name: string): T {
return vscode.workspace
      .getConfiguration("cursorless.hatSettings")
      .get<T>(name) ??
vscode.workspace
      .getConfiguration("cursorless")
      .get<T>(name);
}
```

pokey (changes requested):
On second, thought, though, I'm not sure we want to group the hat settings.  The hat enablement settings are far more common than the other ones.  I think probably to arrange settings by how likely they are to be used, rather than category.  Here's the order I'd suggest:


1. Cursorless: Hat Size Adjustment
2. Cursorless: Hat Vertical Offset
3. Cursorless › Hat Enablement: Colors
4. Cursorless › Hat Enablement: Shapes
5. Cursorless › Colors: Light
6. Cursorless › Colors: Dark

Then after that I think they're all less common so doesn't matter as much

paul-schaaf:
@pokey done!

pokey (approved):
Great thanks!

Issue #819, opened 2022-07-04
Can't set selection in the source/changes working tree


AndreasArvidsson:
The function [focusEditor](https://github.com/cursorless-dev/cursorless/blob/bc24adcb984fd0da6845985a8ecff4982884fcf7/src/util/setSelectionsAndFocusEditor.ts#L53-L66) dost not work when `editor.viewColumn` is null

Cursorless throws the following error message `Couldn't find notebook editor for given document`

PR #820, opened 2022-07-04
Only focus editor if needed

AndreasArvidsson:
* Saves performance/faster testing
* Mitigates #819 if the correct editor is already focused

## Checklist


pokey (changes requested):
This breaks the ability to select something in an editor (and focus it) when you have the terminal focused.  `activeEditor === editor` doesn't mean that the correct editor is _focused_, just that it is the active editor.  In VSCode there is no way to determine whether it is actually focused

AndreasArvidsson:
Ah that is really unfortunate. #819 is really problematic for me at the moment. I'm going to have the rethink this.

pokey:
I believe this solution also won't work if you are in an opposite split.  

Could maybe iterate the [tab groups](https://code.visualstudio.com/api/references/vscode-api#TabGroup:~:text=tabGroups%3A%20TabGroups) and then look at uri's inside any [diff tab](https://code.visualstudio.com/api/references/vscode-api#Tab:~:text=%3A%20TabInputText%20%7C-,TabInputTextDiff,-%7C%20TabInputCustom%20%7C)?  Then you ask the tab group for its view column and focus it.  Idk maybe there's a cleaner solution but that might work

pokey:
Keeping in mind the tab api's are relatively new so you'll need to some type gymnastics if we want to keep maintaining backwards compatibility for VSCode

PR #821, opened 2022-07-04
init-launch-sandbox: Stop capturing stdout/stderr

auscompgeek:
This stops capturing and buffering stdout and stderr when installing extensions in the sandbox, so live progress is visible.

## Checklist

### DIFF
--- a/src/test/initLaunchSandbox.ts
+++ b/src/test/initLaunchSandbox.ts
@@ -4,7 +4,6 @@
  */
 import * as path from "path";
 import * as cp from "child_process";
-import { promisify } from "util";
 import { extensionDependencies } from "./extensionDependencies";
 import { mkdir } from "fs/promises";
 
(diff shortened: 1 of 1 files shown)


pokey (commented):
What platforms have you tried this one on?  The reason we originally switched to exec and capturing stdout was because it didn't work on Windows.  cc/ @AndreasArvidsson

AndreasArvidsson:
Yes there was problem with the output on windows. I can try this on when I get back home

pokey:
@AndreasArvidsson when you get a minute it would be awesome if you could try this out on Windows; I believe that's what's blocking this PR

AndreasArvidsson:
stdout appears to be working on windows

pokey:
@AndreasArvidsson do you remember what the problem was originally?  Unfortunately we didn't leave any comment.  [Here's](https://github.com/cursorless-dev/cursorless/pull/691/commits/a044fd0f3009c00e0c24ffa48c7f43ddf3587c32) the commit where you fixed Windows originally fwiw

AndreasArvidsson:
I think it was that stdout wasn't showing. I can't remember exactly. Now it appears to be working at least.

pokey (approved):

Issue #725, opened 2022-06-02
Use data classes on talon side in order to get better type hints


pokey:

auscompgeek:
I'll probably pick this up this weekend, assuming the massive refactor PR gets merged by then.

pokey:
cool!  we'll ping when it's merged

PR #822, opened 2022-07-04
Fix type hints

pokey:
I fixed all the easy ones.  There were a few that I couldn't fix due to the way Talon works, eg not having `self` arg in action classes

## Checklist

### DIFF
--- a/cursorless-talon/src/actions/actions_makeshift.py
+++ b/cursorless-talon/src/actions/actions_makeshift.py
@@ -1,4 +1,5 @@
 from dataclasses import dataclass
+from typing import Any, Optional
 
 from talon import Module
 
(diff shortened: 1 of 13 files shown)


cursorless-talon/src/actions/actions_makeshift.py
```diff
@@ -56,13 +57,13 @@ class MakeshiftAction:
...
 class TalonOptions:
-    post_command_sleep_ms: int = None
+    post_command_sleep_ms: Optional[int] = None
     await_command: bool = True
 
 
 def get_parameters(action: MakeshiftAction):
     command = action.vscode_command_id
-    command_options = {
+    command_options: dict[str, Any] = {
```
auscompgeek, line 66: Can the values be `Callable`?

auscompgeek: Ah. I'll see if I can't turn this into a dataclass or something at some point then.

pokey: Ah right yeah that sounds good.  You mean as part of #725 I assume

PR #828, opened 2022-07-04
Fix/rust followup

purpleP:
Adding missing pieces that have been found after testing.


src/test/suite/fixtures/recorded/languages/rust/changeNameDrum2.yml
```diff
@@ -0,0 +1,30 @@
...
+  targets:
+    - type: primitive
+      mark: {type: decoratedSymbol, symbolColor: default, character: d}
+      modifiers:
+        - type: containingScope
+          scopeType: {type: name}
+  usePrePhraseSnapshot: true
+  action: {name: clearAndSetSelection}
+initialState:
+  documentContents: "let user = User {\n\tfirst_name: \"John\",\n    last_name: \"Doe\",\n}\n"
```
auscompgeek, line 14: These YAML files aren't particularly readable with hard tabs, mind re-recording with spaces?

auscompgeek (commented):
heh, I noticed this the other day too, thanks for the fixes

purpleP:
> heh, I noticed this the other day too, thanks for the fixes

Lest me know if something else is missing!

src/test/suite/fixtures/recorded/languages/rust/changeNameDrum2.yml
```diff
@@ -0,0 +1,38 @@
+languageId: rust
```
pokey, line 1: this should also work for "key", as it looks a lot like a key-value pair

Fwiw we're planning to support "name" everywhere we support "key" in the future

src/test/suite/fixtures/recorded/languages/rust/changeNameDrum3.yml
```diff
@@ -0,0 +1,36 @@
+languageId: rust
```
pokey, line 1: I _think_ I'd also support "key" on this one, just because of the way it looks, though tbh I'm not totally sure what this code actually does 😅

src/test/suite/fixtures/recorded/languages/rust/changeValueCap.yml
```diff
@@ -0,0 +1,32 @@
+languageId: rust
```
pokey, line 1: Oh ha man how did we miss this one 😂.  Good catch

Worth also adding a "chuck" test for this one to make sure it removes the `=`, as that will just leave you with a variable declaration, right?

src/test/suite/fixtures/recorded/languages/rust/chuckNameDrum.yml
```diff
@@ -0,0 +1,38 @@
+languageId: rust
```
pokey, line 1: Not sure we need a "chuck" test here if you're not removing the `:` or doing anything interesting

src/test/suite/fixtures/recorded/languages/rust/chuckNameJury.yml
```diff
@@ -0,0 +1,36 @@
+languageId: rust
```
pokey, line 1: Likewise

src/test/suite/fixtures/recorded/languages/rust/chuckValueFine.yml
```diff
@@ -0,0 +1,38 @@
...
+    - anchor: {line: 3, character: 2}
+      active: {line: 3, character: 2}
+  marks:
+    default.f:
+      start: {line: 1, character: 4}
+      end: {line: 1, character: 14}
+finalState:
+  documentContents: |
+    let user = User {
+        first_name,
```
pokey, line 29: Just to check: is this valid code?

pokey (commented):
Good stuff!  Left a few minor comments

src/test/suite/fixtures/recorded/languages/rust/chuckValueFine.yml
auscompgeek: Perfectly valid assuming there's a `first_name` variable in scope.

src/test/suite/fixtures/recorded/languages/rust/changeNameDrum3.yml
auscompgeek: Given key-value literals aren't really a thing in Rust I'd be okay with letting `key` match struct field names.

auscompgeek:
Hmm, on the note of one of the above threads, this'd be an interesting test:

```rust
if let User { first_name: name, surname: "Doe" } = user {
}
```

I'd expect the scopes when targeting within `first_name: name` to behave the same way as destructuring in JavaScript.

src/test/suite/fixtures/recorded/languages/rust/chuckValueFine.yml
pokey: Got it.  So in that case, I'd remove the `:` for chuck key / name and chuck value.  We do that kinda thing all the time in typescript; quite convenient

src/test/suite/fixtures/recorded/languages/rust/changeNameDrum3.yml
purpleP: > Hmm, on the note of one of the above threads, this'd be an interesting test:
> 
> ```rust
> if let User { first_name: name, surname: "Doe" } = user {
> }
> ```
> 
> I'd expect the scopes when targeting within `first_name: name` to behave the same way as destructuring in JavaScript.

Which is how exactly?

src/test/suite/fixtures/recorded/languages/rust/changeValueCap.yml
purpleP: Not really. You can’t declare a `const ` without providing some value for it.

pokey: Ah right.  So we just support "chuck value" removing `=` for non-const then?  makes sense to me

pokey:
@purpleP I just updated this branch to get the latest on `main` which fixes CI

src/test/suite/fixtures/recorded/languages/rust/chuckNameDrum.yml
```diff
@@ -0,0 +1,38 @@
...
+      active: {line: 3, character: 2}
+  marks:
+    default.d:
+      start: {line: 2, character: 16}
+      end: {line: 2, character: 19}
+finalState:
+  documentContents: |
+    let user = User {
+        first_name: "John",
+        : "Doe",
```
pokey, line 30: As mentioned in https://github.com/cursorless-dev/cursorless/pull/828#discussion_r913665423, I'd remove the leading `:` here.  I'd also remove trailing `:` when they say "chuck name" or "chuck key".  I know this looks awkward with this code, but in the following code it's perfectly reasonable:

```rust
let user = User {
        first_name: "John",
        last_name: last_name,
}
```

In this case you could say either "chuck key" or "chuck value" to switch to using [field initialiser shorthand](https://rust-lang.github.io/rfcs/1682-field-init-shorthand.html)

src/test/suite/fixtures/recorded/languages/rust/chuckNameJury.yml
```diff
@@ -0,0 +1,36 @@
...
+    - anchor: {line: 3, character: 0}
+      active: {line: 3, character: 0}
+  marks:
+    default.j:
+      start: {line: 1, character: 24}
+      end: {line: 1, character: 28}
+finalState:
+  documentContents: |
+    match user {
+        User { : "John", last_name: "Doe" } => {},
```
pokey, line 28: Is there a similar shorthand supported here?  Eg

```rust
match user {
        User { first_name, last_name: "Doe" } => {},
}
```

If so I'd remove leading / trailing `:` on "chuck"

auscompgeek: That would be a name binding, but I think that'd make sense.

purpleP:
I think I have addressed all the comments, haven’t I?

src/test/suite/fixtures/recorded/languages/rust/chuckNameDrum.yml
pokey: How come this one still has the leading `:`?

src/test/suite/fixtures/recorded/languages/rust/chuckNameJury.yml
pokey: @purpleP how come this one also has leading `:` still?

pokey (changes requested):
Almost there; left a couple more comments.  Also:

Would be good to add a test for "chuck value" with one of those pattern matchers:

```rust
match user {
        User { first_name: first_name, last_name: "Doe" } => {},
}
```

src/test/suite/fixtures/recorded/languages/rust/chuckNameDrum.yml
pokey: Ah looks like you just forgot to delete this test; I see it's failing as expected

purpleP:
> > Lest me know if something else is missing!
> 
> I think the only thing I found was not being able to grab the nearest `attribute` from the field/Rust item it's on, but `name` being weird is annoying enough that we can follow up later.

Can you show an example? Maybe we can fix that right away.

purpleP:
> > How about treating list of structure fields as arguments? I would like to `chuck` and `carve` them with their type to use e.g. in constructor function etc. Is that possible without such feature?
> 
> You should already be able to do that today using "item". I am neutral about whether these should also be considered "arg"; I'll defer to you and @auscompgeek on that

Yes, makes sense! Let’s forget about what I said than 😀

purpleP:
@pokey take a look at the errors in the tests on windows.

`UnsupportedError: Please use "trailing" modifier; removal is not supported for "end"` This is the same error I got at some point when moving arguments.

src/test/suite/fixtures/recorded/languages/rust/chuckNameAir.yml
```diff
@@ -0,0 +1,36 @@
...
+      mark: {type: decoratedSymbol, symbolColor: default, character: a}
+      modifiers:
+        - type: containingScope
+          scopeType: {type: name}
+  usePrePhraseSnapshot: true
+  action: {name: remove}
+initialState:
+  documentContents: |
+    match user {
+        User { first_name: "John", last_name: last_name } => {}
```
auscompgeek, line 16: Can you change this test so the name binding differs from the field name? It's unclear from this test which one it picked to delete.

src/test/suite/fixtures/recorded/languages/rust/chuckValueAir.yml
```diff
@@ -0,0 +1,36 @@
...
+      mark: {type: decoratedSymbol, symbolColor: default, character: a}
+      modifiers:
+        - type: containingScope
+          scopeType: {type: value}
+  usePrePhraseSnapshot: true
+  action: {name: remove}
+initialState:
+  documentContents: |
+    match user {
+        User { first_name: "John", last_name: last_name } => {}
```
auscompgeek, line 16: Ditto - I know it should be deleting `: last_name` but the test doesn't explicitly test that!

src/languages/rust.ts
```diff
@@ -186,24 +186,36 @@ const nodeMatchers: Partial<
...
-    "trait_item[name]",
-    "const_item[name]",
-    "meta_item.identifier!",
-    "let_declaration[pattern]",
-    "constrained_type_parameter[left]",
-    "where_predicate[left]",
-  ],
+  collectionKey: cascadingMatcher(
+    patternMatcher("field_pattern[name]"),
+    trailingMatcher(["field_initializer[name]"], [":"])
```
pokey, line 191:
```suggestion
    trailingMatcher(["field_initializer[name]", "field_pattern[name]"], [":"])
```

pokey (approved):
Looks good!

PR #832, opened 2022-07-05
Fall back to creating a new editor if it can't be reused

AndreasArvidsson:
## Checklist


AndreasArvidsson:
@pokey I had to temporary disable reuse of editor and always default to opening a new one. I tried checking if it was closed, I tried opening a new on error but nothing works reliably in CI.

pokey (approved):
Bummer, but yeah let's get CI green again for now 🤷

PR #836, opened 2022-07-05
Leading/trailing infers token on empty range

### LINKED ISSUE #835: leading/trailing should infer token on empty range
If you say `"chuck trailing"` with your selection on a token you now get an error because your current selection does not have a trailing. We could use the same logic that we do for sub ranges and infer token on empty range.

AndreasArvidsson:
`"chuck trailing"` with empty selection now works

Fixes #835 

## Checklist


src/processTargets/modifiers/LeadingTrailingStages.ts
```diff
@@ -22,10 +23,25 @@ export class TrailingStage implements ModifierStage {
   constructor(private modifier: TrailingModifier) {}
 
   run(context: ProcessedTargetsContext, target: Target): Target[] {
-    const trailing = target.getTrailingDelimiterTarget();
+    const trailing = getTargetToUse(
```
pokey, line 26: if we just make it so that non-empty ranges aren't weak, then can't we just use a weak expander of token?

src/processTargets/marks/CursorStage.ts
```diff
@@ -4,17 +4,21 @@ import { ProcessedTargetsContext } from "../../typings/Types";
 import { isReversed } from "../../util/selectionUtils";
 import { MarkStage } from "../PipelineStages.types";
 import WeakTarget from "../targets/WeakTarget";
```
AndreasArvidsson, line 6: I tried replacing these lines with `import { TokenTarget, WeakTarget } from "../targets";` but that gives me the following error on extension activation: `Activating extension 'pokey.cursorless' failed: Class extends value undefined is not a constructor or null.`. I tried replacing all references to targets to go through the `target/index` file but no go.

pokey: Huh. I can take a look in a follow-up

src/processTargets/targets/WeakTarget.ts
```diff
@@ -1,5 +1,5 @@
 import { Range } from "vscode";
-import { BaseTarget } from ".";
+import { BaseTarget } from "./";
```
pokey, line 2: What's with the `/`?

pokey (commented):
LGTM with one question

src/processTargets/targets/WeakTarget.ts
pokey: Does it break if you remove it?

AndreasArvidsson: No it was just a typo

pokey (approved):
Looks good!

PR #844, opened 2022-07-07
Add cursorless settings command

pokey:
It's a bit awkward to access Cursorless VSCode settings by voice today.  I was hoping to show how to do so in the new tutorial, but would be nice if it were an easier way to show them

## Checklist

### DIFF
--- a/cursorless-talon/src/cursorless.talon
+++ b/cursorless-talon/src/cursorless.talon
@@ -25,7 +25,6 @@ cursorless docks:          user.cursorless_open_instructions()
 cursorless reference:      user.cursorless_cheat_sheet_show_html()
 cursorless cheat sheet:    user.cursorless_cheat_sheet_show_html()
 cursorless settings:
-    sleep(50ms)
     user.vscode("workbench.action.openGlobalSettings")
     sleep(250ms)
     insert("cursorless")


cursorless-talon/src/cursorless.talon
```diff
@@ -24,3 +24,8 @@ cursorless instructions:   user.cursorless_open_instructions()
 cursorless docks:          user.cursorless_open_instructions()
 cursorless reference:      user.cursorless_cheat_sheet_show_html()
 cursorless cheat sheet:    user.cursorless_cheat_sheet_show_html()
+cursorless settings:
```
pokey, line 27: One concern is that we also have Talon-side configuration, eg spoken forms, so I'm slightly hesitant for this voice command to mean VSCode settings

cursorless-talon/src/cursorless.talon
```diff
@@ -24,3 +24,8 @@ cursorless instructions:   user.cursorless_open_instructions()
...
+    sleep(50ms)
+    user.vscode("workbench.action.openGlobalSettings")
+    sleep(250ms)
```
pokey, line 30: Not in love with this and the other arbitrary sleep 🤔

AndreasArvidsson: We already have those for other commands so I guess it's fine until we have something better, but why the leading sleep?

cursorless-talon/src/cursorless.talon
AndreasArvidsson: We could of course open a little imgui with buttons/voice commands that will open a csv file or vscode settings

cursorless-talon/src/cursorless.talon
pokey: Yeah good question.  I have it in my personal "show settings" command.  Presuming I added it for a reason 🤷

cursorless-talon/src/cursorless.talon
pokey: Yeah that's fair; we could always do that later, so prob fine to ship it like this for now

cursorless-talon/src/cursorless.talon
AndreasArvidsson: Feels like a left over from something else maybe? I would remove it.

AndreasArvidsson (approved):

PR #850, opened 2022-07-09
Remove empty trailing line for "chuck item" with trailing comma

### LINKED ISSUE #834: `"chuck item"` leaves empty line when there is trailing comma
In the following typescript:

```typescript
[
    foo,
    bar,
]
```

Saying chuck item bat results in the following:

```typescript
[
    foo,
    
]
```

Note that if `bar` has no trailing comma, it removes the line, as desired.

AndreasArvidsson:
Fixes #834

This actually broke 45 tests. Basically we get off by one errors on the that mark.
`const value = "Hello world";` `"chuck vest"`. At the moment the that mark will be an empty selection adjacent the equal sign. If we prefer the leading range it will be adjacent the const instead. 

## Checklist

### DIFF
--- a/src/processTargets/modifiers/ItemStage/ItemStage.ts
+++ b/src/processTargets/modifiers/ItemStage/ItemStage.ts
@@ -8,6 +8,7 @@ import {
 } from "../../../typings/targetDescriptor.types";
 import { ProcessedTargetsContext } from "../../../typings/Types";
 import { getInsertionDelimiter } from "../../../util/nodeSelectors";
+import { getRangeLength } from "../../../util/rangeUtils";
 import { ModifierStage } from "../../PipelineStages.types";
 import ScopeTypeTarget from "../../targets/ScopeTypeTarget";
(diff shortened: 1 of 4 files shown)


pokey:
Huh. I guess it will also affect the highlight? Might look a bit nicer the way it is now with the highlight extending after the token rather than before?

I don't feel super strongly; could try chucking a couple tokens and see what it feels like?

I guess we could use trailing if leading and trailing are equal; otherwise leading?

Or maybe "item" should just use a removal range in addition to leading and trailing. I believe scope type target will then use that for removal, but we'd still support "trailing" / "leading" modifiers

AndreasArvidsson:
Correct highlight is of course also changed to reflect the new range

It's different by I couldn't say it's worse.

Leveraging removal range is not a bad idea I'm gonna try that.

AndreasArvidsson:
The solution with removal range turned out great

pokey (approved):
Nice!

PR #851, opened 2022-07-10
Added support for html element interior

AndreasArvidsson:
`"chuck inside element"`

Started on #254 

## Checklist


src/util/nodeMatchers.ts
```diff
@@ -188,3 +190,27 @@ export function createPatternMatchers(
   });
   return nodeMatchers as Record<SimpleScopeTypeType, NodeMatcher>;
 }
+
+export const getElementMatcher = (...elementTypes: string[]): NodeMatcher => {
```
pokey, line 194: I don't think you need a full-blown matcher here. An extractor should be cleaner. Can also prob just name it `excludeFirstAndLastChildrenExtractor`

You can make an `elementMatcher` shorthand function that wraps it, but I think the core abstraction here we're looking for is the selector

pokey (changes requested):
Looks good with minor tweak requested

src/util/nodeMatchers.ts
AndreasArvidsson: I can turn it into an extractor, but I do find this is very specific for element because its adding an optional interior range. It's not actually an extractor that focuses on the content range.

pokey: That is precisely what an extractor is for. It takes the node found by the finder and constructs the selection context

AndreasArvidsson: It actually constructs the entire selection but I get your point. I just think that this will never be reused by anything other than an element.

AndreasArvidsson: Done

pokey (approved):
Looks good!

PR #868, opened 2022-07-15
Added glossary docs

AndreasArvidsson:
1. Should this be called vocabulary or maybe glossary?
2. Should this be in the user or contributors docs?

## Checklist


pokey:
> 1. Should this be called vocabulary or maybe glossary?

I think I'd go for glossary

> 2. Should this be in the user or contributors docs?

User makes sense to me. If we start adding more internal stuff we can have a separate page in contributor docs

Also, I'm not sure a table is what we want here. I'd just do a markdown section for each term. Then we can easily link to them

AndreasArvidsson:
Agreed. Sections turned out better.

pokey (approved):
Ok I made some tweaks.  If you're happy feel free to merge

docs/user/glossary.md
```diff
@@ -0,0 +1,55 @@
...
+
+The [range](#range) containing the content of a target. This is the range that will be used when selecting a target, copying a target, as the source of a `"bring"` command, etc.
+
+## Domain
+
+The [range](#range) within which a [scope](#scope) is valid/active. For example, the scope type `"key"` is valid anywhere inside its containing key/value pair, and `"funk name"` is valid anywhere inside its containing function.
+
+## Iteration scope
+
+The parental [scope](#scope) in within to look for new instances of the current [scope](#scope). `"take every line"` will look for lines in its `iteration scope` which is `document`.
```
pokey, line 19:
```suggestion
A canonical [range](#range) for a given [scope type](#scope-type) that defines a list of sibling [scopes](#scope).  For example, the iteration scope for `"arg"` is an argument list.  The iteration scope is used for things like `"every"`, to select all sibling scopes.  This allows us to say things like `"take every arg"`, `"pre every line"`, etc.
```

pokey (commented):
Ok made a couple tweaks; look ok?

pokey (approved):

PR #870, opened 2022-07-15
Fix CI failures by adding sleep backoff

### LINKED ISSUE #845: Within cell set selection fails non deterministically on ci
Seems to happen only on windows

https://github.com/cursorless-dev/cursorless/runs/7238130016?check_suite_focus=true#step:6:2236

pokey:
Instead of sleeping the same amount of time every time we run a test, we double the sleep every time

Fixes #845

## Checklist

### DIFF
--- a/src/test/suite/backwardCompatibility.test.ts
+++ b/src/test/suite/backwardCompatibility.test.ts
@@ -1,16 +1,11 @@
 import * as assert from "assert";
 import * as vscode from "vscode";
-import * as sinon from "sinon";
 import { getCursorlessApi } from "../../util/getExtensionApi";
 import { openNewEditor } from "../openNewEditor";
+import { standardSuiteSetup } from "./standardSuiteSetup";
 
 suite("Backward compatibility", async function () {
-  this.timeout("100s");
-  this.retries(3);
-
-  teardown(() => {
-    sinon.restore();
-  });
+  standardSuiteSetup(this);
 
   test("Backward compatibility", runTest);
 });
--- a/src/test/suite/intraCellSetSelection.test.ts
+++ b/src/test/suite/intraCellSetSelection.test.ts
@@ -1,18 +1,13 @@
 import * as assert from "assert";
 import * as vscode from "vscode";
-import * as sinon from "sinon";
 import { getCursorlessApi } from "../../util/getExtensionApi";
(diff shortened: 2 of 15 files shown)


src/test/suite/backwardCompatibility.test.ts
```diff
@@ -1,16 +1,11 @@
...
+import { standardSuiteSetup } from "./standardSuiteSetup";
 
 suite("Backward compatibility", async function () {
-  this.timeout("100s");
-  this.retries(3);
-
-  teardown(() => {
-    sinon.restore();
-  });
+  standardSuiteSetup(this);
```
pokey, line 8: Notice how for some tests we were retrying 3 times, other times 5.  Put this standard preamble into a utility function to encourage consistency

src/test/suite/intraCellSetSelection.test.ts
```diff
@@ -1,18 +1,12 @@
...
-import * as sinon from "sinon";
 import { getCursorlessApi } from "../../util/getExtensionApi";
 import { openNewNotebookEditor } from "../openNewEditor";
-import sleep from "../../util/sleep";
+import { sleepWithBackoff, standardSuiteSetup } from "./standardSuiteSetup";
 
 // Check that setSelection is able to focus the correct cell
 suite("Within cell set selection", async function () {
-  this.timeout("100s");
-  this.retries(3);
```
pokey, line 11: In particular, notice that the test that was failing in CI was only retrying 3 times, instead of the usual 5

src/test/suite/intraCellSetSelection.test.ts
```diff
@@ -24,7 +18,7 @@ async function runTest() {
 
   // FIXME: There seems to be some timing issue when you create a notebook
   // editor
-  await sleep(1000);
+  await sleepWithBackoff(1000);
```
pokey, line 21: This will now cause it to sleep 250ms, then 500ms, 1s, 2s, 4s.  Hopefully that will both speed things up locally, and resolve the CI issues on Mac / Win

AndreasArvidsson (approved):

pokey:
Ok green 8 times in a row (the two "failures" in that streak are pre-commit complaining that there's a newer commit).  Hopefully this one will last us a while 🤞

PR #872, opened 2022-07-17
eslint: Use consistent-type-imports

auscompgeek:
[esbuild will parse and ignore type imports](https://esbuild.github.io/content-types/#typescript), so let's actually give it this information.

[@typescript-eslint/consistent-type-imports](https://typescript-eslint.io/rules/consistent-type-imports)

Ref: #714

## Checklist


pokey:
Seems fine, but a couple questions: 

- [ ] Will VSCode automatically start using this style when it does an auto-import for us, both via
  - [ ] autocomplete, and
  - [ ] quick fix?
- [ ] Also, will pre-commit automatically fix this for us? I can't remember if we have eslint --fix set up

auscompgeek:
This is ESLint autofixable, so will show up as a quick fix in VSCode. I'd love to know how to make autoimport do the right thing though.

We currently don't have pre-commit running ESLint, I recall you had concerns around pulling in the relevant plugin versions?

pokey:
> This is ESLint autofixable, so will show up as a quick fix in VSCode.

Ah no I don't mean once it's already imported. I mean if you have a symbol that you haven't yet imported and you run quick fix to do the auto import

> I'd love to know how to make autoimport do the right thing though.

Does it not? It might be smart enough to use `type` if we have this eslint config

> We currently don't have pre-commit running ESLint, I recall you had concerns around pulling in the relevant plugin versions?

Ah right yeah I remember now. I think the issue was we had nothing to auto fix so didn't seem worth it. Maybe we should revisit now we have more auto fixers?

pokey:
@auscompgeek curious why you closed this one.  I'm overall supportive of this direction, modulo the questions above

auscompgeek:
I'm not really sure the questions are addressable in any straightforward manner. There's also (unsurprisingly) a lot of merge conflicts by this point, so needs a rebase.

Happy for someone else to tackle this, but this probably doesn't win too much for us right now.

pokey:
I think basically it would involve just pulling this branch, then trying to do an auto-import and see what happens.  But I agree this rule probably isn't a massive win for us, so probably not super high priority

Issue #491, opened 2022-01-17
Add ability to show new features in a webview


pokey:
### The problem

There is currently no way for a user to know when we release new features in cursorless. We'd like some way to highlight new features for users.

### The solution

We'd like to display release notes in a webview when we release new features.  We don't want to do so on every release, especially once we move to continuous delivery (#17), but we should have some mechanism to highlight new features when we want to. We could do something roughly as follows  

- [ ] Have a directory of markdown files, one per each feature that we want to highlight
- [ ] Keep track of which feature highlights the user has seen using global storage, making sure to [sync across instances](https://code.visualstudio.com/api/extension-capabilities/common-capabilities#setkeysforsync-example)
- [ ] Stitch together the highlights they haven't seen and show them in a webview on extension update

PR #881, opened 2022-07-19
Add a status bar item for cursorless

### LINKED ISSUE #837: Add cursorless icon to status bar
Would be nice if we could have a cursorless icon in the status bar. Clicking it could take you to documentation for example.

brollin:
Closes #837. Here is what it looks like so far:

image

Currently when the status bar item is clicked, it just does a `quickOpen` , which I was thinking of changing to a quick pick. Here is what that looks like:

image

Thought I would stop here to solicit feedback. What should happen when we click the button? I was thinking of a quick pick menu for easy access to all of the cursorless palette commands.


brollin:
Also, I looked into opening up the cheat sheet, but from my understanding the cheat sheet needs data from the talon side of things, and triggering it to open from the Vscode-side wouldn't be possible (at least as written). If there's an easy way around that, I can work it into the quick pick menu.

src/core/StatusBarItem.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  init() {
+    const commandId = "cursorless.showQuickPick";
+    const statusBarItem = vscode.window.createStatusBarItem(
+      vscode.StatusBarAlignment.Right,
+      100
+    );
+    statusBarItem.command = commandId;
+    statusBarItem.text = "$(cursorless-icon) Cursorless";
```
pokey, line 18: I think I might just have the icon without text?  Idk, curious what other people (especially with bigger screens 😄) think cc/ @AndreasArvidsson @auscompgeek

src/core/StatusBarItem.ts
```diff
@@ -0,0 +1,33 @@
...
+    statusBarItem.show();
+
+    this.disposables.push(
+      vscode.commands.registerCommand(commandId, this.showQuickPick),
+      statusBarItem
+    );
+  }
+
+  private showQuickPick = () =>
+    vscode.commands.executeCommand("workbench.action.quickOpen", ">cursorless");
```
pokey, line 28: I might capitalise this one
```suggestion
    vscode.commands.executeCommand("workbench.action.quickOpen", ">Cursorless");
```

pokey (commented):
Nice!  A custom quick pick would be slightly slicker, but I think just showing quick pick you have is more than fine for a first version.  In addition to the comments on the code:

- [ ] We should probably add a "Cursorless: Show documentation" command, so that it appears in the menu, and is also generally useful

Btw a nice follow-up might be #491, making it so that this icon changes colour or something when we have new features 🤔

src/core/StatusBarItem.ts
AndreasArvidsson: I prefer the text. We could probably add the option to hide the text in the menu when you click it?

src/core/StatusBarItem.ts
```diff
@@ -0,0 +1,38 @@
...
+    statusBarItem.command = commandId;
+    statusBarItem.text = "$(cursorless-icon) Cursorless";
+    statusBarItem.show();
+
+    this.disposables.push(
+      vscode.commands.registerCommand("cursorless.showDocumentation", () =>
+        vscode.env.openExternal(
+          vscode.Uri.parse("https://www.cursorless.org/docs/")
+        )
+      ),
```
brollin, line 26: Not sure if there is a better place to register this command :thinking:

src/core/StatusBarItem.ts
brollin: For now, the menu is just showing all user-facing Cursorless commands. Having a command to hide that text feels a little funny to me since it would show up in the command palette all the time. 🤔 

As long as it's not taking up too much space, I have a minor preference for the text as well, since I think it's very easy to miss things in the clutter of icons/text in the status bar.

pokey (approved):
Looks good thanks!

PR #888, opened 2022-07-24
Reworked weak target

### LINKED ISSUE #882: "take inside" Does not work when part of the string is already selected
Suppose I have the string  `"hello world"` With a dot on the w.   If I say take whale, the word world is selected. If I then say take inside, the command is recognized but nothing happens. If I say go left  so that now nothing is selected then say take inside,  both hello and world are selected.

### LINKED ISSUE #883: Re-evaluate decision to make non-empty selections non-weak
We're not sure whether the following is a good idea:

https://github.com/cursorless-dev/cursorless/blob/b639e42c6776cf758b758d5f9affbc0f1abb47a5/src/processTargets/marks/CursorStage.ts#L21-L23

## Pros

- https://github.com/cursorless-dev/cursorless/issues/797
- #836

## Cons

- #882

# The solution

- [x] Rename `WeakTarget` to `UntypedRangeTarget` and add a constructor argument `hasExplicitRange`
- [x] Rename `isWeak` to `hasExplicitScopeType`, but need to negate all references to it as it is the opposite
- [x] Check for other references to the term "weak" to make sure we're not missing something
- [x] Introduce `hasExplicitRange` as attribute on `Target`
- [x] For eg `ScopeTypeTarget`, `hasExplicitScopeType = true` and `hasExplicitRange = true`
- [x] For each target in [spreadsheet](https://docs.google.com/spreadsheets/d/1f0xEuE6wXMLJyr9O883ppWmHGGGgcdbubvIiQQob6Ko/edit?usp=sharing), return the proposed type of target as indicated in the **Proposed type** row
- [x] For each action / modifier in [spreadsheet](https://docs.google.com/spreadsheets/d/1f0xEuE6wXMLJyr9O883ppWmHGGGgcdbubvIiQQob6Ko/edit?usp=sharing), depending on the value of column **Checks for implicit**, check either `!hasExplicitRange` or `!hasExplicitScopeType`
- [x] Add tests for every cell in [spreadsheet](https://docs.google.com/spreadsheets/d/1f0xEuE6wXMLJyr9O883ppWmHGGGgcdbubvIiQQob6Ko/edit?usp=sharing)

AndreasArvidsson:
Fixes #883
Fixes #882

## Checklist

- [x] Add checkboxes to issues that will unlock additional test cases for cells in our [spreadsheet](https://docs.google.com/spreadsheets/d/1f0xEuE6wXMLJyr9O883ppWmHGGGgcdbubvIiQQob6Ko/edit#gid=1018390187)


AndreasArvidsson:
@pokey Tests added. Some test or not currently possible. For example `"inside funk"`. I also don't have `"funk wrap` and `"snip funk"` wired up. Could you please add those tests?

src/processTargets/modifiers/ItemStage/ItemStage.ts
```diff
@@ -40,10 +40,10 @@ export default class ItemStage implements ModifierStage {
   private getEveryTarget(context: ProcessedTargetsContext, target: Target) {
     const itemInfos = getItemInfosForIterationScope(context, target);
 
-    // If weak expand to all items in iteration scope
-    const filteredItemInfos = target.isWeak
-      ? itemInfos
-      : filterItemInfos(target, itemInfos);
+    // If target has explicit scope type filter to items in that scope. Otherwise expand to all items in iteration scope.
+    const filteredItemInfos = target.hasExplicitScopeType
```
pokey, line 44: I thought this was where we wanted to use `hasExplicitRange`?

src/processTargets/targetUtil/createContinuousRange.ts
```diff
@@ -67,21 +67,23 @@ export function createSimpleContinuousRangeTarget(
...
     editor: startTarget.editor,
     isReversed,
-    contentRange: createContinuousRange(
-      startTarget,
-      endTarget,
-      includeStart,
-      includeEnd
-    ),
+    contentRange,
+    hasExplicitRange: !contentRange.isEmpty,
```
pokey, line 87: I'd be tempted to just go `true` here

src/processTargets/modifiers/ModifyIfImplicitScopeTypeStage.ts
```diff
@@ -0,0 +1,37 @@
+import { Target } from "../../typings/target.types";
+import { ModifyIfImplicitScopeTypeModifier } from "../../typings/targetDescriptor.types";
+import { ProcessedTargetsContext } from "../../typings/Types";
+import getModifierStage from "../getModifierStage";
+import { ModifierStage } from "../PipelineStages.types";
+
+/**
+ * Runs {@link ModifyIfImplicitScopeTypeModifier.modifier} if the target is lacking an explicit modifier.
+ */
+export default class ModifyIfImplicitScopeTypeStage implements ModifierStage {
```
pokey, line 10: I don't love the term "implicit scope type", because that implies it has a scope type. I think the clearest would be `ModifyIfNoExplicitScopeTypeStage`, but that's starting to get a bit verbose even for me, esp when you get to `containingSurroundingPairIfNoExplicitScopeTypeStage` 😅. So maybe we go with something like `ModifyIfUntypedStage`, and then eg `containingSurroundingPairIfUntypedStage`?

pokey (commented):
Nice! Haven't reviewed tests yet but overall looks like what we agreed on. Left a few comments

AndreasArvidsson:
okay updated

src/processTargets/modifiers/ModifyIfImplicitScopeTypeStage.ts
```diff
@@ -0,0 +1,37 @@
+import { Target } from "../../typings/target.types";
```
pokey, line 1: Don't forget to rename file

src/typings/targetDescriptor.types.ts
```diff
@@ -203,10 +203,10 @@ export interface HeadTailModifier {
 }
 
 /**
- * Runs {@link modifier} if the target is weak.
+ * Runs {@link modifier} if the target is lacking an explicit scope type.
  */
-export interface ModifyIfWeakModifier {
-  type: "modifyIfWeak";
+export interface ModifyIfImplicitScopeTypeModifier {
```
pokey, line 208:
```suggestion
export interface ModifyIfUntypedModifier {
```

src/typings/targetDescriptor.types.ts
```diff
@@ -239,7 +239,7 @@ export type Modifier =
   | LeadingModifier
   | TrailingModifier
   | RawSelectionModifier
-  | ModifyIfWeakModifier
+  | ModifyIfImplicitScopeTypeModifier
```
pokey, line 242:
```suggestion
  | ModifyIfUntypedModifier
```

src/core/commandRunner/CommandRunner.ts
```diff
@@ -113,6 +113,12 @@ export default class CommandRunner {
         getNodeAtLocation: this.graph.getNodeAtLocation,
       };
 
+      // Process targets before the `testCaseRecorder.preCommandHook()` to make sure all marks actually exist.
+      const targets = processTargets(
```
pokey, line 117: How come this move wasn't necessary before?

pokey (commented):
Yeah looking better. Few more minor comments then I'll have a look at the tests and we should be good to go!

pokey:
> @pokey Tests added. Some test or not currently possible. For example `"inside funk"`, or modifiers after range. I also don't have `"funk wrap` and `"snip funk"` wired up. Could you please add those tests?

sure happy to record the snippet ones.

For "inside funk", we're actually looking for a scope type that doesn't have an "inside" to ensure that it doesn't expand, right?  I don't see why we can't test that today.  Would just be an error test

src/core/commandRunner/CommandRunner.ts
AndreasArvidsson: This is just a bug I found.  has nothing to do with these changes actually. If you try reference a mark that does not exist to while you recording a test you get an undefined exception.

AndreasArvidsson:
I think I've added all the test I can now
1. The function scope doesn't really have inside boundaries or proper leading/trailing delimiters
2. Our test recorded doesn't record breakpoints
3. The implementation of that mark is the same as cursor.

src/processTargets/marks/ThatStage.ts
```diff
@@ -1,16 +1,49 @@
...
 }
+
+export class SourceStage implements MarkStage {
+  constructor(_modifier: SourceMark) {
+    // Empty
+  }
+
+  run(context: ProcessedTargetsContext): Target[] {
+    if (context.sourceMark.length === 0) {
+      throw Error("No available that marks");
```
AndreasArvidsson, line 32: This should be source marks

src/processTargets/marks/ThatStage.ts
```diff
@@ -1,16 +1,49 @@
...
+import { isReversed } from "../../util/selectionUtils";
 import { MarkStage } from "../PipelineStages.types";
-import { cursorSelectionsToTarget } from "./CursorStage";
+import { UntypedTarget } from "../targets";
 
-export default class implements MarkStage {
-  constructor(private modifier: ThatMark) {}
+export class ThatStage implements MarkStage {
+  constructor(_modifier: ThatMark) {
+    // Empty
```
AndreasArvidsson, line 13: Is this really a better syntax than just having private?

src/processTargets/modifiers/scopeTypeStages/ParagraphStage.ts
```diff
@@ -21,10 +21,9 @@ export default class implements ModifierStage {
 
   private getEveryTarget(target: Target): ParagraphTarget[] {
     const { contentRange, editor } = target;
-    const { isEmpty } = contentRange;
     const { lineCount } = editor.document;
-    const startLine = isEmpty ? 0 : contentRange.start.line;
-    const endLine = isEmpty ? lineCount - 1 : contentRange.end.line;
+    const startLine = target.hasExplicitRange ? contentRange.start.line : 0;
+    const endLine = target.hasExplicitRange ? contentRange.end.line : lineCount;
```
pokey, line 26: Why was there a `- 1` here?  It caused a bug when the selection ended within the first line of a block

pokey: Oh ok wow this is all messed up.  One sec fixing

pokey (approved):

PR #891, opened 2022-07-25
[WIP] Add Elm

Gauteab:
## What

Adds support for the Elm programming language

## Checklist

- [ x ] Recorded tests for the new language
- [ x ] Used `"change"` / `"clear"` instead of` "take"` for selection tests to make recorded tests easier to read
- [ x ] Added a few specific tests that use `"chuck"` instead of `"change"` to test removal behaviour when it's interesting, especially:
  - [ x ] `"chuck arg"` with single argument in list
  - [ x ] `"chuck arg"` with multiple arguments in list
  - [ x ] `"chuck item"` with single argument in list
  - [ x ] `"chuck item"` with multiple arguments in list
- [ x ] Supported` "type"` both for type annotations (eg `foo: string`) and declarations (eg `interface Foo {}`) (and added tests for this behaviour 😊)
- [ x ] Supported` "item"` both for map pairs and list entries (with tests of course)

## Currently not working

- Strings
- "take every arg"

### DIFF
--- a/src/languages/elm.ts
+++ b/src/languages/elm.ts
@@ -1,0 +1,55 @@
+import { SimpleScopeTypeType } from "../typings/targetDescriptor.types";
+import { NodeMatcherAlternative } from "../typings/Types";
+import { createPatternMatchers } from "../util/nodeMatchers";
+
+const nodeMatchers: Partial<
+  Record<SimpleScopeTypeType, NodeMatcherAlternative>
+> = {
+  map: ["record_expr", "record_pattern"],
+  list: ["list_expr", "list_pattern"],
+  statement: [
+    // not all of these are "statements" in Elm, but they fit the usage pattern in Cursorless
+    "let_in_expr",
+    "case_of_expr",
+    "port_annotation",
+    "infix_declaration",
+    "if_else_expr",
+  ],
+  string: "string_constant_expr",
+  collectionItem: ["exposed_value"],
+  collectionKey: ["field_type[name]", "record_expr.field![name]"],
+  ifStatement: "if_else_expr",
+  anonymousFunction: "anonymous_function_expr",
+  functionCall: "function_call_expr",
+  functionCallee: "function_call_expr[target][name]",
+  comment: ["block_comment", "line_comment"],
+  namedFunction: "value_declaration",
+  functionName: "value_declaration[functionDeclarationLeft][0]",
+  condition: "if_else_expr[0]",
+  type: [
+    "type_ref",
+    "type_expression",
+    "type_declaration",
+    "type_annotation[typeExpression]!",
+    "type_alias_declaration",
+    // TODO: select type from within function (needs to find the previous sibling of the parent `value_decoration`)
+  ],
+  name: [
+    "type_annotation[name]",
+    "port_annotation[name]",
+    "type_declaration[name]",
+    "type_alias_declaration[name]",
+    "value_declaration.function_declaration_left.lower_case_identifier!",
+  ],
+  value: [
+    "record_expr[field][expression]",
+    "value_declaration[body]",
+    "case_of_branch[expr]",
+    "let_in_expr[body]",
+  ],
+  argumentOrParameter: [
+    "value_declaration.function_declaration_left[pattern]!",
+  ],
+};
+
+export default createPatternMatchers(nodeMatchers);
--- a/src/test/suite/fixtures/recorded/languages/elm/changeFunkName2.yml
+++ b/src/test/suite/fixtures/recorded/languages/elm/changeFunkName2.yml
@@ -1,0 +1,30 @@
+languageId: elm
+command:
+  spokenForm: change funk name
+  version: 2
+  targets:
+    - type: primitive
+      modifiers:
+        - type: containingScope
+          scopeType: {type: functionName}
+  usePrePhraseSnapshot: false
+  action: {name: clearAndSetSelection}
+initialState:
+  documentContents: |
+    f x =
+        x
+  selections:
+    - anchor: {line: 1, character: 4}
+      active: {line: 1, character: 4}
+  marks: {}
+finalState:
+  documentContents: |2
+     x =
+        x
+  selections:
+    - anchor: {line: 0, character: 0}
+      active: {line: 0, character: 0}
+  thatMark:
+    - anchor: {line: 0, character: 0}
+      active: {line: 0, character: 0}
+fullTargets: [{type: primitive, mark: {type: cursor}, modifiers: [{type: containingScope, scopeType: {type: functionName}}]}]
--- a/src/test/suite/fixtures/recorded/languages/elm/changeIfStateInk.yml
+++ b/src/test/suite/fixtures/recorded/languages/elm/changeIfStateInk.yml
@@ -1,0 +1,32 @@
+languageId: elm
+command:
+  spokenForm: change if state ink
+  version: 2
+  targets:
+    - type: primitive
+      mark: {type: decoratedSymbol, symbolColor: default, character: i}
+      modifiers:
+        - type: containingScope
+          scopeType: {type: ifStatement}
+  usePrePhraseSnapshot: true
+  action: {name: clearAndSetSelection}
+initialState:
+  documentContents: |
+    _ = if True then 1 else 0
+  selections:
+    - anchor: {line: 1, character: 0}
+      active: {line: 1, character: 0}
+  marks:
+    default.i:
+      start: {line: 0, character: 4}
+      end: {line: 0, character: 6}
+finalState:
+  documentContents: |
+    _ = 
+  selections:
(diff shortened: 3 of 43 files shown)


src/languages/elm.ts
```diff
@@ -0,0 +1,55 @@
...
+  namedFunction: "value_declaration",
+  functionName: "value_declaration[functionDeclarationLeft][0]",
+  condition: "if_else_expr[0]",
+  type: [
+    "type_ref",
+    "type_expression",
+    "type_declaration",
+    "type_annotation[typeExpression]!",
+    "type_alias_declaration",
+    // TODO: select type from within function (needs to find the previous sibling of the parent `value_decoration`)
```
pokey, line 35: Did you want help with this one?

src/languages/elm.ts
```diff
@@ -0,0 +1,55 @@
...
+    "type_alias_declaration[name]",
+    "value_declaration.function_declaration_left.lower_case_identifier!",
+  ],
+  value: [
+    "record_expr[field][expression]",
+    "value_declaration[body]",
+    "case_of_branch[expr]",
+    "let_in_expr[body]",
+  ],
+  argumentOrParameter: [
```
pokey, line 50: This should support both formal parameters and function call arguments. Looks like you're missing function call arguments?

src/languages/elm.ts
```diff
@@ -0,0 +1,55 @@
...
+  statement: [
+    // not all of these are "statements" in Elm, but they fit the usage pattern in Cursorless
+    "let_in_expr",
+    "case_of_expr",
+    "port_annotation",
+    "infix_declaration",
+    "if_else_expr",
+  ],
+  string: "string_constant_expr",
+  collectionItem: ["exposed_value"],
```
pokey, line 19: Fwiw for comma-separated lists, if you just don't support them you'll get them for free.  You only need to implement "item" for cases where either of the following is true:

- The list is not surrounded by a paired delimiter such as `[]`, `()` etc, or
- The list is not comma-separated

If the list is comma-separated and surrounded by delimiters, it's supported automatically via textual magic

I'd still leave your tests, but try removing this line and seeing if they still pass

```suggestion
```

src/test/suite/fixtures/recorded/languages/elm/changeFunkName2.yml
```diff
@@ -0,0 +1,30 @@
+languageId: elm
```
pokey, line 1: This looks the same as the other "change funk name".  Am I missing something?

src/test/suite/fixtures/recorded/languages/elm/changeIfStateInk.yml
```diff
@@ -0,0 +1,32 @@
+languageId: elm
```
pokey, line 1: This looks functionally identical to the other if state test.  What is this one supposed to be testing?

src/languages/elm.ts
```diff
@@ -0,0 +1,55 @@
...
+    "type_annotation[typeExpression]!",
+    "type_alias_declaration",
+    // TODO: select type from within function (needs to find the previous sibling of the parent `value_decoration`)
+  ],
+  name: [
+    "type_annotation[name]",
+    "port_annotation[name]",
+    "type_declaration[name]",
+    "type_alias_declaration[name]",
+    "value_declaration.function_declaration_left.lower_case_identifier!",
```
pokey, line 42: How come this is different from `functionName` above?

src/test/suite/fixtures/recorded/languages/elm/changeKey.yml
```diff
@@ -0,0 +1,30 @@
+languageId: elm
```
pokey, line 1: This one should actually be "name".  You could prob also support "key" if you'd like, but "name" is more typical when the thing after the `:` is a type

src/test/suite/fixtures/recorded/languages/elm/changeType2.yml
```diff
@@ -0,0 +1,30 @@
+languageId: elm
```
pokey, line 1: Hmm.  I think I'd go "value" rather than "type" here

src/test/suite/fixtures/recorded/languages/elm/changeTypeRed.yml
```diff
@@ -0,0 +1,30 @@
+languageId: elm
```
pokey, line 1: Mixed feelings on this one.  We do typically support "type" for return type, but not usually when it's part of a single type annotation

In Cursorless, "type" should select type declarations and type annotations

But I don't feel super strongly in this case, esp if this is the only way to declare a function return type?

src/test/suite/fixtures/recorded/languages/elm/changeKey.yml
```diff
@@ -0,0 +1,30 @@
...
+  targets:
+    - type: primitive
+      modifiers:
+        - type: containingScope
+          scopeType: {type: collectionKey}
+  usePrePhraseSnapshot: true
+  action: {name: clearAndSetSelection}
+initialState:
+  documentContents: |
+    f : { x : Int }
```
pokey, line 14: Should support "type" for `Int` here as well.  Maybe you already do, but either way a test for that would be nice 😊

src/languages/elm.ts
```diff
@@ -0,0 +1,55 @@
...
+  ],
+  string: "string_constant_expr",
+  collectionItem: ["exposed_value"],
+  collectionKey: ["field_type[name]", "record_expr.field![name]"],
+  ifStatement: "if_else_expr",
+  anonymousFunction: "anonymous_function_expr",
+  functionCall: "function_call_expr",
+  functionCallee: "function_call_expr[target][name]",
+  comment: ["block_comment", "line_comment"],
+  namedFunction: "value_declaration",
```
pokey, line 26: Is this not too generic?

src/languages/elm.ts
```diff
@@ -0,0 +1,55 @@
+import { SimpleScopeTypeType } from "../typings/targetDescriptor.types";
+import { NodeMatcherAlternative } from "../typings/Types";
+import { createPatternMatchers } from "../util/nodeMatchers";
+
+const nodeMatchers: Partial<
+  Record<SimpleScopeTypeType, NodeMatcherAlternative>
+> = {
+  map: ["record_expr", "record_pattern"],
+  list: ["list_expr", "list_pattern"],
+  statement: [
```
pokey, line 10: don't you want `value_declaration` here?

pokey (commented):
Wow super solid first PR!  Looks like you've done your homework 😄.  I left a bunch of minor inline comments. 

In addition, it would be great if you could add the following tests:

- [ ] "change arg"
- [ ] "change value" for values in a map
- [ ] "change state" for various statement types
- [ ] "change string"
- [ ] "change comment"
- [ ] "change condition"

pokey:
Oh yeah also lmk if you want help with getting strings to work, and with getting "change round" to work inside strings

src/test/suite/fixtures/recorded/languages/elm/changeKey.yml
Gauteab: If the cursor is on `x`, should `type` select `Int` (the type of `x`) or `{ x : Int }` (the closest enclosing type)? 🤔 
Currently it does the latter

src/languages/elm.ts
Gauteab: Not sure, but tree-sitter-elm does not distinguish functions and other values (see image).
I kinda got used to thinking that values are functions with zero arguments.

I suppose it's not right if you should be able to select this with `take func` and with the cursor within `z = 0`:
`f x = let z = 0 in x`

So I could add `"value_declaration.anonymous_function_expr"` to cover `g = \x -> x`, and then something to select "value_declaration with at least one pattern in function_declaration_left".

image

src/test/suite/fixtures/recorded/languages/elm/changeKey.yml
pokey: Should be the former (ie `Int`)

src/languages/elm.ts
pokey: Huh. Yeah I guess that makes sense?

Feel free to drop into one of the meet-ups if you want to discuss; see [the calendar](https://tinyurl.com/talon-events)

src/test/suite/fixtures/recorded/languages/elm/changeTypeRed.yml
auscompgeek: What do we do for Haskell?

pokey: We don't support Haskell yet

pokey:
Closing this one as it looks like it's been stalled for a bit.  Filed #1088 to track.  Feel free to re-open if you get a chance to hack on it again 😊

PR #892, opened 2022-07-25
Improve snapshot update with expected-error tests

pokey:
This PR adds support for the Update fixtures launch configuration to change an error test to a success test, and vice versa.  Also adds support for Update fixtures to change the error name 

## Checklist

### DIFF
--- a/src/test/suite/recorded.test.ts
+++ b/src/test/suite/recorded.test.ts
@@ -111,26 +111,40 @@ async function runTest(file: string) {
   // Assert that recorded decorations are present
   checkMarks(fixture.initialState.marks, readableHatMap);
 
-  if (fixture.thrownError != null) {
-    await assert.rejects(
-      async () => {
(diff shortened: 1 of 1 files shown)


src/test/suite/recorded.test.ts
```diff
@@ -111,26 +111,40 @@ async function runTest(file: string) {
   // Assert that recorded decorations are present
   checkMarks(fixture.initialState.marks, readableHatMap);
 
-  if (fixture.thrownError != null) {
-    await assert.rejects(
-      async () => {
-        await vscode.commands.executeCommand(
-          "cursorless.command",
-          fixture.command
```
pokey, line 119: Note that the fact we had two callsites for executing the cursorless command resulted in this one rotting: we override `usePrePhraseSnapshot` below but not here 😳

AndreasArvidsson (approved):

PR #895, opened 2022-07-27
Fix inside ranges

pokey:
Fixes `"take inside air past bat"`.  Need to wait for #888 to merge

## Checklist


src/processTargets/targets/InteriorTarget.ts
```diff
@@ -32,4 +38,34 @@ export default class InteriorTarget extends BaseTarget {
...
+
+      return new constructor({
+        ...this.getCloneParameters(),
+        isReversed,
+        fullInteriorRange: createContinuousRangeFromRanges(
+          this.fullInteriorRange,
+          endTarget.fullInteriorRange,
+          includeStart,
+          includeEnd
+        ),
```
pokey, line 59: The issue was that we were letting the base target construct the range, but base target is ignorant of `fullInteriorRange` so it just ended up spreading it from the start of the range

src/test/suite/fixtures/recorded/surroundingPair/takeCoreRiskPastJustHarp.yml
```diff
@@ -0,0 +1,45 @@
+languageId: plaintext
```
pokey, line 1: This test actually worked before, but seemed worth having a test for the code that downgrades to weak target

src/processTargets/targets/InteriorTarget.ts
AndreasArvidsson: I would argue that we don't want full interior range in this case. If you make arrange from one inside to the other I would assume that that only used the content range.

pokey: Functionally speaking, what's the difference? See the test: it's not including whitespace

AndreasArvidsson: Then what's the point?

AndreasArvidsson: I guess if you do a chuck there will a difference.

AndreasArvidsson: That is probably the test to add

AndreasArvidsson: btw take tests are so hard to read. Why not use change?

pokey: Yeah good call to add a "chuck" test

InteriorRange ignores contentRange so I have to use the full range

I used "take" because "change" would have resulted in a fairly unclear result imo. Ie where are the spaces and parens coming from?

AndreasArvidsson: I understand. A proper test fewer can't come fast enough imo.

AndreasArvidsson (approved):

Issue #81, opened 2021-07-07
Decorations break font ligatures


brxck:
Ligatures between characters are broken when they are decorated.

![ligatures](https://user-images.githubusercontent.com/28877984/124823092-a500a900-df25-11eb-8fd6-a34cbe6a4b13.png)

However, a possible solution also appears in this image: the ligature is preserved when only the last character is decorated.

pokey:
duplicate of #71

PR #899, opened 2022-07-28
Add grapheme Unicode normalization support for hats

### LINKED ISSUE #276: Add setting to strip unicode characters
We now have support for unicode characters in tokenizer regex. If the user doesn't have these characters bound to a voice command we could add a setting to just strip them. Turning `å` into `a`

`text = text.normalize("NFD").replace(/\p{Diacritic}/gu, "");`

### LINKED ISSUE #896: Allow case-sensitive targets (lowercase AND uppercase)
For anybody with a lowercase and uppercase alphabet (air = a, ace = A, bus = b, Bill = B...) it would be nice to be able to toggle target case sensitivity as an additional way to expand target options rather than just colours or shapes.

pokey:
- Fixes #896
- Fixes #276

## Checklist

- [x] Fix ordinal char modifier to handle NFD
- [x] Add `preserveSymbolsRegex` setting that runs before every other normalisation except for NFC
- [x] Add "special" to csv
- [x] Update docs and cheatsheet
  - [x] "special"
  - [x] Add Unicode section to docs
  - [x] #896
- [x] Stop splitting tokens twice
- [x] Record a couple "clear" tests with unicode characters
- [x] Add normaliser for unspeakable characters? (eg emoji etc).  Maybe user can have regex that defines list of speakable characters.   Unspeakable characters would all get mapped to single special lexeme that then has spoken form for entire class Talon-side


src/util/addDecorationsToEditor.ts
```diff
@@ -169,3 +168,33 @@ export function addDecorationsToEditors(
...
+  text: string;
+
+  /** The start offset of the lexeme within its containing token */
+  tokenStartOffset: number;
+
+  /** The end offset of the lexeme within its containing token */
+  tokenEndOffset: number;
+}
+
+function getTokenLexemes(
```
pokey, line 183: Note how this generic function will make it easier to handle unicode with diacritics, as well as #81

src/util/getTokensInRange.ts
```diff
@@ -12,7 +12,7 @@ export function getTokensInRange(
   editor: vscode.TextEditor,
   range: vscode.Range
 ): PartialToken[] {
-  const text = editor.document.getText(range).toLowerCase();
+  const text = editor.document.getText(range);
```
pokey, line 15: This function really shouldn't have been lower-casing, as it is supposed to be a generic function

AndreasArvidsson: Definitely

src/util/addDecorationsToEditor.ts
```diff
@@ -169,3 +168,33 @@ export function addDecorationsToEditors(
...
+  text: string,
+  tokenHatSplittingMode: TokenHatSplittingMode
+): Lexeme[] {
+  switch (tokenHatSplittingMode) {
+    case "standard":
+    case "preserveCase": {
+      const normalisedText =
+        tokenHatSplittingMode === "preserveCase" ? text : text.toLowerCase();
```
AndreasArvidsson, line 191: Doing a fall through and then doing a check for the switch value inside case to do different behavior is a bit weird. Why not just remove the switch?

pokey: Was trying to plan ahead.  But I think I'll just switch to boolean anyway

src/util/addDecorationsToEditor.ts
```diff
@@ -169,3 +169,45 @@ export function addDecorationsToEditors(
...
+  text: string,
+  tokenHatSplittingMode: TokenHatSplittingMode
+): Lexeme[] {
+  const { preserveCase, removeAccents } = tokenHatSplittingMode;
+
+  if (removeAccents) {
+    return matchAll<Lexeme>(text, /\p{L}\p{M}*|\P{L}/gu, (match) => {
+      const matchTextNoAccents = match[0]
+        .normalize("NFD")
+        .replace(/\p{M}/gu, "");
```
pokey, line 194: Tbh, we should probably always split this way and do `NFD` normalisation, even if they want to preserve accents.  And then do the same normalisation when they say "air".  Otherwise they might have weird behaviour where it only sometimes works depending whether the char uses combining mark or single codepoint.  Does that make sense?

package.json
```diff
@@ -155,6 +155,16 @@
...
+				"cursorless.tokenHatSplittingMode.preserveCase": {
+					"type": "boolean",
+					"default": false,
+					"description": "Whether to distinguished between uppercase and lower case letters for hat"
+				},
+				"cursorless.tokenHatSplittingMode.removeAccents": {
+					"type": "boolean",
+					"default": false,
+					"description": "Whether to remove accents when constructing hats"
+				},
```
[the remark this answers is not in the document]
AndreasArvidsson: I agreed that removing accents should be the default. `preserveAccents` sounds good

src/core/TokenGraphemeSplitter.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  private updateTokenHatSplittingMode() {
+    const { accentsToPreserve, symbolsToPreserve, ...rest } =
+      this.graph.ide.configuration.getOwnConfiguration<TokenHatSplittingMode>(
+        "tokenHatSplittingMode"
+      )!;
+
+    this.tokenHatSplittingMode = {
+      accentsToPreserve: accentsToPreserve.map((grapheme) =>
+        grapheme.toLowerCase().normalize("NFC")
```
pokey, line 86: This is a bit awkward.  We still letters to lowercase here even if they have set `preserveCase=true`.  I wasn't entirely sure what to do in that case.  To me it seemed that even if they want to preserve case, it's unlikely they'd want to preserve accents for lowercase version of letter but not uppercase

Fwiw we can get around this issue entirely by only paying attention to the accents themselves in this string.  Basically do NFD normalisation and then keep only the diacritics (kind of the opposite of what we do to strip the marks below).  That would mean that they can't keep a particular diacritic for one char but not another, ie preserve accent for `ö` but not `ä`

cc/ @AndreasArvidsson @auscompgeek

AndreasArvidsson:
Appears to be working for my user case :)

src/test/suite/runTestSubset.ts
```diff
@@ -4,7 +4,7 @@
  * configuration.
  * See https://mochajs.org/#-grep-regexp-g-regexp for supported syntax
  */
-export const TEST_SUBSET_GREP_STRING = "implicit";
+export const TEST_SUBSET_GREP_STRING = "Grapheme";
```
AndreasArvidsson, line 7: This bastard gets updated all the time apparently :D

pokey: haha yeah I wonder if it's worth coming up with some alternative.  Doesn't seem like too big of a deal tho i guess 🤷‍♂️

AndreasArvidsson (dismissed):

src/test/suite/runTestSubset.ts
AndreasArvidsson: No it's fine

src/core/TokenGraphemeSplitter.ts
auscompgeek: In principle I agree too. I wonder if there are any characters that have accents where `upper(lower(char)) != char`.

pokey: @auscompgeek what would be the problem if such characters existed? I don't believe I'm relying on that fact anywhere, but definitely might be missing something

src/test/suite/runTestSubset.ts
auscompgeek: Would probably be nicer to make it purely a CLI thing, but we can kick the can down the road a bit.

pokey: How would we make it a CLI thing?  This variable only has an impact when you run a particular VSCode launch config

src/core/TokenGraphemeSplitter.ts
auscompgeek: I'm probably overthinking it, but just thinking of characters that behave like "ß" for example (but in the other direction for this particular scenario). In such a case it'd probably be a mistake to treat the lowercase and uppercase of such characters as the same.

src/test/suite/runTestSubset.ts
auscompgeek: Launch configs can accept input, see https://code.visualstudio.com/docs/editor/variables-reference#_input-variables

pokey: that's an interesting idea.  Makes it a bit harder to run the same set of tests over and over again, but does avoid this git fight

src/core/TokenGraphemeSplitter.ts
pokey: huh interesting.  I guess for these cases one could just use the other setting (symbols to preserve)? An example use case for that was `Σ` vs `σ`, where you'd probably want different spoken forms for those if you're just using them for math, even if you prefer case normalisation for Latin alphabet

auscompgeek: Ah yeah, makes sense. :+1:

auscompgeek (dismissed):
For the most part I'll defer to Andreas for this one :)

src/ide/ide.types.ts
```diff
@@ -0,0 +1,25 @@
...
+   * Register disposables to be disposed of on IDE exit.
+   *
+   * @param disposables A list of {@link Disposable}s to dispose when the IDE is exited.
+   * @returns A function that can be called to deregister the disposables
+   */
+  disposeOnExit(...disposables: Disposable[]): () => void;
+}
+
+export interface Configuration {
+  getOwnConfiguration<T>(key: string): T | undefined;
```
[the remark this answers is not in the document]
pokey: Ooh good call. That's a nice advantage of having our own ide interface

src/core/TokenGraphemeSplitter.ts
```diff
@@ -0,0 +1,212 @@
...
+   * Splits {@link token} into a list of graphemes, normalised as per
+   * {@link normalizeGrapheme}.
+   * @param token The token to split
+   * @returns A list of normalised graphemes in {@link token}
+   */
+  getTokenGraphemes = (token: string): Grapheme[] =>
+    matchAll<Grapheme>(token, SPLIT_REGEX, (match) => ({
+      text: this.normalizeGrapheme(match[0]),
+      tokenStartOffset: match.index!,
+      tokenEndOffset: match.index! + match[0].length,
```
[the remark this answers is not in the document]
pokey: Nope; didn't work 😕

src/ide/ide.types.ts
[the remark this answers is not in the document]
pokey: Done; lmk what you think

src/processTargets/modifiers/OrdinalRangeSubTokenStage.ts
```diff
@@ -37,17 +37,12 @@ export default class OrdinalRangeSubTokenStage implements ModifierStage {
...
-        start: index,
-        end: index + 1,
-      }));
-    }
+    const regex =
+      this.modifier.scopeType.type === "word" ? SUBWORD_MATCHER : SPLIT_REGEX;
+    pieces = [...tokenText.matchAll(regex)].map((match) => ({
+      start: match.index!,
+      end: match.index! + match[0].length,
+    }));
```
pokey, line 45: This seemed worth fixing while we were here; we weren't properly handling combining chars.  @AndreasArvidsson you'll be happy to see we've replaced another scope type by a regex 😄

AndreasArvidsson: Nice :D

src/core/TokenGraphemeSplitter.ts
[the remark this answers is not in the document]
auscompgeek: Huh, wonder why TypeScript is unhappy with this.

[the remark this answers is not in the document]
pokey: From some basic testing, looks like that `!` assertion just applies at the place where you put it.  Doesn't influence inference elsewhere

src/ide/ide.types.ts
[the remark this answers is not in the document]
auscompgeek: Nice!

PR #900, opened 2022-07-28
Updated how positional target calculates line indentation

### LINKED ISSUE #833: Don't use final line indentation for move after
When determining how to indent an inserted target for `"pour"` or `"move after"`, we look at the final line of the target that is being poured or inserted after. This results in `"move funk air after bat"` in python having the first line of the new function indented. For example:

```python
def foo():
    pass

    def foo():
    pass
```

We should instead use either the first line of the target or the line with minimum indentation to determine desired indentation for new target

AndreasArvidsson:
Fixes #833

Changed how line indentation/padding is calculated

## Checklist


src/actions/BringMoveSwap.ts
```diff
@@ -163,12 +163,13 @@ class BringMoveSwap implements Action {
...
-            getSelectionInfo(
-              editor.document,
-              originalTarget.contentSelection,
-              DecorationRangeBehavior.OpenOpen
-            )
+          const editSelectionInfos = edits.map(
+            ({ edit: { range }, originalTarget }) =>
+              getSelectionInfo(
+                editor.document,
+                selectionFromRange(originalTarget.isReversed, range),
```
pokey, line 170: This should always use the edit range.  Before, the edit range and `originalTarget.contentSelection` were equal, so it didn't matter, but now that the edit range is at the start of the line for line targets, the range updater wasn't actually updating the edit range 😳.  It appears that the way you implemented `updateRange` allowed you to work around this bug, but when I tried to simplify things, the bug became apparent

src/processTargets/targets/PositionTarget.ts
```diff
@@ -102,37 +105,23 @@ export default class PositionTarget extends BaseTarget {
   }
 
   private getEditText(text: string) {
-    if (this.isLineDelimiter) {
-      return this.isBefore
-        ? this.linePadding + text + this.insertionDelimiter
-        : this.insertionDelimiter + this.linePadding + text;
-    }
+    const insertionText = this.indentationString + text;
```
pokey, line 108: We don't need the special casing, because `indentationString` is just empty when `!this.isLineDelimiter`

src/processTargets/targets/PositionTarget.ts
```diff
@@ -102,37 +105,23 @@ export default class PositionTarget extends BaseTarget {
...
-      return this.isBefore
-        ? startOffset
-        : startOffset + this.insertionDelimiter.length;
-    })();
+    const baseStartOffset = this.editor.document.offsetAt(range.start);
+    const startIndex = this.isBefore
+      ? baseStartOffset + this.indentationString.length
+      : baseStartOffset +
+        this.insertionDelimiter.length +
+        this.indentationString.length;
```
pokey, line 121: As above, we don't need the special casing, because `indentationString` is just empty when `!this.isLineDelimiter`

pokey (dismissed):
Looks good!  This is going to make Python editing much more pleasant.  I made some tweaks in https://github.com/cursorless-dev/cursorless/pull/900/files/3ba8a67ea786a43f69624ad93c3758090594c38b..cb2469c4899dcbd464401bf9cc7edc5778662625; feel free to merge if you're happy with them

src/processTargets/targets/PositionTarget.ts
```diff
@@ -97,14 +90,53 @@ export default class PositionTarget extends BaseTarget {
...
+        const startOffset = this.editor.document.offsetAt(
+          this.isBefore ? line.range.start : line.range.end
+        );
+        return this.isBefore
+          ? startOffset - this.insertionDelimiter.length - text.length
+          : startOffset +
+              this.insertionDelimiter.length +
+              this.indentationString.length;
+      }
+      const startOffset = this.editor.document.offsetAt(range.start);
```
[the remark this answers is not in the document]
pokey: Ok I figured out how to make things more consistent; see my changes

AndreasArvidsson:
I'm happy

pokey (approved):

PR #904, opened 2022-07-30
Added support for custom regex modifiers

AndreasArvidsson:
In `experimental/regex_modifiers.csv` users can now define their own custom regex modifiers
```csv
Spoken form, Regex

id, [\w/_.]+
```

I also updated our csv parser to properly support quoted lines
`id, "a,b"`

## Checklist


cursorless-talon/src/csv_overrides.py
```diff
@@ -255,37 +256,44 @@ def read_file(
...
+                i,
+                f"Malformed csv entry. Expected {len(headers)} columns.",
+                create_line(*row),
+            )
             continue
 
-        key = parts[0].strip()
-        value = parts[1].strip()
+        key = row[0]
+        value = row[1]
```
auscompgeek, line 296: nit:
```suggestion
        key, value = row
```

cursorless-talon/src/csv_overrides.py
```diff
@@ -255,37 +256,44 @@ def read_file(
...
+        # Remove trailing whitespaces for each cell
+        row = [x.rstrip() for x in row]
+        # Exclude empty or comment rows
+        if len(row) == 0 or (len(row) == 1 and row[0] == "") or row[0].startswith("#"):
             continue
 
         if not seen_headers:
             seen_headers = True
-            if line != expected_headers:
+            if create_line(*row) != expected_headers:
```
auscompgeek, line 279: Let's compare the row directly now instead of round-tripping them through `create_line`?
```suggestion
            if row != headers:
```

auscompgeek (approved):
Talon side LGTM

cursorless-talon/src/csv_overrides.py
AndreasArvidsson: Ah. I'm not a python developer and it shows :D

cursorless-talon/src/csv_overrides.py
AndreasArvidsson: If you compare to list do you actually compare the content and not that they are the exact same list?

auscompgeek: Yes.

AndreasArvidsson: Apparently it does :)

src/processTargets/modifiers/BoundedNonWhitespaceStage.ts
```diff
@@ -25,7 +18,7 @@ export type BoundedNonWhitespaceSequenceModifier = (
 export default class BoundedNonWhitespaceSequenceStage
   implements ModifierStage
 {
-  constructor(private modifier: BoundedNonWhitespaceSequenceModifier) {}
+  constructor(private modifier: ContainingScopeModifier | EveryScopeModifier) {}
```
pokey, line 21: How come you weakened this type?

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
```diff
@@ -103,39 +94,45 @@ class RegexStage implements ModifierStage {
...
-  | EveryScopeModifier
-) & {
-  scopeType: { type: "nonWhitespaceSequence" };
-};
-
-export class NonWhitespaceSequenceStage extends RegexStage {
-  constructor(modifier: NonWhitespaceSequenceModifier) {
-    super(modifier, /\S+/g, "Non whitespace sequence");
+export class NonWhitespaceSequenceStage extends RegexStageBase {
+  constructor(modifier: ContainingScopeModifier | EveryScopeModifier) {
```
pokey, line 104: As above re weaker type

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
```diff
@@ -70,7 +65,7 @@ class RegexStage implements ModifierStage {
     contentRange: Range
   ): ScopeTypeTarget {
     return new ScopeTypeTarget({
-      scopeTypeType: this.modifier.scopeType.type,
+      scopeTypeType: this.modifier.scopeType.type as SimpleScopeTypeType,
```
pokey, line 68: Hmm.  Not in love with the cast.  Why do we need `scopeTypeType` on `ScopeTypeTarget` anyway?

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
```diff
@@ -82,11 +77,7 @@ class RegexStage implements ModifierStage {
       range.contains(position)
     );
     if (match == null) {
-      if (this.name) {
-        throw new Error(`Couldn't find containing ${this.name}`);
-      } else {
-        throw new Error(`Cannot find sequence defined by regex: ${this.regex}`);
-      }
+      throw new NoContainingScopeError(this.modifier.scopeType.type);
```
pokey, line 80: What if we just passed in the whole `scopeType` to `NoContainingScopeError`, and then we have a function that converts a scope type to text when we display the error message.  So eg for simple scope types it just converts camel case to regular text (can just use [`lodash.lowerCase`](https://lodash.com/docs/4.17.15#lowerCase), and for regex scope types it prints out `regex ${regex}`

src/test/suite/fixtures/recorded/selectionTypes/clearCustomRegex.yml
```diff
@@ -0,0 +1,26 @@
+languageId: plaintext
```
pokey, line 1: Nice.  Would be good to also have an "every" test

pokey (commented):
Neat!  Left some comments below

src/processTargets/modifiers/BoundedNonWhitespaceStage.ts
AndreasArvidsson: Because there is really no gain. In every other modifier we are not this specific. Look at line for example.
https://github.com/pokey/cursorless-vscode/blob/6c54a17b9439917e890090f8eea7e3d9f2456ba3/src/processTargets/modifiers/scopeTypeStages/LineStage.ts#L12

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
AndreasArvidsson: See above

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
AndreasArvidsson: Well we use it for creating continuous ranges and default delimiters. Not really something we can just remove without consequence.

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
AndreasArvidsson: I would be fine with that. But I'm not sure if it fits in this pr

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
pokey: Ah right yeah default delimiters.

I still don't love the cast here. It's not actually a simple scope type type in the case of a custom regex right? I think a cast is ok when the compiler just isn't smart enough to do inference. In this case we're actually lying to the compiler right?

AndreasArvidsson: Yeah somewhat

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
pokey: I guess the main issue I'd like to avoid is the blanket error catching below in custom regex; the approach I proposed strikes me as the right approach, but if you can find something simpler, I guess we can consider the above out of scope 🤷.  Seems related enough to fit into this PR, but I don't feel strongly

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
pokey: This is the only real showstopper for me.  There are a few ways to handle it.  Any preference?

src/processTargets/modifiers/BoundedNonWhitespaceStage.ts
pokey: I would argue that we should be more strongly typed there as well.  I don't think you need to go and add it everywhere, but I don't think we should be removing it.  Time and time again we've found that the more you give Typescript the more it gives you.  I'm not in love with the changes in this PR that subvert and weaken our types

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
AndreasArvidsson: Not really

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
AndreasArvidsson: I don't see how this is worse than what was there before?

src/processTargets/modifiers/BoundedNonWhitespaceStage.ts
AndreasArvidsson: Since the regex class was the only one that did this strongly typed and it created a lot of additional boilerplate I prefer to not use it

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
pokey: Maybe just pass in the `scopeType` itself instead of  the `scopeTypeType`?

Alternately don't pass in the scopeTypeType at all and move the switch statement that uses it into the syntactic scope stage

src/processTargets/modifiers/BoundedNonWhitespaceStage.ts
pokey: Fine let's leave it. We can revisit later

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
pokey: The problem is the blanket error catching. So something else could go wrong and we report it as couldn't find scope type

AndreasArvidsson: If something else goes wrong we should not use that error. In this case we only throw this error if we can't find a match.  I don't quite follow the problem here?

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
AndreasArvidsson: I guess we could do that. 


We use it in multiple places so that would not be an easy fix.

pokey: Do you have a simpler solution? I don't think that lying to the type system is a good idea, so we need to figure something out

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
```diff
@@ -103,39 +94,45 @@ class RegexStage implements ModifierStage {
...
-  constructor(modifier: UrlModifier) {
-    super(modifier, URL_REGEX, "URL");
+export class CustomRegexStage extends RegexStageBase {
+  constructor(modifier: CustomRegexModifier) {
+    super(modifier, new RegExp(modifier.scopeType.regex, "g"));
+  }
+  run(context: ProcessedTargetsContext, target: Target): ScopeTypeTarget[] {
+    try {
+      return super.run(context, target);
+    } catch (error) {
```
pokey, line 133: This is the blanket error catching I'm referring to

src/processTargets/modifiers/scopeTypeStages/RegexStage.ts
pokey: I think the simplest solution is prob to make the scopeTypeType arg be nullable, ie `ScopeTypeType | null`. And then just pass in `null` here. Maybe not the most precise but at least it's not a lie 😅

AndreasArvidsson: If we're not going to rely on anything it has to do with the scope type maybe we should use a different target instead?

pokey: Fine by me

pokey (approved):

PR #907, opened 2022-08-03
Fix special chars without accents

pokey:
We weren't handling non-`[a-z]` letters that didn't have accents well.  They would pass through normalisation unmodified, making them impossible to refer to if you don't have them in `any_alphanumeric_key`.  This PR does the following:

- Makes regex for mapping to `[unk]` stricter, to only include `[a-zA-Z]` for letters
- Renames `accentsToPreserve` to `lettersToPreserve` for clarity
- If a letter is in `lettersToPreserve`, returns it directly after case normalisation, skipping regex check so that it can bypass the new stricter regex
- Removes `preserveAccents` because it was tough to see a use case for that

## Checklist


src/test/suite/tokenGraphemeSplitter.test.ts
```diff
@@ -14,280 +14,265 @@ import { FakeIDE } from "./fakes/ide/FakeIDE";
...
+        "\u00D1", // Ñ as single codepoint
+        [["n", 0, 1]],
+      ],
+      [
+        "\u004E\u0303", // Ñ using combining mark
+        [["n", 0, 2]],
+      ],
+      ["ꝏ", [[UNKNOWN, 0, 1]]],
+      ["ø", [["o", 0, 1]]],
+      ["æ", [[UNKNOWN, 0, 1]]],
```
pokey, line 62: I've been playing with these a bit locally while editing the test cases.  It feels pretty good, though part of me wants to say `"air"` for `æ` and `"special"` for `ø`, whereas this PR makes them `"special"` and `"odd"`, respectively.

AndreasArvidsson (approved):

src/test/suite/tokenGraphemeSplitter.test.ts
AndreasArvidsson: Yeah it is a bit in inconsistent how these are treated. These are proper characters in the danish/Norwegian alphabet
danish/norwegian | swedish
| - | - |
ø | ö
æ | ä

PR #909, opened 2022-08-04
Migrated makeshift actions from talon to extension

AndreasArvidsson:
Also merged implementation with some extension side actions that only called a vscode command

## Checklist


phillco:
Wow, that was fast :)

cursorless-talon/src/actions/actions_simple.py
```diff
@@ -46,6 +53,7 @@
 # Don't wait for these actions to finish, usually because they hang on some kind of user interaction
 no_wait_actions = [
     "generateSnippet",
+    "rename",
```
AndreasArvidsson, line 56: Since we are not waiting for the rename action any sleep extension side doesn't really matter. Maybe we should turn this into an dict with the value being the sleep time out that we run talon side for that no waiting actions?

src/actions/ActionsMakeshift.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export class ShowReferences extends MakeshiftAction {
+  command = "references-view.find";
+  ensureSingleTarget = true;
+}
+
+export class Rename extends MakeshiftAction {
+  command = "editor.action.rename";
+  ensureSingleTarget = true;
+  postCommandSleepMs = 200;
```
AndreasArvidsson, line 63: This sleep doesn't matter since talon is not waiting for this command

AndreasArvidsson:
> Wow, that was fast :)

Well it was a small update :)

cursorless-talon/src/actions/actions_simple.py
pokey: Wait so does this mean that in your PR, rename no longer has any sleep?

src/actions/ActionsMakeshift.ts
pokey: How come it is here then?

src/actions/ActionsMakeshift.ts
```diff
@@ -0,0 +1,86 @@
+import { Target } from "../typings/target.types";
```
pokey, line 1: Why not just call this `MakeshiftActions`?

src/actions/ActionsMakeshift.ts
```diff
@@ -0,0 +1,86 @@
...
+import sleep from "../util/sleep";
+import CommandAction from "./CommandAction";
+
+abstract class MakeshiftAction extends CommandAction {
+  abstract command: string;
```
pokey, line 6: Wow TIL.  Didn't realise you could declare non-function properties abstract 😄

src/actions/ActionsMakeshift.ts
```diff
@@ -0,0 +1,86 @@
...
+  async run(targets: [Target[]]) {
+    return super.run(targets, {
+      command: this.command,
+      commandArgs: this.commandArg ? [this.commandArg] : [],
+      ensureSingleTarget: this.ensureSingleTarget ?? false,
+      restoreSelection: this.restoreSelection ?? true,
+    });
+    if (this.postCommandSleepMs) {
+      await sleep(this.postCommandSleepMs);
+    }
```
pokey, line 21: This code never runs

pokey:
Overall looks good!  Left a few minor comments; biggest issues for me are around sleeps

src/actions/ActionsMakeshift.ts
AndreasArvidsson: Cool isn't it? :D

src/actions/ActionsMakeshift.ts
AndreasArvidsson: Because I wanted to have a discussion about it. This sleep is necessary for mac os afaik. So we have to put it somewhere.

cursorless-talon/src/actions/actions_simple.py
AndreasArvidsson: Well it sleeps extension side but since the client is not waiting for it so in practice no.

pokey: Maybe just have a separate dict mapping from action id to sleep time. Not sure we want to fold it into the no-wait dict

AndreasArvidsson: Done

cursorless-talon/src/actions/actions.py
```diff
@@ -40,22 +40,10 @@ def cursorless_command(action_id: str, target: dict):
...
-
-            if talon_options.post_command_sleep_ms:
-                actions.sleep(f"{talon_options.post_command_sleep_ms}ms")
-
-            return return_value
         elif action_id in no_wait_actions:
-            return actions.user.cursorless_single_target_command_no_wait(
-                action_id, target
-            )
+            actions.user.cursorless_single_target_command_no_wait(action_id, target)
```
pokey, line 44: How come you're dropping the return value?

cursorless-talon/src/actions/actions.py
```diff
@@ -40,22 +40,10 @@ def cursorless_command(action_id: str, target: dict):
...
+            if action_id in no_wait_actions_post_sleep:
+                actions.sleep(no_wait_actions_post_sleep[action_id])
```
pokey, line 46: How come we only sleep for actions that we don't wait for?  Seems somewhat orthogonal

pokey (commented):
Ok getting there; left a couple more minor comments

cursorless-talon/src/actions/actions.py
AndreasArvidsson: Action we wait for we sleep extension side.

cursorless-talon/src/actions/actions.py
AndreasArvidsson: Since we don't wait it cannot return a value. In fact the action it's calling does not have a return.

pokey (approved):
lgtm!

PR #913, opened 2022-08-15
Fix status bar icon

pokey:
The assets for the status bar icon were being excluded by our .vscodeignore, so this is what we currently see in status bar in production:

image

## Checklist

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -4,9 +4,11 @@
 !NOTICE.md
 !README.md
 !build-info.json
-!cursorless-snippets/**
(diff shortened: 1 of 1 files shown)


pokey:
@brollin fyi for future reference

AndreasArvidsson (approved):

brollin:
Whoops, thanks for the fix. TIL about the `.vscodeignore` file :)

auscompgeek:
Ah, I was wondering what that was meant to look like :sweat_smile:

PR #915, opened 2022-08-16
Tweak docs search ranking

### LINKED ISSUE #644: Tweak search rankings in documentation
Currently, if you search for "copy" in the [cursorless docs](https://www.cursorless.org/docs/), the top results are all from the contributor side, rather than the user documentation for the "copy" action.  In addition, the top-level sections in the search box are strange, with "For users" being on the same level as subsections of "For contributors", such as "Classes" and "Interfaces":

image

We'd like to do the following:

- [ ] Prioritise the user docs, especially if the user is currently in the user docs
- [ ] Change the hierarchy so that "For users" and "For contributors" are the top levels of the hierarchy

This issue does beg the question of whether we want to have a harder separation between user docs and contributor docs

pokey:
Tweaks docs search so that we

- prefer to search within the same lvl0 (eg "For users" vs "For contributors"), and
- prefer docs that are not api docs, as these are a bit noisy.

Note that this PR ["swizzles"](https://docusaurus.io/docs/swizzling) a Docusaurus core class, because there doesn't seem to be any other way to create context-aware search rankings.  For the actual change we made to the docusaurus core class, see [`5a2a365`](https://github.com/cursorless-dev/cursorless/pull/915/commits/5a2a365ad35de713adec4661b3bc1c291c3db4a9).  Note that we would like to un-swizzle the component so that we can un-pin Docusaurus; see #931 

This PR also captures the crawler / index settings that live in the Algolia console today.  See #917 for how we plan to make the files in source control the source of truth in the future.  You can see the actual changes that were made in the search console to improve search ranking in [`b37cd16`](https://github.com/cursorless-dev/cursorless/pull/915/commits/b37cd1619509d175087058a7f63df4483d2163bc)

To try out the new search rankings, pop up the search widget in the deploy preview at https://deploy-preview-915--cursorless.netlify.app/docs/

Fixes #644

## Before

164753938-cde8c7e0-2db0-4e30-b68b-3a531345482d

## After

image

## Checklist


website/src/theme/SearchBar/index.js
```diff
@@ -0,0 +1,208 @@
+import React, {useState, useRef, useCallback, useMemo} from 'react';
```
pokey, line 1: Note that this file was ["swizzled"](https://docusaurus.io/docs/swizzling) from a Docusaurus core class, because there doesn't seem to be any other way to create context-aware search rankings.  For the actual change we made to the docusaurus core class, see https://github.com/cursorless-dev/cursorless/pull/915/commits/5a2a365ad35de713adec4661b3bc1c291c3db4a9

website/config/algolia/crawler-settings.js
```diff
@@ -13,23 +13,29 @@ new Crawler({
       recordExtractor: ({ $, helpers }) => {
         // priority order: deepest active sub list header -> navbar active item -> 'Documentation'
         const lvl0 =
-          $(
-            ".menu__link.menu__link--sublist.menu__link--active, .navbar__item.navbar__link--active"
-          )
-            .last()
-            .text() || "Documentation";
+          $(".navbar__item.navbar__link--active").last().text() ||
```
pokey, line 16: The menu selector that came with the default setup was doing really weird things, and we prefer to just use "For users" vs "For contributors" as the lvl0 of the hierarchy so that we can use it for search faceting, eg to downrank contributor docs when viewing user docs and vice-versa

website/config/algolia/crawler-settings.js
```diff
@@ -13,23 +13,29 @@ new Crawler({
...
+          "Documentation";
+        const isApi = $(".menu__link.menu__link--active")
+          .toArray()
+          .some((element) => $(element).text() === "API")
+          ? "yes"
+          : "no";
```
pokey, line 22: This allows us to down-rank api docs

website/config/algolia/crawler-settings.js
```diff
@@ -13,23 +13,29 @@ new Crawler({
...
         return helpers.docsearch({
           recordProps: {
             lvl0: {
               selectors: "",
               defaultValue: lvl0,
             },
-            lvl1: "header h1",
+            lvl1: "article h1",
```
pokey, line 29: For some reason the default selector here was buggy, and only worked if title was defined in frontmatter, not when it was just an h1 tag, as is the case for most of our docs

website/config/algolia/crawler-settings.js
```diff
@@ -0,0 +1,129 @@
+new Crawler({
```
pokey, line 1: Note that this file was dumped from the search console after a bit of tweaking to improve rankings.  To see the actual changes made, see https://github.com/cursorless-dev/cursorless/pull/915/commits/b37cd1619509d175087058a7f63df4483d2163bc

AndreasArvidsson (approved):

PR #923, opened 2022-08-21
Use dataclass in wrap action

auscompgeek:
This just changes the wrap action to internally use a dataclass. This was one of the modules that doesn't require any changes to other modules.

Part of #725

## Checklist


pokey (approved):
Nice!  Great to start getting stronger types

PR #930, opened 2022-09-05
Tweak recording calibration display

pokey:
For recording Cursorless demo videos.  For lower framerates, 30ms is too short, and sometimes the display just doesn't appear in any frames

Also attempts to ensure that timestamp is last frame of calibration display

## Checklist


AndreasArvidsson (dismissed):

AndreasArvidsson:
@pokey Looks like we have a merge conflict here. Since I don't use this code it's probably better if you do the merging so you can try it out.

pokey:
Ok @AndreasArvidsson I resolved merge conflict and confirmed that the screen recorder works as expected, and that I haven't broken normal test case recording

AndreasArvidsson (approved):

PR #942, opened 2022-09-08
Fix test fixture upgrade scripts

SimeonC:
Fixed the v1 to v2 script to preserve order and not drop any values.

- [ ] ~~Use normalisation code for updating fixtures in https://github.com/cursorless-dev/cursorless/blob/a4cee9611f4c4c6bbb5add250ce8c113f5efc918/src/test/suite/recorded.test.ts#L130-L138~~ (filed https://github.com/cursorless-dev/cursorless/issues/1168 to follow up on this one)


.vscode/settings.json
```diff
@@ -22,5 +22,6 @@
 		"pojo",
 		"subword"
 	],
-	"files.eol": "\n"
+	"files.eol": "\n",
+	"typescript.tsdk": "node_modules/typescript/lib"
```
pokey, line 26: This is nice; let's upgrade Typescript in this PR as well though I think

src/definitions/utils.d.ts
```diff
@@ -0,0 +1,19 @@
...
+  [K in keyof T as Pick<T, K> extends Required<Pick<T, K>> ? never : K]: T[K];
+};
+
+type GetRequired<T> = {
+  [K in keyof T as Pick<T, K> extends Required<Pick<T, K>> ? K : never]: T[K];
+};
+
+type UnionUndefined<T> = {
+  [K in keyof T]: T[K] | undefined;
+};
```
pokey, line 11: Did we want to expose these globally?

src/scripts/transformRecordedTests/transformations/reorderFields.ts
```diff
@@ -1,13 +1,19 @@
...
     command: fixture.command,
     marksToCheck: fixture.marksToCheck,
     initialState: fixture.initialState,
     finalState: fixture.finalState,
+    decorations: fixture.decorations,
     returnValue: fixture.returnValue,
     fullTargets: fixture.fullTargets,
+    thrownError: fixture.thrownError,
+    postCommandSleepTimeMs: fixture.postCommandSleepTimeMs,
+    postEditorOpenSleepTimeMs: fixture.postEditorOpenSleepTimeMs,
```
pokey, line 17:
```suggestion
    decorations: fixture.decorations,
    returnValue: fixture.returnValue,
    thrownError: fixture.thrownError,
    postEditorOpenSleepTimeMs: fixture.postEditorOpenSleepTimeMs,
    postCommandSleepTimeMs: fixture.postCommandSleepTimeMs,
    fullTargets: fixture.fullTargets,
```

pokey (commented):
Looks great thank you!  Left a few minor comments

src/definitions/utils.d.ts
SimeonC: Personally, I've never really bothered thinking about that. If that's an issue we could just change it to a manual import, or I could use a namespace for the utils - whichever you prefer.

With the namespace way `Utils` would be globally available but it does clutter it less, we could put all the types in the `Utils` namespace so the only "globally" available type is the `Utils` variable.

```ts
declare namespace Utils {
  type GetOptional ...
}
type EnforceUndefined = Utils.GetOptional...

// some other file
EnforcedUndefined<... // This is OK
GetOptional<... // would error
Utils.GetOptional<... // This is also OK
```

.vscode/settings.json
SimeonC: I upgraded TS as well, but I left the website at 4.5 as the typedocs only supports up to 4.5 and I didn't want to risk breaking the website!

pokey: Hmm I think that will be a problem, because we run typedocs on all the code in cursorless to generate api docs. We should prob pin both to the same version

When we switch to nx we'll only have one set of versions for everything fwiw

SimeonC: I can try upgrade it but I'm not familiar with docasaurus and it's typedocs so I'd need someone to double check that. It may end up being a much bigger upgrade - if it does need more than a simple dependency upgrade do we want to just do that here or make a different PR for upgrading the website deps.

src/testUtil/cleanUpTestCaseCommand.ts
```diff
@@ -1,16 +1,12 @@
...
-    ...rest,
-    action: {
-      ...action,
-      args: args == null ? undefined : args.length === 0 ? undefined : args,
-    },
-  };
+  const { args } = command.action;
+  const result = merge({}, command);
+  result.action.args =
+    args == null ? undefined : args.length === 0 ? undefined : args;
```
auscompgeek, line 10: This might be clearer?
```suggestion
  if (args == null || args.length === 0) {
    result.action.args = undefined;
  }
```
I'm assuming lodash merge means that `result.action !== command.action` so we're not accidentally mutating it

auscompgeek (commented):
Looking pretty good to me too!

src/testUtil/cleanUpTestCaseCommand.ts
SimeonC: Yea, lodash merge does a deep merge of the objects.

Thanks for pointing out the nested ternary. I'd prefer `result.action.args = !args?.length ? undefined : args;` just because it's a bit shorter, but is that clear enough?

src/definitions/utils.d.ts
SimeonC: @pokey What approach do you want me to take here? `import { EnforceUndefined } from '...'` or the Utils namespace or just leave it as is?

src/testUtil/cleanUpTestCaseCommand.ts
SimeonC: On second thought, yours is clearer, let's go with that.

.vscode/settings.json
SimeonC: I upgraded TS in website, it seems working to me as far as I can tell. The types are getting output in the contributor docs but I'm not sure they're 100% correct.

auscompgeek: We should probably also make VSCode actually ask whether we want to use it too, otherwise nobody will actually switch to it.
```suggestion
	"typescript.enablePromptUseWorkspaceTsdk": true,
	"typescript.tsdk": "node_modules/typescript/lib"
```

src/definitions/utils.d.ts
pokey: Tbh I'd be tempted to just require users to import `EnforceUndefined`.  With auto-import, that's not particularly onerous, and it makes the code a bit better self-documenting, as you can see in the imports where it's coming from.  I'm not a huge fan of globally available symbols

.vscode/settings.json
pokey: @auscompgeek that sounds reasonable to me.  Any objections @SimeonC ?

Re types in website; I just kicked off a deploy preview.  I'll take a look at docs to see if they look sane

pokey: Huh.  Unfortunately this PR breaks the api docs.  For example, in the docs for `runCommand`, see how on [production](https://www.cursorless.org/docs/contributing/api/classes/core_commandrunner_commandrunner.commandrunner/#runcommand) there are a bunch of links in the description, but these seem to have disappeared for some reason in the [deploy preview](https://deploy-preview-942--cursorless.netlify.app/docs/contributing/api/classes/core_commandRunner_CommandRunner.CommandRunner/#runcommand) 😕

pokey:
I think we should split this PR into two:

- one to fix the fixture upgrade scripts, and
- one to upgrade typescript and refactor type hints in `makeGraph.ts`

I believe these two pieces are completely orthogonal, and the former is more imperative, while the latter is causing the problems.  Make sense?

SimeonC:
Sure we could do that. I haven't had the time to come back to look at whether the fix for the docs is just some missing new config or it needs a refactor.
Sorry I'm taking so long to get round to finishing this -  don't mind if someone else who has more time than me picks this up and splits it into 2 PRs.

pokey:
ok @SimeonC I factored out the typescript stuff into https://github.com/cursorless-dev/cursorless/pull/1166 and https://github.com/cursorless-dev/cursorless/pull/1167, and made some tweaks.  lmk if you have any objections

AndreasArvidsson (approved):

SimeonC:
Thanks for updating and completing it, looks good to me!

pokey (approved):

PR #948, opened 2022-09-09
Fix focus editor in changes working tree

### LINKED ISSUE #819: Can't set selection in the source/changes working tree

AndreasArvidsson:
Fixes #819

## Checklist

### DIFF
--- a/src/util/setSelectionsAndFocusEditor.ts
+++ b/src/util/setSelectionsAndFocusEditor.ts
@@ -1,8 +1,10 @@
+import * as semver from "semver";
 import {
   commands,
   NotebookDocument,
   Selection,
   TextEditor,
+  version,
   ViewColumn,
   window,
 } from "vscode";
@@ -72,6 +74,9 @@ function getViewColumn(editor: TextEditor): ViewColumn | undefined {
   }
   // TODO: tabGroups is not available on older versions of vscode we still support.
   // Remove any cast as soon as version is updated.
+  if (semver.lt(version, "1.67.0")) {
+    return undefined;
+  }
   const viewColumn: ViewColumn | undefined = (window as any)?.tabGroups
     ?.activeTabGroup?.viewColumn;
   if (viewColumn != null) {
--- a/src/util/setSelectionsAndFocusEditor.ts
+++ b/src/util/setSelectionsAndFocusEditor.ts
@@ -77,8 +77,15 @@ function getViewColumn(editor: TextEditor): ViewColumn | undefined {
   if (semver.lt(version, "1.67.0")) {
     return undefined;
   }
-  const viewColumn: ViewColumn | undefined = (window as any)?.tabGroups
-    ?.activeTabGroup?.viewColumn;
+  const uri = editor.document.uri.toString();
+  const tabGroups: any[] = (window as any)?.tabGroups?.all ?? [];
+  const tabGroup = tabGroups.find(
+    (tabGroup: any) =>
+      !!tabGroup?.tabs.find(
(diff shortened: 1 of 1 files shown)


src/util/setSelectionsAndFocusEditor.ts
```diff
@@ -65,6 +68,26 @@ export async function focusEditor(editor: TextEditor) {
...
+  if (editor.viewColumn != null) {
+    return editor.viewColumn;
+  }
+  // TODO: tabGroups is not available on older versions of vscode we still support.
+  // Remove any cast as soon as version is updated.
+  if (semver.lt(version, "1.67.0")) {
+    return undefined;
+  }
+  const viewColumn: ViewColumn | undefined = (window as any)?.tabGroups
+    ?.activeTabGroup?.viewColumn;
```
pokey, line 81: Why are you using the view column of the active tab group? Shouldn't you look for the tab group containing `editor`?

AndreasArvidsson: As far as I can tell that information is not available on each tab. I have to dig deeper in the api and see if I can find that data.

AndreasArvidsson: There is no editor or document on the tab groups, but I managed to find a uri

src/util/setSelectionsAndFocusEditor.ts
```diff
@@ -78,21 +78,10 @@ function getViewColumn(editor: TextEditor): ViewColumn | undefined {
     return undefined;
   }
   const uri = editor.document.uri.toString();
-  const tabGroups: any[] = (window as any)?.tabGroups?.all ?? [];
-  const tabGroup = tabGroups.find(
-    (tabGroup: any) =>
-      !!tabGroup?.tabs.find(
-        (tab: any) => tab?.input?.modified?.toString() === uri
-      )
+  const tabGroup = (window as any)?.tabGroups?.all?.find((tabGroup: any) =>
```
pokey, line 81: Do we need all these `?.`s?

src/util/setSelectionsAndFocusEditor.ts
```diff
@@ -78,21 +78,10 @@ function getViewColumn(editor: TextEditor): ViewColumn | undefined {
...
+    tabGroup?.tabs.find((tab: any) => tab?.input?.modified?.toString() === uri)
```
pokey, line 82: I think we prob want to support `original` and `uri` in addition to `modified`

pokey (commented):
Looks good with a couple comments

src/util/setSelectionsAndFocusEditor.ts
AndreasArvidsson: Probably not but since it's untyped I think it's better to be safe than sorry.

src/util/setSelectionsAndFocusEditor.ts
AndreasArvidsson: I decided to leave them out. Since original and modified is in the same tab focusing that tab would always focus the modified editor for some reason. I think it's a better behavior to throw an error than to focus the wrong editor. Including uri only gave us false positives like notebooks and I haven't found a use case where it's needed.

pokey (approved):

PR #952, opened 2022-09-12
Fixed bug where ternary condition skipped parentheses

### LINKED ISSUE #928: Bug with ternary condition selection.
Considering this code;

```ts
const systemDir = (process.env._ || '').match(
  // this check is for any scripts running from the tablecheck-frontend package.json which are absolutely referenced
  // all other cases should be running via NPM and in node_modules
  /^\.\/packages\/(quality|audit)\/bin\/main\.ts/gi
)
  ? path1
  : path2;
// <-- cursor here
```

If you try to bring the "condition" of the ternary eg `bring condition red made` the whole condition should be selected.

Expected;
```ts
const systemDir = (process.env._ || '').match(
  /^\.\/packages\/(quality|audit)\/bin\/main\.ts/gi
)
  ? path1
  : path2;
(process.env._ || '').match(
  /^\.\/packages\/(quality|audit)\/bin\/main\.ts/gi
);
```

Actual (note the missing `(` and `)` at the beginning and the end of the newly inserted section):
```ts
const systemDir = (process.env._ || '').match(
  /^\.\/packages\/(quality|audit)\/bin\/main\.ts/gi
)
  ? path1
  : path2;
process.env._ || '').match(
  /^\.\/packages\/(quality|audit)\/bin\/main\.ts/gi
```

AndreasArvidsson:
Fixes #928 

## Checklist

### DIFF
--- a/src/languages/typescript.ts
+++ b/src/languages/typescript.ts
@@ -200,7 +200,10 @@ const nodeMatchers: Partial<
     )
   ),
   statement: STATEMENT_TYPES.map((type) => `export_statement?.${type}`),
-  condition: conditionMatcher("*[condition]"),
+  condition: cascadingMatcher(
+    patternMatcher("ternary_expression[condition]"),
+    conditionMatcher("*[condition]")
+  ),
   class: [
     "export_statement?.class_declaration", // export class | class
     "export_statement?.abstract_class_declaration", // export abstract class | abstract class
--- a/src/test/suite/fixtures/recorded/languages/typescript/clearCondition.yml
+++ b/src/test/suite/fixtures/recorded/languages/typescript/clearCondition.yml
@@ -1,0 +1,26 @@
+languageId: typescript
+command:
+  spokenForm: clear condition
+  version: 2
+  targets:
+    - type: primitive
+      modifiers:
+        - type: containingScope
+          scopeType: {type: condition}
+  usePrePhraseSnapshot: true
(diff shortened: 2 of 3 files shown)


src/languages/typescript.ts
```diff
@@ -200,7 +200,10 @@ const nodeMatchers: Partial<
     )
   ),
   statement: STATEMENT_TYPES.map((type) => `export_statement?.${type}`),
-  condition: conditionMatcher("*[condition]"),
+  condition: cascadingMatcher(
+    patternMatcher("ternary_expression[condition]"),
+    conditionMatcher("*[condition]")
```
pokey, line 205: I'm a bit nervous about relying on order here.  I'm also nervous about using `*` for something that blindly removes parens.  Maybe better to use the exact scope type(s) here?

pokey (commented):
Looks good with one comment

src/languages/typescript.ts
AndreasArvidsson: Isn't order the point of the cascading matcher?

We can do that

pokey: Yeah I guess that's a fair point.  But the I think the blindly removing parens thing holds, esp given the error we saw that led to this PR

AndreasArvidsson: Have now added specific types

pokey (approved):

PR #954, opened 2022-09-12
Added member scope type

AndreasArvidsson:
## Checklist

### DIFF
--- a/cursorless-talon/src/modifiers/containing_scope.py
+++ b/cursorless-talon/src/modifiers/containing_scope.py
@@ -60,6 +60,7 @@ scope_types = {
     "short paint": "boundedNonWhitespaceSequence",
     "link": "url",
     "token": "token",
+    "member": "member",
     # LaTeX
     "part": "part",
     "chapter": "chapter",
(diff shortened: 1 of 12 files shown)


pokey:
Interesting. A few questions:

### paren matching
What happens with something like the following?

```
foo.bar() + baz()
```

If cursor is on `foo`, won't it take the whole thing, because it eats until it hits a closing `)`? I think we want just `foo.bar()` in that case

### multi line / whitespace
Does this support line breaks inside a function call, eg

```
foo(
  bar + 5
)
```

With cursor on `foo`. Looks like it does, but test would be good

Does this support line breaks / spaces between the accesses? Eg for a long chain I'll do

```
foo
.bar
.baz
```

### other
What happens in C++ for `::` and `->`?

What about square brackets?

What about js template strings? Eg

```javascript
foo.bar`baz bongo`
```

Does this work with Unicode identifiers?

Does this work with snake case identifiers?

AndreasArvidsson:
This needs to be implemented using the syntax tree

PR #955, opened 2022-09-12
make `shellscript` use the new CSS tokenizer

phillco:
Dashed identifiers are valid tokens inside of shell scripts, so let's use the benefits of the new language-specific tokenizer added in https://github.com/cursorless-dev/cursorless/pull/718.

Also refactor a bit since this is no longer a CSS-specific tokenized.

## Checklist


pokey (commented):
Looks good! Please add a couple tests. Can have a look at #718 for inspiration

phillco:
Done.

phillco:
I could make them use change instead of chuck if that would be better

pokey:
I'd make one of them "change"

phillco:
Done.

pokey (commented):
Ok almost there!  Would be good to add

- [ ] tests for the tokeniser itself, like we have [for css](https://github.com/cursorless-dev/cursorless/blob/4b38d4cefaba4a39ee44460d020c3b950d310427/src/test/suite/tokenizer.test.ts#L80-L89).  Can just use that same command line you're using here
- [ ] A "chuck first word" test with something like `--foo-bar`.  Curious to see what will happen there

phillco:
The latter works great! `--foo-bar` becomes `--bar`.

phillco:
Hm, looks like I actually need to update `textFragmentExtractors` as well because I'm adding a new language.

phillco:
This may actually be harder to merge incrementally than I thought. In order for us to add the custom tokenizer, we have to add `shellscript` to `SupportedLanguageId`. But to do that it has to have a node matcher and text fragment extractor -- we effectively have to build language support for it. Unless there's an easier way.

phillco:
Unless we want to make

```
const languageTokenizerOverrides: Partial<
  Record<SupportedLanguageId, LanguageTokenizerOverrides>
```

Not use `SupportedLanguageId` as a key, but instead a wider list of language IDs (`AllLanguageId`). Think it's reasonable to want to customize tokenization before having full language support (as is the case with shell scripts currently), so that makes a little bit of sense to me.

src/languages/constants.ts
```diff
@@ -24,4 +25,36 @@ export const supportedLanguageIds = [
   "xml",
 ] as const;
 
+/**
+ * All language IDs that we might reference but don't have full support.
+ * (e.g., for token customization).
+ */
+export const allLanguageIds = [
```
phillco, line 32: I tried to make this the result of adding just ["shellscript"] to `supportedLanguageIds` but the compiler didn't like it because it's constant.

phillco:
I've pushed what this would look like using the `allLanguageIds` approach. Let me know your thoughts.

src/languages/constants.ts
phillco: e.g. 

```typescript
/**
 * Other language IDs that we might reference (e.g., for token customization)
 * but don't have full tree-sitter support for yet.
 */
export const otherLanguageIds = [
  "shellscript",
] as const;

export const allLanguageIds = supportedLanguageIds.concat(otherLanguageIds);
```

auscompgeek: What about a spread?
```
export const allLanguageIds = [...supportedLanguageIds, ...otherLanguageIds];
```

phillco: ah thanks, that's much cleaner

phillco:
Basically, if the `allLanguageIds` approach is acceptable to you @pokey, I'll finish it off by writing the tests for the tokenizer; if not I'll abandon the PR and bring it up as part of Bash language support. But I'd prefer to land this incrementally.

src/languages/constants.ts
```diff
@@ -24,4 +25,13 @@ export const supportedLanguageIds = [
   "xml",
 ] as const;
 
+/**
+ * Other language IDs that we might reference (e.g., for token customization)
+ * but don't have full tree-sitter support for yet.
+ */
+export const otherLanguageIds = ["shellscript"] as const;
```
auscompgeek, line 32: I wonder, is this array actually used? Can it just live in the land of types instead?

src/languages/constants.ts
```diff
@@ -24,4 +25,13 @@ export const supportedLanguageIds = [
...
+
+export const allLanguageIds = [...supportedLanguageIds, ...otherLanguageIds];
+
 export type SupportedLanguageId = typeof supportedLanguageIds[number];
+export type AllLanguageId = typeof allLanguageIds[number];
```
pokey, line 37: I think I would just call this `LanguageId`

pokey (commented):
Yeah this approach seems fine to me 👍

phillco:
OK great, I'll finish it up. Is there an easy way to just run the tokenizer tests?

src/languages/constants.ts
phillco: Hm, possibly not. I was just mirroring the existing code. `supportedLanguageIds` is used inside of `isLanguageSupported()`. But that function itself is unused...

pokey:
> OK great, I'll finish it up. Is there an easy way to just run the tokenizer tests?

https://www.cursorless.org/docs/contributing/#running-a-subset-of-tests

phillco:
Thanks, easy enough! Done!

pokey (approved):

Issue #466, opened 2022-01-12
Smarter "that" mark


pokey:
Today, the "that" mark doesn't keep contextual information about its target.  For example, if you refer to "arg", and then say "chuck that", it forgets that "that" was an argument, and thus doesn't clean up the trailing comma.  We'd like to preserve the full target, including context, when we store the "that" mark

This functionality will be even more powerful when combined with #69, because it will enable chains to be broken up by stopping mid-way through a long chain, and using "that" as the mark to start the next section of the chain

See also #191; this functionality could be baked into the navigation map

Question: do we want to do the same for "this"?

PR #957, opened 2022-09-13
Initial work for smarter `that` marks

pokey:
This PR is the first step towards adding smarter `that` marks (#466).  It adds the infrastructure so that actions can return smarter `that` marks, and allows us to test this behaviour, but for readability, it doesn't actually modify any targets to return smarter `that` marks.  Thus, this PR doesn't actually change Cursorless functionally at all.

See also #965, which builds on this PR to add rich targets to a few easy actions

This PR builds on #967

## Checklist

- [x] Stop removing fields equal to their defaults
- [x] Add new PR which removes `that` mark from all tests except for tests in `action/` directory and makes that mark optional in recorded tests
- [x] Add new PR which adds support for excluding / including that marks in tests (defaults to excluding), adds ability to include recording test config in a directory, adding config to include that marks in action directory.  Make it recursive with parents

### DIFF
--- a/src/actions/actions.types.ts
+++ b/src/actions/actions.types.ts
@@ -56,10 +56,41 @@ export type ActionType =
   | "wrapWithPairedDelimiter"
   | "wrapWithSnippet";
 
+/**
+ * To be returned by {@link Action.run}
+ */
 export interface ActionReturnValue {
+  /**
+   * The value that should be returned to the caller of the command
+   */
   returnValue?: any;
-  thatMark?: SelectionWithEditor[];
-  sourceMark?: SelectionWithEditor[];
+
+  /**
+   * A list of selections that will become the `that` mark for the next command.
+   * The given selections will be wrapped in {@link UntypedTarget}s. This
+   * attribute is provided for convenience. Mutually exclusive with
+   * {@link thatTargets}
+   */
+  thatSelections?: SelectionWithEditor[];
+
+  /**
+   * A list of targets that will become the `that` mark for the next command.
+   * Mutually exclusive with {@link thatSelections}
+   */
+  thatTargets?: Target[];
+
+  /**
+   * A list of selections that will become the `source` mark for the next command.
+   * The given selections will be wrapped in {@link UntypedTarget}s. This
+   * attribute is provided for convenience. Mutually exclusive with {@link sourceTargets}
+   */
+  sourceSelections?: SelectionWithEditor[];
+
+  /**
+   * A list of targets that will become the `source` mark for the next command.
+   * Mutually exclusive with {@link sourceSelections}
+   */
+  sourceTargets?: Target[];
 }
 
 export interface Action {
--- a/src/core/ThatMark.ts
+++ b/src/core/ThatMark.ts
@@ -1,10 +1,10 @@
-import { SelectionWithEditor } from "../typings/Types";
+import { Target } from "../typings/target.types";
 
 export class ThatMark {
-  private mark?: SelectionWithEditor[];
+  private mark?: Target[];
 
-  set(value?: SelectionWithEditor[]) {
-    this.mark = value;
+  set(targets: Target[] | undefined) {
+    this.mark = targets;
   }
 
   get() {
--- a/src/core/commandRunner/CommandRunner.ts
+++ b/src/core/commandRunner/CommandRunner.ts
@@ -3,13 +3,19 @@ import { ActionType } from "../../actions/actions.types";
 import { OutdatedExtensionError } from "../../errors";
 import processTargets from "../../processTargets";
 import isTesting from "../../testUtil/isTesting";
-import { Graph, ProcessedTargetsContext } from "../../typings/Types";
+import { Target } from "../../typings/target.types";
+import {
+  Graph,
+  ProcessedTargetsContext,
+  SelectionWithEditor,
+} from "../../typings/Types";
 import { isString } from "../../util/type";
 import { canonicalizeAndValidateCommand } from "../commandVersionUpgrades/canonicalizeAndValidateCommand";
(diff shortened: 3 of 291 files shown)


src/actions/actions.types.ts
```diff
@@ -56,10 +56,41 @@ export type ActionType =
   | "wrapWithPairedDelimiter"
```
pokey, line 56: This file is important

src/core/ThatMark.ts
```diff
@@ -1,10 +1,10 @@
-import { SelectionWithEditor } from "../typings/Types";
+import { Target } from "../typings/target.types";
```
pokey, line 1: Note that now the `ThatMark` class doesn't know anything about selections; felt specific to the `CommandRunner`

src/core/commandRunner/CommandRunner.ts
```diff
@@ -210,3 +220,20 @@ export default class CommandRunner {
     this.disposables.forEach(({ dispose }) => dispose());
   }
 }
+
+function constructThatTarget(
```
pokey, line 224: This is now where the logic lives for wrapping selections in a `UntypedTarget`.  Seemed better than spreading it out

src/processTargets/marks/CursorStage.ts
```diff
@@ -1,9 +1,9 @@
-import { Target } from "../../typings/target.types";
+import type { Target } from "../../typings/target.types";
```
pokey, line 1: I went through and added `type` in a bunch of places

src/processTargets/marks/CursorStage.ts
```diff
@@ -1,9 +1,9 @@
...
 import type { CursorMark } from "../../typings/targetDescriptor.types";
-import { ProcessedTargetsContext } from "../../typings/Types";
+import type { ProcessedTargetsContext } from "../../typings/Types";
 import { isReversed } from "../../util/selectionUtils";
-import { MarkStage } from "../PipelineStages.types";
-import UntypedTarget from "../targets/UntypedTarget";
+import type { MarkStage } from "../PipelineStages.types";
+import { UntypedTarget } from "../targets";
```
pokey, line 6: I kept getting import cycles, so I just went through and fixed all the `targets` imports

src/processTargets/marks/ThatStage.ts
```diff
@@ -16,7 +11,7 @@ export class ThatStage implements MarkStage {
       throw Error("No available that marks");
     }
 
-    return selectionsToTarget(context.thatMark);
+    return context.thatMark;
```
pokey, line 14: `thatMark` is now a target, so we just return it.  Note that in this PR, it's always just an `UntypedTarget`, but now actions have the option to return something richer

src/scripts/transformRecordedTests/index.ts
```diff
@@ -3,11 +3,12 @@ import { identity } from "./transformations/identity";
 import { upgrade } from "./transformations/upgrade";
 import { transformFile } from "./transformFile";
 import { FixtureTransformation } from "./types";
+import { upgradeThatMarks } from "./upgradeThatMarks";
 
 const AVAILABLE_TRANSFORMATIONS: Record<string, FixtureTransformation> = {
   upgrade,
   autoFormat: identity,
-  // custom: MY_CUSTOM_TRANSFORMER,
+  custom: upgradeThatMarks,
```
pokey, line 11: we'll revert this file before merging; it just imports the transformation I used on test cases

src/scripts/transformRecordedTests/upgradeThatMarks.ts
```diff
@@ -0,0 +1,48 @@
+import { TestCaseFixture } from "../../testUtil/TestCase";
```
pokey, line 1: This file will be reverted.  It transforms all `that` marks to be the `UntypedTarget` that they would have become via `ThatMarkStage` before.

src/test/suite/fixtures/recorded/actions/bringArgueOxAndZipToAfterJustLeper.yml
```diff
@@ -50,11 +50,14 @@ finalState:
     - anchor: {line: 3, character: 0}
       active: {line: 3, character: 0}
   thatMark:
-    - anchor: {line: 4, character: 15}
-      active: {line: 4, character: 39}
+    - contentRange:
```
pokey, line 53: Tons of test case changes; worth a quick look at a few to get a flavour

src/test/suite/recorded.test.ts
```diff
@@ -78,18 +79,17 @@ async function runTest(file: string) {
...
   if (fixture.initialState.thatMark) {
-    const initialThatMark = fixture.initialState.thatMark.map((mark) => ({
-      selection: createSelection(mark),
-      editor,
-    }));
-    cursorlessApi.thatMark.set(initialThatMark);
+    const initialThatTargets = fixture.initialState.thatMark.map((mark) =>
+      plainObjectToTarget(editor, mark)
+    );
+    cursorlessApi.thatMark.set(initialThatTargets);
```
pokey, line 85: Note how we now rehydrate the `that` mark to create a proper target

src/typings/Types.ts
```diff
@@ -41,8 +42,8 @@ export interface ProcessedTargetsContext {
   currentSelections: SelectionWithEditor[];
   currentEditor: vscode.TextEditor | undefined;
   hatTokenMap: ReadOnlyHatMap;
-  thatMark: SelectionWithEditor[];
-  sourceMark: SelectionWithEditor[];
+  thatMark: Target[];
+  sourceMark: Target[];
```
pokey, line 46: Note that these are rich targets now

phillco (approved):
lgtm

PR #959, opened 2022-09-14
Updated regex stage with support for text range and multiple line pat…

AndreasArvidsson:
…terns

Came out of the work on #954 
Not really used for anything at the moment but gives us the opportunity to clamp the text range and support multiline regex patterns if we want to.

## Checklist


pokey (approved):
Ok I made some tweaks, because this PR was only partially adding support for multi-line regexes.  After these tweaks, a modifier stage can just override `expandRangeForSearch` and multi-line regexes should work

If you're happy with the tweaks merge away

Issue #13, opened 2021-06-23
Add ordinal scope types


pokey:
So could say "take next dollar fine" to select the first `$` after the given token.  Could also say "`<ordinal>` `<symbol>`", eg "second dollar".  

In addition, could support "next line fine", "third line fine", etc, as well as "next funk fine", etc

For convenience, potentially "down `<number>`" could be alias for "`<ordinal>` line".  

Would need to support other direction as well, eg "take last dollar fine", or "take second to last dollar fine", tho the latter is a bit verbose

Use "prior" for opposite of "next"

We should also figure out compact grammar for the following:

- [x] https://github.com/cursorless-dev/cursorless/issues/982
- [ ] Scope and next scope
- [ ] Scope and next two scopes
- [ ] Every scope from scope through end of iteration scope
- [ ] Every scope from scope through start of iteration scope
- [ ] Scope and previous scope
- [ ] Scope and previous two scopes
- [x] Target past next scope
- [x] This past next scope
- [x] Target past previous scope
- [x] This past previous scope
- [x] Target past next two scopes (might not need to be compact)
- [ ] How to make these work with `"next <symbol>"` eg `"next comma"`
- [ ] How to make these work with `"next instance <target>"` for selecting next occurrence of a target
  - [ ] Does this work for any target or just mark?
  - [ ] Where does it start searching?  At `<target>` or cursor or able to search from a different target?

## `"next <mark>"`

It should always be next instance of given mark's text relative to input.  So eg if there is a token `"hello"` with a hat over the `h`, then `"next harp"` will select the next instance of `hello` after the cursor

Should support `"every"`, `"first"`, etc for these as well

Note that for things like `:`, `,`, `)`, etc, we don't really care about the token that has the hat because it will always just be the one character, so it will behave like we want anyway.  For failure modes of this assumption (eg `<` and `<=`), we can add something like `glyph` modifier or something in the future if that becomes a problem

## Plurals

Where plurals make sense (eg `"three funks"`), vendor in the following sections of `inflection`:

- https://github.com/jpvanhal/inflection/blob/master/inflection/__init__.py#L1-L141
- https://github.com/jpvanhal/inflection/blob/master/inflection/__init__.py#L284-L306
- https://github.com/jpvanhal/inflection/blob/master/inflection/__init__.py#L419-L426

And have csv overrides code generate a list called `foo_plural` if you pass in a flag asking for plurals

## Some proposed spoken forms:

[...]

PR #960, opened 2022-09-14
Added support for previous and next containing scope type

AndreasArvidsson:
`"take next funk"`
`"change previous state"`

For now you can't use this when not in the current containing scope. This is a limitation of our current implication of every. I still feel it could be really useful.

Fixes https://github.com/cursorless-dev/cursorless/issues/982

## Checklist


cursorless-talon/src/modifiers/containing_scope.py
```diff
@@ -84,16 +84,24 @@ def cursorless_custom_regex_scope_type(m) -> dict[str, str]:
 
 
 @mod.capture(
-    rule="[every] (<user.cursorless_scope_type> | <user.cursorless_custom_regex_scope_type>)"
+    rule="[every | previous | next] (<user.cursorless_scope_type> | <user.cursorless_custom_regex_scope_type>)"
```
AndreasArvidsson, line 87: We really should move these into a csv file. Any recommendation for which file?

src/typings/targetDescriptor.types.ts
```diff
@@ -171,6 +171,16 @@ export interface ContainingScopeModifier {
...
 
+export interface PreviousContainingScopeModifier {
+  type: "previousContainingScope";
+  scopeType: ScopeType;
+}
+
+export interface NextContainingScopeModifier {
+  type: "nextContainingScope";
+  scopeType: ScopeType;
+}
```
auscompgeek, line 182: Is having these as separate interfaces useful? Maybe we can simplify this to be a single interface. Something like this

```ts
export interface SiblingContainingScopeModifier {
  type: "previousContainingScope" | "nextContainingScope";
  scopeType: ScopeType;
}
```

AndreasArvidsson:
@phillco @auscompgeek I have now updated previous/next to use existing ordinal range modifier. For now I have removed all changes Talon side and is going to do a follow up pr regarding #13 where I'm going to implement all the grammar and thoroughly test these.

I think we need discussion about how to handle relative sub tokens with regards to explicit range. For now relative is not enabled for sub tokens.

src/processTargets/modifiers/scopeTypeStages/SubTokenStages.ts
```diff
@@ -0,0 +1,139 @@
...
+    return [this.findSingleTarget(target, targets)];
+  }
+
+  private findSingleTarget(target: Target, targets: Target[]): Target {
+    let intersectingTargets = targets
+      .map((t) => ({
+        target: t,
+        intersection: t.contentRange.intersection(target.contentRange),
+      }))
+      .filter((it) => it.intersection != null);
```
pokey, line 65: why do you need the `map` here?  why not just filter directly
```suggestion
      .filter((t) => t.contentRange.intersection(target.contentRange) != null);
```

src/processTargets/modifiers/scopeTypeStages/SubTokenStages.ts
```diff
@@ -0,0 +1,139 @@
...
+    const targets = this.createTargets(target, contentRanges);
+
+    if (this.modifier.type === "everyScope") {
+      return targets;
+    }
+
+    return [this.findSingleTarget(target, targets)];
+  }
+
+  private findSingleTarget(target: Target, targets: Target[]): Target {
```
pokey, line 59: I'd rename `targets` to something like `allTargets` or something.  Maybe look at your `ItemStage` code for naming inspiration

src/processTargets/modifiers/scopeTypeStages/SubTokenStages.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    // Empty range utilize single leftmost target
+    if (target.contentRange.isEmpty) {
+      return intersectingTargets[0].target;
+    }
+
+    // On non empty range utilize all non-empty intersecting targets
+    intersectingTargets = intersectingTargets.filter(
+      (it) => !it.intersection!.isEmpty
+    );
```
pokey, line 75: is this what we do for eg "token", "block" and "line"?  I'd be tempted to expand on adjacent but idk.  Maybe worth discussing at meet-up

I see now why you're doing the `map` above.  Might be able to remove it depending what we decide here

src/processTargets/modifiers/scopeTypeStages/SubTokenStages.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  protected abstract createTargets(target: Target, ranges: Range[]): Target[];
+}
+
+export class WordStage extends SubTokenStage {
+  constructor(modifier: ContainingScopeModifier | EveryScopeModifier) {
+    super(modifier, SUBWORD_MATCHER);
+  }
+
+  protected createTargets(target: Target, ranges: Range[]): Target[] {
```
pokey, line 97: I would call this `inputTarget`.  We're throwing the word "target" around a lot here so a bit of verbosity might aid clarity

I'd also add a docstring; not immediately obvious what this does and what `ranges` is supposed to be

src/processTargets/processTargets.ts
```diff
@@ -246,3 +235,9 @@ function calcIsReversed(anchor: Target, active: Target) {
 function uniqTargets(array: Target[]): Target[] {
   return uniqWith(array, (a, b) => a.isEqual(b));
 }
+
+function ensureSingleEditor(anchorTarget: Target, activeTarget: Target) {
+  if (anchorTarget.editor !== activeTarget.editor) {
+    throw new Error("AnchorTarget and activeTarget must be in same document");
```
pokey, line 241:
```suggestion
    throw new Error("Cannot form range between targets in different editors");
```

src/test/suite/fixtures/recorded/selectionTypes/clearCar.yml
```diff
@@ -0,0 +1,23 @@
...
+  usePrePhraseSnapshot: true
+  action: {name: clearAndSetSelection}
+initialState:
+  documentContents: aaaBbbCcc
+  selections:
+    - anchor: {line: 0, character: 4}
+      active: {line: 0, character: 4}
+  marks: {}
+finalState:
+  documentContents: aaabbCcc
```
pokey, line 19: huh.  I think I'd go right here.  That's what we do for surrounding pair

pokey (commented):
Wow this turned out super clean!  Left a million little comments and a couple things to discuss at meet-up 😅

cursorless-talon/src/modifiers/ordinals.py
```diff
@@ -0,0 +1,75 @@
...
+            m.cursorless_scope_type, m.ordinal_or_last_list[1]
+        )
+        return {
+            "type": "range",
+            "anchor": anchor,
+            "active": active,
+            "excludeAnchor": not include_anchor,
+            "excludeActive": not include_active,
+        }
+    except AttributeError:
```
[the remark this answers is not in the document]
AndreasArvidsson: Agreed

src/core/commandVersionUpgrades/upgradeV1ToV2/upgradeV1ToV2.ts
```diff
@@ -155,7 +156,7 @@ function upgradePrimitiveTarget(
     // Cursor token is just cursor position but treated as a token. This is done in the pipeline for normal cursor now
     mark: mark?.type === "cursorToken" ? undefined : mark,
     // Empty array of modifiers is not allowed
-    modifiers: modifiers.length > 0 ? modifiers : undefined,
+    modifiers: modifiers.length > 0 ? (modifiers as Modifier[]) : undefined,
```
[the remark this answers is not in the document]
AndreasArvidsson: Because they are of different types. We had to do a new version of modifier when we retired ordinal range.

src/typings/targetDescriptor.types.ts
```diff
@@ -248,19 +266,29 @@ export interface CascadingModifier {
   modifiers: Modifier[];
 }
 
+export interface RangeModifier {
+  type: "range";
+  anchor: Modifier;
+  active: Modifier;
+  excludeAnchor?: boolean;
+  excludeActive?: boolean;
+}
```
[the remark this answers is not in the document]
AndreasArvidsson: Sounds good. I will be there.

src/core/commandVersionUpgrades/upgradeV2ToV3/upgradeV2ToV3.ts
```diff
@@ -0,0 +1,100 @@
...
+    case "primitive":
+      return upgradePrimitiveTarget(target);
+  }
+}
+
+function upgradePrimitiveTarget(
+  target: PartialPrimitiveTargetDescriptor
+): PartialPrimitiveTargetDescriptor {
+  if (target.modifiers != null) {
+    const modifiers = target.modifiers as ModifierV2[];
```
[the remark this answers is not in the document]
AndreasArvidsson: The old modifier is only available in version two and the new one only in version three

src/core/commandVersionUpgrades/upgradeV2ToV3/upgradeV2ToV3.ts
```diff
@@ -0,0 +1,100 @@
...
+    for (let i = 0; i < modifiers.length; ++i) {
+      const oldModifier = modifiers[i];
+      if (oldModifier.type === "ordinalRange") {
+        target.modifiers[i] = createNewModifier(oldModifier);
```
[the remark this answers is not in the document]
AndreasArvidsson: I can use a map() if you prefer? I don't really see a problem with the current approach thou

src/processTargets/modifiers/scopeTypeStages/SubTokenStages.ts
```diff
@@ -0,0 +1,139 @@
+import { Range } from "vscode";
```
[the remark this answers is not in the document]
AndreasArvidsson: Because sub words has special logic around delimiters that the regexStage is lacking

src/processTargets/modifiers/scopeTypeStages/SubTokenStages.ts
AndreasArvidsson: Because I require the intersection at a later stage

src/processTargets/modifiers/scopeTypeStages/SubTokenStages.ts
AndreasArvidsson: That would mean that if you have a single character selected we would expand to all three of them. Probably not to the desired behavior

src/test/suite/fixtures/recorded/selectionTypes/clearCar.yml
AndreasArvidsson: I follow the logic we use with token. There we prefer left.

src/processTargets/modifiers/OrdinalStages.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+}
+
+export class RelativeOrdinalStage implements ModifierStage {
+  constructor(private modifier: RelativeOrdinalScopeModifier) {}
+
+  run(context: ProcessedTargetsContext, target: Target): Target[] {
+    const targets = getTargets(
+      context,
+      createTargetWithoutExplicitRange(target),
```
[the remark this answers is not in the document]
AndreasArvidsson: I feel the same

src/test/suite/fixtures/recorded/selectionTypes/clearWord.yml
```diff
@@ -0,0 +1,23 @@
+languageId: plaintext
+command:
```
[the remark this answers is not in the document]
AndreasArvidsson: Could you elaborate on the first test?

src/processTargets/modifiers/OrdinalStages.ts
```diff
@@ -0,0 +1,112 @@
...
+  run(context: ProcessedTargetsContext, target: Target): Target[] {
+    const targets = getTargets(context, target, this.modifier.scopeType);
+    const isForward = this.modifier.direction === "forward";
+
+    const containingIndices = targets
+      .map((t, i) => ({
+        index: i,
+        intersection: t.contentRange.intersection(target.contentRange),
+      }))
+      .filter((t) => t.intersection != null)
```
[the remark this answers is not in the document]
AndreasArvidsson: In case of an empty selection we would get no containing indices. ie empty selection in the middle of a function will give no result which is not what we want.

src/processTargets/modifiers/OrdinalStages.ts
```diff
@@ -0,0 +1,112 @@
+import { Target } from "../../typings/target.types";
```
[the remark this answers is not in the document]
AndreasArvidsson: It is less than 150 lines so I think it's fine in a single file? Deep directory hierarchies can also be annoying.

pokey (commented):
- [ ] make sure that "token" goes to the right when the tokens are equivalent

src/test/suite/fixtures/recorded/selectionTypes/takeEveryCarBlock.yml
```diff
@@ -0,0 +1,51 @@
...
+    ccc
+  selections:
+    - anchor: {line: 0, character: 0}
+      active: {line: 0, character: 1}
+    - anchor: {line: 0, character: 1}
+      active: {line: 0, character: 2}
+    - anchor: {line: 0, character: 2}
+      active: {line: 0, character: 3}
+    - anchor: {line: 0, character: 3}
+      active: {line: 1, character: 0}
```
pokey, line 36: For some reason "char" seems to be yielding a newline character here

pokey (commented):
- [ ] Add migration and talon side for getting "row" to use new representation; don't bother with ranges for now

src/test/suite/fixtures/recorded/subtoken/takeFirstChar5.yml
```diff
@@ -16,6 +16,6 @@ initialState:
 finalState:
   documentContents: //**
   selections:
-    - anchor: {line: 0, character: 0}
-      active: {line: 0, character: 1}
+    - anchor: {line: 0, character: 2}
```
AndreasArvidsson, line 19: @pokey Notice that we now select the token to the right that will change current behavior

AndreasArvidsson: The problem is that `/\p{L}\p{M}*|\P{L}/gu` matches whitespace and new lines as well

src/test/suite/fixtures/recorded/selectionTypes/takeEveryCarBlock.yml
AndreasArvidsson: The problem is that `/\p{L}\p{M}*|\P{L}/gu` matches whitespace and new lines as well

pokey: That is so strange. Why on earth would whitespace be considered a letter or a diacritic?

AndreasArvidsson: Sounds reasonable. Can we get by just with? `/\p{L}\p{M}*/` or do you have any better suggestions?

pokey: I think that will miss symbols like brackets. I'd look at the token regex and see what we can steal from there. If you come up blank I can take a look a bit later

AndreasArvidsson: This seems to work both grapheme and sub token `/\p{L}\p{M}*|[^\p{L}\p{Z}\p{C}]/gu`

AndreasArvidsson:
@pokey Ready for another review! :)

pokey (commented):
- [ ] Change `type` fields to `"relativeScope"` and `"ordinalScope"`

AndreasArvidsson:
@pokey Modifiers are renamed

src/processTargets/modifiers/ordinalStages/RelativeScopeStage.ts
```diff
@@ -0,0 +1,99 @@
...
+  const targetsWithIntersection = targets
+    .map((t, i) => ({
+      index: i,
+      intersection: t.contentRange.intersection(inputTargetRange),
+    }))
+    .filter((t) => t.intersection != null);
+
+  // Content range is empty. Use rightmost target and accept weak containment.
+  if (inputTargetRange.isEmpty) {
+    // FIXME: Handle case where no targets intersect with the input range
```
pokey, line 82: This breaks when the input is between targets.  For example:

```
hello | world
```

In this case `"next token"` breaks

pokey (changes requested):
Ok I pushed a bunch of minor changes, and was hoping to just do everything myself, but ran into an issue with the logic for relative scopes, and I think it may require a rethinking of the way you handle relative scopes, so I'm going to put this one back into your court

src/processTargets/modifiers/ordinalStages/RelativeScopeStage.ts
AndreasArvidsson: I know. My reasoning for leaving it as it is was that most of the scopes wouldn't even support `every` in that case and therefore relative scope has no targets to work with.

pokey: Right, but "next token" should work, and is pretty useful

Also I think it indicates a flaw in the logic of the code as it violates an assumption that is baked in. Hesitant to merge the code with this issue

AndreasArvidsson: Okay.  I will have a look at it tomorrow.

AndreasArvidsson: done

pokey (commented):
Ok I made a bunch of minor changes, and left one question.  Lmk if you have a problem with any of the changes, and curious to hear your thoughts on my comment, but otherwise good to go!

src/processTargets/modifiers/ordinalStages/RelativeScopeStage.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    const containingIndices = (() => {
+      const containingIndices = getContainingIndices(
+        target.contentRange,
+        targets
+      );
+
+      // The content range isnt containing any scopes, but they should not be
+      // included anyhow. Find a non containing relative index.
+      if (containingIndices.length === 0 && !includeContaining) {
```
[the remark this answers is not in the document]
AndreasArvidsson: I'm not against is. Looks like you are on a roll so please go ahead and make these changes :)

pokey (approved):
Ok I think this is good to go!  I added a bunch of tests and did some cleanup.  Have a quick look before you merge to make sure you're happy

Fwiw I made a super basic grammar for `"two funk"` so I could easily add tests; doesn't have pluralisation tho so I'm not going to merge it.  [Here](https://github.com/cursorless-dev/cursorless/commit/6820a2104cd03a62a019dd970405498c5c286f4d) it is tho in case helpful

src/processTargets/modifiers/ordinalStages/RelativeScopeStage.ts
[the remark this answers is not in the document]
pokey: done :)

AndreasArvidsson:
I think it looks good :)

PR #965, opened 2022-09-18
Smarter `that` marks for easy actions

pokey:
This PR partially implements #466 by adding smarter `that` marks for actions where it is easy to do so, because they don't modify the target.  It adds rich targets to the following actions:

- `"highlight"`
- `"take"`
- Any action that uses `CommandAction` that doesn't modify the document, eg `"hover"`

## Checklist

### DIFF
--- a/src/test/suite/fixtures/recorded/actions/copyVest.yml
+++ b/src/test/suite/fixtures/recorded/actions/copyVest.yml
@@ -32,5 +32,5 @@ finalState:
         start: {line: 1, character: 6}
         end: {line: 1, character: 11}
       isReversed: false
-      hasExplicitRange: true
+      hasExplicitRange: false
 fullTargets: [{type: primitive, mark: {type: decoratedSymbol, symbolColor: default, character: v}, selectionType: token, position: contents, modifier: {type: identity}, insideOutsideType: inside}]
--- a/src/actions/CommandAction.ts
+++ b/src/actions/CommandAction.ts
@@ -1,5 +1,6 @@
 import { flatten } from "lodash";
 import { commands, window } from "vscode";
+import { selectionToThatTarget } from "../core/commandRunner/selectionToThatTarget";
 import { callFunctionAndUpdateSelections } from "../core/updateSelections/updateSelections";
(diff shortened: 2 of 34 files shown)


src/test/suite/fixtures/recorded/actions/copyVest.yml
```diff
@@ -30,4 +30,5 @@ finalState:
     - contentRange:
         start: {line: 1, character: 6}
         end: {line: 1, character: 11}
+      hasExplicitRange: false
```
pokey, line 33: note that "copy" uses `CommandAction` and doesn't change the document, so it gets rich targets now.  This one loses its explicit range because it is a decorated mark, and they don't have explicit range.  So now you can say "copy air", then "take inside that"

src/test/suite/fixtures/recorded/compoundTargets/postBatVerticalUntilFine.yml
```diff
@@ -44,10 +44,12 @@ finalState:
     - anchor: {line: 3, character: 1}
       active: {line: 3, character: 1}
   thatMark:
-    - contentRange:
+    - type: PlainTarget
```
pokey, line 47: vertical ranges return plain targets, and "post" leverages `setSelection`, so it now preserves this information

src/test/suite/fixtures/recorded/actions/copyVest.yml
```diff
@@ -32,5 +32,5 @@ finalState:
         start: {line: 1, character: 6}
         end: {line: 1, character: 11}
       isReversed: false
-      hasExplicitRange: true
```
AndreasArvidsson, line 35: Why was this true before?

AndreasArvidsson (approved):

src/test/suite/fixtures/recorded/actions/copyVest.yml
pokey: https://github.com/cursorless-dev/cursorless/pull/965#discussion_r973741366

PR #970, opened 2022-09-19
Support default recorder config per directory

pokey:
This allows us to make it so that all tests in the `actions` directory capture the `that` mark.

Thanks to @AndreasArvidsson for the idea.

Requires #968 because it uses the flag introduced there

## Checklist

- [x] Add config for highlights dir and maybe others
- [x] Add link to test case recorder arg type to docs for default dir config and for recorder config


AndreasArvidsson (approved):
Turned out great :)

PR #974, opened 2022-09-21
Disable issues on cursorless-talon

pokey:
## Checklist

### DIFF
--- a/cursorless-talon/.github/ISSUE_TEMPLATE/config.yml
+++ b/cursorless-talon/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Create issue on cursorless-dev/cursorless
+    url: https://github.com/cursorless-dev/cursorless/issues/new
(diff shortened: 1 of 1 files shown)


auscompgeek:
TIL this is a thing. Although it'd be simpler to just disable issues entirely in the repo settings. Hmm.

Is there an example repo somewhere set up like this? I wonder whether this config actually has any effect with no templates.

pokey:
Yeah I was hesitant to disable issues entirely, because then people might not know what to do and just give up.  I haven't seen a repo do this, but figured it was worth a try 🤷‍♂️

AndreasArvidsson (approved):

AndreasArvidsson:
I'm not against disabling issues on Cursorless Talon and then just put an instruction at the top of the readme

pokey:
@auscompgeek seems to work well I think.  Try pressing New Issue on https://github.com/cursorless-dev/cursorless-talon/issues

AndreasArvidsson:
I think this turned out really well :)

Issue #867, opened 2022-07-14
Restructure documentation to reflect new compositional modifiers


pokey:
We'd like to restructure out documentation to more accurately reflect Cursorless's new internal representation.  We should have sections for:

- Scope types
- Modifiers
- Actions

They can link to each other.  So for example the "containing scope" modifier can link to list of scope types

- [ ] Also discuss frontmatter thing for auto-generating docs

Scope types should include:

- [ ] Languages supported
- [ ] Examples from different languages
- [ ] Iteration scope
- [ ] Domain
- [ ] Removal range
- [ ] Content range

Could include info from test cases as per #866?

See also some (very rough 😅) images from a brainstorming session:

![IMG_8536](https://user-images.githubusercontent.com/755842/179037752-1a2ea2b2-1ac8-493a-9a15-5e639962fb9c.jpg)
![IMG_8535](https://user-images.githubusercontent.com/755842/179037759-dd8197ce-18d0-44a5-98f8-0c1630dd5ae2.jpg)
![IMG_8534](https://user-images.githubusercontent.com/755842/179037766-1b3ebba6-4d0c-4de9-9a55-78199818a272.jpg)
![IMG_8539](https://user-images.githubusercontent.com/755842/179037697-17f5960a-f604-403b-b202-5876a047d7f0.jpg)
![IMG_8538](https://user-images.githubusercontent.com/755842/179037727-8a7b67ed-bf68-4afd-80f5-475461d1d0a5.jpg)
![IMG_8537](https://user-images.githubusercontent.com/755842/179037741-83f0f320-2202-4fe0-b98a-3f120ef5d901.jpg)

AndreasArvidsson:
The doc page now have separate sections for actions modified scopes and so on with examples.

Issue #843, opened 2022-07-07
Add developer docs for cheatsheet


pokey:
Basic docs introduced in #1718. In addition, we'd like to:

- [ ] How CI works
- [ ] Add note about the fact that the cheatsheet doesn't get rolled up when you run extension locally and how to fix that / run locally
- [x] Switch cheatsheet json update to use cursorless repo root
- [x] add note that you need to regen cheatsheet from development instance
- [ ] switch to using spaces instead of tabs
- [x] document / look into the fact that for some reason you need to have root-level typescript watcher running for live reload to work effectively
- [x] file issue to use default spoken forms for cheatsheet
  - [ ] need to have version of shapes / colors lists that are not restricted to what appears in vscode `settings.json`
  - [ ] need to have version of section heading lists that exclude the ones that are off by default
  - [x] add test for cheatsheet that compares payload to `defaults.json`
  - [ ] mention to run spoken form test in cheatsheet developer docs
- [x] How to build locally
- [x] How to update the example json for local development
- [x] How to run locally

PR #987, opened 2022-10-02
Support complex relative grammar

### LINKED ISSUE #979: Add a few terms to overrides csv
- [ ] "last"
- [ ] "previous"
- [ ] "next"
- [ ] other stuff?

_Originally posted by @pokey in https://github.com/cursorless-dev/cursorless/pull/960#discussion_r981181083_

AndreasArvidsson:
Partial #13 
Fixes #979

image

## Checklist

- [x] Update relative modifier grammar 
- [x] Add new keywords to csv files
- [x] Update cheat sheet
- [x] Update docs


cursorless-talon/src/modifiers/relative_scope.py
```diff
@@ -5,13 +5,77 @@
...
+def cursorless_relative_backward() -> str:
+    """Backward"""
+    return "backward"
+
+
+@mod.capture(
+    rule="[<user.ordinals_small>] <user.cursorless_relative_direction> <user.cursorless_scope_type>"
+)
+def cursorless_relative_scope_singular(m) -> dict[str, Any]:
+    """Relative previous/next singular scope. `next funk`"""
```
pokey, line 24:
```suggestion
    """Relative previous/next singular scope, eg `"next funk"` or `"third next funk"`."""
```

AndreasArvidsson:
The grammar is now updated to support every thing from the image in issue #13. 

image

cursorless-talon/src/modifiers/relative_scope.py
```diff
@@ -5,13 +5,77 @@
...
+    return create_relative_scope_modifier(
+        m.cursorless_scope_type,
+        getattr(m, "ordinals_small", 1),
+        1,
+        m.cursorless_relative_direction,
+    )
+
+
+@mod.capture(
+    rule="<user.cursorless_relative_direction> <number_small> <user.cursorless_scope_type_plural>"
```
pokey, line 34: Do we want to support a rule that combines this with the above, so that you can say eg `"third next two funks"`?  Our api supports that and I don't think there's any way to express it

cursorless-talon/src/modifiers/ordinal_scope.py
```diff
@@ -59,7 +59,7 @@ def cursorless_ordinal_scope(m) -> dict[str, Any]:
     return m[0]
```
pokey, line 59: shall we modify "last three funk" to be plural, ie "last three funks"?

cursorless-talon/src/modifiers/scopes.py
```diff
@@ -0,0 +1,128 @@
...
+
+mod = Module()
+
+
+mod.list("cursorless_scope_type", desc="Supported scope types")
+mod.list("cursorless_scope_type_plural", desc="Supported scope types")
+mod.list(
+    "cursorless_custom_regex_scope_type",
+    desc="Supported custom regular expression scope types",
+)
```
pokey, line 13: don't you need to declare a plural for this one as well?

cursorless-talon/src/vendor/inflection.py
```diff
@@ -0,0 +1,174 @@
+"""Copied from inflection. https://github.com/jpvanhal/inflection/blob/master/inflection/__init__.py"""
```
pokey, line 1: Worth including a link to the license as well

pokey, line 1: Worth including a link to the license as well

pokey (commented):
Cool!! 😍

cursorless-talon/src/modifiers/relative_scope.py
AndreasArvidsson: I'm leaning against it. First I think the grammar is becoming way to verbose to be able to be remembered easily and you can always just use a hat. eg `"two funks air"`. And since we are talking about something that is close to the selection we will probably always have a good hat.

src/processTargets/modifiers/RelativeScopeStage.ts
```diff
@@ -1,6 +1,5 @@
 import { findLastIndex } from "lodash";
 import { Range } from "vscode";
-import { NoContainingScopeError } from "../../errors";
```
auscompgeek, line 3: Heh, was about to make a PR to fix this lint too. :+1:

cursorless-talon/src/modifiers/relative_scope.py
pokey: Hmm yeah I'm slightly on the fence as well.  Tbh the easiest way to approximate this behaviour is to just chain: `"two funks third next funk"`, which will work with multiple cursors, unlike using a mark

The main reason I'm leaning towards this direction is that I kind of like the idea of orthogonalising position and quantity.  Then we have these ways of specifying position:

- `"first"` / `"second"` / `"last"`
- `"next"` / `"previous"` / `"second next"` / `"second previous"`

And we have these ways of specifying quantity:

- `"funk"`
- `"two funks"`

And we can combine any position with any quantity

But I'm not sure practically speaking how much we'd use this expressivity 🤷‍♂️

pokey (commented):
One thing that's missing in this PR from the diagram is `"second last"` (equivalent to "second to last" in English)

cursorless-talon/src/modifiers/relative_scope.py
```diff
@@ -5,13 +5,76 @@
...
+
+
+@mod.capture(rule="backward")
+def cursorless_relative_backward() -> str:
+    """Backward"""
+    return "backward"
+
+
+@mod.capture(
+    rule="[<user.ordinals_small>] <user.cursorless_relative_direction> <user.cursorless_scope_type>"
```
pokey, line 21: I don't believe we have any tests for `"third next"`

pokey (commented):
- [ ] Oh yeah we should also do a quick DFA test before merge just in case

cursorless-talon/src/modifiers/relative_scope.py
AndreasArvidsson: No tests are still on my to do list that's why the task isn't completed in this pr :)

pokey: Right sounds good.  Fwiw I'm pretty sure those are the only tests necessary; I believe I added tests for everything else

AndreasArvidsson:
@pokey I have updated the legend extension side, but I'm not sure how to actually test it. That cheat sheet doesn't work in development host for me.

Since I know that you are going to rewrite all of the documentation any how do you mind just doing it originally the way you want it?

pokey:
> @pokey I have updated the legend extension side, but I'm not sure how to actually test it. That cheat sheet doesn't work in development host for me.

I can test it out.  You're off the hook until I get around to https://github.com/cursorless-dev/cursorless/issues/843 😊

> Since I know that you are going to rewrite all of the documentation any how do you mind just doing it originally the way you want it?

I will probably wait until https://github.com/cursorless-dev/cursorless/issues/867 to rework, as we can't really do anything too proper until we rework the docs to separate the notion of scope from containing scope, which I think is out of scope (ha) for this PR.  So if you wanna just drop something rough that would be great, so that users at least know this exists, even if it's not the most polished

AndreasArvidsson:
Looks like we don't have a section in the documentation for general modifiers? I'm a bit unsure where to add this.

AndreasArvidsson:
Ok documentation updated. Ready for review :)

pokey:
I just tested DFA, and unfortunately it is above 1s with this PR 😕.  DFA caching does help, so when there is a cache hit it is ~1ms, but there are enough cache misses to make the UX not great.  Worth testing on your machine @AndreasArvidsson to see how it feels

AndreasArvidsson:
Is this for public or beta?  
I do have a long compile time the first time I select vscode. But once that is cached I can toggle back and forth with no problem. Do you have a trick to get cache misses?

AndreasArvidsson:
Yeah but that's only the first time isn't it? Switching back and forth will use the cache?

AndreasArvidsson:
Sounds good

AndreasArvidsson:
@pokey I found an optimization for `number_small`. Basically instead of using an advance capture I will just turn it into a single list. I made a few different performance measurements and it's very clear that a list is definitely the fastest.
```python
@ctx.capture("number_small", rule=f"({alt_digits} | {alt_teens} | {alt_tens} [{alt_digits}])") # 1710ms
@ctx.capture("number_small", rule=f"one") # 710ms
@ctx.capture("number_small", rule="{user.number_small}") # 667ms
@ctx.capture("number_small", rule=f"{'|'.join(number_small)}") # 1794ms
def number_small(m): 
    return int(parse_number(list(m)))
```

pokey:
oh yeah good call.  yeah like you suggest we could probably inline a bunch of compound captures into big lists.  certainly for things that don't include customisable spoken forms.  for lists that are customisable we could just regenerate the lists if part of it changes.  but yeah I'll try out the `number_small` optimisation and see if that gets us to a good enough spot

pokey:
Wow just tested locally.  Absolute game changer.  Filed PR on knausj to save you the trouble https://github.com/knausj85/knausj_talon/pull/993.  I'll finish reviewing this PR and I think we can get this one in and just tell people they need to update their knausj once that PR is merged

pokey:
Just tried it on my personal fork, and it's <600ms, so that's good.  Still slower than I'm used to, but good enough for now.

AndreasArvidsson:
Nice. I saved over a second in compile time on this pr. That is quite massive for such a small change  :)

pokey (approved):
Ok I'm happy with it, but I made a bunch of changes, so have a look.  I changed all the capture formatting in cheatsheet; see screenshot in description

pokey:
Actually one more thing let me drop diagram into docs

pokey:
Ok good to go; have a look to make sure it all seems good to you and then feel free to merge if so

pokey (approved):

PR #998, opened 2022-10-02
Added support for line number ranges

### LINKED ISSUE #438: Feature request: refer to a range of rows without saying "row" twice

AndreasArvidsson:
fixes #438 

## Checklist

### DIFF
--- a/cursorless-talon/src/marks/lines_number.py
+++ b/cursorless-talon/src/marks/lines_number.py
@@ -3,6 +3,7 @@ from dataclasses import dataclass
 from typing import Any
 
 from talon import Context, Module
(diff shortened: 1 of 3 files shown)


AndreasArvidsson:
@pokey Ready for review

AndreasArvidsson:
@wolfmanstout This is now merged :)

wolfmanstout:
Great, thanks!

PR #1008, opened 2022-10-04
Filter modifiers

### LINKED ISSUE #1005: `"next line"` doesn't behave as expected with empty lines
Our `"line"` modifier doesn't count empty / whitespace-only lines as lines, so `"next line"` doesn't allow you to target empty lines.  My primary use case is `"chuck next line air"`.

Tbh I'm not exactly sure what the fix should be, because only considering lines that are not all whitespace is useful in other situations 🤔.  Probably worth discussing

### LINKED ISSUE #1006: Add filter modifiers
Returns `[]` if input passes filter, otherwise `[inputTarget]`

- [x] `"empty"`: `keepEmptyFilter` zero length or whitespace content only
- [x] `"content"` `keepContentFilter` nonzero length and nonwhitespace content (`=!keepEmptyFilter`)
- [x] Don't use `"whitespace"`

Can be used for eg `"take content every line"`

### Filter viewpoint
- is it nonzero length?
- does it have visible characters selected?

empty
empty and not adjacent to visible characters
empty and adjacent to visible characters
nonempty
nonempty and only containing whitespace
nonempty and containing visible chars

AndreasArvidsson:
Fixes #1005 
Fixes #1006 

## Checklist

- [x] Add filter modifiers
- [x] `"every line"` should include empty and whitespace lines


cursorless-talon/src/modifiers/modifiers.py
```diff
@@ -14,6 +14,9 @@
     "just": "toRawSelection",
     "leading": "leading",
     "trailing": "trailing",
+    "content": "content",
+    "empty": "empty",
+    "whitespace": "whitespace",
```
pokey, line 19: Now that I see the impl for these two, I'm kind of struggling to see the use case.  I wonder if we should have them disabled by default, or just remove them for now?  Not sure it's worth polluting our spoken forms with things that we may never use, even if they are trivial to implement 🤔

I think `"content"` is great.  The term is a bit unintuitive, because just the word `"content"` on its own doesn't imply to me that it is filtering.

I wonder if default spoken form should be a bit longer, given it's unlikely to be a super common use case.  Eg `"only content"` or `"keep content"`

We could by default prefix all of them with `"keep"` or `"only"`, then it's not such a big deal that we're polluting the spoken forms as it's a bit longer

src/processTargets/getModifierStage.ts
```diff
@@ -5,6 +5,11 @@ import {
   Modifier,
 } from "../typings/targetDescriptor.types";
 import CascadingStage from "./modifiers/CascadingStage";
+import {
+  ContentStage,
+  EmptyStage,
+  WhitespaceStage,
```
pokey, line 11: I wonder if we should be a bit more verbose, and call these eg `ContentFilterStage`, `EmptyFilterStage`, and `WhitespaceFilterStage`.  Ie all filter stages are suffixed with `FilterStage`

pokey (commented):
Super cool to see that this was actually as trivial as we hoped.  Great sign.  Left a couple minor comments

src/processTargets/getModifierStage.ts
AndreasArvidsson: Done. Should we still keep the modifier type eg "content" or do you prefer "contentFilter"?

AndreasArvidsson:
It is an interesting idea. Maybe we should do that in addition to the ones we have here? In some sense I think that content fits my mental image better than `only characters`

pokey:
I'm just hesitant to add something to the csv if we aren't sure about it

AndreasArvidsson:
I think we should add those in addition.

pokey:
Let's chat tomorrow

AndreasArvidsson:
@pokey Updated. I'm still a bit hesitant if we should add the word "keep" before the default spoken form? This is really only a modifier for multiple cursors so maybe it's fine?

pokey (approved):

PR #1010, opened 2022-10-04
Improve cheatsheet update machinery

pokey:
- Sorts `defaults.json` for diff stability
- Adds Talon-side action for updating `defaults.json`

## Checklist


cursorless-nx/libs/cheatsheet/src/lib/data/sampleSpokenFormInfos/defaults.json
```diff
@@ -5,252 +5,266 @@
 			"id": "actions",
 			"items": [
 				{
-					"id": "scrollToBottom",
```
pokey, line 8: alphabetised / added things which had been forgotten from previous updates.  Remember this file is just for convenience during local cheatsheet development

src/core/Cheatsheet.ts
```diff
@@ -111,10 +111,38 @@ export default class Cheatsheet {
       "defaults.json"
     );
 
-    await writeFile(defaultsPath, JSON.stringify(spokenFormInfo, null, "\t"));
+    const outputObject = produce(spokenFormInfo, (draft) => {
+      draft.sections = sortBy(draft.sections, "id");
+      draft.sections.forEach((section) => {
+        section.items = sortBy(section.items, "id");
+      });
+    });
```
pokey, line 119: I think [immer](https://immerjs.github.io/immer/) is a nice way to handle immutable data structures.  Planning to start using it if no objections

thanks for the pointer @SimeonC

AndreasArvidsson (approved):

src/core/Cheatsheet.ts
auscompgeek: 👍 It does seem like it's more suited to our use case than immutable.js which seems to be the other choice here. I found an article comparing the two: https://blog.logrocket.com/immer-and-immutable-js-how-do-they-compare/

pokey: That's what they do [in the docs](https://immerjs.github.io/immer/#with-immer) 🤷‍♂️

PR #1011, opened 2022-10-05
Use natural sorting order for numbers

### LINKED ISSUE #699: Sort is incorrect when sorting numbers
According to @AndreasArvidsson, sort is implemented by casting items as strings and then sorting based on the string representation of the numeral, resulting in the following bug:

Case:
```let list = [9, 3, 11, 90];```

Result:
```let list = [11, 3, 9, 90];```

Expected:
```let list = [3, 9, 11, 90];```

### PR DESCRIPTION
**Pull request by AndreasArvidsson**
Fixes #699 

## Checklist

### CODE REVIEW

File: src/test/suite/fixtures/recorded/actions/sortEveryItem.yml [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+  selections:
+    - anchor: {line: 0, character: 0}
+      active: {line: 0, character: 0}
+  marks: {}
+finalState:
+  documentContents: 3, 9, 11, 90
+  selections:
+    - anchor: {line: 0, character: 0}
+      active: {line: 0, character: 0}
+  thatMark:
```
pokey, line 23: I do wonder if we want to start having specific tests for the `thatMark`.  Even limiting it to actions results in a lot of noise 🤔

Obv out of scope here but just a thought

pokey (approved):
I made a few changes, but otherwise looks good to me

I was tempted to have a Unicode test, but worry that might vary by locale, so result in non-deterministic failures.  If we wanted to test that properly, we should probably mock the locale or something, but I think this is ok for now

src/test/suite/fixtures/recorded/actions/sortEveryItem.yml
AndreasArvidsson: Definitely worth thinking about.

AndreasArvidsson:
> I was tempted to have a Unicode test, but worry that might vary by locale, so result in non-deterministic failures. If we wanted to test that properly, we should probably mock the locale or something, but I think this is ok for now

I tried that, but it failed on ci due to different locale. I tested it locally and everything appears to be working as expected.

PR #1028, opened 2022-10-10
Token removal range based upon merging of tokens or not

### LINKED ISSUE #800: Remove leading / trailing whitespace when next to paired delimiter
## The problem
Today, we are quite cautious about removing leading or trailing whitespace when we delete a token, to avoid unintentionally merging two tokens.  The current behaviour only removes whitespace if the token is not directly adjacent to another token on one side.  This behaviour is a bit bothersome when we actually do want the whitespace removed.

## The solution

For this first step, we would like to relax this behaviour and delete whitespace in the following cases:

- the token has leading whitespace, and the token to the right of the removed token is a closing delimiter
- the token has trailing whitespace, and the token to the left of the removed token is an opening delimiter

This PR will likely entail a modification to [`getTokenRemovalRange`](https://www.cursorless.org/docs/contributing/api/modules/processTargets_targetUtil_insertionRemovalBehaviors_TokenInsertionRemovalBehavior/#gettokenremovalrange)

This behaviour will be especially useful when combined with #737 

### Caveats

We'll need to think a bit about how to handle quotation marks, as they can be opening or closing delimiter depending on context.  Ideally, we'd leverage the logic for yielding delimiters that is used by our core surrounding pair algorithm, but it will require a bit of work to separate that code.  For now, we should either:

- Always remove space next to quote
- Never remove space next to quote

Neither is ideal, because we probably don't want to remove space when we're outside the string and probably do want to remove it if we're inside the string

AndreasArvidsson:
Fixes #800

## Checklist

### DIFF
--- a/src/processTargets/targetUtil/insertionRemovalBehaviors/TokenInsertionRemovalBehavior.ts
+++ b/src/processTargets/targetUtil/insertionRemovalBehaviors/TokenInsertionRemovalBehavior.ts
@@ -67,8 +67,8 @@ export function getTokenRemovalRange(
   target: Target,
   contentRange?: Range
 ): Range {
-  const actualContentRange = contentRange ?? target.contentRange;
   const { document } = target.editor;
+  const actualContentRange = contentRange ?? target.contentRange;
   const removalRange = getDelimitedSequenceRemovalRange(target, contentRange);
 
   if (!actualContentRange.isEqual(removalRange)) {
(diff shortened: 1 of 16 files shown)


pokey (approved):
Nice!  Let's ship it and see how it feels for a few days; not too destructive so can always tweak / rollback if we find it to be strange / awkward in some way

Issue #797, opened 2022-06-28
Implement generic "every" logic


pokey:
- For `"every <scope> this"` with empty selection, expand to iteration scope and yield every scope in that range
- For `"every <scope> this"` with non-empty selection, yield every scope that has non-empty overlap with selection
- For `"every <scope> air"`, expand to iteration scope and yield every scope in that range
- For `"every <scope> block"`, yield every scope that has non-empty overlap with selection

For reference, this is what weakness looks like:

- For `"if wrap this"` with empty selection, wrap containing statement
- For `"if wrap this"` with non-empty selection, wrap containing statement
- For `"if wrap air"`, wrap containing statement
- For `"if wrap block"`, just wrap exact block

Possible solution is to change `CursorMarkStage` to only return weak target if selection is empty.  In that case, weakness for wrap now looks like:

- For `"if wrap this"` with empty selection, wrap containing statement
- For `"if wrap this"` with non-empty selection, just wrap exact selection
- For `"if wrap air"`, wrap containing statement
- For `"if wrap block"`, just wrap exact block

Then we would make it so that for `"every"`, we expand to iteration scope for weak targets and for non weak targets we yield every scope that overlaps. Then our proposal at the top works because the first in the third are week so they expand to iteration scope but the second in the fourth do not the because they are not weak

Ideally as much as possible is implemented in new `EveryScopeStage`, but will need to think a bit about how much autonomy scope handlers need

We should probably let `EveryScopeStage` decide whether to expand to iteration scope, and then just pass input target content range and Boolean indicating whether to expand

- [ ] Don't allow input range start and end to be in different iteration scopes
- [ ] https://github.com/cursorless-dev/cursorless/issues/894

PR #1031, opened 2022-10-12
Generic scope handler interfaces

### LINKED ISSUE #484: Update scope type modifier to expand on both ends

AndreasArvidsson:
- Implements generic scope handlers
- Adds `"identifier"` scope type
- Closes https://github.com/cursorless-dev/cursorless/issues/797
- Closes #484 

## Checklist

- [x] Add "identifier" scope type
- [x] Switch "word" and "char" to use this new setup, with "identifier" as their parent scope type
- [x] Add test for `"take next word"` when cursor is in the whitespace between two tokens, eg `"foo | barBaz"`
- [x] Add test for `"one token backward"`
- [x] Link issues that this PR closes and clean up related issues
- [x] Remove old modifier stages
- [x] Figure our grand scopes. Just pass ancestorIndex to `getScopesTouchingPosition`
- [x] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet_html)


src/core/tokenizer.ts
```diff
@@ -123,10 +123,20 @@ const tokenMatchersForLanguage: Partial<Record<LanguageId, Matcher>> =
...
+  const matcher =
     tokenMatchersForLanguage[languageId as SupportedLanguageId] ??
-    defaultMatcher
-  );
+    defaultMatcher;
+
+  return {
+    ...matcher,
+    // TODO Why is this necessary???????????
+    identifierMatcher: new RegExp(matcher.identifierMatcher),
```
AndreasArvidsson, line 133: I can't for the life of me understand why I need to create a new instance of this regex. As far as I can tell the old and new one are identical. Could we be running into a concurrency problem?

src/processTargets/getModifierStage.ts
```diff
@@ -63,8 +39,9 @@ export default (modifier: Modifier): ModifierStage => {
     case "trailing":
       return new TrailingStage(modifier);
     case "containingScope":
```
AndreasArvidsson, line 41: Noticed that containing and every are two separate modifiers. The different scope types are not modifiers per se.

src/processTargets/modifiers/ContainingScopeStage.ts
```diff
@@ -0,0 +1,120 @@
...
+      startTarget.createContinuousRangeTarget(
+        target.isReversed,
+        endTarget,
+        true,
+        true
+      ),
+    ];
+  }
+
+  private runLegacy(
```
AndreasArvidsson, line 75: This should of course be removed ones all the scope types are migrated into scope type handlers

src/processTargets/modifiers/scopeHandlers/TokenScopeHandler.ts
```diff
@@ -0,0 +1,79 @@
...
+      (contentRange) =>
+        new TokenTarget({
+          editor,
+          isReversed,
+          contentRange,
+        })
+    );
+  }
+
+  protected getContainingIndicesForPosition(
```
AndreasArvidsson, line 35: Notice how token overwrites the base class behavior for containing scope for a position(empty range)

src/processTargets/modifiers/scopeHandlers/scopeHandler.types.ts
```diff
@@ -0,0 +1,194 @@
...
+      const intersection = r.intersection(iterationScope);
+      return intersection != null && !intersection.isEmpty;
+    });
+  }
+
+  protected abstract getEveryTarget(
+    editor: TextEditor,
+    contentRange: Range,
+    isReversed: boolean,
+    hasExplicitRange: boolean
```
pokey, line 192: I'm not sure this is what we want to be passing in here; won't "next" want to fake it to `false`?  I think it's confusing to have something that is an attribute of the target and then to hack it to be false.  Could we maybe just name the variable something else?

I think it would be useful to see "next" implemented, as this impl is still doing the hacky faked explicit range thing via calling "every".  Shouldn't be too hard, no?

AndreasArvidsson: Yeah probably. As I said at the top names could probably be improved. 

I can do that. I'm just a bit has content of having these different implementations: new/legacy without support from the type system.

AndreasArvidsson: Ok now relative and ordinal scope stages use the new scope handler for token

src/processTargets/modifiers/scopeHandlers/TokenScopeHandler.ts
```diff
@@ -0,0 +1,79 @@
...
+  ): Target[] {
+    const tokenRanges = getTokensInRange(
+      editor,
+      expandToFullLine(editor, contentRange)
+    ).map(({ range }) => range);
+
+    const filteredTokenRanges = hasExplicitRange
+      ? this.filterRangesByIterationScope(contentRange, tokenRanges)
+      : tokenRanges;
+
```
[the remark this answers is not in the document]
AndreasArvidsson: Maybe` but I'm not sure if that is going to be consistent across all different scope types.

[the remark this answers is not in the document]
AndreasArvidsson: We might have multiple iteration scopes. Functions inside/outside classes for example. I'm not sure it will differ but it night since we haven't thought about that implementation yet it's hard to say.

Also this line is in the middle of another function I'm unsure about how much code we save by extracting this.

src/processTargets/modifiers/scopeHandlers/TokenScopeHandler.ts
```diff
@@ -0,0 +1,79 @@
...
+    return filteredTokenRanges.map(
+      (contentRange) =>
+        new TokenTarget({
+          editor,
+          isReversed,
+          contentRange,
+        })
```
pokey, line 31: I'd be tempted to have this construction be a separate function in the scope handler.  Then we can avoid having to construct targets that will end up getting filtered out when the above filter step is in the base class.  Idk; I don't feel super strongly on this one tho. Might be simpler to just have one function that returns targets rather than one that returns the ranges and another that constructs the targets

One thought though is that returning something other than a target would allow us to add domains to each range, so that the base class can use that for intersection

AndreasArvidsson: I've had that thought as well. The problem is that the constructor for some of these targets has a lot more information than this. Basically this is the untyped clone parameter issue all over again.

pokey: Another advantage of separating the construction is that we can remove another arg from this function signature; don't need `isReversed` here

Would be kinda cool if this function just took an editor and a range

AndreasArvidsson: Agreed. If we can just find a good enough solution for the constructor parameters then I'm totally with you.

src/processTargets/modifiers/scopeHandlers/TokenScopeHandler.ts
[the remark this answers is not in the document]
pokey: Also allows us to make `filterRangesByIterationScope` private rather than protected, which is nice.  Just trying to simplify the contract between base class and scope class as much as possible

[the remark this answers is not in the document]
pokey: I know you view it as an implementation detail, but I think it should be a first-class api.  For example, we could imagine third parties might implement their own scope types via extensions.  I know that sounds far out, but the more we can view separate components of cursorless as independent pieces with well-defined api contracts, the better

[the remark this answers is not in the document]
AndreasArvidsson: I will give it a try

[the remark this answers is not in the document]
pokey: Looking at the code more, I'm actually tempted to move the filter stage into the modifiers.  It's only "every" and ordinal that need to do this; containing and relative don't filter

If you're really concerned about the duplication, we could revert to having ordinal call every stage; I actually don't think that's too crazy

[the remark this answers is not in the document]
AndreasArvidsson: Updated now. Definitely needs to have a look at the typing system, but something general is in place now at least.

pokey:
Ok super helpful to see the code written out; great call to just bang it out.

Here's what I'm thinking.  Scope type exposes two functions:

- A function that takes a position and returns containing iteration range, just as a range
- A function that takes an iteration range and returns a list of scopes

I think it would look as follows

```ts
interface ScopeDefinition {
   getIterationRange(editor: TextEditor, position: Position): Range;
   getScopesInIterationRange(editor: TextEditor, range: Range): Scope[];
}

interface Scope {
   domain: Range;
   getTarget(isBackward: boolean): Target;
   getPriority?(): number;
}
```

Then the base class / intermediary has an interface as follows:

```ts
/** For use as return type of {@link ScopeHandler.getTargetsForRange} */
interface IterationScope {
   targets: Target[]
   containingStartIndex: number;
   containingEndIndex: number;
}

class ScopeHandler {
   getTargetsForRange(editor: TextEditor, range: Range, isBackward: boolean): IterationScope {
      const startIterationRange = self.scopeDefinition.getIterationRange(editor, range.start)
      const iterationRange = contains(startIterationRange, range.end) ? startIterationRange : startIterationRange.union(self.scopeDefinition.getIterationRange(editor, range.end)

      const scopes = self.scopeDefinition.getScopesInRange(editor, iterationRange)

      // Then compute containing scope by intersection with domain for non-empty range
      // For empty range, pick weakly containing domain, breaking ties using `scope.getPriority()` if it exists, else prefer right
      // Then call `getTarget()` on the ones we want to return for `targets`
   }
}
```

Then

- containing scope modifier stage is implemented by grabbing the `iterationScope.containingIndices` and constructing range target
- `every` scope modifier stage is implemented by grabbing the `iterationScope.targets` and just doing an intersection internally if `hasExplicitRange`
- `ordinal` scope modifier stage is implemented by calling the `every` stage, like we do today.  I think it is actually good that there is a hard dependency here, because we want `every` and `ordinal` to behave identically
- `relative` scope modifier stage is implemented by looking at containing scope and list of targets and doing its magic from there


Note that the only place that we ever refer to `hasExplicitRange` is in the `every` scope modifier stage

wdyt?

pokey:
One other question is how we handle "every round". I guess maybe we just have it return an empty range for getIterationRange? I'm not sure we want to do any expansion

Then for getIterationScope, if any pairs start or end inside the iteration range, it returns all top-level pairs that it it intersects with. Top-level here means not contained by any other pair intersecting with iteration range.

If no pairs start or end inside the iteration range, we return the smallest pair containing the iteration range

So I think this works within the api I defined above; just wanted to write it out to see

pokey:
For functions, we expand to class in getIterationRange

In getScopesInRange, I think we just check that the range is just one class, and return its functions if so? Otherwise error? Idk

src/core/tokenizer.ts
AndreasArvidsson: @pokey Any ideas?

pokey: What happens if you don't do that?

AndreasArvidsson: It will fail to match in some of the test cases and we get no identifiers back causing the entire test to fail.

pokey:
Ok what should the following do?

```
aaa bbb|. ccc
```

Cursor is the `|`; saying `"two tokens"`?  Should it refer to `bbb.` or `. ccc`?  I think probably the consistent thing is `bbb.`, because it's two tokens starting from `"token"`.  One could also argue it should always just look to the right, though that breaks down when the cursor is in the middle of a token, which should obviously be that token and the following.

Unfortunately, I don't think it's as simple as `containing` and then adding as many scopes as necessary.  For example:

```
{
    aaa {bbb} ccc {ddd} eee
    {fff}
}
```

In this case, if the user says `"three curlies line air"`,  I think it should be `{bbb}` past `{fff}`, but if we treat `"<number> <scope>s"` as first expanding to containing, it will expand to the big curly containing everything and then have nothing to add

So my thinking is as follows:

If the input target is empty, then index 0 should refer to containing scope.  If the input target is not empty, then index 0 should refer to all intersecting scopes.  Wdyt?

pokey:
@AndreasArvidsson fyi https://github.com/cursorless-dev/cursorless/pull/1031/commits/969bc720e98b89a7526656622b40375285b9c5f7.  I accidentally imported the legacy v2 `ScopeType` in a couple places when I did auto-import.  Shall we make it a policy not to export any legacy types without a `vN` suffix?

pokey (commented):
Ok I just need to add some jsdocs / comments to `RelativeScopeStage` but otherwise I think this is good to go.  Feel free to have a peek before meet-up if you get a minute

src/test/suite/fixtures/recorded/relativeScopes/clearTwoTokenBackward2.yml
```diff
@@ -18,9 +18,5 @@ initialState:
     - anchor: {line: 0, character: 8}
       active: {line: 0, character: 8}
   marks: {}
-finalState:
-  documentContents: "  ccc"
-  selections:
-    - anchor: {line: 0, character: 0}
-      active: {line: 0, character: 0}
 fullTargets: [{type: primitive, mark: {type: cursor}, modifiers: [{type: relativeScope, scopeType: {type: token}, offset: 0, length: 2, direction: backward}]}]
+thrownError: {name: NoContainingScopeError}
```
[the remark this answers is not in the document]
pokey: Update from meet-up: agreed

src/processTargets/modifiers/scopeHandlers/LineScopeHandler.ts
```diff
@@ -0,0 +1,111 @@
...
+import type { IterationScope, TargetScope } from "./scope.types";
+import type { ScopeHandler } from "./scopeHandler.types";
+
+export default class LineScopeHandler implements ScopeHandler {
+  public readonly scopeType: ScopeType = { type: "line" };
+  public readonly iterationScopeType: ScopeType = { type: "document" };
+
+  constructor(_scopeType: ScopeType, _languageId: string) {
+    // empty
+  }
```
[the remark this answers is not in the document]
pokey: Decision: Make them optional

pokey:
fwiw this is what it looks like to add a scope type now: https://github.com/cursorless-dev/cursorless/pull/1031/commits/08188162a8e5b40c96dbc3c55ebb49008b606806 cc/ @AndreasArvidsson

pokey (approved):
Ok I think it's good to go!

pokey:
Filed #1052 to track migrating remaining scope types

PR #1062, opened 2022-10-22
Support matching pair inside jsx text nodes

### LINKED ISSUE #1049: Matching pair inside jsx strings not matchable
`"take pair"` inside `(category.count)` will select the entire return value. 

The reason for this is that tree sitter gives us this node:
type: `jsx_text`
text:  `" (category.count) "`

```tsx
return (
  <div>
    (category.count)
  </div>
);
```

AndreasArvidsson:
Fixes #1049

## Checklist


pokey (approved):
👍👍

PR #1063, opened 2022-10-22
Support abstract methods

### LINKED ISSUE #1042: Typescript function definition doesn't work for abstract functions
In the following code:

```typescript
abstract class Foo {
  protected abstract bar(): void;
}
```

The line containing `bar` is not considered a function by `"funk"`.

AndreasArvidsson:
Fixes #1042

## Checklist


src/test/suite/fixtures/recorded/languages/typescript/clearFunk3.yml
```diff
@@ -0,0 +1,29 @@
...
+        protected abstract bar(): void;
+      }
+  selections:
+    - anchor: {line: 1, character: 26}
+      active: {line: 1, character: 26}
+  marks: {}
+finalState:
+  documentContents: |-
+    abstract class Foo {
+        ;
```
pokey, line 24: Shall we use trailing `;` matcher here as well?

pokey (commented):
looks good with minor comment

src/test/suite/fixtures/recorded/languages/typescript/clearFunk3.yml
AndreasArvidsson: sure

src/test/suite/fixtures/recorded/languages/typescript/clearFunk3.yml
```diff
@@ -0,0 +1,29 @@
...
+        protected abstract bar(): void;
+    }
+  selections:
+    - anchor: {line: 1, character: 26}
+      active: {line: 1, character: 26}
+  marks: {}
+finalState:
+  documentContents: |-
+    abstract class Foo {
+
```
pokey, line 24: It's more consistent with our other statement code to always remove the `;`

pokey (approved):
Hmm I tried this one out with "chuck", and I'm not sure that trailing matcher is what we want here.  It is super aggressive about whitespace, removing until the next nonwhitespace char no matter how far down it is.  That ends up pulling the class's final `}` up, or merging blocks, depending what follows the function.  I also don't think it's consistent with our other statement code to leave the `;` for clear

I made some changes and pushed.  Lmk what you think

I think the same statement applies to #1065, so if you like this one let's go ahead and use the new `extendForwardPastOptional` there as well once this one merges

Issue #1047, opened 2022-10-20
Create a smarter iteration scope for "item"


pokey:
Today, the parse tree code can takeover "item" if it wants, but can't just indicate to `ItemStage` the iteration scope and let it do the rest.  We'd like to enable that behaviour, so we can have uniformity of implementation of `"item"` while still supporting cases where there is no surrounding pair, eg the items `foo` and `bar` in `var foo, bar = ...`.  

## Approach

It will be a cascading scope type (#1045) combining the following:

One of the constituent scope types will be just be surrounding pair, but where we ask pair for `inside` before returning it in `getTarget()`; domain is just the surrounding pair

The other constituent scope type is a tree-sitter syntactic scope handler with a tree-sitter pattern that languages can define however they want.  For example, Typescript could define it to be `foo, bar` in `var foo, bar = ...`

See also #1046, as this approach is very similar

AndreasArvidsson:
We do support language specific implementations for collection item using a oneOf scope handler

PR #1065, opened 2022-10-22
Supports states scope for interface properties

### LINKED ISSUE #43: Support type / interface fields in typescript

AndreasArvidsson:
Fixes #43

## Checklist

### DIFF
--- a/src/languages/typescript.ts
+++ b/src/languages/typescript.ts
@@ -199,7 +199,9 @@ const nodeMatchers: Partial<
       childRangeSelector(["arguments"], [])
     )
   ),
-  statement: STATEMENT_TYPES.map((type) => `export_statement?.${type}`),
+  statement: STATEMENT_TYPES.map((type) => `export_statement?.${type}`).concat([
+    "property_signature",
+  ]),
   condition: cascadingMatcher(
     patternMatcher("ternary_expression[condition]"),
     conditionMatcher(
--- a/src/test/suite/fixtures/recorded/languages/typescript/clearState.yml
+++ b/src/test/suite/fixtures/recorded/languages/typescript/clearState.yml
@@ -1,0 +1,31 @@
+languageId: typescript
+command:
+  spokenForm: clear state
+  version: 3
+  targets:
+    - type: primitive
+      modifiers:
+        - type: containingScope
+          scopeType: {type: statement}
+  usePrePhraseSnapshot: true
+  action: {name: clearAndSetSelection}
+initialState:
+  documentContents: |-
+    interface Foo {
+      bar: Bar;
+      baz: Baz;
+    }
+  selections:
+    - anchor: {line: 1, character: 2}
+      active: {line: 1, character: 2}
+  marks: {}
+finalState:
+  documentContents: |-
+    interface Foo {
+      ;
+      baz: Baz;
+    }
+  selections:
+    - anchor: {line: 1, character: 2}
+      active: {line: 1, character: 2}
+fullTargets: [{type: primitive, mark: {type: cursor}, modifiers: [{type: containingScope, scopeType: {type: statement}}]}]


src/test/suite/fixtures/recorded/languages/typescript/clearState.yml
```diff
@@ -0,0 +1,31 @@
...
+      baz: Baz;
+    }
+  selections:
+    - anchor: {line: 1, character: 2}
+      active: {line: 1, character: 2}
+  marks: {}
+finalState:
+  documentContents: |-
+    interface Foo {
+      ;
```
pokey, line 25: Could go either way on keeping this semicolon for "change", but it should definitely go for "chuck"

AndreasArvidsson: That makes it a bit more complicated than just treating it like any other statement but I do agree

AndreasArvidsson: done

auscompgeek:
We should add tests with regular object types too.

pokey:
> We should add tests with regular object types too.

Yeah, probably class fields as well

AndreasArvidsson:
> > We should add tests with regular object types too.
> 
> Yeah, probably class fields as well

Maybe? Not sure.

pokey:
Why not?

AndreasArvidsson:
> Why not?

Probably because I never felt a need for it. On the other hand I guess it wouldn't hurt.

src/languages/typescript.ts
```diff
@@ -199,7 +199,12 @@ const nodeMatchers: Partial<
       childRangeSelector(["arguments"], [])
     )
   ),
-  statement: STATEMENT_TYPES.map((type) => `export_statement?.${type}`),
+  statement: cascadingMatcher(
+    trailingMatcher(["property_signature", "public_field_definition"], [";"]),
```
pokey, line 203: See https://github.com/cursorless-dev/cursorless/pull/1063#pullrequestreview-1152043776

pokey (commented):
Looks good but see comment below, and I think we're still missing the test that @auscompgeek was asking for:

```typescript
type foo = {
   bar: string;
}
```

Saying "clear state bat"

AndreasArvidsson:
> Looks good but see comment below, and I think we're still missing the test that @auscompgeek was asking for:
> 
> ```ts
> type foo = {
>    bar: string;
> }
> ```
> 
> Saying "clear state bat"

All ready added.
https://github.com/cursorless-dev/cursorless/pull/1065/files#diff-6b323569c0a7d841d62e954a6a5eaf111d4c3405b27113856d75feaeb4f233bb

auscompgeek:
That's an interface rather than a simple object type 😅

Should also consider what happens when the type is inline in a variable declaration, e.g.
```ts
let foo: { a: string };
let bar: {
  a: string,
  b: string,
};

auscompgeek:
My 2c: it'd likely be more useful if `state` _didn't_ target properties in an object type. But then we need to make sure `item` works in object types when the properties are separated by semicolons.

AndreasArvidsson:
Inline types works fine. Added test for that.

AndreasArvidsson:
I would also prefer to use item in this case. We can do that after we have fixed #1047

pokey:
- [ ] Let's add support for referring to abstract function declarations as "state"

pokey:
@auscompgeek When I try your comma example, prettier changes the commas to semicolons

auscompgeek:
> prettier changes the commas to semicolons

Heh. At work we have [dprint](https://dprint.dev) configured to use commas in object types.

pokey:
@AndreasArvidsson won't "item" just work out of the box when there are commas?  I'd be inclined to just add a couple tests for that and also support "state" for now, and deprecate "state" for inlines if it becomes too annoying.  My guess is it's not super common to get trapped there

AndreasArvidsson:
Yes item will work when you use commas. No extra tests are needed imho; we have plenty of plain text tests for item

AndreasArvidsson:
> * [ ]  Let's add support for referring to abstract function declarations as "state"

I just checked on main an functions in classes no matter if they are abstract or not cannot be referenced as states.

pokey (approved):

pokey:
> No extra tests are needed imho; we have plenty of plain text tests for item.

I went ahead and added these tests.  Just wanted to have a test in case Typescript tries to override `collectionItem` and drops these.  Maybe a bit remote?

Either way I think it's good to go

PR #1067, opened 2022-10-22
Moved global state into ide definition

### LINKED ISSUE #1066: User global state makes test fail
Tests like `takePastStartOfToken` that tests the `deprecatedPositionInference` warning relies on user global state. If the user has set 
`hideInferenceWarning` those tests will fail. The solution is to mock global state during testing.

AndreasArvidsson:
Fixes #1066

## Checklist


auscompgeek (commented):
The set of state keys we need is static, right? Could the interface just be this:
```ts
interface State {
  fontRatios: {
    widthRatio: number;
    heightRatio: number;
    fontFamily: string;
  } | undefined;

  hideInferenceWarning: boolean;
}
```

pokey:
yeah @auscompgeek I was thinking something along these lines as well, though I'd be tempted to have the knowledge of which keys need to be globally synced incorporated in some way as well, rather than hardcoded into VSCode impl

pokey:
@auscompgeek the only issue with an interface as simple as that is that VSCode makes the setter async, which I don't think your version allows

auscompgeek:
Yeah, if awaiting on the setter is required then we'd need separate getter/setter methods. It doesn't look like we're awaiting any of those right now though, so we could probably get away with the simple interface for now.

pokey:
True, tho probably useful to be able to wait until set has happened. Also, if we want to let the ide handle the global state, we might need want it to be async as it will be going across the wire from cursorless lsp

pokey:
But even if we don't switch to that simplified interface we could still make the key strongly typed

AndreasArvidsson:
Making the key strongly typed is probably my preference.
Unsure we want to add a field for `widthRatio` and the up their hat parameters. They feel kinda vscode specific. Maybe they should just use the old implementation and not the ide abstraction?

AndreasArvidsson:
Could we just merged this one and then do a refactor of the interface in a follow up pr? Because at the moment I can't run tests locally.

pokey:
- [ ] as discussed, let's move default value to definition of state key rather than passing it in

Should prob do that for our settings code as well but out of scope

auscompgeek (commented):
Nice. The fake and VSCode impl just need the same dependent type treatment.

src/core/commandVersionUpgrades/canonicalizeAndValidateCommand.ts
```diff
@@ -128,11 +126,9 @@ export async function checkForOldInference(
...
 
   if (hasOldInference) {
-    const hideInferenceWarning =
-      graph.extensionContext.globalState.get<boolean>(
-        globalStateKeys.hideInferenceWarning,
-        false
-      );
+    const hideInferenceWarning = graph.ide.globalState.get(
+      "hideInferenceWarning"
+    );
```
pokey, line 131: I improved the type inference a bit, so you don't need to put the `<boolean>` anymore.  I think it's now pretty slick

src/core/commandVersionUpgrades/canonicalizeAndValidateCommand.ts
```diff
@@ -142,10 +138,7 @@ export async function checkForOldInference(
       );
 
       if (pressed) {
-        graph.extensionContext.globalState.update(
-          globalStateKeys.hideInferenceWarning,
-          true
-        );
+        graph.ide.globalState.set("hideInferenceWarning", true);
```
pokey, line 141: Note that the value is now strongly typed, so you have to put a boolean here for this key

pokey (approved):
Made some tweaks to strengthen the type inference.  Feel free to merge if it looks good to you

src/core/commandVersionUpgrades/canonicalizeAndValidateCommand.ts
AndreasArvidsson: Oh that is neat!

PR #1069, opened 2022-10-23
One of compound scope type

### LINKED ISSUE #1045: Add new `oneOf` compound scope type
Not sure if `oneOf` is the right name, but we should support a way to compose scope types like the following:

```typescript
interface OneOfScopeType {
    type: "oneOf";
    scopeTypes: ScopeType[]
}
```

Then we can construct a `OneOfScopeHandler` that will behave the same as a hierarchical scope where each scope can be any of `scopeTypes`.  So basically:

- `getScopesTouchingPosition`: return smallest scopes of any of `scopeTypes` touching position
- `getScopesOverlappingRange`: call `getScopesOverlappingRange` on all `scopeTypes` and then remove any scopes that are contained by another scope
- `getScopeRelativeToPosition`: pick first scope of any of the `scopeTypes`, then skip it for offset > 1, etc
- `iterationScope`: apply `oneOf` to the iteration scopes of `scopeTypes`

## Use cases

- #1046
- #1047 
- defining "name" as a union of "class name", "function name", etc
- to define iteration scope of surrounding pair to be union of "inside" and "file"
- users could map a custom scope type Talon side that constructs an `oneOf` scope type to enable them to construct their own unions of other scope types
- we could maybe map a spoken form `"<scope> or <scope>"` to construct these on the fly, eg `"take next funk or class"`

## Possible names

- `cascading`
- `union`
- `oneOf`
- `anyOf`
- `or`

AndreasArvidsson:
Fixes #1045

## Checklist


AndreasArvidsson:
@pokey Added this draft so we can discuss it on the meet up

pokey (commented):
Ok I rebased onto #1075 and tweaked your implementation.  Lmk what you think

Not totally sold on the name `oneOf`; let's discuss at meet-up

src/processTargets/modifiers/scopeHandlers/OneOfScopeHandler.ts
```diff
@@ -0,0 +1,113 @@
...
+    return keepOnlyBottomLevelScopes(
+      this.scopeHandlers.flatMap((scopeHandler) =>
+        scopeHandler.getScopesTouchingPosition(editor, position, ancestorIndex)
+      )
+    );
+  }
+
+  /** Return all scopes overlapping range not contained by another scope */
+  getScopesOverlappingRange(editor: TextEditor, range: Range): TargetScope[] {
+    return keepOnlyTopLevelScopes(
```
pokey, line 55: I can't decide if this is right. I think we may want to make it so that intersecting scopes only count if they start or end within the range. Then I think this one would be right. Let's discuss tomorrow or Tues

src/processTargets/modifiers/scopeHandlers/OneOfScopeHandler.ts
```diff
@@ -0,0 +1,113 @@
...
+      return handler;
+    }
+  );
+
+  public iterationScopeType: OneOfScopeType = {
+    type: "oneOf",
+    scopeTypes: this.scopeHandlers.map(
+      ({ iterationScopeType }) => iterationScopeType
+    ),
+  };
```
pokey, line 28: I'm worried I introduced an infinite recursion here. Should be able to work around it by making it lazy if necessary

pokey (approved):

pokey (approved):
I guess let's merge this in for now

PR #1073, opened 2022-10-23
Migrating scope stages into handlers

### LINKED ISSUE #1054: Migrate `document` to use a scope handler

AndreasArvidsson:
Fixes #1054

## Checklist


src/processTargets/modifiers/scopeHandlers/LineScopeHandler.ts
```diff
@@ -12,7 +12,7 @@ export default class LineScopeHandler implements ScopeHandler {
   public readonly iterationScopeType = { type: "document" } as const;
 
   constructor(
-    public readonly scopeType: { type: "line" },
+    public readonly scopeType: ScopeType,
```
pokey, line 15: I'm thinking of using this type information for generating our docs.  Do you mind if we leave it?

AndreasArvidsson: Why not turn it into a proper type instead? I mean in the types file we have made them all `SimpleScopeType` if we actually want to have it more fine grain I think we should do it at that end so we don't have to do casts when calling the constructor.
`public readonly scopeType: LineScopeType`

pokey: Why would giving it a name allow us to avoid casts?

pokey:
Folded into #1075

Issue #640, opened 2022-04-20
Sort Typescript imports using `prettier-plugin-sort-imports`


pokey:
See https://github.com/trivago/prettier-plugin-sort-imports.  Would just need to add it as a prettier plugin

phillco:
From the meetup: @pokey's main concern with doing this now is version skew between pre-commit's `eslint` and what's in `package.json`, as they should match, but pre-commit.ci only bumps the pre-commit configuration. Semi relatedly, we're also not thrilled with pre-commit.ci's lack of respect for semver.

Ideally https://github.com/renovatebot/renovate would be used to bump both locations together, but we're not quite ready to pull the trigger on that, because it looks like it wants to bump everything, and things like the cheatsheet and website do not have good test coverage currently.

In the short term, we agreed that if we could add eslint to pre-commit config but tell pre-commit.ci not to bump it (and thus require that people bump both locations together manually), that would be fine for now to unblock this to get the benefits of automatic fixing on PRs / locally, until we set up something like renovate. It will be a bit more annoying when we go to bump `eslint`, but hopefully we run fix way more often than we bump `eslint`.

auscompgeek:
> In the short term, we agreed that if we could add eslint to pre-commit config but tell pre-commit.ci not to bump it (and thus require that people bump both locations together manually), that would be fine for now to unblock this. This looks pretty easy to do.
>
> That way, we get the benefits of automatic fixing on PRs

Isn't that `ci#skip` configuration to skip the hook entirely on pre-commit.ci, hence we _won't_ get autofixing on PRs?

phillco:
Oops, you're right. It doesn't look like they separate skipping fixups from skipping version bumping for a particular hook. I guess we could maybe move to bumping versions of everything manually?

auscompgeek:
In case we're unhappy with prettier, and alphabetical sorting works for us, [dprint]'s [typescript plugin] will sort imports alphabetically by default.

[dprint]: https://dprint.dev
[typescript plugin]: https://dprint.dev/plugins/typescript/

PR #1076, opened 2022-10-23
`eslint`: make `--fix` remove unused imports

phillco:
I noticed that `eslint --fix` didn't remove unused imports and `eslint src --ext ts` complained with a warning about an existing unused one:

```
$ eslint src --ext ts
.../src/core/commandVersionUpgrades/canonicalizeAndValidateCommand.ts
  1:10  warning  'window' is defined but never used. Allowed unused vars must match /^_/u  @typescript-eslint/no-unused-vars
```

I imagine this is because most people are fixing this up using the Visual Studio Code action, but it's nice to have everything working from the command line. Also JetBrains will only do it if `eslint --fix` does, hence my motivation :)

This adds the configuration such that `eslint src --ext ts --fix` removed unused imports, and fixes the existing unused imports in the codebase. Let me know if I made a mistake.

## Checklist


src/typings/target.types.ts
```diff
@@ -5,17 +5,8 @@ import type { Range, Selection, TextEditor } from "vscode";
 // - https://github.com/microsoft/TypeScript/issues/43869
 // - https://github.com/microsoft/TypeScript/issues/43950
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-import type ModifyIfUntypedStage from "../processTargets/modifiers/ModifyIfUntypedStage";
```
pokey, line 8: These are here so that links get properly resolved in our api docs.  Maybe we need to add `eslint/no-unused-imports` to the ignore on the previous line?  See the links in comments leading up to this line

pokey (commented):
I am supportive of this direction, but I think we need to tweak the ignore comments for most of these unused imports; see inline comment.

Also see https://github.com/cursorless-dev/cursorless/issues/640; would be good to get this into pre-commit if easy

package.json
```diff
@@ -586,6 +586,7 @@
 		"esbuild": "^0.11.12",
 		"eslint": "^8.13.0",
 		"eslint-config-prettier": "^8.5.0",
+		"eslint-plugin-unused-imports": "^2.0.0",
```
auscompgeek, line 589: Curious, why this instead of https://github.com/import-js/eslint-plugin-import?

src/typings/target.types.ts
phillco: Guessing this applies to anywhere you've tagged them with `// eslint-disable-next-line @typescript-eslint/no-unused-vars`? Didn't notice that before :P

phillco:
Yeah, I definitely did take a look at #640 :)

Will update the tagging on the existing imports.

package.json
pokey: That one does look pretty cool.  Would be interesting to try that one instead and see what it complains about

src/typings/target.types.ts
phillco: Something something, Chesterton's Fence.

phillco:
I've tagged the existing unused imports with `unused-imports/no-unused-imports` and restored them, and updated the branch. Will take a look at https://github.com/import-js/eslint-plugin-import next to see if it's better, but otherwise this version is ready to go.

package.json
phillco: Does https://github.com/import-js/eslint-plugin-import actually fix unused imports? I've added it and it looks like it's linting a bunch of other stuff, but not that. Maybe I missed it.

phillco:
From the meetup: we looked at https://github.com/import-js/eslint-plugin-import briefly but didn't see a rule for removing unused imports. [no-unused-modules](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-unused-modules.md) seems to be more about finding exports in your own modules that aren't used. @auscompgeek let me know if I missed it.

If there is, we can use it. Otherwise, this should be ready to merge and we can take a look at adding some of [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import)'s rules as a follow up change, as I imagine that would be valuable in their own right.

phillco:
@auscompgeek @pokey can we merge this?

package.json
pokey: I took a look and agree with your assessment @phillco.  I did notice that `eslint-plugin-import` has lots of cool rules, tho, so am definitely up for activating it in a follow-up.  In particular, the [`caseSensitive` option](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-unresolved.md#casesensitive) of `no-unresolved` would have saved @AndreasArvidsson some headache on one of his PRs

pokey: https://github.com/cursorless-dev/cursorless/issues/1146

pokey (approved):
👍👍

PR #1078, opened 2022-10-24
Add word separator setting

### LINKED ISSUE #999: Add user setting for tokeniser
The user should be able to override the values of https://github.com/cursorless-dev/cursorless/blob/2098df631cdb9fb597b0f8eb90f5bd210a47186f/src/core/tokenizer.types.ts#L3-L9 in their settings.  This would be similar to the way that languages override them, eg https://github.com/cursorless-dev/cursorless/blob/4906e2b24845a89557851ba0cca8136d12b8d32e/src/core/languageTokenizers.ts#L9

Note that users will be able to use language-specific overrides just via [VSCode language settings](https://code.visualstudio.com/docs/getstarted/settings#_language-specific-editor-settings)

This setting could be used to accomplish https://github.com/cursorless-dev/cursorless/pull/986 on a per-user basis

AndreasArvidsson:
Fixes #999 

## Checklist

- [x] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet_html)


AndreasArvidsson:
@pokey Unsure what to do with all their language specific tests in our tokenizer?
https://github.com/pokey/cursorless-vscode/blob/8295c710966b11e3c56f23e52521aefb1738d9a2/src/test/suite/tokenizer.test.ts#L103-L109

Should we just remove these?

src/extension.ts
```diff
@@ -42,6 +43,11 @@ export async function activate(context: vscode.ExtensionContext) {
   // TODO: Do this using the graph once we migrate its dependencies onto the graph
   new CommandRunner(graph, thatMark, sourceMark);
 
+  // TODO: Remove this once tokenizer has access to graph
+  if (isTesting()) {
```
AndreasArvidsson, line 47: This keeps backwards comparability with existing recorded tests. We could of course remove this and update existing tests if we so choose.

src/core/tokenizerConfiguration.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * TODO: This is just an ugly mock since that tokenizer doesn't have access to the graph/ide
+ * Remove this once https://github.com/cursorless-dev/cursorless/issues/785 is implemented
+ */
+
+import * as vscode from "vscode";
+
+const defaultGetWordSeparators = (languageId: string) => {
+  return vscode.workspace
+    .getConfiguration("cursorless", { languageId })
```
pokey, line 10: Should also add a comment to https://github.com/cursorless-dev/cursorless/blob/c914d477c9624c498a47c964088b34e484eac494/src/core/Decorations.ts#L58 indicating that we can't use a fine-grained settings listener until we move to graph

src/core/tokenizer.ts
```diff
@@ -91,12 +91,13 @@ function generateMatcher(
   };
 }
 
-const matchers = new Map<string[], Matcher>();
+const matcherCache: Record<string, Matcher> = {};
 
 export function getMatcher(languageId: string): Matcher {
   const wordSeparators = tokenizerConfiguration.getWordSeparators(languageId);
+  const cacheKey = JSON.stringify(wordSeparators);
```
AndreasArvidsson, line 98: I don't think this is an improvement at all. You're just wasting resources with something that maps are super optimized to do. The amount of times this function gets called in in a single second is quite high.

pokey (commented):
Yikes the mocking hack here gets smeared all over the place.  I think it would be good to file an issue to track the proper fix, and then add comments with `#XXXX` of that issue number everywhere that needs to get cleaned up.  Otherwise I worry we'll miss one of these places

Otherwise looks good to me with a couple minor comments.  Also make sure to pull before fixing these, because I fixed your caching code.  Maps compare keys by equality so I don't think your cache was doing anything

I tested this one locally with language-specific overrides and no overrides and it works fine for me

cc/ @phillco fyi.  I think it's cleaner if by default there is one tokeniser for all languages.  I would recommend users stick to a single tokeniser for all languages, even if they decide that tokeniser should include `-` (as I imagine you will do).  But this PR does make it possible to use language-specific overrides if a user wants to

src/core/tokenizer.ts
pokey: See my comment; maps compare keys by equality

pokey: Feel free to switch it back to a map, but you need to stringify.  See eg https://stackoverflow.com/a/69219091

AndreasArvidsson:
> Yikes the mocking hack here gets smeared all over the place. I think it would be good to file an issue to track the proper fix, and then add comments with `#XXXX` of that issue number everywhere that needs to get cleaned up. Otherwise I worry we'll miss one of these places
> 
> Otherwise looks good to me with a couple minor comments. Also make sure to pull before fixing these, because I fixed your caching code. Maps compare keys by equality so I don't think your cache was doing anything
> 
> I tested this one locally with language-specific overrides and no overrides and it works fine for me
> 
> cc/ @phillco fyi. I think it's simpler if by default there is one tokeniser for all languages. I would recommend users stick to a single tokeniser for all languages, even if they decide that tokeniser should include `-` (as I imagine you will do). But this PR does make it possible to use language-specific overrides if a user wants to

There is already an issue and it is referenced from the tokenizer configuration file.

The cache code was working perfectly when I tested it

pokey:
> The cache code was working perfectly when I tested it

How do you know?

AndreasArvidsson:
> > The cache code was working perfectly when I tested it
> 
> How do you know?

Because I put breakpoints and prints and tested it?

pokey:
> There is already an issue and it is referenced from the tokenizer configuration file.

We should have a more specific issue, and link to it everywhere that needs to change; there's like 4 or 5 places; best if we can just grep for issue number

src/core/tokenizer.ts
AndreasArvidsson: I know

AndreasArvidsson:
> > There is already an issue and it is referenced from the tokenizer configuration file.
> 
> We should have a more specific issue, and link to it everywhere that needs to change; there's like 4 or 5 places; best if we can just grep for issue number



> > There is already an issue and it is referenced from the tokenizer configuration file.
> 
> We should have a more specific issue, and link to it everywhere that needs to change; there's like 4 or 5 places; best if we can just grep for issue number

There is one implementation that is going to be removed and three references that will give errors once the source is gone

pokey:
> There is one implementation that is going to be removed and three references that will give errors once the source is gone

What about listening for settings changes?  What if we want to make decorations listener more fine-grained?

pokey:
> Because I put breakpoints and prints and tested it?

Can you explain why it was working?  In my tests, maps compare keys by equality

image

AndreasArvidsson:
> > Because I put breakpoints and prints and tested it?
> 
> Can you explain why it was working? In my tests, maps compare keys by equality
> 
> image

Yes they do but vscode takes care of this internally in its configuration. Every time the configuration changes there is a new object. I have tested always I can it works fine.

pokey:
Yes but if it returns a new object every time, then you will never get a cache hit, so why are you bothering with a cache at all?

AndreasArvidsson:
> Yes but if it returns a new object every time, then you will never get a cache hit, so why are you bothering with a cache at all?

It doesn't return in new object every time. Only when the configuration actually changes.

pokey:
But what if you have multiple languages that use the same separator setting?  I just feel that the cache should be based on the actual list rather than equality, as that is what determines what the regex should be.  Feel free to switch it back to a map if the performance bothers you but I don't like the idea of indexing the map with something that checks strict equality

pokey:
I can roll it back to a map if you'd like, as I made that change, but I'm still not convinced that we should pass the list directly in as the key

AndreasArvidsson:
> I'm not sure we want to rely on that. Ie what happens when we switch to lsp?

If it makes you that anxious I have switched to a string :D

src/core/tokenizer.ts
```diff
@@ -109,24 +91,25 @@ function generateMatcher(
...
-  );
+const matchers = new Map<string, Matcher>();
 
 export function getMatcher(languageId: string): Matcher {
-  return (
-    tokenMatchersForLanguage[languageId as SupportedLanguageId] ??
-    defaultMatcher
-  );
+  const wordSeparators = tokenizerConfiguration.getWordSeparators(languageId);
+  const key = wordSeparators.join();
```
pokey, line 98: This is going to result in subtle bugs in certain extremely rare cases.  Eg `["--", "_"]` and `["-", "-_"]` will map to the same key.  I know it's extremely unlikely that this will happen, but do we really want to knowingly introduce a subtle bug that will probably be really confusing for that one user that hits it?

pokey: Note that if you're worried about performance from the json stringify, we can do more effective caching once we switch to graph impl, as then we can just cache the language's regex and listen for settings changes directly to invalidate it

AndreasArvidsson: No it still joins with a `,`

pokey: That still has potential for bugs, though they're even rarer.  How about

```suggestion
  const key = wordSeparators.join("\u0000");
```

AndreasArvidsson: I guess it's fine but the chances that someone's going to have a comma in the middle of an identifier must be close to zero.

pokey (approved):
LGTM!

@phillco once this PR merges, just add

```json
   "cursorless.wordSeparators": ["_", "-"],
```

To your `.vscode/settings.json` and you'll have dash-separated tokens as far as the eye can see

phillco:
Cool! Imagine I could use a language-specific setting if I did want to customize by language?

pokey:
> Cool! Imagine I could use a language-specific setting if I did want to customize by language?

Yep!  Would be 

```json
  "[css]": {
    "cursorless.wordSeparators": ["_", "-"]
  },
```

instead.  That info will be in the docs under customization once this ships

Issue #1072, opened 2022-10-23
Make scope type `character` prefer iteration scope `identifier` over `token`


AndreasArvidsson:
Since the latest change where scope type `identifier` was created I feel we have a regression in character. 
eg given the text `hello|.` `"clear last car"` will now remove `.` instead of `o`

This could probably be solved using a compound scope type

pokey:
I tried this initially.  The problem is that the way the hierarchy of iteration scopes works, that would cause "every char block" to ignore non-alphanumeric characters.  In addition, it felt a bit confusing to have this kind of precedence logic.  I'd argue for just using "last char identifier" when you want this behaviour

pokey:
But maybe we should discuss at meet-up

pokey:
I think just getting in the habit of using "identifier" is not necessarily a bad thing.  Less tricky little rules to learn

AndreasArvidsson:
We can definitely have a discussion about it. Personally I don't care if character only works inside an identifier. But I think we can be smarter than that if we want to. For example empty selection could have preference of identifier.

pokey:
Yes but I think those sorts of little tricky rules are something we should be getting rid of.  I'm not strictly opposed to using a compound scope type, though I don't think we can use `oneOf`.  It would need to be something like `fallback` where it has a preference.  `oneOf` has no preference.  Fwiw `fallback` is what we'd use to replace the cascading scopes we use for our `insertionScopeType` that allows trying function, then statement, then line.

AndreasArvidsson:
No it would be another compound scope.

PR #1087, opened 2022-10-24
Support whitespace characters for `character`

### LINKED ISSUE #1071: Create a scope type that allows white space characters
There is currently no way to target individual spaces. For example if I wanted to remove one of the extra spaces in the following:

```
foo  bar
```

I'd like to be able to say something like `"chuck first char leading bat"`, but `"char"` ignores whitespace chars.

pokey:
The `character` scope type now supports any characters, including newlines, but has preferences for identifiers, then non-whitespace, then non-newline

Fixes #1071
Fixes https://github.com/cursorless-dev/cursorless/issues/1072

This one was also a nice workout for our scope handlers

## Checklist

- [ ] Add test with a sequence of characters with lots of modifiers
- [ ] Add test for "clear char" with `hello|.`


src/typings/targetDescriptor.types.ts
```diff
@@ -128,6 +128,7 @@ export type SimpleScopeTypeType =
   | "paragraph"
   | "document"
   | "character"
+  | "nonWhitespaceCharacter"
```
AndreasArvidsson, line 131: Since we have the filter stages I'm not sure we actually need this scope type.

src/test/suite/fixtures/recorded/selectionTypes/clearCar5.yml
```diff
@@ -16,8 +16,8 @@ initialState:
       active: {line: 0, character: 1}
   marks: {}
 finalState:
-  documentContents: " "
+  documentContents: a
```
AndreasArvidsson, line 19: I would like it to prefer non white space characters

src/processTargets/modifiers/scopeHandlers/CharacterScopeHandler.ts
```diff
@@ -1,21 +1,48 @@
...
+import { getDocumentRange } from "../../../util/rangeUtils";
+import { generateMatchesInRange, testRegex } from "../../../util/regex";
 import { PlainTarget } from "../../targets";
+import BaseScopeHandler from "./BaseScopeHandler";
+import { expandPosition, expandRange } from "./expandRange";
 import type { TargetScope } from "./scope.types";
+import { ScopeIteratorRequirements } from "./scopeHandler.types";
 
-export default class CharacterScopeHandler extends NestedScopeHandler {
+const EXPANSION_CHARACTERS = 4;
```
AndreasArvidsson, line 13: Do we really need this? Could we not just yield a line at the time?

AndreasArvidsson: There is range including line break on each line that we probably could use. This would also help with breaking in the middle of a diacritic sends these can't pass lines breaks.

pokey: Yeah maybe that is simplest; unfortunate because this was a lot of work 😅. Might be useful to retain some of that machinery for more complex regex-based scopes. Tbh might be useful for expansion when doing pairs

I'll have a look at switching this to use line with breaks and migrating the range generator to a utility

pokey: How important do you think it is that "char" includes newlines?  I tried to impl your solution by just using a nested scope handler with `line` as the search scope, but because the domain of `line` doesn't include the line break, the nested scope handler will skip the line if `position` is at the end of the line, because it thinks there's nothing left in the domain, even though there is the line break

If we just bail on supporting newlines as "char", then we can just use a nested scope handler with search scope of `line`.  I'd be inclined to do this until we have a use case for using "char" to target line

Alternately, we can duplicate the logic from `LineScopeHandler` https://github.com/cursorless-dev/cursorless/blob/c9e06d765cdb74c9d137d07133904711f0955a1f/src/processTargets/modifiers/scopeHandlers/LineScopeHandler.ts#L21-L30 Not too much code, but not sure it's worth it 🤷‍♂️

wdyt?

AndreasArvidsson: I don't think it's important at all. It might actually be confusing to include line break.

AndreasArvidsson (approved):

Issue #1015, opened 2022-10-06
Filter / forgiving modifiers


pokey:
- [ ] Keep only cursors in scope, but don't expand ("only in funk")
   - eg is it weakly contained by visible characters? ("only chars")
- [x] Error ("funk") (default behaviour)
- [ ] Drop target ("only funk")
- [ ] Passthrough unmodified ("soft funk")

Wait for generic scope handlers from #621

josharian:
Cross-linking: This came up as useful in combination with https://github.com/cursorless-dev/cursorless/issues/1631.

pokey:
I do wonder if the error boundary approach is too fancy. Having modifiers which have long-range impacts on other modifiers feels slightly strange. Also, how does the data flow work? Does it flow out of the original modifier or the error boundary? And how would it interact with https://github.com/cursorless-dev/cursorless/issues/756? and https://github.com/cursorless-dev/cursorless/issues/2000?

The advantage of the original proposal is that the semantics are very clear, and no special sauce needs to appear in the pipeline. In addition, for keyboard, the pipeline is not used at all, so they wouldn't be able to use these error boundaries

cc/ @AndreasArvidsson @josharian

AndreasArvidsson:
I was just thinking that if one stage of the pipeline throws an error we check if the next stage can handle that. Nothing fancy. How this interact with future features and specifically the keyboard I have no idea.

pokey:
I thought you were planning to walk multiple stages, as you suggested it was an alternative to having to think bout whether we swallow multiple modifiers

AndreasArvidsson:
That is a possibility, but not necessary.

PR #1094, opened 2022-10-26
Allow using "this" to refer to current token with empty selection

### LINKED ISSUE #1043: Allow using "this" to refer to current token with empty selection
## Use cases
"take this" / "chuck this" / "bring this to bat" with cursor at end of token.  Note that we don't want to auto-expand for implicit destination for a bring command

## Approach
After final modifier stages, but right before running action, run containing token stage if target is empty, weak, and has implicit range

Note that we're rely on the fact that implicit (ie unspoken) target for "bring" automatically has an explicit range to avoid expanding when you say eg "bring air" with an empty cursor

### LINKED ISSUE #1141: Migrate `isImplicit` boolean into its own target type
Instead of
```ts
{
  type: "primitive";
  mark?: Mark;
  modifiers?: Modifier[];
  isImplicit?: boolean;
}
```

Were `isImplicit: true` can't be combined with `mark` or `modifiers` I'd suggest:
```ts
{
  type: "implicit"
}
```

AndreasArvidsson:
- Fixes #1043
- Fixes #1141 
- Depends on https://github.com/cursorless-dev/cursorless/pull/1187

## Checklist


src/processTargets/processTargets.ts
```diff
@@ -198,6 +208,10 @@ function processPrimitiveTarget(
     ...context.actionFinalStages,
   ];
 
+  if (!isRangeTarget) {
+    modifierStages.push(new ModifyUnTypedEmptyToTokenStage());
+  }
```
pokey, line 213: How come you skip this on range targets?

src/processTargets/modifiers/ModifyIfUntypedStage.ts
```diff
@@ -1,38 +1,77 @@
...
 import { ProcessedTargetsContext } from "../../typings/Types";
 import getModifierStage from "../getModifierStage";
 import { ModifierStage } from "../PipelineStages.types";
 
-/**
- * Runs {@link ModifyIfUntypedModifier.modifier} if the target has no explicit
- * scope type, ie if {@link Target.hasExplicitScopeType} is `false`.
- */
-export default class ModifyIfUntypedStage implements ModifierStage {
+abstract class ModifyIfBaseStage implements ModifierStage {
```
pokey, line 10: I'd call this `ConditionalModifierBaseStage`

src/processTargets/modifiers/ModifyIfUntypedStage.ts
```diff
@@ -1,38 +1,77 @@
...
   private get nestedStage() {
     if (this.nestedStage_ == null) {
-      this.nestedStage_ = getModifierStage(this.modifier.modifier);
+      this.nestedStage_ = getModifierStage(this.modifier);
     }
 
     return this.nestedStage_;
   }
+
+  protected abstract modifyIf(target: Target): boolean;
```
pokey, line 42: I'd call this `shouldModify`

src/processTargets/modifiers/ModifyIfUntypedStage.ts
```diff
@@ -1,38 +1,77 @@
...
-/**
- * Runs {@link ModifyIfUntypedModifier.modifier} if the target has no explicit
- * scope type, ie if {@link Target.hasExplicitScopeType} is `false`.
- */
-export default class ModifyIfUntypedStage implements ModifierStage {
+abstract class ModifyIfBaseStage implements ModifierStage {
   private nestedStage_?: ModifierStage;
 
-  constructor(private modifier: ModifyIfUntypedModifier) {}
+  constructor(private modifier: Modifier, private suppressErrors?: boolean) {}
```
pokey, line 13: Do you ever use `suppressErrors`?

src/test/suite/fixtures/recorded/hatTokenMap/bringHarpAndPointToThisAndStartOfWhaleTakeWhale.yml
```diff
@@ -35,10 +35,10 @@ initialState:
       start: {line: 0, character: 7}
       end: {line: 0, character: 12}
 finalState:
-  documentContents: hello. hello.world
+  documentContents: hello. .hello
```
pokey, line 38: This test needs to be re-recorded to use "just this" instead of "this"

src/test/suite/fixtures/recorded/hatTokenMap/bringPointAndHarpToThisAndStartOfWhaleTakeWhale.yml
```diff
@@ -35,10 +35,10 @@ initialState:
       start: {line: 0, character: 7}
       end: {line: 0, character: 12}
 finalState:
-  documentContents: hello. .helloworld
+  documentContents: hello. hello.
```
pokey, line 38: This one too

pokey (commented):
Looks good with some minor comments.  Also worth adding tests for:

- [x] "bring this to bat"
- [ ] "bring bat to this"
- [ ] "chuck this"

src/processTargets/processTargets.ts
AndreasArvidsson: Because we run this on both the primaries before the range is constructed. I don't think I want `take air past this` or `this slice air` to expand to token?

src/processTargets/modifiers/ModifyIfUntypedStage.ts
AndreasArvidsson: Yes the new modifier passes it in the constructor

pokey: Ah right I see now; that line is a bit hard to read with the inline obj

pokey, line 13: Also, I think I'd call the nested modifier `nestedModifier`, because usually `modifier` refers to the modifier for which we are constructing a stage

src/processTargets/processTargets.ts
pokey: Huh.  This one feels a bit inconsistent, but I guess I agree.  Are there tests for this case?

src/processTargets/modifiers/ModifyIfUntypedStage.ts
```diff
@@ -1,38 +1,77 @@
...
+  }
+}
+
+/**
+ * Tries to convert target into token if:
+ * {@link Target.hasExplicitScopeType} is `false` and
+ * {@link Target.hasExplicitRange} is `false` and
+ * {@link Target.contentRange.isEmpty} is `true`.
+ */
+export class ModifyUnTypedEmptyToTokenStage extends ModifyIfBaseStage {
```
pokey, line 65: Capitalisation here is inconsistent with `modifyIfUntyped`: `UnTyped` should be `Untyped`.  Actually this name is a bit wonky overall.  Maybe `ContainingTokenIfUntypedEmptyStage`?

src/processTargets/processTargets.ts
AndreasArvidsson: There is now :D

AndreasArvidsson:
@pokey Ready!

pokey:
Thinking about it more, I don't think we should have special cases for ranges.  It is confusing if "bring this to bat" behaves differently from "take this past bat".  Happy to discuss at meet-up if you prefer the special case for ranges

src/processTargets/modifiers/ConditionalModifierStages.ts
```diff
@@ -0,0 +1,80 @@
...
+  Modifier,
+  ModifyIfUntypedModifier,
+} from "../../typings/targetDescriptor.types";
+import { ProcessedTargetsContext } from "../../typings/Types";
+import getModifierStage from "../getModifierStage";
+import { ModifierStage } from "../PipelineStages.types";
+
+abstract class ConditionalModifierBaseStage implements ModifierStage {
+  private nestedStage_?: ModifierStage;
+  protected suppressErrors = false;
```
pokey, line 12: This feels orthogonal to conditional modifier.  Should really just use #1015 once we have that.  I'm ok with leaving this as is for now, as that machinery isn't there yet, but prob worth adding a checkbox to #1015 to use that machinery here

AndreasArvidsson:
> Thinking about it more, I don't think we should have special cases for ranges. It is confusing if "bring this to bat" behaves differently from "take this past bat". Happy to discuss at meet-up if you prefer the special case for ranges

I'm not sure either. Lets every discussion.

pokey:
Update based on discussion:

- [ ] No special cases for range targets
- [ ] Make sure that "take past air" uses implicit target for anchor
- [ ] Update tests, potentially splitting some of them into up to three tests:
   - [ ] using `"just"` to get old behaviour
   - [ ] using implicit target to get old behaviour
   - [ ] new behaviour

AndreasArvidsson:
@pokey ready!

src/processTargets/processTargets.ts
```diff
@@ -229,7 +231,13 @@ export function processModifierStages(
...
 
 function calcIsReversed(anchor: Target, active: Target) {
-  return anchor.contentRange.start.isAfter(active.contentRange.start);
+  if (anchor.contentRange.start.isAfter(active.contentRange.start)) {
+    return true;
+  }
+  if (anchor.contentRange.start.isBefore(active.contentRange.start)) {
+    return false;
+  }
+  return anchor.contentRange.end.isAfter(active.contentRange.end);
```
pokey, line 243: how come this had to change?

src/test/suite/fixtures/recorded/compoundTargets/clearJustThisPastEndOfToken.yml
```diff
@@ -0,0 +1,36 @@
...
+  marks: {}
+finalState:
+  documentContents: hello th
+  selections:
+    - anchor: {line: 0, character: 8}
+      active: {line: 0, character: 8}
+fullTargets: [{type: range, excludeAnchor: false, excludeActive: false, rangeType: continuous, anchor: {type: primitive, mark: {type: cursor}, modifiers: [{type: toRawSelection}]}, active: {type: primitive, mark: {type: cursor}, modifiers: [{type: containingScope, scopeType: {type: token}}], positionModifier: {type: position, position: end}}}]
+ide:
+  messages:
+    - {type: warning, id: deprecatedPositionInference}
```
pokey, line 36: how come we're using this deprecated command form?  presumably we should switch to "past end of its token"?

src/test/suite/fixtures/recorded/compoundTargets/clearJustThisPastStartOfToken.yml
```diff
@@ -0,0 +1,36 @@
...
+  marks: {}
+finalState:
+  documentContents: hello ere
+  selections:
+    - anchor: {line: 0, character: 6}
+      active: {line: 0, character: 6}
+fullTargets: [{type: range, excludeAnchor: false, excludeActive: false, rangeType: continuous, anchor: {type: primitive, mark: {type: cursor}, modifiers: [{type: toRawSelection}]}, active: {type: primitive, mark: {type: cursor}, modifiers: [{type: containingScope, scopeType: {type: token}}], positionModifier: {type: position, position: start}}}]
+ide:
+  messages:
+    - {type: warning, id: deprecatedPositionInference}
```
pokey, line 36: as above

src/test/suite/fixtures/recorded/compoundTargets/clearPastStartOfToken.yml
```diff
@@ -1,11 +1,11 @@
...
 command:
   version: 1
-  spokenForm: take past start of token
-  action: setSelection
+  spokenForm: clear past start of token
+  action: clearAndSetSelection
   targets:
     - type: range
-      start: {type: primitive}
+      start: {type: primitive, isImplicit: true}
```
pokey, line 8: do we need to do a migration?  what happens when they issue the command with new extension / old Talon?  Does this no longer work as it did before?

src/test/suite/fixtures/recorded/compoundTargets/clearThisSliceBatt.yml
```diff
@@ -0,0 +1,38 @@
...
+    - anchor: {line: 0, character: 0}
+      active: {line: 0, character: 1}
+  marks:
+    default.b:
+      start: {line: 1, character: 0}
+      end: {line: 1, character: 2}
+finalState:
+  documentContents: |-
+    a
+    b
```
pokey, line 32: huh.  is this what we want?  seems inconsistent

pokey (commented):
Looks pretty good.  I fixed merge conflicts and tweaked comments, but see questions below

src/processTargets/processTargets.ts
AndreasArvidsson: This was incorrectly implemented, but only for very niche cases. If I remember correctly it had to do with reverse selection `this past start of token` was inconsistent with `this past end of token`

src/test/suite/fixtures/recorded/compoundTargets/clearJustThisPastEndOfToken.yml
AndreasArvidsson: Doesn't matter for this test.

src/test/suite/fixtures/recorded/compoundTargets/clearPastStartOfToken.yml
AndreasArvidsson: Should work as before

src/test/suite/fixtures/recorded/compoundTargets/clearThisSliceBatt.yml
AndreasArvidsson: The selection is non empty and therefore does not expand

src/test/suite/fixtures/recorded/compoundTargets/clearPastStartOfToken.yml
pokey: Needs a migration 😕.  I can knock it out

pokey:
Unfortunately, [`"square switch past whale"`](https://github.com/cursorless-dev/cursorless/blob/main/src/test/suite/fixtures/recorded/implicitExpansion/squareSwitchPastWhale.yml) breaks, because this PR makes that command equivalent to `"square switch just this past whale"`, which means that the `square switch` action won't do implicit expansion on `this` because of the `just` 😕.

Not immediately obvious how to fix this one cc/ @AndreasArvidsson

AndreasArvidsson:
Isn't that the desired behavior? If you say this it will expand but implicit targets do not?

pokey:
I guess maybe it's ok?  It's causing 4 CI failures...

pokey (approved):
Ok I think this one is good to go; I'll give you a chance to have a quick look if you get a minute @AndreasArvidsson

pokey:
Update from meet-up:
- Let's guard this one with a setting: `cursorless.experimental.featureFlags.emptyCursorTokenExpansion`
- Let's add tests that exercise this setting

AndreasArvidsson:
@pokey Looks good

src/core/commandVersionUpgrades/upgradeV3ToV4/upgradeV3ToV4.ts
```diff
@@ -0,0 +1,56 @@
...
+      };
+    }
+    case "list": {
+      const { elements, ...rest } = target;
+
+      return {
+        ...rest,
+        elements: elements.map(upgradeTarget) as [
+          PartialPrimitiveTargetDescriptor | PartialRangeTargetDescriptor,
+        ],
```
auscompgeek, line 42: This should be an array rather than a 1-tuple, right?

pokey:
I found myself wanting this one again today; I say let's ship it without feature flag.  If we decide it's a bad idea down the road, we can always:

1. move it behind a feature flag, then
2. have default behaviour (no feature flag) be to show a warning, pointing them to doc for enabling feature flag

src/core/commandVersionUpgrades/upgradeV3ToV4/upgradeV3ToV4.ts
pokey: Nice catch 😊

pokey (approved):

PR #1095, opened 2022-10-26
Added paragraph scope handler

### LINKED ISSUE #1055: Migrate `paragraph` to use a scope handler

AndreasArvidsson:
Fixes #1055

## Checklist


AndreasArvidsson:
@pokey Updated with new generator format

pokey (commented):
Looks good!  Worth migrating all block tests to `scopes/paragraph` directory and then seeing what's missing.  Upon a quick glance I didn't see any "next", "previous", or "n blocks" tests.  Worth having some with edge cases too.  I think probably want cross product of:

- empty cursor at start of block
- empty cursor in middle of block
- empty cursor at end of block

Cross with

- "block"
- "next block"
- "previous block"
- "two blocks"
- "two blocks backward"

It looks like we already have a few "every" tests, so probably ok on that front?  Worth quick review tho

A bit annoying we still have to do all these cross products.  If we tested the scope handler directly we could do many fewer tests 🤔

AndreasArvidsson:
Tests added
Yeah all of these cross products are really annoying. Would be nice if we could get abstraction to a point where we could do this as unit tests.

pokey (approved):
Solid

PR #1096, opened 2022-10-27
Iteration-based scope handler interface

pokey:
## Checklist

- [x] Remove `getPreferredScope` function from `ScopeHandler` interface
- [x] Add `isPreferredOver` as optional function to `ScopeHandler` interface
- [x] Implement token / char / word preferences

## Decided not to do the following in this PR
- Implement surrounding pairs
- Implement tree-sitter scopes
- Implement fallback handler? (just yield everything from first subhandler, then everything from second, etc)


pokey:
Need to make a few tweaks

pokey:
ok good to go

src/processTargets/modifiers/ContainingScopeStage.ts
```diff
@@ -93,3 +112,71 @@ export class ContainingScopeStage implements ModifierStage {
...
+  // If there is no backward scope, or if the backward scope is an ancestor of
+  // forward scope, return forward scope
+  if (
+    backwardScope == null ||
+    backwardScope.domain.contains(forwardScope.domain)
+  ) {
+    return forwardScope;
+  }
+
+  return scopeHandler.isPreferredOver(backwardScope, forwardScope) ?? false
```
AndreasArvidsson, line 179: `isPreferredOver` can't be null here. You shouldn't need the nullish operator.

pokey: Yes it can.  See implementation of token preferences in TokenScopeHandler

AndreasArvidsson: Ok so it can return undefined. Gotcha

src/processTargets/modifiers/scopeHandlers/TokenScopeHandler.ts
```diff
@@ -23,4 +24,40 @@ export default class TokenScopeHandler extends NestedScopeHandler {
         }),
     }));
   }
+
+  isPreferredOver(
```
AndreasArvidsson, line 28: Maybe we should simplify this one. 
* We know that you can't have two adjacent identifiers. That means that we basically only have to check if the left one is an identifier. 
* I'm not sure if I care about length to be honest
* Returning false or undefined will give the same result. I would always return a boolean.
What about:
```js
const textA = document.getText(scopeA.domain);
return identifierMatcher.test(textA);
```

src/test/suite/fixtures/recorded/scopes/token/clearTokenBackward.yml
```diff
@@ -0,0 +1,26 @@
...
+      modifiers:
+        - type: relativeScope
+          scopeType: {type: token}
+          offset: 0
+          length: 1
+          direction: backward
+  usePrePhraseSnapshot: true
+  action: {name: clearAndSetSelection}
+initialState:
+  documentContents: aaa.
```
AndreasArvidsson, line 16: This test should probably be done with the tokens in reverse order to make sure we are ignoring preference?

src/processTargets/modifiers/scopeHandlers/shouldYieldScope.ts
```diff
@@ -0,0 +1,102 @@
...
+    previousScope != null &&
+    compareTargetScopes(direction, position, previousScope, scope) >= 0
+  ) {
+    // Don't yield any scopes that are considered prior to a scope that has
+    // already been yielded
+    return false;
+  }
+
+  // Simple containment checks
+  switch (containment) {
```
AndreasArvidsson, line 40: Do we have need for a `requiredStrict`?

src/processTargets/modifiers/scopeHandlers/getScopesOverlappingRange.ts
```diff
@@ -0,0 +1,40 @@
...
+  scopeHandler: ScopeHandler,
+  editor: TextEditor,
+  { start, end }: Range,
+): TargetScope[] {
+  let seenScope = false;
+  const scopes: TargetScope[] = [];
+  for (const scope of scopeHandler.generateScopes(editor, start, "forward", {
+    distalPosition: end,
+  })) {
+    if (seenScope && scope.domain.contains(start)) {
```
AndreasArvidsson, line 22: Would you mind adding a comment about this logic?

src/processTargets/modifiers/scopeHandlers/getScopesOverlappingRange.ts
```diff
@@ -0,0 +1,40 @@
...
+      continue;
+    }
+
+    seenScope = true;
+
+    if (scope.domain.start.isAfterOrEqual(end)) {
```
AndreasArvidsson, line 28: Would you mind adding a comment about this logic?

src/processTargets/modifiers/scopeHandlers/getScopeRelativeToPosition.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export default function getScopeRelativeToPosition(
+  scopeHandler: ScopeHandler,
+  editor: TextEditor,
+  position: Position,
+  offset: number,
+  direction: Direction,
+): TargetScope {
+  let scopeCount = 0;
+  for (const scope of scopeHandler.generateScopes(editor, position, direction, {
```
AndreasArvidsson, line 25: Probably looks cleaner if we save the iterator in a constant first. Multiline function call inside the loop header is a bit cluttery for my taste tout

src/processTargets/modifiers/scopeHandlers/compareTargetScopes.ts
```diff
@@ -0,0 +1,101 @@
...
+  // Otherwise `b` starts before position, but `a` does not.  Apply reverse
+  // logic to last `if` statement above
+  const value = a.start.compareTo(b.end);
+
+  return value !== 0 ? value : a.isEmpty ? -1 : 1;
+}
+
+// FIXME: Unify this function with compareTargetScopesForward by constructing
+// distal / proximal versions of these ranges
+function compareTargetScopesBackward(
```
AndreasArvidsson, line 72: This function could definitely need some comments

src/processTargets/modifiers/scopeHandlers/BaseScopeHandler.ts
```diff
@@ -0,0 +1,167 @@
...
+ */
+export default abstract class BaseScopeHandler implements ScopeHandler {
+  public abstract readonly scopeType: ScopeType;
+  public abstract readonly iterationScopeType: ScopeType;
+
+  /**
+   * Indicates whether scopes are allowed to contain one another.  If `true`, we
+   * can optimise the algorithm by making certain assumptions.
+   */
+  protected abstract readonly isHierarchical: boolean;
```
AndreasArvidsson, line 25: I mean even if this one is false we can optimize by using `canStopEarly`

AndreasArvidsson:
General looks really good. Am definitely on board with moving to a single generator function. Are left a few comments about minor stuff.

src/processTargets/modifiers/getContainingScope.ts
```diff
@@ -0,0 +1,30 @@
...
+) {
+  return getOne(
+    scopeHandler.generateScopes(editor, position, direction, {
+      containment: "required",
+    }),
+  );
+}
+
+function getOne<T>(iterable: Iterable<T>): T | undefined {
+  const { value, done } = iterable[Symbol.iterator]().next();
```
AndreasArvidsson, line 28: Depending on if the generator is using a return statement or not the very last value could have `done: true`
Calling next after the generator is done will just make the value undefined.
In this case I would just return the value and don't care about `done`

src/processTargets/modifiers/scopeHandlers/shouldYieldScope.ts
pokey: No

src/processTargets/modifiers/scopeHandlers/BaseScopeHandler.ts
pokey: Whoops; meant to say `false` there, today we only do optimisations for non-hierarchical scopes

src/processTargets/modifiers/scopeHandlers/TokenScopeHandler.ts
pokey: nice

pokey:
Ok should be ready now; thanks for the comments.  You may see some weird diffs in the recorded tests; I moved a few tests around and had to do some weird renaming because eg we had a test named `clearTwoTokensBackward` in two different directories.  Probably should have done that in a separate PR, but noticed that we had a bunch of relative tests in `containingScope` directory while I was looking for a test and decided to move them to `relativeScope`

src/processTargets/modifiers/scopeHandlers/TokenScopeHandler.ts
```diff
@@ -23,4 +24,17 @@ export default class TokenScopeHandler extends NestedScopeHandler {
...
+    scopeA: TargetScope,
+    _scopeB: TargetScope,
+  ): boolean | undefined {
+    const {
+      editor: { document },
+    } = scopeA;
+    const { identifierMatcher } = getMatcher(document.languageId);
+
+    const textA = document.getText(scopeA.domain);
+    return identifierMatcher.test(textA) ? true : undefined;
```
AndreasArvidsson, line 38: I'm still not sure what's the different between false and undefined is?

AndreasArvidsson (approved):

src/processTargets/modifiers/scopeHandlers/TokenScopeHandler.ts
pokey: Undefined lets the containing scope stage decide; didn't want to bake the fact that we go right into this function.  Tho on second thought, your optimisations already rely heavily on that fact.  How about now?

AndreasArvidsson: Looks good

PR #1099, opened 2022-10-29
Switch `NestedScopeHandler` to generator

pokey:
## Checklist

### DIFF
--- a/src/util/regex.ts
+++ b/src/util/regex.ts
@@ -1,4 +1,6 @@
+import { imap } from "itertools";
 import { Range, TextEditor } from "vscode";
+import { Direction } from "../typings/targetDescriptor.types";
 
 function _rightAnchored(regex: RegExp) {
   const { source, flags } = regex;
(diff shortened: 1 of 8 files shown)


src/util/regex.ts
```diff
@@ -74,3 +76,28 @@ export function getMatchesInRange(
...
+      editor.document.positionAt(offset + match.index! + match[0].length),
+    );
+
+  // Reset the regex to start at the beginning of string, in case the regex has
+  // been used before.
+  // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#finding_successive_matches
+  regex.lastIndex = 0;
+
+  return direction === "forward"
+    ? imap(text.matchAll(regex), matchToRange)
```
pokey, line 101: Note that we are now lazy if we're going forward, because regexes support that

AndreasArvidsson (approved):

PR #1103, opened 2022-10-31
ide: introduce `getActiveTextEditor()`

phillco:
Per Sunday's discussion, we'd like the ability to override the lookup of the "active" editor to better aid Cursorless Everywhere's implementation of splits.

Ideally, this would be part of the `ide` abstraction, but that would require us to define interface replacements for _all_ aspects of the `TextEditor`, and we don't want to rush that.

Thus, we decided the simplest possible solution would be to simply abstract out this specific lookup into its own global singleton function (`activeEditor.ts`). That way we can override the logic of that file in the fork (it's unlikely to change very much upstream, until we replace it with `ide`), rather than accumulating dozens of future merge conflicts and code changes.

~~Ideally we'll also have a lint rule preventing new references to `window.activeTextEditor`, but for now I can just check them manually and fix any that get introduced.~~ lint rule added ✅

Semi related: https://github.com/cursorless-dev/cursorless/pull/1102

## Checklist


AndreasArvidsson (approved):

pokey:
@phillco ok I pushed a lint rule; seems to be working as expected.  Reported an error in the one place you use it and then I added an ignore and it went away

pokey (approved):
lgtm!  I did a minor rename and added a lint rule.  otherwise looks good!

phillco:
Cool! Do I hit the "squash and merge" button or do you?

AndreasArvidsson:
You snooze, you loose!

PR #1109, opened 2022-11-01
Fixed bug with reusing regex in token preference

### LINKED ISSUE #1108: Format command doesn't work with the token scope
In the following code:

```
ExtensionContext)
```

With the cursor just before the `)`, saying `"form hammer at token"` results in

```
)
```

AndreasArvidsson:
Fixes #1108

## Checklist

### DIFF
--- a/src/processTargets/modifiers/scopeHandlers/TokenScopeHandler.ts
+++ b/src/processTargets/modifiers/scopeHandlers/TokenScopeHandler.ts
@@ -43,6 +43,8 @@ export default class TokenScopeHandler extends NestedScopeHandler {
     } = scopeA;
     const { identifierMatcher } = getMatcher(document.languageId);
 
+    // NB: Don't directly use `test` here because global regexes are stateful
+    // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#finding_successive_matches
     return testRegex(identifierMatcher, document.getText(scopeA.domain))
       ? true
       : testRegex(identifierMatcher, document.getText(scopeB.domain))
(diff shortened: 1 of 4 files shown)


src/test/suite/containingTokenTwice.test.ts
```diff
@@ -0,0 +1,41 @@
+import { assert } from "chai";
+import * as vscode from "vscode";
+import { CommandV3 } from "../../core/commandRunner/command.types";
+import { getCursorlessApi } from "../../util/getExtensionApi";
+import { openNewEditor } from "../openNewEditor";
+import { standardSuiteSetup } from "./standardSuiteSetup";
+
+suite("Take token twice", async function () {
```
pokey, line 8:
```suggestion
// Check that we don't run afoul of stateful regex craziness
// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#finding_successive_matches
// When this fails, the regex that checks if something is an identifier will start at the wrong place the second time it is called
suite("Take token twice", async function () {
```

src/processTargets/modifiers/scopeHandlers/TokenScopeHandler.ts
```diff
@@ -43,9 +43,9 @@ export default class TokenScopeHandler extends NestedScopeHandler {
     } = scopeA;
     const { identifierMatcher } = getMatcher(document.languageId);
 
-    return identifierMatcher.test(document.getText(scopeA.domain))
+    return testRegex(identifierMatcher, document.getText(scopeA.domain))
```
pokey, line 46:
```suggestion
    // NB: Don't directly use `test` here because global regexes are stateful
    // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#finding_successive_matches
    return testRegex(identifierMatcher, document.getText(scopeA.domain))
```

pokey (approved):

PR #1114, opened 2022-11-02
Added shell:true to init launch sandbox

AndreasArvidsson:
Doesn't work on windows otherwise
If this has any problem on mac or Linux we could make this boolean only be true on windows

## Checklist


pokey (approved):
Works for me; thanks!

PR #1124, opened 2022-11-09
Dependency injection

AndreasArvidsson:
I was a bit bored so I looked into how dependency injection is done in typescript. Actually for our needs it's not that much code so I took the liberty of throwing together a working proof of concept. The upside with doing it ourself is that our special needs with regards to end to end test can be easily met. I'm not sure this is a good idea and we might just end up using `tsyringe`, but I wanted to experiment a bit.

Look in `extension.ts` for the example

```ts
class A {
  constructor() {
    console.log("constructor A");
  }
}

@singleton()
class B {
  constructor(a: A, a2: A) {
    console.log("constructor B");
  }
}

class C {
  constructor(value: string) {
    console.log("constructor C", value);
  }
}

@transient()
class D {
  constructor(b: B, b2: B, c: C) {
    console.log("constructor D");
  }
}

container.registerTransient(A);
container.registerValue(C, new C("foo"));
const instance = container.resolve(D);

console.log("clearInstances");
container.clearInstances();
container.resolve(B);
```
output:
```
constructor C foo // Value with given constructor argument
constructor A // Transient
constructor A
constructor B // Singleton
constructor D
clearInstances
constructor A
constructor A
constructor B
```

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
 	"activationEvents": [
 		"*"
 	],
-	"main": "./dist/extension.js",
+	"main": "./out/extension.js",
 	"capabilities": {
 		"untrustedWorkspaces": {
 			"supported": true
@@ -588,6 +588,7 @@
 		"@types/js-yaml": "^4.0.2",
 		"@types/mocha": "^8.0.4",
 		"@types/node": "^16.11.3",
+		"@types/reflect-metadata": "^0.1.0",
 		"@types/semver": "^7.3.9",
 		"@types/sinon": "^10.0.2",
 		"@types/vscode": "~1.61.0",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -46,7 +46,7 @@
 	"activationEvents": [
 		"*"
 	],
-	"main": "./dist/extension.js",
+	"main": "./out/extension.js",
```
AndreasArvidsson, line 49: This was needed without it the metadata was missing. Probably need to update some additional tsconfig.json file/s

AndreasArvidsson:
Cursorless probably doesn't need a full dependency injection framework. Closing for now.

PR #1126, opened 2022-11-09
Update editor config indent

AndreasArvidsson:
## Checklist


.editorconfig
```diff
@@ -11,10 +11,9 @@ max_line_length = 80
 trim_trailing_whitespace = true
 
 [*.json]
-indent_style = tab
+indent_size = 4
```
pokey, line 14: How come 4?

auscompgeek:
Curious why? FWIW Prettier has an open issue to switch the default to tabs.

.editorconfig
AndreasArvidsson: Personally I prefer that in all languages. I can go with two if you prefer it?

AndreasArvidsson:
> Curious why? FWIW Prettier has an open issue to switch the default to tabs.

The main reason is this
https://github.com/cursorless-dev/cursorless/pull/1126/files#diff-ce616deccd5f516e1b391ea2e6dd39c415e0959a3c72cf8e17a4635023437a6b

The left hand side is more whitespace than actual content. Also I prefer to just have this same indentation settings no matter the language

AndreasArvidsson:
This is how it looks for me. I'm pretty sure @pokey had a similar problem.

image

auscompgeek:
image

Still looks reasonable to me in split diff mode.

pokey:
Looks like it's a GitHub setting https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-personal-account-settings/managing-your-tab-size-rendering-preference

The advantage of tabs is that you can decide how wide you want them to be

AndreasArvidsson:
> Looks like it's a [GitHub setting](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-personal-account-settings/managing-your-tab-size-rendering-preference)
> 
> The advantage of tabs is that you can decide how wide you want them to be

Cool thanks, but since 8 is the default spaces have the advantage of being consistent. Personally I don't se why you would want tabs in some languages and spaces in other.

AndreasArvidsson:
> The argument would be to use tabs everywhere

I much rather prefer that than mixing

pokey:
I think tabs are better in theory, but I worry that in practice they may be a bit cumbersome.  Eg

- by default our code will look crazy spaced out on Github, as most users won't have tweaked their tab render setting
- VSCode kind of handles configuring tab display width, but it is a bit wonky (see eg https://github.com/microsoft/vscode/issues/10339 and https://github.com/microsoft/vscode/issues/42740)

@auscompgeek if you are planning to make it to meet-up today maybe we can discuss then

.editorconfig
```diff
@@ -5,14 +5,10 @@ root = true
 charset = utf-8
 end_of_line = lf
 insert_final_newline = true
-indent_size = 2
-indent_style = space
+indent_style = tab
```
pokey, line 8: don't forget to override yaml

phillco:
+1 on just converting everything to spaces.

Also, lol at that list of merge conflicts :)

AndreasArvidsson:
> +1 on just converting everything to spaces.
> 
> Also, lol at that list of merge conflicts :)

After today's session we went other route and actually changed everything into tabs :D

phillco:
Ah well. As long as it's consistent.

pokey:
Huh https://code.visualstudio.com/updates/v1_74#_new-indent-size-setting cc/ @auscompgeek @AndreasArvidsson

auscompgeek:
Oh goodie, I can finally convert my vim config to VSCode

pokey:
For now, let's stick to spaces, as the tooling hasn't really caught up with tabs yet, eg:

- yaml heredocs converting to inline strings,
- yaml not supporting tab indentation
- VSCode disallowing auto-indent detection,
- GitHub large spaces by default

AndreasArvidsson:
@pokey Updated once again with space indentation

pokey (approved):

PR #1128, opened 2022-11-11
Implement our own types instead of vscode 

AndreasArvidsson:
## Checklist

- [x] Add tests for `Selection`
- [x] Add tests for `TextLine`


src/libs/common/ide/Location.ts
```diff
@@ -0,0 +1,33 @@
+import { URI } from "vscode-uri";
+import Position from "./Position";
+import Range from "./Range";
+
+/**
+ * Represents a location inside a resource, such as a line
+ * inside a text file.
+ */
+export default class Location {
```
pokey, line 9: I wonder if we want a version of this that has a document pointer. Also where do you need this one?

src/testUtil/TestCaseRecorder.ts
```diff
@@ -186,7 +185,7 @@ export class TestCaseRecorder {
             undefined,
             ["clipboard"],
             this.active ? this.extraSnapshotFields : undefined,
-            getActiveTextEditor()!,
+            ide().activeTextEditor!,
```
pokey, line 188: Tempted to let snapshot code get this off ide obj

src/typings/Types.ts
```diff
@@ -38,21 +40,21 @@ export interface ProcessedTargetsContext {
    */
   actionFinalStages: ModifierStage[];
   currentSelections: SelectionWithEditor[];
-  currentEditor: vscode.TextEditor | undefined;
+  currentEditor: TextEditor | undefined;
   hatTokenMap: ReadOnlyHatMap;
   thatMark: Target[];
   sourceMark: Target[];
-  getNodeAtLocation: (location: Location) => SyntaxNode;
+  getNodeAtLocation: (location: vscode.Location) => SyntaxNode;
```
pokey, line 47: Hmm. I guess we revisit this once we have parse tree in monorepo

Tempted to wrap it tho for now to contain VSCode dep. 

Maybe add it to document?

pokey (commented):
Seems like a solid direction

.eslintrc.json
```diff
@@ -52,7 +52,7 @@
 		"no-restricted-properties": [
 			2,
 			{
-				"property": "activeTextEditor"
+				"property": "activeTextEditor TODO: "
```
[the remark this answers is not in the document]
AndreasArvidsson: That just my own mark that I'm not done with this line :D

src/libs/common/ide/Location.ts
AndreasArvidsson: It's used by Tree sitter getNodeAtLocation

src/testUtil/TestCaseRecorder.ts
AndreasArvidsson: Right now I just did a find and replace

src/typings/Types.ts
AndreasArvidsson: I was thinking of adding it to ide for now

src/actions/Scroll.ts
```diff
@@ -18,12 +17,12 @@ class Scroll implements Action {
...
     });
 
-    const originalEditor = getActiveTextEditor();
+    const originalEditor = ide().activeEditableTextEditor;
 
     for (const lineWithEditor of lines) {
       // For reveal line to the work we have to have the correct editor focused
-      if (lineWithEditor.editor !== getActiveTextEditor()) {
-        await focusEditor(lineWithEditor.editor);
+      if (lineWithEditor.editor !== ide().activeTextEditor) {
```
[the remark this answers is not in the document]
AndreasArvidsson: That entire thing is going to be folded into the editor. For now I have not moved `executeCommand` from the actions into the actual editor. Wants that is done the entire focus thing will be inside the vscode implementation.

src/apps/cursorless-vscode-e2e/suite/recorded.test.ts
```diff
@@ -177,7 +177,7 @@ async function runTest(file: string, spyIde: SpyIDE) {
     cursorlessApi.sourceMark,
     excludeFields,
     [],
-    vscode.window.activeTextEditor!, // eslint-disable-line no-restricted-properties
+    spyIde.activeTextEditor!,
```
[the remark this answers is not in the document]
AndreasArvidsson: We could if we only wanted to be able to snap shot the active editor. The wait is now you can give it whatever editor you want.

src/apps/cursorless-vscode-e2e/suite/groupByDocument.test.ts
```diff
@@ -1,9 +1,11 @@
+import { splitKey } from "@cursorless/common";
 import { getCursorlessApi } from "@cursorless/vscode-common";
 import * as assert from "assert";
 import * as vscode from "vscode";
-import { splitKey } from "@cursorless/common";
-import { runCursorlessCommand } from "../runCommand";
+// eslint-disable-next-line @typescript-eslint/no-restricted-imports
+import { toVscodeEditor } from "../../../ide/vscode/VscodeUtil";
```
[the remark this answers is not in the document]
AndreasArvidsson: That is exactly what that function does, but since its used in multiple places I wanted an import

True. I will have to have a look at that

yarn.lock
```diff
@@ -4,14 +4,14 @@
 
 "@babel/runtime@^7.15.4":
   version "7.20.0"
-  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.20.0.tgz#824a9ef325ffde6f78056059db3168c08785e24a"
+  resolved "https://registry.npmjs.org/@babel/runtime/-/runtime-7.20.0.tgz"
```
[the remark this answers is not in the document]
AndreasArvidsson: weird

src/util/setSelectionsAndFocusEditor.ts
```diff
@@ -1,124 +1,27 @@
...
-  await focusEditor(editor);
+  await editor.focus();
 }
 
 export function setSelectionsWithoutFocusingEditor(
-  editor: TextEditor,
+  editor: EditableTextEditor,
   selections: Selection[],
 ) {
   editor.selections = uniqDeep(selections);
```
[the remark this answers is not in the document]
AndreasArvidsson: We should definitely move away from that but I don't think that should be in this pr.

src/libs/common/ide/Range.ts
```diff
@@ -0,0 +1,161 @@
...
+    endLine: number,
+    endCharacter: number,
+  );
+
+  constructor(...args: any[]) {
+    const [start, end] = (() => {
+      // Arguments are one positions
+      if (args.length === 1) {
+        return [args[0], args[0]];
+      }
```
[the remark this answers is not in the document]
AndreasArvidsson: We could do that if you prefer it?

src/libs/common/ide/Range.ts
```diff
@@ -0,0 +1,161 @@
+import { Position, Selection } from ".";
+
+export default class Range {
+  /**
+   * The start position. It is before or equal to {@link Range.end end}.
```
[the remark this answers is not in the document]
AndreasArvidsson: Copied directly from vscode. No idea why they do it like that.

src/libs/common/ide/Range.ts
```diff
@@ -0,0 +1,161 @@
...
+   * @param endCharacter A zero-based character value.
+   */
+  constructor(
+    startLine: number,
+    startCharacter: number,
+    endLine: number,
+    endCharacter: number,
+  );
+
+  constructor(...args: any[]) {
```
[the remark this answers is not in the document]
AndreasArvidsson: No. It's actually only the above constructors that are usable. The syntax is a bit weird but the ones without implementation is like the actual declaration and the one with implementation is not actually declaring a new constructor.

src/apps/cursorless-vscode-e2e/suite/groupByDocument.test.ts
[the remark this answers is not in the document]
AndreasArvidsson: That is what that utility function is doing. The reason for having utility function is that I want to make sure that it is of the correct type before unwrapping the editor.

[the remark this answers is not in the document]
pokey: Fine then please move that utility function into `vscode-common`

src/util/setSelectionsAndFocusEditor.ts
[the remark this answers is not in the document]
pokey: https://github.com/cursorless-dev/cursorless/issues/1131

src/libs/common/ide/Range.ts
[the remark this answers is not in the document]
pokey: Fine not a big deal

src/libs/common/ide/Range.ts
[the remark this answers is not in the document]
pokey: ah weird ok

src/ide/vscode/VscodeFocusEditor.ts
```diff
@@ -0,0 +1,102 @@
...
+      end: desiredEditorIndex + 1,
+    }),
+  ];
+  desiredNotebookEditor.selections = desiredSelections;
+  desiredNotebookEditor.revealRange(desiredSelections[0]);
+
+  // Issue a command to tell VSCode to focus the cell input editor
+  // NB: We don't issue the command if it's already focused, because it turns
+  // out that this command is actually a toggle, so that causes it to de-focus!
+  if (ide().activeTextEditor?.id !== editorId) {
```
pokey, line 99: I think I'd avoid the singleton here and just use `vscode.window.activeEditor`; this code is extremely VSCode-specific, so I'm not sure I see the downside of that, and the upside is removing references to the singleton, and that I believe you can remove `editorId` from the signature of every function in this file.

src/ide/vscode/vscodeIdeUtil.ts
```diff
@@ -0,0 +1,24 @@
+import { TextDocument, TextEditor, TextLine } from "@cursorless/common";
+import * as vscode from "vscode";
+import { VscodeTextDocumentImpl } from "./VscodeTextDocumentImpl";
+import { VscodeTextEditorImpl } from "./VscodeTextEditorImpl";
+import VscodeTextLineImpl from "./VscodeTextLineImpl";
+
+const editorMap = new WeakMap<vscode.TextEditor, TextEditor>();
```
pokey, line 7: I'm skeptical of these module-level variables.  Shall we just add this to the vscode ide object?  Or we could make a `window` / `editorService` that is part of the `ide` 🤷‍♂️

src/libs/common/ide/fake/FakeIDE.ts
```diff
@@ -17,7 +19,7 @@ export default class FakeIDE implements IDE {
   clipboard: FakeClipboard;
   private disposables: Disposable[] = [];
 
-  constructor() {
+  constructor(private original?: IDE) {
```
pokey, line 22: Not a fan of this setup.  I would prefer a normalized IDE in this PR.  I think it is in scope because you are being forced to break the `FakeIDE` abstraction.  Lmk if you want me to do that for you; I don't mind either way

src/libs/vscode-common/index.ts
```diff
@@ -1,7 +1,8 @@
 export * from "./getExtensionApi";
 export * from "./notebook";
-export * from "./toPlainObject";
+export * from "./testUtil/toPlainObject";
 export * from "./testUtil/testConstants";
+export * from "./vscodeUtil";
```
pokey, line 5: fwiw i lean towards specific exports rather than `*`, but this setup is still pretty new to me so I don't have a strong feeling yet

src/processTargets/marks/LineNumberStage.ts
```diff
@@ -35,7 +35,7 @@ const getLineNumber = (
     case "absolute":
       return lineNumber;
     case "relative":
-      return editor.selection.active.line + lineNumber;
+      return editor.selections[0].active.line + lineNumber;
```
pokey, line 38: Huh.  This feels kinda VSCode-specific 🤔.  I wonder if relative line numbers should be part of the IDE abstraction.  Not sure.  We should prob discuss with maintainers of other sidecar impls to see what makes sense in their editors

src/processTargets/modifiers/scopeHandlers/LineScopeHandler.ts
```diff
@@ -58,14 +58,11 @@ export function createLineTarget(
...
-  const endCharacterIndex =
-    endLine.range.end.character -
-    (endLine.text.length - endLine.text.trimEnd().length);
 
   return new Range(
     startLine.lineNumber,
     startLine.firstNonWhitespaceCharacterIndex,
     endLine.lineNumber,
-    endCharacterIndex,
+    endLine.lastNonWhitespaceCharacterIndex,
```
pokey, line 66: 👍

src/util/targetUtils.ts
```diff
@@ -49,12 +49,12 @@ export function groupTargetsForEachEditor(targets: Target[]) {
...
 }
 
-export function groupForEachEditor<T>(
+function groupForEachEditor<T>(
   targets: T[],
   getEditor: (target: T) => TextEditor,
 ): [TextEditor, T[]][] {
   // Actually group by document and not editor. If the same document is open in multiple editors we want to perform all actions in one editor or an concurrency error will occur.
-  const getDocument = (target: T) => getEditor(target).document;
+  const getDocument = (target: T) => getEditor(target).document.uri;
```
pokey, line 57: *nit*: this should prob now be called `getDocumentUri`

src/util/setSelectionsAndFocusEditor.ts
```diff
@@ -1,124 +1,27 @@
...
-  editor: TextEditor,
+  editor: EditableTextEditor,
   selections: Selection[],
   revealRange: boolean = true,
 ) {
   setSelectionsWithoutFocusingEditor(editor, selections);
 
   if (revealRange) {
-    editor.revealRange(editor.selection);
+    editor.revealRange(editor.selections[0]);
```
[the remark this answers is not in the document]
pokey: Curious why you marked this one resolved?  Did you fix it but not push or something?

Not sure I'm a fan of marking comments resolved without a response or fix.  If you think this is out of scope that's fine, but probably something we should discuss here and then potentially file follow-up to track

Tbh I'd be tempted to have a policy where the person who posts the comment is the one that should resolve it, tho we'd prob relax that for comments from non-maintainers, as they are liable to post a comment and then disappear 😅

I think an exception would be for comments that are non-actionable, eg "👍".  Maybe we should implement something like [conventional comments](https://conventionalcomments.org/)

src/libs/common/ide/types/Events.ts
```diff
@@ -0,0 +1,46 @@
+import type { TextDocument, Range } from "@cursorless/common";
```
pokey, line 1: Given the amount of interface definitions / docstrings we've stolen from VSCode in this directory, we might want to put a note in the directory to that effect, and pointing to the VSCode license

src/libs/common/ide/types/ide.types.ts
```diff
@@ -24,17 +26,49 @@ export interface IDE {
...
+  getEditableTextEditor(editor: TextEditor): EditableTextEditor;
+
+  /**
+   * An event that is emitted when a {@link TextDocument text document} is changed. This usually happens
+   * when the {@link TextDocument.getText contents} changes but also when other things like the
+   * {@link TextDocument.isDirty dirty}-state changes.
+   */
+  onDidChangeTextDocument(
+    listener: (event: TextDocumentChangeEvent) => void,
+  ): Disposable;
```
pokey, line 71: Tbh I'm a bit skeptical about mirroring their whole `onDidXxx` returning a `Disposable` setup, but I think ok to parrot at least some of their interface for the first draft of our api spec

src/libs/common/types/Location.ts
```diff
@@ -0,0 +1,33 @@
+import type { URI } from "vscode-uri";
+import type { Position } from "@cursorless/common";
+import { Range } from "@cursorless/common";
+
+/**
+ * Represents a location inside a resource, such as a line
+ * inside a text file.
+ */
+export class Location {
```
pokey, line 9: Hmm.  I thought we decided we didn't want to mirror their `Location` interface.  Do we use this in a lot of places?  Might be nice if we could avoid it for now

src/libs/common/types/TextDocument.ts
```diff
@@ -0,0 +1,84 @@
...
+   * @return A {@link TextLine line}.
+   */
+  lineAt(line: number): TextLine;
+
+  /**
+   * Returns a text line denoted by the position. Note
+   * that the returned object is *not* live and changes to the
+   * document are not reflected.
+   *
+   * The position will be {@link TextDocument.validatePosition adjusted}.
```
pokey, line 49: This link will be dead, right?  Do we want to reflect this function as well so that we have something to link to?

src/libs/common/types/TextDocument.ts
```diff
@@ -0,0 +1,84 @@
...
+   * Converts a zero-based offset to a position.
+   *
+   * @param offset A zero-based offset.
+   * @return A valid {@link Position}.
+   */
+  positionAt(offset: number): Position;
+
+  /**
+   * Get the text of this document. A substring can be retrieved by providing
+   * a range. The range will be {@link TextDocument.validateRange adjusted}.
```
pokey, line 78: as above; dead link

pokey (commented):
Ok made it through another few dozen files, so here's another round of comments.  Not too much to go

src/util/setSelectionsAndFocusEditor.ts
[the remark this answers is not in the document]
AndreasArvidsson: That is exactly the rules I'm following. `Probably out of scope` I interpreted as you were going to do this in a follow up.

src/libs/common/ide/Range.ts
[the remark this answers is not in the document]
AndreasArvidsson: Went ahead and did this

.eslintrc.json
```diff
@@ -48,13 +48,7 @@
 			}
 		],
 		"no-throw-literal": "warn",
-		"semi": "off",
-		"no-restricted-properties": [
-			2,
-			{
-				"property": "activeTextEditor"
-			}
-		]
```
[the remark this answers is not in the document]
AndreasArvidsson: Because now we are using `ide().activeTextEditor` all over the place. But as we talked about none of the Cursorless code is now excepting a pure vscode editors they can't really do anything with it. ie if you incorrectly fetch the wrong editor it will become very apparent

src/ide/vscode/VscodeEvents.ts
```diff
@@ -0,0 +1,43 @@
...
+      reason: fromVscodeReason(e.reason),
+    });
+  });
+}
+
+function fromVscodeContentChange(
+  change: vscode.TextDocumentContentChangeEvent,
+): TextDocumentContentChangeEvent {
+  return {
+    ...change,
```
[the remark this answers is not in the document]
AndreasArvidsson: That was spreading a class. This one is a proper object. Do want me to deconstruct it manually anyhow?

src/libs/common/ide/types/Events.ts
AndreasArvidsson: Do just mean put a read me there or a license file or?

src/ide/vscode/vscodeIdeUtil.ts
AndreasArvidsson: moved it into ide

src/libs/common/ide/fake/FakeIDE.ts
AndreasArvidsson: If you have a good idea for a normalized ide, please go ahead.

src/libs/vscode-common/index.ts
AndreasArvidsson: I think I'm on the opposite side. Things that should be public should be in specific files that are exported. And things that should be private should be in its own files. Instead of having to look through files and figure out what's exported and not.

src/processTargets/marks/LineNumberStage.ts
AndreasArvidsson: The whole notion of having a main selection is potentially vscode specific as well. We definitely need to have a discussion about this.

src/libs/common/ide/types/ide.types.ts
AndreasArvidsson: Agree that this could probably be cleaned up. Just for now I try to get this pr to not balloon anymore.

src/libs/common/types/Location.ts
AndreasArvidsson: We don't but for now we are required because it's used by `getNodeAtLocation`. I was thinking of moving that from the graph and into the document instead in a follow up pr.

src/libs/common/types/TextDocument.ts
AndreasArvidsson: I don't think so. Of course we should not return lines and text that are actually outside of the document. 
I have updated the comment for now.

src/libs/common/types/Location.ts
AndreasArvidsson: Okay I went a here then actually moved it. The nice thing about this change is that our scope handlers now have access to `getNodeAtLocation`

src/ide/vscode/VscodeEvents.ts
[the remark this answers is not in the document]
pokey: The type signature just indicates an interface, so we don't really know what it is going to be.  Ie they could change it to a class in the future.  If that is potentially dangerous, then we should maybe avoid spreading this input.  Does that make sense?

src/libs/vscode-common/index.ts
pokey: That sounds like a reasonable position.  Let's start with that

src/util/setSelectionsAndFocusEditor.ts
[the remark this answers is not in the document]
pokey: Ah I see.  That would be "definitely out of scope" 😅.  Which means we should probably at least briefly discuss comment convention / resolution policies at meet-up to make sure we're on the same page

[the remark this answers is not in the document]
pokey: (and https://github.com/cursorless-dev/cursorless/discussions/1135)

.eslintrc.json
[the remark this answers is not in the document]
pokey: https://github.com/cursorless-dev/cursorless/issues/1136

src/libs/common/types/Location.ts
pokey: Hmm.  I thought about it more, and I don't think we want the whole world having access to the parse tree, as it is pretty volatile and low-level.  How hard is it to roll back that change?

src/actions/Scroll.ts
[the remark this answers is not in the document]
pokey: Fine, but is there any reason not to use `!lineWithEditor.editor.isActive` here?

src/libs/common/types/Location.ts
pokey: Sounds good

src/actions/Scroll.ts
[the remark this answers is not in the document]
AndreasArvidsson: Already are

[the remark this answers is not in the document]
pokey: ah right.  looks like my VSCode pull requests extension was being too smart and automatically showing me the source commit because this comment was outdated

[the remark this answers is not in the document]
AndreasArvidsson: Ah

src/libs/common/types/Location.ts
pokey: Update: we hesitantly decided to roll it back, and to keep an eye on how much more painful all of this "good engineering practice" makes it to change our code base

src/ide/vscode/VscodeFocusEditor.ts
```diff
@@ -0,0 +1,98 @@
...
+  [ViewColumn.Five]: "workbench.action.focusFifthEditorGroup",
+  [ViewColumn.Six]: "workbench.action.focusSixthEditorGroup",
+  [ViewColumn.Seven]: "workbench.action.focusSeventhEditorGroup",
+  [ViewColumn.Eight]: "workbench.action.focusEighthEditorGroup",
+  [ViewColumn.Nine]: "workbench.action.focusNinthEditorGroup",
+  [ViewColumn.Active]: "",
+  [ViewColumn.Beside]: "",
+};
+
+export default async function focusVscodeEditor(editor: TextEditor) {
```
pokey, line 31: Switch this to use `VSCodeTextEditorImpl`

AndreasArvidsson:
@pokey Once again updated and ready :)

src/util/notebookLegacy.ts
```diff
@@ -1,30 +1,31 @@
...
 export function isVscodeLegacyNotebookVersion() {
   return semver.lt(version, "1.68.0");
 }
 
-export async function focusNotebookCellLegacy(editor: TextEditor) {
-  const activeTextEditor = getActiveTextEditor();
+export async function focusNotebookCellLegacy(
+  editor: VscodeEditableTextEditorImpl,
+) {
+  const activeTextEditor = editor.ide.activeTextEditor;
```
pokey, line 16: Notice that I now get this from the editor instead of using the `ide` from ambient context

src/ide/vscode/VscodeIDE.ts
```diff
@@ -1,51 +1,92 @@
...
+    return vscodeOnDidChangeTextDocument(listener);
+  }
+
+  public fromVscodeEditor(
+    editor: vscode.TextEditor,
+  ): VscodeEditableTextEditorImpl {
+    if (!this.editorMap.has(editor)) {
+      this.editorMap.set(
+        editor,
+        new VscodeEditableTextEditorImpl(uuid(), this, editor),
```
pokey, line 84: This is now the only place where we instantiate editors.  There was a bug before where one editor could have two different ids, because `activeEditableTextEditor` was instantiating editors directly, ignoring the editor map

src/util/notebookLegacy.ts
AndreasArvidsson: Am not a big fan of this solution. I guess it's only for vscode notebooks but still feels like a anti pattern.

src/ide/vscode/VscodeIDE.ts
AndreasArvidsson: That means that we could just remove `VscodeTextEditorImpl` because it's never actually used. Maybe we should keep the name `VscodeTextEditorImpl` as the single implementation and instead remove the verbose `VscodeEditableTextEditorImpl`?

AndreasArvidsson: This of course means that any text editor could just be cast to an editable text editor anywhere in the code.

src/util/notebookLegacy.ts
pokey: Why is this an anti-pattern?

src/ide/vscode/VscodeIDE.ts
pokey: > Maybe we should keep the name VscodeTextEditorImpl as the single implementation and instead remove the verbose VscodeEditableTextEditorImpl?

I don't feel strongly one way or another.  Would be tempted to just leave it for now, but if you want to merge them I'm not opposed; does get rid of the weird property override thing 🤷‍♂️

> This of course means that any text editor could just be cast to an editable text editor anywhere in the code.

I mean sure there's all kinds of horrible things the client code could do, but that's not really the fault of this design decision; the fact that there is one implementation is hidden behind the interface.  I think it's actually a good idea to allow one implementation to appear in different places via different interfaces in order to hide irrelevant details from the client

src/util/notebookLegacy.ts
AndreasArvidsson: ide is a singleton and responsible for creating editors. Putting that as a member of an editor feels really weird.

src/ide/vscode/VscodeIDE.ts
AndreasArvidsson: Ok will go ahead and do this.

AndreasArvidsson: btw why is passing uuid in better then creating it in the editor constructor?

pokey: i don't feel super strongly bout that one, just felt like we wanted one centralised place managing the id's.  then eg we could switch from uuid to sequential, could easily inject id's during a test, etc

AndreasArvidsson: Ok I will leave the id as a parameter.

src/util/notebookLegacy.ts
AndreasArvidsson: Ok but today ide is a global. Exactly everywhere else in the code we fetch from the same resource except here that why I think this is a weird pattern. I don't feel strongly enough about it but I still think it's a bit weird.

src/libs/common/types/TextLine.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * The offset of the first character which is not a whitespace character as defined
+   * by `/\s/`. **Note** that if a line is all whitespace the length of the line is returned.
+   */
+  readonly firstNonWhitespaceCharacterIndex: number;
+
+  /**
+   * The offset of the last character which is not a whitespace character as defined
+   * by `/\s/`. **Note** that if a line is all whitespace the length of the line is returned.
```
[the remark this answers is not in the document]
AndreasArvidsson: Correct. We kinda had both of these flipped around, but because the way the range constructed works it turned out correctly. I have now made a commit to fix this.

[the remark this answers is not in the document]
pokey: Tbh I think I'd go with the reverse: if it is all whitespace then firstNonWhitespace should be length, and lastNonWhitespace should be 0.  That seems more logical: eg if client forgets to check if all whitespace and tries to make a range of the line's "content", then your current impl will end up with them getting a range full of whitespace, whereas this proposal will result in empty range, which seems like what you'd want.  It also means we don't need special case logic in the implementation; that will just happen automatically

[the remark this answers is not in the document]
pokey: That's not true.  Let's say I do something like:

```typescript
const contentRange = line.range.with(new Position(line.lineNumber, line.firstNonWhitespaceCharacterIndex));
const content = document.getText(contentRange)
```

In your current impl, you'll get all whitespace.  With my impl, you'd get an empty string.  Same goes if you do the corresponding thing with `lastNonWhitespaceCharacterIndex`:

```typescript
const contentRange = line.range.with(undefined, new Position(line.lineNumber, line.lastNonWhitespaceCharacterIndex));
const content = document.getText(contentRange)
```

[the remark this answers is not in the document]
pokey: I think you're talking about

```typescript
new Range(line.lineNumber, line.firstNonWhitespaceCharacterIndex, line.lineNumber, line.lastNonWhitespaceCharacterIndex)
```

which is fair, but I don't think detracts from my example above.  Tbh it's tempting not to support auto-swapping start and end for that reason, but it is useful in other scenarios

I wonder if we should have a property `trimmedRange` that just would return empty range at start of line if line is all whitespace.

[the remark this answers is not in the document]
AndreasArvidsson: Please go ahead

src/libs/common/types/Selection.ts
```diff
@@ -0,0 +1,62 @@
+import { Position, Range } from "@cursorless/common";
+
+export class Selection extends Range {
```
pokey, line 3: How do we want to handle equality for selections?  On the one hand, it doesn't seem right for two selections to be equal if their anchor and active are swapped.  On the other hand, a client might only know that they're instances of `Range`, and be surprised when they have the same start and end, but are not equal 🤔

Kinda argues for selection not to derive from range, and instead just support a `selection.range` accessor

AndreasArvidsson: We should jot compare anchor and active

pokey: why?

AndreasArvidsson: Talon that can't hear the word just apparently

pokey: Ok, I saw your change.  You don't think the following would be surprising:

```typescript
function foo(rangeA: Range, rangeB: Range) {
    if (rangeA.isEqual(rangeB)) {
        ...
    }
}

...

foo(selectionA, selectionB)
```

And then even though they are the same range, that check returns false

pokey: I'm going to quickly remove that `extends Range` and just see how much breaks.  If it's too much stuff, then let's call it out of scope, because we're mirroring VSCode behaviour for now

pokey: Ok well that was a disaster.  What if we just rename `isEqual` to `isRangeEqual`?  It is used in exactly 2 places in our entire codebase 🤔

AndreasArvidsson: Of course it can be but so far we haven't really run into that. In Cursorless we are barely using selections at all since we prefer a range and a boolean instead. 

Sounds reasonable

pokey: Yeah I think it is useful, but the semantics are pretty odd in this case.  I'd be inclined to use `isRangeEqual` for now and then when we need it we can add a better `isEqual`, with a JSDoc comment indicating why this function might not do what you expect.  Make sense?

AndreasArvidsson: Go ahead

AndreasArvidsson:
We can if you like but with two and a half thousand tests I'm feeling pretty comfortable

pokey (approved):
Ok I made a few tweaks; have a look, and if it seems good to you let's merge!

AndreasArvidsson:
Looks good.

PR #1140, opened 2022-11-15
Fix bug with displayLineMap for adding decorations

pokey:
If you have a selection spanning more than one line, and your selection is not reversed, and you scroll offscreen, Cursorless won't prioritise hats on lines closer to your cursor

Unfortunately this one is hard to test until we have better mocking capabilities

Noticed this while reviewing #1128

## Checklist


src/util/addDecorationsToEditor.ts
```diff
@@ -58,8 +67,8 @@ export function addDecorationsToEditors(
 
       tokens.sort(
         getTokenComparator(
-          displayLineMap.get(editor.selections[0].active.line)!,
```
pokey, line 61: Notice how we were using `active` here...

src/util/getDisplayLineMap.ts
```diff
@@ -3,20 +3,22 @@ import { concat, flatten, flow, range, uniq } from "lodash";
...
+  editor: TextEditor,
+  extraLines: number[] = [],
+): Map<number, number> {
   return new Map(
     flow(
       flatten,
       uniq,
     )(
       concat(
-        [[editor.selections[0].start.line]],
```
pokey, line 19: ...but `start` here 😳

This means that if `active` is not on the same line as `start`, and not visible, our token prioritisation breaks

AndreasArvidsson (approved):

PR #1143, opened 2022-11-17
Free actions from vscode dependencies

AndreasArvidsson:
## Checklist


AndreasArvidsson:
@pokey ready

src/libs/common/types/DecorationRangeBehavior.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Describes the behavior of decorations when typing/editing at their edges.
+ */
+export type DecorationRangeBehavior =
```
auscompgeek, line 4: Would it be worth making this a string `const enum`? It seems a bunch of changes are just changing `DecorationRangeBehavior` references to strings.

AndreasArvidsson: We are not really using enums in cursorless and are iinstead using typed strings. I'm not against enums, but then I think we should use them in all places we can. Ie I prefer one solution or the other.

src/ide/vscode/VscodeFold.ts
```diff
@@ -1,43 +1,53 @@
...
 }
 
 async function foldOrUnfold(
   ide: VscodeIDE,
   editor: VscodeTextEditorImpl,
-  lineNumbers: number[] | undefined,
+  ranges: Range[] | undefined,
   command: "editor.fold" | "editor.unfold",
 ): Promise<void> {
+  ranges = ranges ?? editor.selections;
```
AndreasArvidsson, line 28: This is probably complex enough to go in its own file

src/ide/vscode/VscodeEdit.ts
```diff
@@ -7,12 +7,12 @@ import {
...
 import type * as vscode from "vscode";
 
-export default function vscodeEdit(
+export default async function vscodeEdit(
   editor: vscode.TextEditor,
   callback: (editBuilder: TextEditorEdit) => void,
   options?: { undoStopBefore: boolean; undoStopAfter: boolean },
-): Thenable<boolean> {
-  return editor.edit((editBuilder) => {
+): Promise<boolean> {
```
pokey, line 14: Why are we using `Promise` instead of `Thenable`?

src/ide/vscode/VscodeTextEditorImpl.ts
```diff
@@ -65,29 +76,125 @@ export class VscodeTextEditorImpl implements TextEditor {
...
+
+  public fold(ranges?: Range[]): Promise<void> {
+    return vscodeFold(this.ide, this, ranges);
+  }
+
+  public unfold(ranges?: Range[]): Promise<void> {
+    return vscodeUnfold(this.ide, this, ranges);
+  }
+
+  public toggleBreakpoint(ranges?: Range[]): Promise<void> {
```
pokey, line 130: As above, I'd be tempted to make all of these `Thenable`, as VSCode does

pokey:
Ok looks good to me; created https://github.com/cursorless-dev/cursorless/issues/1160 and https://github.com/cursorless-dev/cursorless/issues/1161 to track some things we should discuss / rework, and left a couple of comments

src/libs/common/types/DecorationRangeBehavior.ts
pokey: I'm increasingly leaning towards enums, because then we can do renames, and we have proper JSDoc support for the members.  Shall we use these as our first enums, and then start migrating the others?  I don't think we need to migrate all strings to enums at once.  I think we probably want to keep strings on our API surface, unfortunately, though maybe we can use them there as well?

Maybe let's discuss at meet-up

src/ide/vscode/VscodeEdit.ts
AndreasArvidsson: I don't think thenable is part of typescript? There is of course 3rd party, but I don't really see the upside instead of using regular promises?

src/libs/common/types/DecorationRangeBehavior.ts
pokey: Discussed at meet-up and we want to start using enums instead of strings, for the reasons listed above, and we plan to roll back these enum-to-string changes in the PR, and start migrating our other strings to enums in follow up PRs

src/ide/vscode/VscodeToggleBreakpoint.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export async function vscodeToggleBreakpoint(
+  editor: VscodeTextEditorImpl,
+  descriptors: BreakpointDescriptor[] | undefined,
+): Promise<void> {
+  if (descriptors == null) {
+    return await vscode.commands.executeCommand(
+      "editor.debug.action.toggleBreakpoint",
+    );
+  }
```
pokey, line 14: note: this seemed like reasonable behaviour when no ranges are passed in.  fwiw, we don't ever utilise this functionality but 🤷‍♂️

src/libs/common/types/DecorationRangeBehavior.ts
pokey: Ok switched this and `RevealLineAt` to be enums.  I used integers for the former, and strings for the latter.  I think integers are ok internally, but for api surface we should use strings.  Because we pass `RevealLineAt` to the ide api, it will end up in lsp api.  Could be convinced to just use string enums everywhere 🤷‍♂️

src/actions/ToggleBreakpoint.ts
```diff
@@ -30,31 +21,18 @@ export default class ToggleBreakpoint implements Action {
...
-          new vscode.SourceBreakpoint(new vscode.Location(uri, range)),
-        );
-      }
-    });
+        return range;
+      });
 
-    vscode.debug.addBreakpoints(toAdd);
-    vscode.debug.removeBreakpoints(toRemove);
+      await ide().getEditableTextEditor(editor).toggleBreakpoint(ranges);
```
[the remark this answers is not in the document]
AndreasArvidsson: Looks good but I don't think a line breakpoint can span multiple lines.

pokey (approved):

PR #831, opened 2022-07-04
Pattern matching support

purpleP:
I’m working on pattern matching support as discussed [here](https://github.com/cursorless-dev/cursorless/issues/579#issuecomment-1161627378) and planning to add it to rust, java, c, c++, python, javascript, typescript and haskell (languages I know).

So far I don’t think cursorless even recognised new scopes. @pokey Can you point what I’m missing here?

PR #1147, opened 2022-11-18
Improve `condition` support for Python and Rust

pokey:
- Adds `condition` support for Python ternaries
- Adds `condition` support for Python list comprehensions
- Adds `condition` support for Python match statements
- Adds `condition` support for Rust match statements
- Note that `"chuck"` includes the `if` keyword for match statements and list comprehensions
- Split from #831

## Checklist

Co-authored-by: Michael Doronin <warrior2031@mail.ru>


auscompgeek:
Seems reasonable. I presume `if` clauses in Python list comprehensions are treated similarly?

pokey:
> Seems reasonable. I presume `if` clauses in Python list comprehensions are treated similarly?

Hmm I don't believe that they are.  Does make me wonder if it's what we want; feels a bit small for `"if state"`, as it's not a full statement.  Ie `"state"` should be a superset of `"if state"`, and I don't think `"state"` wants to get trapped this far down

Otoh I do sometimes find myself wanting to target the `if` in python comprehensions

cc/ @AndreasArvidsson wdyt

AndreasArvidsson:
I guess it's similar to trying to target a ternary in other languages. I do agree that if statement isn't the best description for any of those, but it probably would be useful to be able to target them.

pokey:
Discussed in meet-up:

- [ ] We'd like to switch this to `condition`
- [ ] For "change" it doesn't include the `if` keyword
- [ ] For `"chuck"` it _**does**_ include the `if` keyword

pokey:
@purpleP fyi we are now using `"condition"` instead of `"if state"` here

src/test/suite/fixtures/recorded/languages/python/ditchCondition.yml
```diff
@@ -0,0 +1,23 @@
...
+  usePrePhraseSnapshot: true
+  action: {name: remove}
+initialState:
+  documentContents: "[aaa for aaa in bbb if ccc]"
+  selections:
+    - anchor: {line: 0, character: 1}
+      active: {line: 0, character: 1}
+  marks: {}
+finalState:
+  documentContents: "[aaa for aaa in bbb ]"
```
auscompgeek, line 19: Hmm, I was expecting the whitespace to disappear too. It looks like it does in the Rust implementation, wonder what's up here.

src/languages/rust.ts
```diff
@@ -146,6 +146,7 @@ const nodeMatchers: Partial<
   ),
   string: ["raw_string_literal", "string_literal"],
   ifStatement: ["if_expression", "if_let_expression"],
+  condition: leadingMatcher(["*.match_pattern![condition]"], ["if"]),
```
auscompgeek, line 149: I'm surprised this wasn't implemented in the first place. We should make sure this handles `if` and `while` conditions too.

Open question for how to handle `if let` and `while let`, though I imagine it should take the whole `let <match_pattern> = <value>`.

auscompgeek (changes requested):

src/test/suite/fixtures/recorded/languages/python/ditchCondition.yml
pokey: We generally consider this whitespace cleanup out of scope for scope types, as a) we have plans improve this stuff in the future in a more generic way, and b) an autoformatter can fix this one easily.  I'd be inclined to not bother fixing unless you feel strongly

src/languages/rust.ts
pokey: Done

src/test/suite/fixtures/recorded/languages/python/ditchCondition.yml
auscompgeek: Fair enough. Just struck me as slightly odd; Rust condition not working as intended was the big one.

auscompgeek (approved):

PR #1149, opened 2022-11-18
Add `"branch"` scope type

pokey:
- Partially addresses https://github.com/cursorless-dev/cursorless/issues/1173
- Split from https://github.com/cursorless-dev/cursorless/pull/831

## Checklist

|             | Python | Rust | Java | Typescript | C++ | C#  |
| ----------- | ------ | ---- | ---- | ---------- | --- | --- |
| Switch-case | ✅     | ✅   | ✅   | ✅         |     |     |
| If-else     | ✅     | ✅   |      | ✅         |     |     |
| Try-catch   | ✅     |   N/A   |      |     ✅       |     |     |
| Ternary     |   ✅     | ✅   |      |     ✅       |     |     |

Co-authored-by: Michael Doronin <warrior2031@mail.ru>


pokey:
Closing until someone has bandwidth to finish this PR

pokey:
After discussion in meet-up:

Just kidding; let's get this thing in for the languages that exist here.  more useful to at least have it functional in some places

src/test/suite/fixtures/recorded/languages/python/clearBranch14.yml
```diff
@@ -0,0 +1,34 @@
...
+      pass
+    else:
+      pass
+  selections:
+    - anchor: {line: 0, character: 0}
+      active: {line: 0, character: 0}
+  marks: {}
+finalState:
+  documentContents: |-
+
```
auscompgeek, line 26: It feels a bit weird for `branch` to work on the `try` block itself 🤷

auscompgeek:
We should also test:

- [ ] Python for/else and while/else
- [ ] Rust `if let` and let/else

src/test/suite/fixtures/recorded/languages/python/clearBranch14.yml
pokey: Yeah I'm not entirely sure what the use case for that is tbh 😅, but seemed consistent, and maybe someone would want to use it with the `copy` action?

auscompgeek:
> Unfortunately that is blocked on #1201 😕

Ah, I was afraid as such.

cursorless-talon/src/modifiers/scopes.py
```diff
@@ -46,6 +46,7 @@
     "selector": "selector",
     "state": "statement",
     "string": "string",
+    "branch": "branch",
```
auscompgeek, line 49: nit: any reason this is listed in this order in particular?

auscompgeek (approved):
I think the implicit "this" change got cherry-picked into this branch for some reason, but otherwise LGTM.

I presume it should be trivial to reuse the if/else matchers for Java and C++ like we did for Rust, and I hope the same for try/catch, so might be worth doing it here anyway.

pokey:
> I think the implicit "this" change got cherry-picked into this branch for some reason, but otherwise LGTM.

Strange.  That's already merged into main anyway, but looks like the parentage is strange here; I'll have a look

> I presume it should be trivial to reuse the if/else matchers for Java and C++ like we did for Rust, and I hope the same for try/catch, so might be worth doing it here anyway.

Unfortunately not 😕.  Java, C#, and C++ don't have a special node for else clauses, so the logic gets more complicated.  Should be doable, but not just a quick afterthought like one would hope.  I'd be inclined to leave for follow-up work

cursorless-talon/src/modifiers/scopes.py
pokey: Not really.  I should prob file a follow-up to alphabetise; didn't want to mess with the diff

auscompgeek (approved):

PR #1157, opened 2022-12-01
Limit the amount of text used in `scout` and `scout all`

### LINKED ISSUE #943: Limit the amount of text used in `scout` and `scout all`
If a lot of text is selected for `scout` or `scout all` it can lock-up talon and your computer for quite a while typing all that text into the find text field. If you select a far too large selection of text you may have to wait a long time before you can do anything again or be forced to crash/quit talon as the text will be inserted into any application/text entry not just cursorless.

I came across this while editing a 1000 line file which had a syntax error in it and due to the invalid code tree-sitter decided that I wanted to search for the entire file, which after about 2 minutes was still not done entering in!

To avoid this kind of situation I think it'd be great if we could limit the amount of text that can be entered in the search fields by 200 characters.

For example; `scout ____` with bad syntax selects the entire document, we would only insert the first 200 characters of the document into the search field to avoid the bad edge case mentioned above.

pokey:
Fixes #943

## Checklist

### DIFF
--- a/src/actions/Find.ts
+++ b/src/actions/Find.ts
@@ -1,4 +1,5 @@
 import { commands } from "vscode";
+import ide from "../libs/cursorless-engine/singletons/ide.singleton";
 import { Target } from "../typings/target.types";
 import { Graph } from "../typings/Types";
 import { ensureSingleTarget } from "../util/targetUtils";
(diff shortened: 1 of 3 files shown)


src/actions/Find.ts
```diff
@@ -1,4 +1,5 @@
 import { commands } from "vscode";
```
pokey, line 1: Tbh I'm not entirely sure whether we need it for "scout all", as that just passes the text as a VSCode command arg, but I figure probably doesn't hurt?  And guessing could result in VSCode getting upset if you do it on a gigantic file

src/actions/Find.ts
```diff
@@ -12,10 +13,21 @@ export class FindInFiles implements Action {
...
+      query = text.substring(0, 200);
+      ide().messages.showWarning(
+        "truncatedSearchText",
+        "Search text is longer than 200 characters; truncating",
+      );
+    } else {
+      query = text;
+    }
 
     await commands.executeCommand("workbench.action.findInFiles", {
```
pokey, line 32: Not sure how easy this one is to test until we have this as part of the IDE api

AndreasArvidsson (dismissed):

src/ide/vscode/textLine.test.ts
```diff
@@ -13,7 +13,9 @@ const whiteSpaceTests: [string, number, number][] = [
   [" foo ", 1, 4],
 ];
 
-suite("TextLine", () => {
+suite("TextLine", function () {
+  this.timeout("100s");
+  this.retries(5);
```
pokey, line 18: This is to fix CI for now.  I think the proper fix will be https://github.com/cursorless-dev/cursorless/issues/1158 so we don't keep forgetting to do this

PR #1165, opened 2022-12-09
Treat tuple type arguments as arguments

purpleP:
## Checklist

- [ x ] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)
- [ x ] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [ x ] I have not broken the cheatsheet


pokey:
Thanks!

Looks like this just adds tests; just checking that's intentional?

Also, in what sense are these arguments?  I'd be tempted to have these be "item" rather than "arg", but I don't feel strongly.  cc/ @auscompgeek

purpleP:
> Thanks!
> 
> Looks like this just adds tests; (not that I'm complaining 😄) just checking that's intentional?
> 
> Also, in what sense are these arguments? I'd be tempted to have these be "item" rather than "arg", but I don't feel strongly. cc/ @auscompgeek

Oh, sorry I forgot to add a change itself.

auscompgeek:
> I'd be tempted to have these be "item" rather than "arg"

I agree. I'd expect `take arg eight` to take the parameter foo with its type in this example:
```rust
fn do_something(foo: (u8, u16)) {}
```

We're not treating tuple type items as arguments in TypeScript or Python, so we probably shouldn't do this in Rust either.

pokey:
@purpleP any thoughts on https://github.com/cursorless-dev/cursorless/pull/1165#issuecomment-1345233099?

purpleP:
> @purpleP any thoughts on [#1165 (comment)](https://github.com/cursorless-dev/cursorless/pull/1165#issuecomment-1345233099)?

Yeah, sorry I’m short on time. I don’t have an opinion on this, so I’ve just changed it do use `item`.

pokey:
@AndreasArvidsson fyi this doesn't work out of the box with textual item, I believe because of the `'`

pokey (approved):
lgtm thanks!

pokey:
> While we’re at it, how should one select a whole body of a function? Should we treat it as a `value`?

That will be `"inside"`, but a bit nontrivial to implement that until we have #629

pokey (approved):

PR #1166, opened 2022-12-09
Upgrade typescript and use it for vscode TS server

pokey:
- Factored out of https://github.com/cursorless-dev/cursorless/pull/942
- Closes https://github.com/cursorless-dev/cursorless/issues/1044

## Checklist

- [ ] Make sure this PR doesn't break doc links; see https://github.com/cursorless-dev/cursorless/pull/942#discussion_r968556276
- [ ] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)
- [ ] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [ ] I have not broken the cheatsheet

PR #1167, opened 2022-12-09
Clean up makeGraph.makeGetter

pokey:
Depends on https://github.com/cursorless-dev/cursorless/pull/1166

## Checklist

### DIFF
--- a/src/util/makeGraph.ts
+++ b/src/util/makeGraph.ts
@@ -12,26 +12,24 @@ function makeGetter<GraphType, K extends keyof GraphType>(
   key: K,
 ): () => GraphType[K] {
   return () => {
-    let returnValue: GraphType[K];
-
-    if (components[key] == null) {
-      if (lockedKeys.includes(key)) {
-        const cycle = [...lockedKeys.slice(lockedKeys.indexOf(key)), key].join(
-          " -> ",
-        );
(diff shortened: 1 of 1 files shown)


src/util/makeGraph.ts
```diff
@@ -12,26 +12,24 @@ function makeGetter<GraphType, K extends keyof GraphType>(
...
-      returnValue = factory(graph);
-      if (lockedKeys.pop() !== key) {
-        throw new Error("Unexpected key at top of graph stack");
-      }
-      components[key] = returnValue;
-    } else {
-      returnValue = components[key] as GraphType[K];
+    const componentValue = components[key];
+    if (componentValue != null) {
+      return componentValue;
```
auscompgeek, line 17: If TypeScript is updated we should be able to still check `components[key]` without an intermediate `const`: https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#control-flow-analysis-for-element-access
```suggestion
    if (components[key] != null) {
      return components[key];
```

pokey: Nice. Fwiw I'll prob just close this PR tho. I factored it out of another PR, but we're moving to pure DI, so this code is all going to go away

pokey:
Closing, as we're moving to pure DI

PR #1188, opened 2022-12-17
Fix State type holes

auscompgeek:
Currently only a single state key is in the `State` interface. However, if another state key were to be added with a non-`boolean` type, it would cause type problems, as all state `value`s would have the type `StateType[keyof StateType]`, which is a _union_ of all types of the values in `StateType`.

For example, if a `string` `foo` were to be added, then the types would allow setting `hideInferenceWarning` to a `boolean | string`, and all calls to `State#get()` would have a return type of `boolean | string`.

This fixes the footgun, and renames a few things to be clearer.

## Checklist


src/libs/common/ide/fake/FakeGlobalState.ts
```diff
@@ -1,14 +1,15 @@
-import { State, StateKey, StateType, STATE_KEYS } from "../types/State";
+import type { State, StateData, StateKey } from "../types/State";
+import { STATE_DEFAULTS } from "../types/State";
 
 export default class FakeGlobalState implements State {
-  private readonly data: Map<string, any> = new Map();
+  private readonly data: StateData = { ...STATE_DEFAULTS };
```
pokey, line 5: How come you switched to a map? Just to make typing easier?

Fwiw I believe @AndreasArvidsson prefers maps to objects, so let's give him a chance to have a look at this one

pokey (approved):
Nice catch! Looks great; see comment below tho

src/libs/common/ide/fake/FakeGlobalState.ts
auscompgeek: Yeah, it wasn't difficult to avoid using `any` here. I don't think there's really any benefit to using a `Map` here since we define the defaults in an object literal anyway.

pokey: Ok I got nerd sniped.  We could keep it a map and maintain strong types by doing something like [this](https://www.typescriptlang.org/play?#code/PTAEF5K6tARApgMwJYDsUBcUHs0GcIZiYAoTATwAcFQBVDPAFRwEk1MEAnfBAY2x4APHQB8EUKVCgAFHVAIAHpzQATQgEM0FUAH5ZAawBc9AJThRANxwpVoE2gSXuphcoRrCMmcdDok3KCs5lY2qq76rPagjs5cpOTUtAAKGihcTEksALIaVEJM4uCgTG4q6qAA2v6BANIANH5oAVygAGoAunqgufkNbeIOTtwA3AmUNKAA8gBGAFb8mCyp6fgFZR4VOPOLRaAA3lLSVbVNoAYIFDhIJR0mlQ0lDx0dpAC+lRdXN0yvpMAAKgBoAAwng4phNGhQNsFgJQJgcKANKB8JguHgAOYAGx0EwQdgAtnlQADgIlJrM4UscL12JweItcGh1kpyoRYbsJAxmSx6dxeAJmUIVhksrS8kIqYtlmkeAVRIqEv8wCRoKAAMooQlUbG0Thoohq41QCm0ABiOCRxX2ACINLaHABXQkzbiNW18R2o9HoTFvUh8PCGmYaLgmaUCHJ5fmMoXCS04PaOADuPTyMlMgeDmFAoYAXhJQ1wAHSYhCYGSe22uEAIpJ+Qhorh+0AAH1ATrUyHQBOVdZNxRBTp4OC4evwTcwGk49ZohEHpCUVDHuaDBFzGqYAEEmABRAD6cD35u3dAAMkwNRJDtIABa2BDsFoePgIADqYYwaExJiQGmxXh6neMZl1XOdaA1adODgGcUWKfFrk1Hd9yPE8z0vDUxnxTVoIQWpLgkL4kKgmcEFg6cxhVeheBiBA00jGk6Q4AUmTwCCzVwsjegkRjoyoWNBUEFlSJguDRGzDcfW4vITFEhAeOKVN0yoTMEiQLt42hctMDaACnQQIRTjZTZCHkgiKFEHxLhMWpTDkvCKI0Z4DiOaQuArEdoTRGSqDLCtrIoCJ9C3XdD2PU8LyvT5LleN4gA).  Not necessarily recommending that but it was a fun exercise 😄

pokey: Ok I think even if we decide we like maps in general, this situation is not a great fit: I think they make more sense when the structure is homogeneous, given how difficult / impossible it is to cajole the type system into supporting fine-grained types for maps

PR #1207, opened 2023-01-03
Fix extract action for python

### LINKED ISSUE #461: Fix extract action in python
Should just be a matter of finding the right quick action identifier, tho we might then need to make it language specific

Will need to tweak the `kind` in https://github.com/cursorless-dev/cursorless/blob/40ff102f8f8b64144e5536e1afd2660b7ab595d8/src/ide/vscode/VscodeTextEditorImpl.ts#L192

pokey:
- Fixes #461

## Checklist


AndreasArvidsson:
@pokey Looks good but I'm wondering if we should reduce the trailing sleep a bit. Might just nocebo, but felt like it was unnecessarily slow now.

AndreasArvidsson:
Agreed

PR #1213, opened 2023-01-05
Fix modal keyboard interface broken by #1094

pokey:
We were using the equivalent of "take this" to set the target to be the current selection when you enter cursorless keyboard mode, but that obviously broke when we made "take this" refer to token.  This PR effectively changes the command we use to `"highlight just this"`

## Checklist


auscompgeek (approved):

auscompgeek:
For posterity, this is fixing the selection being expanded to the current token instead of starting modal keyboard mode, when the cursor is on a token. (Starting modal keyboard mode on an empty line still works before this fix.)

PR #1221, opened 2023-01-10
Remove VSCode from highlights code

pokey:
## Checklist

- [ ] Add new "clear highlights" command that will remove all highlights in active editor


src/libs/common/types/GeneralizedRange.ts
```diff
@@ -0,0 +1,37 @@
...
+   * Last line, inclusive
+   */
+  end: number;
+}
+
+export interface CharacterRange {
+  type: "character";
+  start: Position;
+  end: Position;
+}
```
pokey, line 18: This is where I landed with the two different range types.  I wouldn't say I'm totally in love with it, but it's cleaner than the other things I tried 🤷

src/processTargets/targets/BaseTarget.ts
```diff
@@ -87,7 +87,7 @@ export default abstract class BaseTarget implements Target {
     return "edit";
   }
 
-  getRemovalHighlightRange(): Range | undefined {
+  getRemovalHighlightRange(): Range {
```
pokey, line 90: not sure why this was nullable.  we don't have any implementations that return null, so I made it required to remove noisy null checks

src/scripts/transformRecordedTests/upgradeDecorations.ts
```diff
@@ -0,0 +1,88 @@
+import { TestCaseFixture } from "../../testUtil/TestCaseFixture";
```
pokey, line 1: we could probably remove this, but it is kinda useful to have one sitting around, even if we also have an example in the docs 🤷

src/test/suite/fixtures/recorded/decorations/carveLineHarp.yml
```diff
@@ -0,0 +1,38 @@
...
+  clipboard: "hello world"
+ide:
+  flashedRanges:
+    - style: referenced
+      range:
+        type: character
+        start: {line: 0, character: 4}
+        end: {line: 0, character: 15}
+    - style: pendingDelete
+      range: {type: line, start: 0, end: 0}
```
pokey, line 37: notice the new flash style for `"carve line"`: we highlight the whole line as deleted, but show just the content as referenced

src/libs/common/types/GeneralizedRange.ts
```diff
@@ -0,0 +1,37 @@
+import { Position } from "./Position";
+import { Range } from "./Range";
+import { TextEditor } from "./TextEditor";
+
+export interface LineRange {
```
pokey, line 5: docstring

src/libs/common/ide/types/ide.types.ts
```diff
@@ -173,6 +176,28 @@ export interface IDE {
...
+
+  /**
+   * Set the ranges to which {@link highlightId} should be applied.  Removes
+   * the given highlight from all other ranges in all editors.
+   *
+   * @param highlightId The id of the highlight to apply
+   * @param ranges The ranges to apply the highlight to
+   */
+  setHighlightRanges(
+    highlightId: HighlightId,
```
pokey, line 198: make this an enum with two values

src/actions/Highlight.ts
```diff
@@ -10,12 +12,19 @@ export default class Highlight implements Action {
 
   async run(
     [targets]: [Target[]],
-    styleName: EditStyleName = "highlight0",
+    highlightId: HighlightId = "highlight0",
```
pokey, line 15: remove default value

src/libs/common/ide/types/ide.types.ts
```diff
@@ -173,6 +176,28 @@ export interface IDE {
...
+  flashRanges(flashDescriptors: FlashDescriptor[]): Promise<void>;
+
+  /**
+   * Set the ranges to which {@link highlightId} should be applied.  Removes
+   * the given highlight from all other ranges in all editors.
+   *
+   * @param highlightId The id of the highlight to apply
+   * @param ranges The ranges to apply the highlight to
+   */
+  setHighlightRanges(
```
pokey, line 197: file issue to migrate this to `TextEditor`, as we don't want to remove highlights from other editors when we set in one editor, so this api seems confusing

src/libs/vscode-common/testUtil/toPlainObject.ts
```diff
@@ -13,6 +22,30 @@ export type RangePlainObject = {
...
+}
+
+interface PlainHighlight {
+  highlightId: string;
+  ranges: GeneralizedRangePlainObject[];
+}
+
+export interface PlainSpyIDERecordedValues {
+  messages: Message[] | undefined;
+  flashedRanges: PlainFlashDescriptor[] | undefined;
```
pokey, line 45: call this `flashes`

src/libs/common/ide/spy/SpyIDE.ts
```diff
@@ -1,27 +1,56 @@
...
 import SpyMessages, { Message } from "./SpyMessages";
 
+interface Highlight {
+  highlightId: HighlightId;
+  ranges: EditorGeneralizedRange[];
+}
+
 export interface SpyIDERecordedValues {
   messages?: Message[];
+  flashedRanges?: FlashDescriptor[];
```
pokey, line 15: also change this to `flashes`

src/util/itertools.ts
```diff
@@ -24,3 +24,19 @@ export function groupBy<T, U>(list: T[], func: (element: T) => U): Map<U, T[]> {
 
   return map;
 }
+
+export function partition<T, U>(
```
pokey, line 28: docstring

src/libs/common/ide/types/ide.types.ts
pokey: for now just add `editor` param and have list of generalised ranges (without editors) instead of `EditorGeneralizedRange`

src/actions/Highlight.ts
```diff
@@ -10,12 +15,33 @@ export default class Highlight implements Action {
...
+      const { activeTextEditor } = ide();
+
+      if (activeTextEditor == null) {
+        throw Error(
+          "The `highlight nothing` command requires an active text editor",
+        );
+      }
+
+      await ide().setHighlightRanges(highlightId, activeTextEditor, []);
+    } else {
```
pokey, line 36: I just ended up putting a special case in here, because otherwise we'd break `"highlight nothing"` if we didn't have a migration, which doesn't seem worth the effort for something that is confined to an action that I'm guessing is very uncommonly used

pokey:
@AndreasArvidsson feedback addressed

AndreasArvidsson (approved):

PR #1229, opened 2023-01-15
Add xaml support

fatalerrorx:
I have also got a pull request to update the visual studio code  parse tree  to add support for XAML.  I guess before something like this can be merged  it would need to be merged  first.


pokey:
see https://github.com/cursorless-dev/vscode-parse-tree/pull/59#issuecomment-1384231692

PR #1246, opened 2023-02-03
Remove VSCode from CommandRunner and Cheatsheet

pokey:
## Checklist


src/core/Cheatsheet.ts
```diff
@@ -28,104 +26,76 @@ interface CheatSheetCommandArg {
   outputPath: string;
 }
 
-export default class Cheatsheet {
-  private disposables: vscode.Disposable[] = [];
+export async function showCheatsheet({
```
pokey, line 29: note: we don't bother with a class at all; we just export a couple functions here

also note that this file doesn't include any meaningful changes; just migrate command registration to `extension.ts`

src/ide/vscode/VscodeIDE.ts
```diff
@@ -156,6 +157,29 @@ export default class VscodeIDE implements IDE {
     return this.editorMap.get(editor)!;
   }
 
+  handleCommandError(err: Error) {
```
pokey, line 160: didn't make this part of the `ide` api because we don't really have a need at this point 🤷.  Only accessed from composition root

AndreasArvidsson (approved):

PR #1248, opened 2023-02-03
Remove remaining vscode references

pokey:
## Checklist


src/libs/common/ide/types/Messages.ts
```diff
@@ -15,7 +22,8 @@ export interface Messages {
    * @returns The option selected by the user, or `undefined` if no option was
    * selected
    */
-  showWarning(
+  showMessage(
```
pokey, line 25: Made it so that the interface that each ide impl needs to conform to includes just one function, and then added utility wrappers `showWarning`, `showError`, etc

src/util/unifyRanges.ts
```diff
@@ -1,94 +0,0 @@
-import { Range } from "vscode";
```
pokey, line 1: looks like we don't use this file anymore 🤷

src/util/unifyRanges.ts
```diff
@@ -1,41 +1,7 @@
...
 import { targetsToContinuousTarget } from "../processTargets/processTargets";
 import { Target } from "../typings/target.types";
 import { groupTargetsForEachEditor } from "./targetUtils";
 
-/** Unifies overlapping/intersecting ranges */
```
pokey, line 6: looks like we don't use these functions anymore 🤷

AndreasArvidsson (approved):

PR #1265, opened 2023-02-12
Change `"highlight nothing"` to target all editors

pokey:
The new behaviour broke my video recorder, because it didn't clear highlights if I targeted something on other side of screen

## Checklist


pokey:
Yeah I think ideally these would be separate commands; ie not even Cursorless actions at all.  We can add those in a follow-up if people ask for them.  But for backwards compatibility I think it would be good to just maintain the previous behaviour

AndreasArvidsson (approved):

PR #1268, opened 2023-02-14
Infer line scope from line number marks

### LINKED ISSUE #1247: "row" should infer a scope type of line
"row" does not infer the scope type the way "line" does. If I say e.g. "move line air before bat", the line moves as expected. But if I say "move row 42 before bat", the two lines are merged without a newline in between.

(originally discussed in Slack)

_edit (@pokey)_: we need to be sure we properly handle "move row five past ten after air", as I believe the mark type there will be a `rangeMark`, rather than a `lineNumberMark`

AndreasArvidsson:
Fixes #1247 

## Checklist


pokey (approved):
He's back! 🙌

PR #1270, opened 2023-02-14
Use raw target that mark on remove action

### LINKED ISSUE #1253: `that` target for `"chuck"` should be a raw target
In the following document:

```
aaa bbb
```

Saying `"chuck leading bat"`, followed by `"take that"`, selects `aaabbb`, but it should leave an empty cursor in the middle of `aaabbb`. This bug is due to our empty expansion behaviour. I think a reasonable way to avoid this behaviour is to make the output of `"chuck"` a raw target.

AndreasArvidsson:
Fixes #1253 

## Checklist

### DIFF
--- a/src/actions/Clear.ts
+++ b/src/actions/Clear.ts
@@ -24,17 +24,15 @@ export default class Clear implements Action {
         }),
     );
 
-    const { thatSelections: thatMark } = await this.graph.actions.remove.run([
-      plainTargets,
-    ]);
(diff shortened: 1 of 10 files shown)


pokey (approved):
good stuff!  I'm assuming you tried the example from the description?  I agree we don't want a new test for it; existing tests capture it well.  worth a sanity check tho

AndreasArvidsson:
Yeah I test it

PR #1280, opened 2023-02-26
Disallow type import exceptions in `.eslintrc`

pokey:
This PR cleans up our cross-package dependency graph even more by removing our exceptiosn for type imports, as well as restricting which packages can import each other via `@cursorless/foo` imports

## Checklist


packages/cursorless-vscode/extension.ts
```diff
@@ -64,6 +55,11 @@ export async function activate(
     injectIde(vscodeIDE);
   }
 
+  const commandServerApi =
+    vscodeIDE.runMode === "test"
+      ? getFakeCommandServerApi()
+      : await getCommandServerApi();
```
pokey, line 61: Note that during testing we have a proper fake command server api now, which is used by [`prePhraseSnapshot.test.ts`](https://github.com/cursorless-dev/cursorless/pull/1280/files#diff-aa3d46ba59620c5260c9095f28685e60b9ed868f6092b68038030f6f39e2e13a)

packages/vscode-common/getExtensionApi.ts
```diff
@@ -1,41 +1,42 @@
 import type {
```
pokey, line 1: This file is worth a look, as this was where most of the violations were coming from

packages/vscode-common/getExtensionApi.ts
```diff
@@ -1,41 +1,42 @@
...
-interface TestHelpers {
-  graph: Graph;
+export interface TestHelpers {
   ide: NormalizedIDE;
   injectIde: (ide: IDE) => void;
 
-  // FIXME: Remove this once we have a better way to get this function
-  // accessible from our tests
-  plainObjectToTarget(
+  hatTokenMap: HatTokenMap;
```
pokey, line 22: Note that the `HatTokenMap` is now [an interface](https://github.com/cursorless-dev/cursorless/pull/1280/files#diff-1bef5c2e48c3c43737fa37e4ccbbfe57fa9354e889dd3880d05a13ef845cd79b), whereas before it was the type of [the class](https://github.com/cursorless-dev/cursorless/pull/1280/files#diff-9f94f61c925ef52b2b18b010d6ba7d6a5d8ac7367bf989b55ef242ff47cb01f5) that implemented our hat token map

packages/vscode-common/getExtensionApi.ts
```diff
@@ -1,41 +1,42 @@
...
+
+  commandServerApi: CommandServerApi;
```
pokey, line 24: We now pass along the command server api object, which will be a fake, because we are testing.  Before we got it off the graph

packages/vscode-common/getExtensionApi.ts
```diff
@@ -1,41 +1,42 @@
...
-  TestCaseSnapshot,
-  ThatMark,
-} from "@cursorless/cursorless-engine";
 import * as vscode from "vscode";
 import type { SyntaxNode } from "web-tree-sitter";
-import type { SnippetMap } from "../cursorless-engine/snippets/snippet.types";
-import type { Target } from "../cursorless-engine/typings/target.types";
 
-interface TestHelpers {
-  graph: Graph;
```
pokey, line 23: We no longer expose the graph to tests

packages/vscode-common/getExtensionApi.ts
```diff
@@ -1,41 +1,42 @@
...
   injectIde: (ide: IDE) => void;
 
-  // FIXME: Remove this once we have a better way to get this function
-  // accessible from our tests
-  plainObjectToTarget(
+  hatTokenMap: HatTokenMap;
+
+  commandServerApi: CommandServerApi;
+
+  toVscodeEditor(editor: TextEditor): vscode.TextEditor;
```
pokey, line 26: this is needed by [`groupByDocument.test.ts`](https://github.com/cursorless-dev/cursorless/pull/1280/files#diff-1b735f813e740060570a879c9bad8493961c73dd1a618a9f9934ba9f88020f0a).  I think it's better to pass it out here rather than test harness directly importing it, as uses dangerous implementation details

packages/vscode-common/getExtensionApi.ts
```diff
@@ -48,9 +49,6 @@ interface TestHelpers {
 }
 
 export interface CursorlessApi {
-  thatMark: ThatMark;
-  sourceMark: ThatMark;
```
pokey, line 52: note that we no longer expose `that` marks to tests.  They were used for two reasons:

1. Setting `that` mark at beginning of test.  We now do that by exposing a `setThatMark` function in our test helpers, which accepts a plain object rather than a full target, and does rehydration inside the engine rather than letting test harness do it
2. Checking the final that mark.  We now do that by exposing a modified `takeSnapshot` function that already has its own reference to the `ThatMark` objects

packages/vscode-common/getExtensionApi.ts
```diff
@@ -1,41 +1,42 @@
...
+  setThatMark(
     editor: vscode.TextEditor,
-    plainObject: TargetPlainObject,
-  ): Target;
+    targets: TargetPlainObject[] | undefined,
+  ): void;
+  setSourceMark(
+    editor: vscode.TextEditor,
+    targets: TargetPlainObject[] | undefined,
+  ): void;
```
pokey, line 35: This is 1) from comment above

packages/vscode-common/getExtensionApi.ts
```diff
@@ -1,41 +1,42 @@
...
 
   // FIXME: Remove this once we have a better way to get this function
   // accessible from our tests
   takeSnapshot(
-    thatMark: ThatMark | undefined,
-    sourceMark: ThatMark | undefined,
```
pokey, line 38: See how we've removed these from signature of `takeSnapshot` exposed to test cases; we provide them ourselves here https://github.com/cursorless-dev/cursorless/pull/1280/files#diff-08aaf0534aa1822e2fe091bc20a807b628da66d0dfa64c85629f43fc7f5e2ffbR52-R53

pokey:
Note that it might be easier to review these commits https://github.com/cursorless-dev/cursorless/pull/1280/files/789e19a25319015e164663de1a104cbdb7899f71..4385fdcbd481cd08f38f37b60c5e044f39cae28c, as the first two commits just touch a bunch of `.eslintrc.json` files

packages/common/ide/.eslintrc.json
```diff
@@ -5,7 +5,7 @@
       {
         "patterns": [
           {
-            "group": ["../../*"],
+            "group": ["@cursorless/*", "../../*"],
```
auscompgeek, line 8: Hm, don't nested eslintrcs cascade? This looks identical to the parent one.

Is it explicitly overriding this rule for some reason?

pokey: The relative exclusions (`../../`) need to be different at each directory level

Note that all of these eslint files will be going away in the next few days when we migrate to pnpm. Just added these to help me spot issues before migration

AndreasArvidsson (approved):

Issue #803, opened 2022-06-29
Destinations as first-class objects


pokey:
# The problem

Today, our handling of positions is inconsistent. This problem manifests itself in multiple ways:

- We allow "end of" to appear in places that expect a position target, which means that `"paste end of air"` is a valid command, making it nearly impossible to get Talon to hear `"take air past end of item bat"`. It instead hears two commands: `"take air"` then `"paste end of item bat"`
- We need to have the `edit` action on regular targets, and position targets need all the methods from regular targets that they don't actually support
- Targets have no strongly typed way of indicating that they expect a position
- Position targets are hacked to set delimiter to `""` when position is `"end of"` or `"start of"`

# The solution

We introduce a new type called a `Destination`, both on the Talon and extension side. This type is distinct from a `Target`, and will be used everywhere we expect a `PositionTarget` today (eg `"paste"`, second target of `"bring"`, etc).

`Destination` will be have the following interface:

```typescript
type InsertionMode = "before" | "after" | "to";

interface Destination {
  insertionMode: InsertionMode;
  target: Target;
}
```

It will exist in the following places:

- As a capture Talon side
- As a new descriptor type, similar to today's `TargetDescriptor`
- As a new post-processTargets type, like today's `Target`

It will involve the following steps:

- [ ] https://github.com/cursorless-dev/cursorless/issues/1440
- [x] https://github.com/cursorless-dev/cursorless/issues/1630

PR #1329, opened 2023-03-20
Support inline snippets; add Talon snippet api

### LINKED ISSUE #1324: Support rich object as argument to snippet actions
Something like the following for wrap:

```typescript

interface NamedSnippet {
  type: "named";
  name: string;
  variableName: string;
}
interface CustomSnippet {
  type: "custom";
  body: string;
  variableName?: string;
  scopeType?: ScopeType;
}
type SnippetDescription = NamedSnippet | CustomSnippet;

async run(
    [targets]: [Target[]],
    snippet: SnippetDescription,
  ): Promise<ActionReturnValue> {}
```

insert should be similar but no `variableName`

### LINKED ISSUE #1325: Support talon API for inserting and wrapping with snippets
Something like the following:

```python
def insert_named_snippet(name: str, target: Optional[Target] = None):
    """Inserts a named snippet"""
    pass

def insert_custom_snippet(body: str, target: Optional[Target] = None, scope: Optional[str] = None):
    """Inserts a custom snippet"""
    # Note that we convert scope to a proper scopeType; technically it is a scopeTypeType
    pass

def wrap_with_named_snippet(name: str, variable_name: str, target: Target):
    """Wrap target with a named snippet"""
    pass

def wrap_with_custom_snippet(body: str, target: Target, scope: Optional[str] = None, variable_name: Optional[str] = None):
    """Wrap target with a custom snippet"""
    pass
```

pokey:
- Fixes #1324
- Fixes #1325 

Note that I created a Talon-side insertion api that supports the full set of options, including substitutions, targets, and scopes, but decided to leave it out for now because I'm not sure exactly how it should look and we don't need it for the mathfly support we're planning to use it for.  In particular, we should probably figure out #803 before we implement the more complex api because snippets really want a destination, not a target.  I did use the complex Talon-side api to record some tests of the extension api, though

Here is the complex Talon snippet insertion api in case useful at some point

<details><summary>Complex talon api</summary>

```diff
From e39e03e3a06a6db4f1edb245a5225037d0ea08d3 Mon Sep 17 00:00:00 2001
From: Pokey Rule <755842+pokey@users.noreply.github.com>
Date: Fri, 24 Mar 2023 14:22:40 +0000
Subject: [PATCH] Complex insert snippet Talon api

---
 cursorless-talon/src/snippets.py | 35 ++++++++++++++++++++++++--------
 1 file changed, 27 insertions(+), 8 deletions(-)

diff --git a/cursorless-talon/src/snippets.py b/cursorless-talon/src/snippets.py
index a9f100f58..409512011 100644
--- a/cursorless-talon/src/snippets.py
+++ b/cursorless-talon/src/snippets.py
@@ -91,15 +91,34 @@ class Actions:
             },
         )
 
-    def cursorless_insert_snippet(body: str):
+    def cursorless_insert_snippet(
+        body: str,
+        target: Optional[dict] = None,
+        scope: Optional[str] = None,
+        snippet_variable: Optional[str] = None,
+        text: Optional[str] = None,
+    ):
         """Inserts a custom snippet"""
-        actions.user.cursorless_implicit_target_command(
-            "insertSnippet",
-            {
-                "type": "custom",
-                "body": body,
-            },
-        )
+        snippet_arg: dict[str, Any] = {
+            "type": "custom",
+            "body": body,
+        }
+        if scope:
+            snippet_arg["scopeType"] = {"type": scope}
+        if snippet_variable:
+            snippet_arg["substitutions"] = {snippet_variable: text}
+
+        if target:
+            actions.user.cursorless_single_target_command_with_arg_list(
+                "insertSnippet",
+                target,
+                [snippet_arg],
+            )
+        else:
+            actions.user.cursorless_implicit_target_command(
+                "insertSnippet",
+                snippet_arg,
+            )
 
     def cursorless_wrap_with_snippet_by_name(
         name: str, variable_name: str, target: dict
-- 
2.39.2

```

</details>

## Checklist


cursorless-talon/src/cursorless_snippets.talon
```diff
@@ -8,4 +8,4 @@ tag: user.cursorless
     user.cursorless_single_target_command(cursorless_insert_snippet_action, cursorless_positional_target, cursorless_insertion_snippet)
 
 {user.cursorless_insert_snippet_action} {user.cursorless_insertion_snippet_single_phrase} <user.text> [{user.cursorless_phrase_terminator}]:
-    user.cursorless_insert_snippet_with_phrase(cursorless_insert_snippet_action, cursorless_insertion_snippet_single_phrase, text)
+    user.private_cursorless_insert_snippet_with_phrase(cursorless_insert_snippet_action, cursorless_insertion_snippet_single_phrase, text)
```
pokey, line 11: we should really do this for all of our private actions.  Figured I'd do it here tho to make it super clear that this is not one of the public snippet api functions

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/snippets/customWrapLine2.yml
```diff
@@ -0,0 +1,32 @@
+languageId: plaintext
+command:
+  version: 5
+  spokenForm: custom wrap line
+  action:
+    name: wrapWithSnippet
+    args:
+      - {type: custom, body: 'dummy snippet hole1: ($TM_SELECTED_TEXT), hole2: ($hole2)'}
```
pokey, line 8: Do we want to use `TM_SELECTED_TEXT`, or should we make up our own variable name?

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/testCaseRecorder/takeHarp.yml
```diff
@@ -1,6 +1,6 @@
 languageId: plaintext
 command:
-  version: 4
+  version: 5
```
pokey, line 3: A bit annoying we'll have to bump this one every time we increment command version number, but I think it's not unreasonable: we probably do want to be ensuring that the test case recorder is recording tests using the latest command version

pokey:
Tempted to switch `custom` to `inline` everywhere.  WDYT @AndreasArvidsson ?  We already support "custom" snippets, which means ones that aren't part of the built-in Cursorless snippets, so I think this terminology is confusing

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/testCaseRecorder/takeHarp.yml
AndreasArvidsson: Maybe we should have a test to make sure that this test actually contains the latest version?

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/snippets/customWrapLine2.yml
AndreasArvidsson: I think it's fine.

cursorless-talon/src/cursorless_snippets.talon
AndreasArvidsson: I agree but I think it should be `cursorless_private`. Once we have actual Talon namespaces all of our actions will be prefixed with cursorless.

AndreasArvidsson:
But isn't that more or less the same thing? In one instance you send a custom snippet in the other case you define a custom snippet.

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/testCaseRecorder/takeHarp.yml
pokey: The test case recorder test fails on this one if you don't bump. Are you saying we should add another test?

AndreasArvidsson: Shouldn't this line always be the last person then?
https://github.com/cursorless-dev/cursorless/blob/b07458bdd94e48dcab97f9b961219f5518fe70eb/packages/cursorless-vscode-e2e/src/suite/testCaseRecorder.test.ts#L76

pokey: I believe the command that gets captured by recorder is after we do normalisation / upgrade at the start of running the command. So even if we pass 4 there, the command that gets captured has version 5

As to whether that's the right thing to do, I could go either way. I think slightly better to capture normalised version. If we want to be able to record legacy command, we could add a flag to test case recorder to enable that

AndreasArvidsson (approved):

pokey:
The confusing thing is that if you want to use one of the custom snippets you've defined, you don't use custom snippet arg, you use named snippet

cursorless-talon/src/cursorless_snippets.talon
pokey: I'm pretty sure we had this discussion a while back and decided on `private_cursorless`, so that it was super obvious that it's not part of our public api.  That way our public api is everything that starts with `cursorless_`.  Re using `_cursorless`, I'm pretty sure we're not allowed to prefix our actions with `_` in Talon

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/snippets/customWrapLine2.yml
pokey: It's just a bit confusing because it's not actually selected text that we're targeting; it's the target.  I'd be tempted to do `$CURSORLESS_TARGET`.  But I guess using `$TM_SELECTED_TEXT` means that we'll work out of the box for any wrapper-style snippets people already have

pokey: let's leave it for now

pokey:
- `cursorless_insert_custom_snippet` => `cursorless_insert_snippet`
- `cursorless_insert_named_snippet` => `cursorless_insert_snippet_by_name`
- `cursorless_wrap_with_custom_snippet` => `cursorless_wrap_with_snippet`
- `cursorless_wrap_with_named_snippet` => `cursorless_wrap_with_snippet_by_name`

AndreasArvidsson (approved):

PR #1281, opened 2023-02-27
Migrate to a pnpm monorepo

pokey:
- closes https://github.com/cursorless-dev/cursorless/issues/945
- closes https://github.com/cursorless-dev/cursorless/issues/1044
- depends on https://github.com/cursorless-dev/cursorless/pull/1322
- depends on https://github.com/cursorless-dev/cursorless/pull/1327

## Checklist

- [x] Split into smaller PRs for ease of review
- [x] Try DX tests again https://github.com/cursorless-dev/cursorless/discussions/1289
- [x] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [x] File issue for speeding up cheatsheet / making it not optional locally https://github.com/cursorless-dev/cursorless/issues/1307
- [x] Set package manager in package.json https://nodejs.org/api/packages.html#packagemanager
- [x] Re-add lint rule to prevent api types from importing anything
- [x] Try local install
  - [x] Make sure cheatsheet works
- [x] Make sure local extension build depends on building cheatsheet, or allows no cheatsheet
- [x] Figure out why we're getting so many extra `internal` modules
- [x] Update https://github.com/cursorless-dev/cursorless/issues/931 to indicate we're now patching instead of swizzling
- [x] Make sure lint rules are actually running in CI
- [x] Check that we've addressed everything in https://github.com/cursorless-dev/cursorless/issues/945
- [x] Incorporate changes from https://github.com/cursorless-dev/cursorless/pull/1166
- [x] ~~Make sure this PR doesn't break doc links; see https://github.com/cursorless-dev/cursorless/pull/942#discussion_r968556276~~ There's no way to avoid these links getting broken if we upgrade typedoc; filed https://github.com/cursorless-dev/cursorless/issues/1304 to track
- [x] Revert https://github.com/cursorless-dev/cursorless/pull/1284
- [x] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)
- [x] I have not broken the cheatsheet
- [x] Generate cursorless-vscode package.json automatically, and change the name we use to `@cursorless/cursorless-vscode` in the `package.json` checked into source control
  - [x] Be sure to update meta-updater to check for that instead of name `cursorless` (see fixme in its `index.ts`)

## Desiderata

- From https://github.com/cursorless-dev/cursorless/discussions/1289

### Correctly flagging errors

[...]

PR #1341, opened 2023-03-21
Fix test case recorder

pokey:
The test case recorder broke in #1281 due to the fact we're now launching extension in subdirectory called `dist` rather than top-level repo.  This PR fixes it

## Checklist

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -4,7 +4,7 @@
     "moduleResolution": "nodenext",
     "moduleDetection": "force",
     "target": "es6",
-    "lib": ["es2020"],
+    "lib": ["es2022"],
     "sourceMap": true,
     "declarationMap": true,
     "resolveJsonModule": true,
--- a/packages/common/src/ide/fake/FakeIDE.ts
+++ b/packages/common/src/ide/fake/FakeIDE.ts
@@ -35,6 +35,7 @@ export default class FakeIDE implements IDE {
   workspaceFolders: readonly WorkspaceFolder[] | undefined = undefined;
   private disposables: Disposable[] = [];
   private assetsRoot_: string | undefined;
+  private quickPickReturnValue: string | undefined = undefined;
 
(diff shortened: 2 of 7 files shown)


AndreasArvidsson (dismissed):

pokey:
For some reason that I can't fathom, we're getting a CI failure on Windows because even though the test case recorder is _correctly_ outputting a file with LF endings there,  the _expected_ content has CRLF endings. This is quite strange, because we're just reading a gold file that has LF endings.  It's almost like GitHub action runner is turning on `git config core.autocrlf` or something 🤔 cc/ @AndreasArvidsson @auscompgeek any ideas?

tsconfig.base.json
```diff
@@ -4,7 +4,7 @@
     "moduleResolution": "nodenext",
     "moduleDetection": "force",
     "target": "es6",
-    "lib": ["es2020"],
+    "lib": ["es2021"],
```
pokey, line 7: needed to bump this to use `.replaceAll`.  Seemed to work locally; we'll see if it passes on our legacy VSCode test

pokey: Looks like it's fine even on legacy VSCode

pokey:
> It's probably fairly likely this is enabled in the global config

but how / why?  I don't see anywhere that this is something GitHub does in their runners 🤔

auscompgeek:
Maybe we should try setting text=lf in .gitattributes 🤔

pokey:
I'm hesitant to mess with that setting, as I don't fully understand it. It would be great to get this PR in, as the test case recorder is currently broken on main. Shall I file a follow-up to look into line endings?

Leeft:
> Maybe we should try setting text=lf in .gitattributes 🤔

Yeah, that seems like a good idea: https://dev.to/deadlybyte/please-add-gitattributes-to-your-git-repository-1jld

pokey:
Ok I filed #1348 to handle line endings properly

PR #1348, opened 2023-03-22
Check for CRLF line endings in CI

pokey:
## The problem

We currently don't have any checks to ensure that we don't use CRLF line endings, which could result in hard-to-detect issues in our files, as the difference between LF and CRLF isn't always visible

## The solution

In this PR, we now check that all files that are not binary do not contain CRLF line endings.

This PR does the following:

- Adds a `pre-commit` rule to ensure that files that text files don't have CRLF
- Fixes all files in our repo that had CRLF line endings; looks like we had only missed one file
- Removes an old Rope database file that was floating around in our repo for some reason.  Seemed better to remove it than add a rule for it to our new `.gitattributes`
- Throws an exception in test case recorder when developer tries to record a test with line endings as CRLF

## References

- https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/
- https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings
- https://git-scm.com/docs/gitattributes

## Checklist

- [x] Make sure we're checking line endings for files in CI


pokey:
@AndreasArvidsson @wenkokke @leeft see **Consequences** section in PR description above.  Are you comfortable with that?

.gitattributes
```diff
@@ -0,0 +1,58 @@
+# Note that we have the line below just so that files whose extensions aren't
+# listed explicitly will be flagged by `check-gitattributes` as needing to be
+# added to this file
+* text=auto
```
pokey, line 4: ok thinking on this one a bit more, I'm not sure we want this as the default, even if it is  primarily to ensure that we catch unlisted files in CI.  I believe this line will make it so that the first time someone checks in a file with an new file type, we will let git decide whether it is text, which is what we were trying to avoid by using the long list of extensions below.  We'll immediately complain in CI, but by this time git may have already mangled their binary file.  Probably safer to default to `binary` 🤔

Leeft: with `auto` I think you're at a bigger risk from text files seen as binary rather than the other way around. That said, I've not used this particular feature myself (we have .gitattributes to mark compiled JS and CSS as binary, that's it).

pokey: No text files seen as binary is fine.  In that case, `git` won't touch the file when you commit it, and then we'll catch the lack of a pattern fro the new file in CI with the new `pre-commit` rule.  The other direction is the problem, because git could try to convert `CRLF` to `LF` in a binary file, which would be bad

pokey: Fixed in latest commit

pokey:
ok @auscompgeek this one should be good to go.  Is this the kind of thing you had in mind?

pokey:
ok @AndreasArvidsson @auscompgeek here's a simplified version based on discussion between me and @AndreasArvidsson

AndreasArvidsson (dismissed):

AndreasArvidsson:
I think this looks fine

pokey:
Cool. Yeah I'm pretty happy with where it landed; much simpler.

I don't think it's urgent now that #1341 is merged, so let's give @auscompgeek a couple days to have a look as he's the one who originally proposed we handle line endings more properly

.pre-commit-config.yaml
```diff
@@ -12,6 +12,13 @@ repos:
         entry: Forbid symlinks
         language: fail
         types: [symlink]
+  - repo: local
+    hooks:
+      - id: check-line-endings
+        name: check for CRLF line endings
```
auscompgeek, line 18: Isn't there a pre-commit hook in pre-commit-hooks that does this already?

pokey: Not that I could find 🤷‍♂️

auscompgeek: I see https://github.com/pre-commit/pre-commit-hooks#mixed-line-ending

Does `--fix=lf` only rewrite the file if it has mixed line endings?

pokey: Ooh interesting; missed that.  I had assumed mixed was just checking for mixed.  I wonder how they avoid messing with binary files; the advantage of our approach is that we explicitly mark binary files, and we never actually touch line endings, just flag them, in case someone checks something in with a new file ending not in our list

pokey: Does our approach seem ok here @auscompgeek?  I think I like _not_ running as a fixer, as we do here, so that we don't inadvertently break someone's binary on checkin if they have commit hooks set up

pokey: Ok if no more objections in the next 24 hrs then I'm merging this one in @auscompgeek

auscompgeek: > I had assumed mixed was just checking for mixed.

I did too given its name. But then I took a look behind the curtain. 

> I wonder how they avoid messing with binary files

It filters by type:

https://github.com/pre-commit/pre-commit-hooks/blob/b0f11fe246ae49c7a3791009753d2148b97cb677/.pre-commit-hooks.yaml#L165-L170

`types` are documented here: https://pre-commit.com/#filtering-files-with-types

pokey: yeah I think I'm maybe too paranoid for "looks like a binary file" 😅

auscompgeek: I mean, if it's [good enough for libmagic](https://github.com/pre-commit/identify/blob/d70d842caf3d9b06560de4db74a98260236f592a/identify/identify.py#L121), surely it's good enough for us? I can't imagine anyone checking in any binary files where the first 1KB happens to not have any control characters

auscompgeek: FWIW https://github.com/Lucas-C/pre-commit-hooks looks like it has a non-fixing check too

pokey: 🤝

auscompgeek (approved):
LGTM, although we probably want to set `eol=lf` in gitattributes so new devs on Windows don't have to go through a lot of pain when using `pre-commit`.

pokey:
I was originally hoping we could support Windows users smudging to CRLF, but I guess given we specify LF in editorconfig and VSCode settings, as you mention, and the fact that our pre-commit check is now going to complain about it, maybe makes sense to set it

So basically if the user has autocrlf set, this will force disable it?

What if they don't have pre-commit hooks installed? I personally run without them and just let CI do it. In that case they should be able to run with CRLF just fine, so I'm not sure we want to go out of our way here to stop that from happening 🤔

pokey:
Or maybe we should only run the check in CI? I think that sidesteps the whole issue and just directly accomplishes what we're after. Not exactly sure how to do that while leveraging pre-commit's binary detection / hook definition

pokey:
had discussion, and decided let's keep it simple for now, and if they run into https://github.com/cursorless-dev/cursorless/pull/1395#issuecomment-1500445873, we help them get out of it

PR #1358, opened 2023-03-23
Another attempt to fix windows ci

pokey:
## Checklist


packages/test-harness/src/util/launchVscodeAndRunTests.ts
```diff
@@ -44,8 +44,6 @@ export async function launchVscodeAndRunTests(extensionTestsPath: string) {
       cli,
       [
         ...args,
-        "--disable-gpu",
-        "--disable-software-rasterizer",
```
pokey, line 48: accidentally put these in the invocation where we're installing extensions, rather than in the invocation where we're actually running tests

PR #1359, opened 2023-03-24
Fix bug with injected ide in test case recorder

pokey:
This PR fixes a bug with our test case recorder.  If an exception is thrown during command processing before we inject a spy ide during `TestCaseRecorder.preCommandHook`, we don't set the `originalIde` member variable on `TestCaseRecorder`.  During `TestCaseRecorder.finallyHook`, we unconditionally call `injectIde(this.originalIde)` even in the case of an error, which means that we will inject an `undefined` ide, so Cursorless becomes completely broken, because `ide()` is now `undefined`, so we get an error about not having injected ide for every single subsequent command

This PR just checks that `this.originalIde != null` before trying to inject it

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/testCaseRecorder/TestCaseRecorder.ts
+++ b/packages/cursorless-engine/src/testCaseRecorder/TestCaseRecorder.ts
@@ -432,7 +432,9 @@ export class TestCaseRecorder {
   }
 
   finallyHook() {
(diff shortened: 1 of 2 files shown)


pokey:
@AndreasArvidsson I added a test

AndreasArvidsson (approved):

PR #1375, opened 2023-04-01
Switch from shed to ruff

auscompgeek:
https://github.com/charliermarsh/ruff

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,9 +3,6 @@ target-version = ["py39"]
 
 [tool.ruff]
 select = ["E", "F", "C4", "I001", "UP", "SIM"]
-ignore = ["E501", "SIM105"]
+ignore = ["E501", "SIM105", "UP035"]
 target-version = "py39"
 extend-exclude = ["vendor"]
-
-[tool.ruff.pyupgrade]
-keep-runtime-typing = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,11 @@
+[tool.black]
+target-version = ["py39"]
+
+[tool.ruff]
+select = ["E", "F", "C4", "I001", "UP", "SIM"]
+ignore = ["E501", "SIM105"]
+target-version = "py39"
+extend-exclude = ["vendor"]
+
+[tool.ruff.pyupgrade]
+keep-runtime-typing = true
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,7 +43,7 @@ repos:
     hooks:
       - id: flynt
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: 'v0.0.260'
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -0,0 +1,11 @@
+[tool.black]
+target-version = ["py39"]
+
+[tool.ruff]
+select = ["E", "F", "C4", "I001", "UP", "SIM"]
```
auscompgeek, line 5: This enables the closest set of rules that `shed` fixed without being overly specific.

- `F`: pyflakes/autoflake
- [`C4`: flake8-comprehensions](https://beta.ruff.rs/docs/rules/#flake8-comprehensions-c4) - shed has some custom autofixers
- `I001`: isort
- `UP`: pyupgrade
- [`SIM`: flake8-simplify](https://beta.ruff.rs/docs/rules/#flake8-simplify-sim) - I think shed has a couple of these autofixers

pokey (approved):
Looks great to me! @phillco @AndreasArvidsson ?

pyproject.toml
```diff
@@ -0,0 +1,11 @@
...
+ignore = ["E501", "SIM105"]
+target-version = "py39"
+extend-exclude = ["vendor"]
+
+[tool.ruff.pyupgrade]
+keep-runtime-typing = true
```
pokey, line 11: Does this fix that issue where Talon choked on the newer type annotations?

auscompgeek: I _think_ so, but not 100% certain. Here's the docs for this setting: https://beta.ruff.rs/docs/settings/#keep-runtime-typing

pokey: Huh. Strange that it says:

> Note that this setting is only applicable when the target Python version is below 3.9 and 3.10 respectively

So I'm not sure this rule even applies? Maybe instead we should just disable `UP006`? That corresponds to [PEP 585](https://peps.python.org/pep-0585/), which I'm pretty sure is the troublemaker.

pokey: Huh. Strange that it says:

> Note that this setting is only applicable when the target Python version is below 3.9 and 3.10 respectively

So I'm not sure this rule even applies? Maybe instead we should just disable `UP006`? That corresponds to [PEP 585](https://peps.python.org/pep-0585/), which I'm pretty sure is the troublemaker.

pokey: Ok I did some local testing and looks like these rules would still have caused https://github.com/knausj85/knausj_talon/issues/922.  Disabling `UP035` did the trick, though.  I pushed updated config; lmk if it looks ok and we can merge away

Thanks for this one!  So much faster.  Now prettier is the slow horse 😄

pokey (approved):

auscompgeek:
Yep, I was planning to enable more rules as a follow-up. Was just aiming to match shed to start.

PR #1379, opened 2023-04-03
Automatically generate activation events

pokey:
- Depends on #1380 
- Depends on #1378 
- Depends on https://github.com/cursorless-dev/cursorless/pull/1412 (due to use of new typescript `satisfies` operator not supported by older versions of `prettier`)

## Checklist

- [x] ~~Maybe better to rebase onto #1393?~~ should be easy to merge these changes with that so let's not bother
- [x] ~~Add test for real font measurements~~ Filed follow-up https://github.com/cursorless-dev/cursorless/issues/1420


packages/cursorless-vscode/package.json
```diff
@@ -44,46 +44,13 @@
...
-    "*"
-  ],
   "main": "./extension.js",
   "capabilities": {
     "untrustedWorkspaces": {
       "supported": true
     }
   },
   "contributes": {
-    "commands": [
```
pokey, line 57: I do wonder if it would be better to just use meta-updater to keep these up-to-date here rather than generating them during packaging.  I think it may be a bit surprising to have only some of the `contributes` fields here; makes it look like we don't contribute any commands if you don't know how all the machinery works 🤔

packages/cursorless-vscode/src/registerCommands.ts
```diff
@@ -20,76 +25,73 @@ export function registerCommands(
   commandRunner: CommandRunner,
   testCaseRecorder: TestCaseRecorder,
   keyboardCommands: KeyboardCommands,
+  hats: VscodeHats,
 ): void {
-  const commands = [
+  const commands: Record<CursorlessCommandId, (...args: any[]) => any> = {
```
pokey, line 30: you'll now get a type error if you forget to define a command

packages/common/src/cursorlessCommandIds.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+}
+
+class HiddenCommand extends Command implements CommandDescription {
+  readonly isVisible = false as const;
+}
+
+class VisibleCommand extends Command implements CommandDescription {
+  readonly isVisible = true as const;
+}
```
pokey, line 23: Could probably simplify these to a single class, but was planning to add argument schema, and require that visible commands can be run with no args.  Maybe I should wait until I do that before introducing this complexity?  Idk it's only a few lines

packages/cursorless-vscode/package.json
AndreasArvidsson: Less surprise seems like a good goal to have

pokey: done a78177c788bb2106ccf23090bdeaeea1a9221eb5

AndreasArvidsson (approved):

packages/cursorless-vscode-core/src/ide/vscode/hats/FontMeasurements.ts
pokey, line 1: moved to `FontMeasurementsImpl` with slight changes

packages/cursorless-vscode-core/src/ide/vscode/hats/VscodeHatRenderer.ts
```diff
@@ -42,28 +42,19 @@ const hatConfigSections = [
...
   private decorationMap!: HatDecorationMap;
   private disposables: vscode.Disposable[] = [];
-  private fontMeasurements: FontMeasurements;
   private notifier: Notifier<[]> = new Notifier();
   private lastSeenEnabledHatStyles: ExtendedHatStyleMap = {};
 
   constructor(
     private extensionContext: vscode.ExtensionContext,
     private enabledHatStyles: VscodeEnabledHatStyleManager,
+    private fontMeasurements: FontMeasurements,
```
pokey, line 51: now gets injected so we can use a fake during testing

pokey:
Ok @AndreasArvidsson I had to make some tweaks to get CI passing; [here's](https://github.com/cursorless-dev/cursorless/pull/1379/files/fd25e34af3fc41e6af661566b2aef391cbb448ae..5c4d3f9d6fe5ae3a1ff958dd9de4e203fc5d8ed5) a diff.  See #1420 for more info

PR #1389, opened 2023-04-06
Move `hats` back onto `ide()`

pokey:
This PR partially rolls back #1378. After some discussion, we decided that it is nice to have a unified definition of what we expect from an ide.  The main benefit of #1378 was making the dependencies of our components more granular.  We'd like to still have a unified interface (`ide`) that defines the api surface of an ide, but to only refer to that interface from the composition root.  The composition root will divvy out the pieces that components depend on

## Checklist

### DIFF
--- a/packages/common/src/ide/PassthroughIDEBase.ts
+++ b/packages/common/src/ide/PassthroughIDEBase.ts
@@ -10,6 +10,7 @@ import {
   TextEditorVisibleRangesChangeEvent,
 } from "./types/events.types";
 import { FlashDescriptor } from "./types/FlashDescriptor";
+import { Hats } from "./types/Hats";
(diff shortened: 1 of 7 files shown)


pokey:
@auscompgeek curious to hear your thoughts here

auscompgeek:
Hmm, is the plan to remove the singleton?

pokey:
> Hmm, is the plan to remove the singleton?

Yep!

pokey:
Thought about this some more and I think I have a better approach, implemented in https://github.com/cursorless-dev/cursorless/pull/1393.  We decided that the `@cursorless/cursorless-engine` package should have its own composition root, which is basically a factory function that it exports.  As I was sketching that out, I realised that the parameters to the factory function indicate what's expected of any client that wants to instantiate the Cursorless engine

PR #1394, opened 2023-04-07
Remove graph

pokey:
## Checklist


packages/cursorless-engine/src/actions/EditNew/EditNew.ts
```diff
@@ -1,8 +1,8 @@
+import { RangeUpdater } from "../..";
```
auscompgeek, line 1: Why the inconsistency in the import path?

pokey: Getting auto-imported this way.  A bit annoying but I don't think it's worth the fight to fix it every time 🤷‍♂️

auscompgeek: Huh. Odd it only happened once though.
```suggestion
import { RangeUpdater } from "../../core/updateSelections/RangeUpdater";
```

AndreasArvidsson (approved):

PR #1395, opened 2023-04-07
Run the check-line-endings script in GHA

auscompgeek:
Opening this to see what happens on Windows. I'm not convinced this script won't break dev workflows if this is run in pre-commit, so it should be tested cross-platform.

## Checklist


pokey:
those CI failures are because it expects to be run with a list of paths, eg `bash ./scripts/check-line-endings.sh foo.js bar/baz.txt`

pokey:
also, you'll get feedback a bit faster if you put it earlier in the action 😅

auscompgeek:
Huh, `**` isn't a thing in bash?

pokey:
huh i thought it was 🤔

pokey:
ok well there's your windows failure.  strange, though; not the failure I would have anticipated

auscompgeek:
I believe `pre-commit` runs in the worktree, but we can check its behaviour too

pokey:
huh ok fair enough.  so I guess with their lint rule, it would still fail, but it would fix them?

auscompgeek:
I guess... but given we have tests that depend on the line endings being correct we should probably set the line endings for those particular file types in `.gitattributes` so it's not an issue in the first place

pokey:
> given we have tests that depend on the line endings being correct

That's actually not the case.  See https://github.com/cursorless-dev/cursorless/blob/68ba545520829a21e26d70437c650fb615d73344/packages/cursorless-vscode-e2e/src/suite/testCaseRecorder.test.ts#L162-L164

> and we have otherwise conflicting settings with git defaults:
> 
> https://github.com/cursorless-dev/cursorless/blob/68ba545520829a21e26d70437c650fb615d73344/.editorconfig#L6
> 
> https://github.com/cursorless-dev/cursorless/blob/68ba545520829a21e26d70437c650fb615d73344/.vscode/settings.json#L23
> 
> we should probably set the line endings for particular file types in `.gitattributes` so it's not an issue in the first place

That was the [original approach](https://github.com/cursorless-dev/cursorless/compare/main...4f6f182210a1190ed7976d2b24d86bcf243338b0), but we moved away from that for two reasons:

1. Maintaining the list of text file types was cumbersome and added complexity
2. We didn't want to force Windows users to check files out with LF if they really wanted to use CRLF smudging.  Though maybe the original approach actually would work ok with such a setup?

pokey:
Closing in favour of https://github.com/cursorless-dev/cursorless/pull/1348

PR #1393, opened 2023-04-07
Add composition roots for packages

pokey:
## Checklist

- [ ] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)
- [ ] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [ ] I have not broken the cheatsheet
- [x] https://github.com/cursorless-dev/cursorless/pull/1379/files#r1164301275

PR #1396, opened 2023-04-07
Add factories for stages and scope handlers

pokey:
- Depends on #1393

## Checklist


packages/cursorless-engine/src/compositionRoot/createCursorlessEngine.ts
```diff
@@ -33,7 +36,12 @@ export function createCursorlessEngine(
 
   const testCaseRecorder = new TestCaseRecorder(hatTokenMap);
 
-  const actions = new Actions(snippets, rangeUpdater);
+  const scopeHandlerFactory = new ScopeHandlerFactoryImpl();
```
pokey, line 39: @AndreasArvidsson notice how the `scopeHandlerFactory` is instantiated at the composition root, so there's no "deep drilling".  Anything that a scope handler needs to do its job just gets passed one level through the factory to the scope handler that needs it.  The decoupling actually results in shallower hierarchies, because the code that uses scope handlers isn't responsible for giving them their dependencies.  It also makes testing easier, because we could just pass in a fake scope handler factory to a modifier.  Tbh it would be a bit cleaner if we could just pass in a scope handler directly to a modifier, but the problem is that we might need a different scope handler per target, so it is useful for the modifier to be able to construct the scope handler on the fly

packages/cursorless-engine/src/processTargets/ModifierStageFactoryImpl.ts
```diff
@@ -0,0 +1,134 @@
+import {
```
pokey, line 1: I wonder if we want to move these factories into the composition root; one could argue they're really part of the composition root, but otoh there's something nice about having them close to all the things they instantiate.  Idk wdyt @auscompgeek @AndreasArvidsson ?

packages/cursorless-engine/src/processTargets/TargetPipeline.ts
```diff
@@ -0,0 +1,276 @@
...
+import { MarkStage, ModifierStage } from "./PipelineStages.types";
+import ImplicitStage from "./marks/ImplicitStage";
+import { ContainingTokenIfUntypedEmptyStage } from "./modifiers/ConditionalModifierStages";
+import { PlainTarget, PositionTarget } from "./targets";
+
+export class TargetPipeline {
+  constructor(
+    private modifierStageFactory: ModifierStageFactory,
+    private markStageFactory: MarkStageFactory,
+    private context: ProcessedTargetsContext,
```
pokey, line 21: Notice that putting these functions onto a class allows us to avoid passing the `context` around so much by just making it a member variable.  We'll eventually split up / remove the `context` altogether, but leaving that for later

AndreasArvidsson (approved):

PR #1399, opened 2023-04-08
Bump engine/VSCode TypeScript target to ES2020

auscompgeek:
Our legacy tests currently run against VSCode 1.66.0:

https://github.com/cursorless-dev/cursorless/blob/68ba545520829a21e26d70437c650fb615d73344/packages/test-harness/src/util/launchVscodeAndRunTests.ts#L37

However, we pull in types for VSCode 1.61:

https://github.com/cursorless-dev/cursorless/blob/68ba545520829a21e26d70437c650fb615d73344/packages/cursorless-vscode-core/package.json#L31

and the extension also declares support for the same version:

https://github.com/cursorless-dev/cursorless/blob/68ba545520829a21e26d70437c650fb615d73344/packages/cursorless-vscode/package.json#L32

so 1.61 shall be our lowest common denominator.

VSCode 1.61.0 uses Electron 13: https://github.com/microsoft/vscode/blob/1.61.0/package.json#L134

Per [Electron's docs](https://www.electronjs.org/docs/latest/tutorial/electron-timelines), Electron 13 included Chromium 91 and Node.js v14.16.

Per the trusty [ECMAScript compatibility table](https://kangax.github.io/compat-table/es2016plus/) and [Stack Overflow](https://stackoverflow.com/questions/61305578/what-typescript-configuration-produces-output-closest-to-node-js-14-capabilities), Node 14 fully supports ES2020.

This bumps the TypeScript target to ES2020 to reduce bundle sizes.

## Checklist

### DIFF
--- a/packages/cursorless-engine/tsconfig.json
+++ b/packages/cursorless-engine/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    "target": "es2020",
     "outDir": "out",
     "rootDir": "src"
   },
--- a/packages/cursorless-vscode/tsconfig.json
+++ b/packages/cursorless-vscode/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    "target": "es2020",
     "outDir": "out",
     "rootDir": "src"
   },
--- a/packages/vscode-common/tsconfig.json
+++ b/packages/vscode-common/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    "target": "es2020",
     "outDir": "out",
     "rootDir": "src"
   },


pokey:
Nice research!  But I'm not convinced this setting will have any impact, as we use esbuild to construct the final bundle 🤔

https://github.com/cursorless-dev/cursorless/blob/68ba545520829a21e26d70437c650fb615d73344/packages/cursorless-vscode/package.json#L749

auscompgeek:
Looks like esbuild respects tsconfig target: https://esbuild.github.io/content-types/#tsconfig-json

Good point about the website though, we should think about what browsers we end up dropping first. If anything is using Babel instead my understanding is that the tsconfig won't affect Babel's emit.

auscompgeek:
It looks like the browserslist config for the docs site targets fairly modern browsers anyway which pretty much all support ES2020. If I'm reading the compat table right they also all support ES2021, so it's just the VSCode minimum holding that back if we want to be consistent.

https://github.com/cursorless-dev/cursorless/blob/68ba545520829a21e26d70437c650fb615d73344/packages/cursorless-org-docs/package.json#L34-L39

```console
> pnpx browserslist
and_chr 111
and_uc 13.4
chrome 111
chrome 110
chrome 109
edge 111
edge 110
firefox 111
firefox 110
ios_saf 16.3
ios_saf 16.2
ios_saf 16.1
ios_saf 16.0
ios_saf 15.6
opera 95
safari 16.3
safari 15.6
samsung 20
```

pokey:
Ah another nice catch.  Looks like you're right:

```sh
➜ comm -23 <(pnpx browserslist |sort) <(pnpx browserslist "supports es6" | sort)
```

But I think we should prob switch our query to `"defaults"`, no?  We get a [big warning](https://browsersl.ist/#q=%3E0.5%25%2C+not+dead%2C+not+op_mini+all) for our current query.  With that query, we are hampered by op_mini

```sh
➜ comm -23 <(pnpx browserslist "defaults"|sort) <(pnpx browserslist "supports es6" | sort)
op_mini all
```

I'd be tempted to do the following:

- [ ] In this PR, set `target: "es2020"` in `cursorless-vscode/tsconfig.json`, instead of in `tsconfig.base.json`

And then in a new PR

- [ ] Explicitly set `target: "es5"` for `cursorless-org-docs/tsconfig.json`
- [ ] Switch our browserslist field in `cursorless-org-docs/package.json` to `"defaults"`
- [ ] Switch our browserslist field in `cursorless-org/package.json` to `"defaults"`
- [ ] Potentially set `target: "es2020"` in the base tsconfig 🤷‍♂️.  Not sure how much it matters.  We should prob think about requiring a target in every package that is deployed

Make sense?

auscompgeek:
Sounds good. I've gone ahead and set the target for `cursorless-engine` and `vscode-common` too.

We should also reconsider the contents of `packages/cheatsheet-local/.browserslistrc`. That seems quite restrictive right now.

For anyone else that wants to dive deeper in double-checking my research:

- https://node.green
- https://github.com/tsconfig/bases/blob/main/bases/node14.json

auscompgeek:
Stats:

- vsix on main: 389607 bytes
- vsix in this PR: 388065 bytes

That's a solid saving of 1.6 kB on a zip file :tada:

pokey:
sorry to be so slow on this one 😅

Re adding `es2020` to `cursorless-engine` and `vscode-common`, I think that's fine, but it does make one wonder about how all this stuff works.  We will potentially start importing `cursorless-engine` from the browser to do some fancy docs stuff.  It almost feels like `target` should be part of apps (eg `cursorless-vscode` / `cursorless-org`) rather than libraries.  We do our bundling by going straight to the source of all our library deps rather than their outputs, so in some sense the app already owns the target.  What should our policy be for targets for libraries?  Minimum of all apps it is imported by?  Don't really care because it's ignored when bundling from a downstream app?

pokey (commented):
https://github.com/cursorless-dev/cursorless/pull/1399#issuecomment-1551367719

auscompgeek:
[`target` does change the default of `lib`](https://www.typescriptlang.org/tsconfig#target). I think it'd be worthwhile keeping those two consistent to avoid confusion.

pokey:
> [`target` does change the default of `lib`](https://www.typescriptlang.org/tsconfig#target). I think it'd be worthwhile keeping those two consistent to avoid confusion.

no i don't mean `lib`, I mean our own packages that have no targets, and are just libraries that are consumed by one of our targets

auscompgeek:
That's what I mean though. If a package needs to be consumed by multiple apps, its `lib` needs to be at most the minimum of the `lib`s of the apps.

pokey:
Right makes sense, and its `target` as well then I assume

pokey (approved):

PR #1402, opened 2023-04-09
[talon] Omit custom actions from action defaults

auscompgeek:
Whilst this effectively does nothing, adding the empty defaults to two different CSVs seems to have been an oversight.

## Checklist


pokey:
I *think* this is fine, but worth checking that the custom actions still end up in the cheatsheet

pokey (approved):
nice catch

PR #1400, opened 2023-04-08
pre-commit: Keep prettier version in sync; add eslint fixer

pokey:
- Partially addresses #640

## Checklist

- [ ] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)
- [ ] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [ ] I have not broken the cheatsheet
- [ ] File issue to revert patch working around https://github.com/pnpm/meta-updater/issues/10
- [ ] Check boxes in #640

PR #1404, opened 2023-04-09
pre-commit: Switch to GitHub action; add eslint fixer

pokey:
- Partially addresses #640
- Supercedes #1400

## Checklist

- [x] Disable pre-commit.ci
- [x] Disable pre-commit.ci lite
- [ ] Change branch protection rules
- [ ] Check boxes in #640

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,31 @@
+name: Pre-commit
+permissions: read-all
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types: [opened, synchronize, reopened]
+  merge_group:
+    branches: [main]
+
+jobs:
+  pre-commit:
+    name: Pre-commit
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.x
+      - uses: pnpm/action-setup@v2
+        with:
+          version: 7
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+          cache: pnpm
+      - run: pnpm --color install
+      - uses: pre-commit/action@v3.0.0
+      - uses: pre-commit-ci/lite-action@v1.0.1
+        if: always()
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -20,6 +20,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
           ref: ${{ github.head_ref }}
       - uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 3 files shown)


pokey:
Huh this unfortunately doesn't seem to work.  I tried changing a lint rule to see if it would auto-fix, but nothing happens 😕

pokey:
Ok I switched the push action to https://github.com/stefanzweifel/git-auto-commit-action instead of pre-commit.ci lite, because the former is basically instant, whereas the latter seems to take a while

AndreasArvidsson (dismissed):

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,37 @@
...
+      - uses: pnpm/action-setup@v2
+        with:
+          version: 7
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+          cache: pnpm
+      - run: pnpm --color install
+      - uses: pre-commit/action@v3.0.0
+      - uses: stefanzweifel/git-auto-commit-action@v4
```
auscompgeek, line 34: Does this work for forks?

auscompgeek:
FWIW we don't have to abandon pre-commit.ci entirely, we can skip the local hooks on pre-commit.ci and just run eslint and prettier on GitHub Actions

pokey:
> FWIW we don't have to abandon pre-commit.ci entirely, we can skip eslint/prettier on pre-commit.ci and just run eslint and prettier on GitHub Actions

What's the benefit of having both systems active?

.github/workflows/pre-commit.yml
pokey: Good question. I'll give it a whirl

pokey: Haven't tested yet, but looks like a bit of a hornet's nest and potentially introduces security vulnerability to get it to work 😬 https://github.com/stefanzweifel/git-auto-commit-action/issues/211

So do we switch to pre-commit lite and deal with really slow commit, or switch to the setup from my other attempt with syncing versions. Downside of that is we can't easily run custom fixers like if we wanted to autoformat our recorded tests using our custom yaml writer

auscompgeek: Hmm. It might be possible to get something working with the completion of this workflow triggering another workflow that runs in the `main` context? That'd be super complex though, and sounds like even more of a nightmare.

I've seen workflows that use a personal access token to get this sort of thing to work, but that's probably just as risky.

auscompgeek:
:shrug:

There's a _potential_ cache speed benefit with pre-commit.ci, but I guess we don't know exactly how much caching will help us here since the cache doesn't exist on `main` for the workflow to use yet.

pokey:
> 🤷
> 
> There's a _potential_ cache speed benefit with pre-commit.ci, but I guess we don't know exactly how much caching will help us here since the cache doesn't exist on `main` for the workflow to use yet.

pre-commit.ci is significantly faster, as you suspect, but tbh most of the fixes come from prettier, so I'm not sure the benefit of putting the other fixers in a different CI system justify the complexity / races of running two systems

auscompgeek (approved):

PR #1405, opened 2023-04-09
`PositionTarget`: consider minimum line indentation when pouring multiline target

pokey:
We have functionality that computes minimum indentation when doing a "pour", but we don't use it if the delimiter is `\n` because we switch to using the editor's newline insertion to handle things like indenting after a line ending in `:` in Python.  However, if the user is explicitly pouring something that is multiple lines, then I argue that it is safe to do our own insertion.  This change enables things like properly dedenting after an if statement in Python

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/targets/PositionTarget.ts
+++ b/packages/cursorless-engine/src/processTargets/targets/PositionTarget.ts
@@ -17,7 +17,6 @@ export default class PositionTarget extends BaseTarget {
   private isLineDelimiter: boolean;
   private isBefore: boolean;
   private indentationString: string;
-  private editNewActionType: EditNewActionType;
 
   constructor(parameters: PositionTargetParameters) {
     super(parameters);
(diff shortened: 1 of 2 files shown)


packages/cursorless-engine/src/processTargets/targets/PositionTarget.ts
```diff
@@ -41,11 +48,7 @@ export default class PositionTarget extends BaseTarget {
   getRemovalRange = () => removalUnsupportedForPosition(this.position);
 
   getEditNewActionType(): EditNewActionType {
-    if (this.insertionDelimiter === "\n" && this.position === "after") {
-      return "insertLineAfter";
-    }
-
-    return "edit";
+    return this.editNewActionType;
```
AndreasArvidsson, line 51: Any reason why you moved it from here to the constructor?

AndreasArvidsson (approved):

packages/cursorless-engine/src/processTargets/targets/PositionTarget.ts
pokey: For some reason I thought we no longer had access to the `thatTarget` at this point.  But in retrospect obviously we do 🤷‍♂️.  Nice catch

AndreasArvidsson (approved):

PR #1427, opened 2023-04-15
Add support for domain, leading, trailing, interior

pokey:
- Depends on #629

## Checklist


queries/ruby.scm
```diff
@@ -13,3 +13,4 @@
 (regex) @regularExpression
 
 (class) @class
+(class name: (_) @className) @domain
```
pokey, line 16: Should this shorthand be `@_.domain` instead to make it clearer that it's a related node rather than a different scope type?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler.ts
```diff
@@ -117,29 +117,73 @@ export class TreeSitterScopeHandler extends BaseScopeHandler {
...
+      "leading",
+    );
+
+    const trailingDelimiterRange = getRelatedRange(
+      match,
+      scopeTypeType,
+      "trailing",
     );
 
+    const interiorRange = getRelatedRange(match, scopeTypeType, "interior");
```
pokey, line 143: Should we call this `"inside"`?  That's the term everyone knows.  I'd also be tempted to do a rename `interior` => `inside` in our code base, so maybe we should just do that in a follow-up

PR #1428, opened 2023-04-15
Add integration test setup; check for legacy/new scope duplicates

pokey:
This PR adds a test to ensure that we don't try to define the same scope type both using our legacy node matchers and the new query scope handlers

I have mixed feelings about the way I made this work, but I wasn't sure how else to do it.  The test needs to be defined in the engine, because otherwise we'd need to expose lots of stuff to cursorless-vscode-e2e, but it needs a full tree-sitter instance, so can't be run as a unit test.  This file is the compromise I came up with 🤷‍♂️.  We should maybe revise once we have tree-sitter integrated with this repo

I don't like that this code gets rolled up into our production bundle, but probably ok for this one-off.  If we find this file growing, we should revisit

It's also possible that once we get rid of our `ide` singleton, we can relax the restriction that our test harness doesn't import from `cursorless-engine`.  That would enable us to just import what we need from the engine directly in the test harness to run this test 🤔

- Depends on #629

## Checklist


packages/cursorless-engine/src/cursorlessEngine.ts
```diff
@@ -68,5 +69,7 @@ export function createCursorlessEngine(
     hatTokenMap,
     snippets,
     injectIde,
+    runIntegrationTests: () =>
+      runIntegrationTests(treeSitter, languageDefinitions),
```
auscompgeek, line 73: If we instead expose `treeSitter` and `languageDefinitions`, then we could avoid `runIntegrationTests` ending up in the prod bundle, right?

pokey: No. The issue is that the test needs to be defined inside cursorless engine; see PR description for more detail. Also notice that treeSitter actually comes from the caller, so exposing it wouldn't actually do anything

AndreasArvidsson (approved):

PR #1431, opened 2023-04-18
Python scope handlers

pokey:
Starts to migrate Python scope handlers to tree-sitter queries.  Introduces "inside class" and "inside funk" along the way

## Checklist

- [x] Check that I didn't break "every state" esp inside a function or class. If broken, just roll back "state" for now
- [x] Check performance on a large file


packages/cursorless-engine/src/languages/TreeSitterQuery.ts
```diff
@@ -0,0 +1,47 @@
...
+
+      if (
+        "onlyIfNotDescendantOfType" in setProperties &&
+        captures.some(
+          ({ node }) =>
+            node.parent?.type === setProperties.onlyIfNotDescendantOfType,
+        )
+      ) {
+        return false;
+      }
```
pokey, line 34: Here's where we can start defining custom properties

queries/python.scm
```diff
@@ -0,0 +1,64 @@
...
+    body: (_) @namedFunction.interior
+  )
+) @namedFunction @functionName.domain
+
+(
+  (class_definition
+    name: (_) @className
+    body: (_) @class.interior
+  ) @class @className.domain
+  (#set! onlyIfNotDescendantOfType decorated_definition)
```
pokey, line 49: Here's an example of using a custom property

queries/python.scm
```diff
@@ -0,0 +1,64 @@
...
+)
+(decorated_definition
+  (class_definition
+    name: (_) @className
+    body: (_) @class.interior
+  )
+) @class @className.domain
```
pokey, line 56: Note that this matcher and the above are the really awkward way that we have to implement the `?` construct from our own pattern language.  It's a bummer that it's so much more cumbersome in the native tree-sitter query language, but unfortunately we're kind of stuck with it.  If the language continues to feel too low-level, we can think about generating queries based on a higher level specification language 🤔

queries/python.scm
```diff
@@ -0,0 +1,64 @@
...
+
+(module) @className.iteration @class.iteration
+(module) @statement.iteration
+(module) @namedFunction.iteration @functionName.iteration
```
pokey, line 60: I had to break this up into multiple patterns because you can't seem to have more than 3 annotations on a single pattern; it just silently ignores anything after the third one 😳

queries/python.scm
```diff
@@ -0,0 +1,64 @@
...
+  (try_statement)
+  (while_statement)
+  (with_statement)
+] @statement
+
+(
+  (function_definition
+    name: (_) @functionName
+    body: (_) @namedFunction.interior
+  ) @namedFunction @functionName.domain
```
pokey, line 34: we are able to pack a lot of information in here.  note how we've defined "funk", "inside funk", and "funk name" all in one go

packages/cursorless-engine/src/languages/python.ts
```diff
@@ -97,10 +67,6 @@ const nodeMatchers: Partial<
   functionCall: "call",
   functionCallee: "call[function]",
   comment: "comment",
-  class: "decorated_definition?.class_definition",
-  className: "class_definition[name]",
```
pokey, line 101: fixed a bug along the way; now you can say "class name" from within the decorator of the class

PR #1458, opened 2023-05-13
Simplify ancestor exclusion code

pokey:
The `maxAncestorIndex` code is overkill, because we only ever use a value of 0 or infinity, so it's must simpler to just use a boolean `skipAncestorScopes`.

## Checklist


packages/cursorless-engine/src/processTargets/modifiers/EveryScopeStage.ts
```diff
@@ -121,3 +120,19 @@ export class EveryScopeStage implements ModifierStage {
     return iterationScopeTarget.contentRange;
   }
 }
+
```
pokey, line 123: not related to this PR, but no reason to have this small function in its own file when it's now only used here

AndreasArvidsson (approved):

PR #1462, opened 2023-05-14
Support "every line air past bat"

pokey:
We also support fancier constructs like the following:

- "every line block air past bat"
- "every line air past block bat"
- "every line past bat"
- "every line past block bat"

The approach is to detect ranges of the form `"every <scope> <target> past <target>"` and to turn them into a special range target, distinct from `"continuous"` or `"slice"`.  We remove the `"every <scope>"` modifier as well because that's done after we construct the range.

The reason we need a special range target is that exclusion needs to be smart to handle scopes whose domain is larger than their content range

We also add slight special casing for line targets so that when you're at the beginning / end of a line, that line is considered part of the range

## Checklist

- [x] Only expand to containing scope if the given end is excluded, eg expand end target if "until".  Might need to tweak construction of ranges so that each side is responsible for exclusion to make "every token until line bat" work, because otherwise it gets downgraded so we won't actually skip the right thing.  eg try "every token air past state bat" if "bat" is in the middle of the statement.  
- [x] Add tests for above examples
- [x] Handle case where one side is not in a scope
- [x] Handle case where one side is in a scope but you don't want it to jump to parent (eg "take every state past bat" when we're inside a function but between statements)
- [x] Add test for "every line air until bat"

### DIFF
--- a/packages/cursorless-engine/src/processTargets/targetUtil/createContinuousRange.ts
+++ b/packages/cursorless-engine/src/processTargets/targetUtil/createContinuousRange.ts
@@ -46,27 +46,6 @@ export function createContinuousLineRange(
   return new Range(start, end);
 }
 
-export function createSimpleContinuousRangeTarget(
-  target1: Target,
-  target2: Target,
-  isReversed: boolean,
-  includeStart: boolean = true,
-  includeEnd: boolean = true,
-) {
-  const isForward = target1.contentRange.start.isBefore(
-    target2.contentRange.start,
-  );
-  const anchorTarget = isForward ? target1 : target2;
-  const activeTarget = isForward ? target2 : target1;
-
-  return anchorTarget.createContinuousRangeTarget(
-    isReversed,
-    activeTarget,
-    includeStart,
-    includeEnd,
-  );
-}
-
 export function createContinuousRangeUntypedTarget(
   isReversed: boolean,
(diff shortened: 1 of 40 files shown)


packages/cursorless-engine/src/util/getPrimitiveTargets.ts
```diff
@@ -36,31 +32,6 @@ function getPartialPrimitiveTargetsHelper(
...
   }
 }
-/**
- * Given a list of targets, recursively descends all targets and returns every
- * contained primitive target.
- *
- * @param targets The targets to extract from
- * @returns A list of primitive targets
- */
-export function getPrimitiveTargets(targets: TargetDescriptor[]) {
```
pokey, line 46: this was causing type issues and not used anywhere so I dropped it 🤷‍♂️

packages/cursorless-engine/src/processTargets/targetUtil/createContinuousRange.ts
```diff
@@ -46,27 +46,6 @@ export function createContinuousLineRange(
   return new Range(start, end);
 }
 
-export function createSimpleContinuousRangeTarget(
```
pokey, line 49: not used anywhere

AndreasArvidsson (approved):

PR #1466, opened 2023-05-16
Added typescript language support for jsx fragment node

### LINKED ISSUE #1450: React Fragment syntax unsupported
It seems that cursorless doesn't identify React fragment syntax, ie `<></>`. Ideally it would be treated like any other xml element. It's just syntactic sugar for `<React.Fragment></React.Fragment>`.

AndreasArvidsson:
Fixes #1450

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/languages/typescript.ts
+++ b/packages/cursorless-engine/src/languages/typescript.ts
@@ -71,7 +71,7 @@ const STATEMENT_TYPES = [
   "with_statement",
 ];
 
-function jsxTag(isStartTag: boolean): NodeMatcher {
+function getJsxFragmentTag(isStartTag: boolean): NodeMatcher {
   return matcher(
     typedNodeFinder("jsx_fragment"),
     (editor: TextEditor, node: SyntaxNode) => {
@@ -93,11 +93,11 @@ function jsxTag(isStartTag: boolean): NodeMatcher {
 
 const getStartTag = cascadingMatcher(
   patternMatcher("jsx_element.jsx_opening_element!"),
-  jsxTag(true),
+  getJsxFragmentTag(true),
 );
 const getEndTag = cascadingMatcher(
   patternMatcher("jsx_element.jsx_closing_element!"),
-  jsxTag(false),
+  getJsxFragmentTag(false),
 );
 
 const getTags = (selection: SelectionWithEditor, node: SyntaxNode) => {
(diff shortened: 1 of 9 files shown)


packages/cursorless-engine/src/languages/typescript.ts
```diff
@@ -70,8 +71,34 @@ const STATEMENT_TYPES = [
   "with_statement",
 ];
 
-const getStartTag = patternMatcher("jsx_element.jsx_opening_element!");
-const getEndTag = patternMatcher("jsx_element.jsx_closing_element!");
+function jsxTag(isStartTag: boolean): NodeMatcher {
```
pokey, line 74: I'd rename to `getJsxFragmentTag`

pokey (approved):
He's back! 😊

Looks good to me with minor comment. Nice tests

packages/cursorless-engine/src/languages/typescript.ts
```diff
@@ -70,8 +71,34 @@ const STATEMENT_TYPES = [
   "with_statement",
 ];
 
-const getStartTag = patternMatcher("jsx_element.jsx_opening_element!");
-const getEndTag = patternMatcher("jsx_element.jsx_closing_element!");
+function getJsxFragmentTag(isStartTag: boolean): NodeMatcher {
```
pokey, line 74:
```suggestion
/** Handles jsx fragment start or end tag, eg the `<>` in `<>foo</>` **/
function getJsxFragmentTag(isStartTag: boolean): NodeMatcher {
```

pokey (commented):
Nice. One more thought: do we want to support "name tags"? Might be useful for turning fragment into proper element

AndreasArvidsson:
> Nice. One more thought: do we want to support "name tags"? Might be useful for turning fragment into proper element

I think we wait with that one until we have the new syntax?

pokey:
sure if you don't think it's a big deal to be missing that functionality / consistency.  I don't use jsx too frequently so not a huge deal

pokey (approved):

AndreasArvidsson:
> sure if you don't think it's a big deal to be missing that functionality / consistency. I don't use jsx too frequently so not a huge deal

I don't in this case since there actually isn't a name in the tag. We don't support take type in functions without a type either. Definitely something that we should support in the future, but for now I think this is fine.

PR #1467, opened 2023-05-16
Ensure that all marks are captured in recorded tests

pokey:
- depends on #1468 

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/testUtil/extractTargetKeys.ts
+++ b/packages/cursorless-engine/src/testUtil/extractTargetKeys.ts
@@ -1,10 +1,16 @@
-import { getKey } from "@cursorless/common";
+import {
+  PartialPrimitiveTargetDescriptor,
+  PartialTargetDescriptor,
+  getKey,
+} from "@cursorless/common";
 import {
   PrimitiveTargetDescriptor,
   TargetDescriptor,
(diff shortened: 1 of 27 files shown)


packages/cursorless-engine/src/testUtil/extractTargetKeys.ts
```diff
@@ -18,14 +24,17 @@ export function extractTargetKeys(target: TargetDescriptor): string[] {
         ...extractPrimitiveTargetKeys(target.active),
       ];
 
-    default:
+    case "implicit":
```
pokey, line 27: this line was how we dropped some marks.  I had added a new target type, and forgot to handle it.  Now typescript will complain if that happens again.  lesson learned: avoid `default` if possible 😅

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/surroundingPair/textual/clearBoundsDouble2.yml
pokey, line 1: i wasn't sure which `"` it was supposed to be so I just made two versions of the test.  Both seem useful 🤷‍♂️

AndreasArvidsson (approved):

PR #1484, opened 2023-05-19
Merge stored targets into one `StoredTargetsMap`

pokey:
## Checklist

### DIFF
--- a/packages/cursorless-engine/src/core/StoredTargets.ts
+++ b/packages/cursorless-engine/src/core/StoredTargets.ts
@@ -1,10 +1,25 @@
+import { DefaultMap } from "@cursorless/common";
 import { Target } from "../typings/target.types";
 
 /**
  * Used to store targets between commands.  This is used by marks like `that`
  * and `source`.
  */
-export class StoredTargets {
+export class StoredTargetMap {
+  private targetMap: DefaultMap<string, StoredTargets> = new DefaultMap(
+    () => new StoredTargets(),
+  );
+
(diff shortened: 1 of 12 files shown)


packages/cursorless-engine/src/core/StoredTargets.ts
```diff
@@ -1,10 +1,25 @@
+import { DefaultMap } from "@cursorless/common";
 import { Target } from "../typings/target.types";
 
 /**
  * Used to store targets between commands.  This is used by marks like `that`
  * and `source`.
  */
-export class StoredTargets {
+export class StoredTargetMap {
+  private targetMap: DefaultMap<string, StoredTargets> = new DefaultMap(
```
AndreasArvidsson, line 9: string? STRING?? I assume that is for future bookmarks, but is there anyway we could type our known stored targets?

AndreasArvidsson (approved):

packages/cursorless-engine/src/core/StoredTargets.ts
pokey: Haha yeah that's for custom bookmarks. I could just add separate fields for `that` and `source` and then the string getter would be just for custom use?

AndreasArvidsson: Sounds good!

pokey: Ok for now I just made it strongly typed.  We can handle custom bookmarks when we get there

Issue #802, opened 2022-06-29
Support context-specific snippet definitions


pokey:
We currently support specifying the scopes in which a particular snippet definition is active but this is not yet implemented. We would like to implement this.

Note that we cannot support this until we have #785.

- [ ] To make this useful, we will need to stop returning the first matching definition we find, and instead prefer snippets that have more specific contexts. Specifically, snippets that match a scope should have higher precedence than snippets that only match on language id, which should in turn have higher precedence than ones that have no language id.
- [ ] We need to think about how this interacts with user snippets overriding core / third-party snippets. Eg if user's snippet is less specific, should it still override a core snippet that is more specific? (probably not)

PR #1487, opened 2023-05-19
Support specializing snippet definitions by containing scope

pokey:
I thought this one would be a quick win, but deciding which matching snippet should apply turned out to be complex

- Closes https://github.com/cursorless-dev/cursorless/issues/802

Note that we don't support scope specialisation for custom snippets; should be a fairly easy follow-up PR building on this machinery.

## Checklist

### DIFF
--- a/cursorless-snippets/functionDeclaration.cursorless-snippets
+++ b/cursorless-snippets/functionDeclaration.cursorless-snippets
@@ -8,10 +8,6 @@
             "typescriptreact",
             "javascript",
             "javascriptreact"
-          ],
-          "scopeTypes": [
-            "namedFunction",
-            "document"
           ]
         },
         "body": [
@@ -35,6 +31,9 @@
           ],
           "scopeTypes": [
             "class"
+          ],
+          "excludeDescendantScopeTypes": [
+            "namedFunction"
           ]
         },
         "body": [
--- a/cursorless-snippets/functionDeclaration.cursorless-snippets
+++ b/cursorless-snippets/functionDeclaration.cursorless-snippets
@@ -8,10 +8,37 @@
             "typescriptreact",
             "javascript",
             "javascriptreact"
+          ],
+          "scopeTypes": [
+            "namedFunction",
+            "document"
           ]
         },
         "body": [
           "function $name($parameterList) {",
+          "\t$body",
+          "}"
+        ],
+        "variables": {
+          "name": {
+            "formatter": "camelCase"
+          }
+        }
+      },
+      {
+        "scope": {
+          "langIds": [
+            "typescript",
+            "typescriptreact",
+            "javascript",
+            "javascriptreact"
+          ],
+          "scopeTypes": [
+            "class"
+          ]
+        },
+        "body": [
+          "$name($parameterList) {",
           "\t$body",
           "}"
         ],
--- a/cursorless-snippets/functionDeclaration.cursorless-snippets
+++ b/cursorless-snippets/functionDeclaration.cursorless-snippets
@@ -62,6 +62,28 @@
             "formatter": "snakeCase"
           }
         }
+      },
+      {
+        "scope": {
+          "langIds": [
+            "python"
+          ],
+          "scopeTypes": [
+            "class"
+          ],
+          "excludeDescendantScopeTypes": [
+            "namedFunction"
+          ]
+        },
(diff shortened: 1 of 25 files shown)


cursorless-snippets/functionDeclaration.cursorless-snippets
```diff
@@ -8,6 +8,10 @@
             "typescriptreact",
             "javascript",
             "javascriptreact"
+          ],
+          "scopeTypes": [
+            "namedFunction",
+            "document"
```
pokey, line 14: I had to do this to make sure that this snippet wins when you're inside a function that is inside a class.  Alternately, we could add some new field to the member function snippet definition below that causes it not to match when its nearest ancestor is a function, not a class.  Not exactly sure the mechanism there

The disadvantage of this current approach is that it makes it a bit harder for users to override, because if they don't match these `scopeTypes` exactly then the core snippet here could just win because it's more specific 🤔

cursorless-snippets/functionDeclaration.cursorless-snippets
```diff
@@ -21,6 +25,29 @@
...
+        "body": [
+          "$name($parameterList) {",
+          "\t$body",
+          "}"
+        ],
+        "variables": {
+          "name": {
+            "formatter": "camelCase"
+          }
+        }
```
pokey, line 49: annoying we have to duplicate this one

packages/cursorless-engine/src/core/mergeSnippets.test.ts
```diff
@@ -0,0 +1,390 @@
...
+            },
+          },
+        ],
+        description: "user snippet",
+      },
+    },
+  },
+
+  {
+    name: "should prefer more specific snippets, even if they are from a lower priority origin",
```
pokey, line 205: I think this is the right thing to do, but see comment above

packages/cursorless-engine/src/core/mergeSnippets.test.ts
```diff
@@ -0,0 +1,390 @@
...
+          {
+            body: ["has no langIds"],
+          },
+        ],
+      },
+    },
+  },
+
+  {
+    name: "should prefer snippets with more specific languages, regardless of scopeType",
```
pokey, line 296: this is not really true anymore.  although we sort them this way, we now still prefer the one with the scope when we actually get around to running the scope handler.  I should prob remove the code that looks at scope types at this point because it doesn't actually have any impact

packages/cursorless-engine/src/core/mergeSnippets.test.ts
```diff
@@ -0,0 +1,390 @@
...
+              scopeTypes: ["anonymousFunction"],
+            },
+          },
+        ],
+      },
+    },
+  },
+
+  {
+    name: "should prefer snippets with more specific scope types",
```
pokey, line 336: doesn't really matter anymore either, because we prefer snippets that matched a scope down the line anyway

packages/cursorless-engine/src/processTargets/targets/BaseTarget.ts
```diff
@@ -30,6 +30,7 @@ export interface CloneWithParameters {
 
 export default abstract class BaseTarget implements Target {
   protected readonly state: CommonTargetParameters;
+  isDocument = false;
```
pokey, line 33: I had to add this to handle the case where we are in a document that only contains a class.  In that case we want to prefer the class method function snippet definition even though it's the same range as the whole document.  We could probably remove this if we add a mechanism for the class method snippet to deactivate itself if there's a function in the ancestor chain between it and its containing class

packages/cursorless-engine/src/core/mergeSnippets.test.ts
```diff
@@ -0,0 +1,348 @@
...
+            body: [""],
+            scope: {
+              langIds: ["typescript", "javascript"],
+              scopeTypes: ["anonymousFunction"],
+            },
+          },
+          {
+            body: [""],
+            scope: {
+              scopeTypes: ["anonymousFunction"],
```
pokey, line 312: change precedence:

1. have language at all wins out over having scope type but no language
2. but having scope type with two languages wins out over having no scope type with one language

AndreasArvidsson (approved):

PR #1497, opened 2023-05-24
Support "instance" pseudo-scope

### LINKED ISSUE #29: Support take every instance of token transformation
- Do we also want to support taking a certain number of instances?
- Also support narrowing to instances which appear within another target, eg "take every token fine in class gust"
- Also support narrowing to instances that appear at the beginning of the line (what would syntax be?)

pokey:
- Implements the "instance" pseudo-scope. It functions grammatical kind of like a scope, but hoists over ranges and uses its input in its own way.
- Fixes #29
- Depends on https://github.com/cursorless-dev/cursorless/pull/1496
- See #1490 for some discussion of how we got to this implementation

## Checklist

- [x] Handle word scopes
- [x] Merge with #1496 ?
- [x] check performance on gigantic file
- [x] Come up with story around repeater noises.  Maybe interact with some new knausj quick actions interface, where "instance" establishes that the "forward" quick action will perform action on additional target
- [x] Document how to add one more selection somewhere: "take that and next instance that"
- [x] Capture writing from #1490
- [x] "first three instances"
- [x] Handle narrowing range by selection

### DIFF
--- a/packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/modifiers/instance/preEveryInstanceEveryCarAirPastBat.yml
+++ b/packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/modifiers/instance/preEveryInstanceEveryCarAirPastBat.yml
@@ -1,0 +1,70 @@
+languageId: plaintext
+command:
+  version: 5
+  spokenForm: pre every instance every car air past bat
+  action: {name: setSelectionBefore}
+  targets:
+    - type: range
+      anchor:
+        type: primitive
+        modifiers:
+          - type: everyScope
+            scopeType: {type: instance}
+          - type: everyScope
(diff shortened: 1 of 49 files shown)


packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/modifiers/instance/preEveryInstanceEveryCarAirPastBat.yml
pokey, line 1: I wonder if `setSelection` actions should sort the selections.  I'm sure it would be unexpected to realise these cursors are in this haphazard order, esp since VSCode gives no visual indication of cursor order.  Out of scope for this PR obv, but just a thought

AndreasArvidsson (approved):

PR #1504, opened 2023-05-26
Support for `@foo.start` and `@foo.end`

### LINKED ISSUE #1479: Support for `@foo.start` and `@foo.end`
Eg

```scm
(
  (aaa
    (bbb) @ddd.start
    (ccc) @ddd.end
  )
)
```

Creates `@ddd` node with range between `@ddd.start` and `@ddd.end`.  If one of them is missing, just use the other one's range as the full range.  Could also add exclusion args

Alternately, we could require that the user use a `#make-range!` operator

Eg

```scm
(
  (aaa
    (bbb) @ddd.start
    (ccc) @ddd.end
  )
  (#make-range! @ddd @ddd.start @ddd.end)
)
```

The advantage of this approach is that we could add exclusion args

pokey:
- Fixes #1479

## Checklist

- [x] Remove `.start` and `.end` when computing capture names in https://github.com/cursorless-dev/cursorless/blob/main/packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts#L73-L75

### DIFF
--- a/packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
+++ b/packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
@@ -7,6 +7,7 @@ import { MutableQueryMatch, QueryCapture, QueryMatch } from "./QueryCapture";
 import { parsePredicates } from "./parsePredicates";
 import { predicateToString } from "./predicateToString";
 import { groupBy, uniq } from "lodash";
+import { checkCaptureStartEnd } from "./checkCaptureStartEnd";
 
 /**
  * Wrapper around a tree-sitter query that provides a more convenient API, and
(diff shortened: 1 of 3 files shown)


packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
```diff
@@ -83,7 +84,25 @@ export class TreeSitterQuery {
...
+        // name, for which we'd return a capture with name `foo`.
+        const captures: QueryCapture[] = Object.entries(
+          groupBy(match.captures, ({ name }) =>
+            name.replace(/\.(start|end)$/, ""),
+          ),
+        ).map(([name, captures]) => ({
+          name,
+          range: captures
+            .map(({ range }) => range)
+            .reduce((accumulator, range) => range.union(accumulator)),
```
pokey, line 101: Is this too lenient?  I'm not sure we want to allow people to use multiple `@foo` captures and rely on the fact they're merged.  As discussed in Slack, I don't think the semantics are obvious when you see multiple `@foo` captures; in fact I'd expect multiple targets.  I could add a check that ensures there's only one `@foo` etc

AndreasArvidsson: I'm fine with a check. The waits implemented now you could mix and match start and end and it wouldn't matter.

AndreasArvidsson (approved):

PR #1507, opened 2023-05-26
Support `#allow-multiple!` predicate to enable multiple content ranges

### LINKED ISSUE #1481: Support `#allow-multiple!` predicate to enable multiple content ranges
- [ ] Modify `getTarget: Target` field https://github.com/cursorless-dev/cursorless/blob/417e4f9637d8204e671fa34af788dca87cdd5686/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/scope.types.ts#L38 to `getTargets(): Target[]`
- [ ] Add predicate `#allow-multiple!` that sets a `allowMultiple` field on the node output 
- [ ] When processing scopes in base tree sitter scope handler https://github.com/cursorless-dev/cursorless/blob/417e4f9637d8204e671fa34af788dca87cdd5686/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler/BaseTreeSitterScopeHandler.ts#L43 detect if there are multiple scopes with identical domains.
- [ ] If it happens, check the nodes for an `allowMultiple` field.  If it exists, merge into one scope that has multiple content ranges.  If it doesn't exist, throw an error saying they need to use the predicate if it was intentional

pokey:
- Depends on https://github.com/cursorless-dev/cursorless/pull/1509
- Fixes #1481

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
+++ b/packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
@@ -78,6 +78,7 @@ export class TreeSitterQuery {
             name,
             node,
             range: getNodeRange(node),
+            allowMultiple: false,
           })),
         }),
       )
@@ -108,6 +109,9 @@ export class TreeSitterQuery {
             range: captures
               .map(({ range }) => range)
               .reduce((accumulator, range) => range.union(accumulator)),
+            allowMultiple: captures
+              .map((capture) => capture.allowMultiple)
+              .some(Boolean),
           };
         });
 
--- a/packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
+++ b/packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
@@ -109,9 +109,7 @@ export class TreeSitterQuery {
             range: captures
               .map(({ range }) => range)
               .reduce((accumulator, range) => range.union(accumulator)),
-            allowMultiple: captures
-              .map((capture) => capture.allowMultiple)
-              .some(Boolean),
+            allowMultiple: captures.some((capture) => capture.allowMultiple),
           };
         });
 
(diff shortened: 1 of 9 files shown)


packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler/TreeSitterScopeHandler.ts
```diff
@@ -33,38 +35,49 @@ export class TreeSitterScopeHandler extends BaseTreeSitterScopeHandler {
   protected matchToScope(
```
pokey, line 35: this function turned out pretty ugly.  I wonder if I should have another version of `getRelatedRange` that just returns a `range` like it used to, as don't care about `allowMultiple` for any of the related ranges other than `.iterationScope`

packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
```diff
@@ -99,6 +100,9 @@ export class TreeSitterQuery {
           range: captures
             .map(({ range }) => range)
             .reduce((accumulator, range) => range.union(accumulator)),
+          allowMultiple: captures
```
AndreasArvidsson, line 103: Why not `captures.some(capture => capture.allowMultiple)`?

AndreasArvidsson (approved):

packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
pokey: Oh yeah that's the way to go. Not sure how I landed here 🤔. I blame copilot 😅

AndreasArvidsson: Sure blame the servants :p

PR #1509, opened 2023-05-28
Support multiple targets per scope

pokey:
- Depends on https://github.com/cursorless-dev/cursorless/pull/1506

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/constructScopeRangeTarget.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/constructScopeRangeTarget.ts
@@ -17,13 +17,20 @@ export function constructScopeRangeTarget(
   isReversed: boolean,
   scope1: TargetScope,
   scope2: TargetScope,
-): Target {
+): Target[] {
   if (scope1 === scope2) {
-    return scope1.getTarget(isReversed);
+    return scope1.getTargets(isReversed);
   }
 
-  const target1 = scope1.getTarget(isReversed);
-  const target2 = scope2.getTarget(isReversed);
+  const targets1 = scope1.getTargets(isReversed);
+  const targets2 = scope2.getTargets(isReversed);
+
+  if (targets1.length !== 1 || targets2.length !== 1) {
+    throw Error("Scope range targets must be single-target");
+  }
+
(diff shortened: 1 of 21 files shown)


packages/cursorless-engine/src/processTargets/modifiers/constructScopeRangeTarget.ts
```diff
@@ -17,13 +17,20 @@ export function constructScopeRangeTarget(
...
   }
 
-  const target1 = scope1.getTarget(isReversed);
-  const target2 = scope2.getTarget(isReversed);
+  const targets1 = scope1.getTargets(isReversed);
+  const targets2 = scope2.getTargets(isReversed);
+
+  if (targets1.length !== 1 || targets2.length !== 1) {
+    throw Error("Scope range targets must be single-target");
+  }
```
pokey, line 30: wasn't sure what to do in this case.  seems like a corner case that will never happen.  But basically it's if user says "take tags air past bat", which doesn't really make any sense

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler/BaseTreeSitterScopeHandler.ts
pokey, line 1: This is the punchline.  We merge tree-sitter scopes with the same domain.  See #1507 for a change that requires developer to specifically tag these captures with `#allow-multiple!`

packages/cursorless-engine/src/processTargets/modifiers/surroundingPair/index.ts
```diff
@@ -89,7 +89,7 @@ function processSurroundingPairCore(
       return findSurroundingPairTextBased(
         editor,
         range,
-        containingScope.contentRange,
+        containingScope[0].contentRange,
```
pokey, line 92: wasn't sure what to do here, but I don't think it matters too much: it's for legacy surrounding pair modifier, and would be the situation where there are is a text fragment capture that has multiple targets, which I can't really see being a case we'll need to support

AndreasArvidsson (approved):

PR #1513, opened 2023-06-01
Rename incorrect usage of `targets`

AndreasArvidsson:
## Checklist

### DIFF
--- a/cursorless-talon/src/actions/find.py
+++ b/cursorless-talon/src/actions/find.py
@@ -5,5 +5,5 @@ mod = Module()
 
 @mod.action_class
 class Actions:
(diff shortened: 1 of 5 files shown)


auscompgeek (approved):

pokey (approved):
Our naming is a bit confusing in this regard. Extension-side, these would be plural, because a target is a single range. Technically these are all target descriptors, which could result in multiple targets being output. I wonder if we should start using `targetDescriptor` more here. I've started to move in that direction extension side. But I _think_ this is an improvement 😄

AndreasArvidsson:
Very true. We should also probably have a proper cursorless target descriptor class instead of just a dictionary.

PR #1518, opened 2023-06-07
Migrate `functionName` to just pipe `namedFunction` into `name`

pokey:
I had the thought that with our compositional pipelines, `"funk name"` and `"class name"` should just be equivalent to `"name funk"` and `"name class"`, respectively, as the latter would just expand to function and then get its name.  This would simplify implementation, and lead to more consistency.  

I implemented the migration, and 10 tests fail.

On the bright side, most of the failures reveal inconsistencies in the way we handle `"funk"` vs `"name"` vs `"funk name"`, ie `"funk name"` sees something as a function but `"funk"` misses it, and should be fixed, or `"name"` is missing something that `"funk name"` hits, and should be fixed

Unfortunately, there are a couple of test cases that are more problematic.  The biggest problem is the C++ test https://github.com/cursorless-dev/cursorless/blob/65fab51e0bb76a2555883d9a28c0fbce25da432a/packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/cpp/takeClassName3.yml#L10 In this case `"class name"` selects `ClassName` before the `::`, but of course `"name class"` breaks because we're not in a class.  Not sure what to do about this one

## Checklist

- [ ] Remove `functionName` and `className` from https://github.com/cursorless-dev/cursorless/blob/65fab51e0bb76a2555883d9a28c0fbce25da432a/packages/common/src/types/command/PartialTargetDescriptor.types.ts#L75-L133
- [ ] Figure out how to handle `TODO` comment, and remaining more complex modifiers that contain other modifiers
- [ ] See if users are actually using the CPP example
- [ ] Talon-side, remove scopes from default csv, but allow it to be present if it's already there


packages/cursorless-engine/src/core/commandVersionUpgrades/upgradeV5ToV6/upgradeV5ToV6.ts
```diff
@@ -0,0 +1,135 @@
...
+    case "everyScope":
+    case "ordinalScope":
+    case "relativeScope": {
+      const upgradedScopeType = upgrades[modifier.scopeType.type];
+
+      if (upgradedScopeType == null) {
+        return [modifier];
+      }
+
+      return [
```
pokey, line 50: show error message here, but still do the migration

pokey:
After discussing with users, we've decided that we can't get rid of `"class name"`, but we should be able to get rid of `"funk name"`, once we've fixed the inconsistencies. But we'd need deprecation path for `"funk name"`

For `"class name"` we should probably replace this with `"namespace"` so that people don't expect `"funk name"` to work because `"class name"` works.

AndreasArvidsson:
Replaced by https://github.com/cursorless-dev/cursorless/pull/3038

Issue #707, opened 2022-05-30
Create scope type for link in a chain of field accesses


pokey:
In the following code:

```typescript
foo.bar.baz(whatever).bongo;
```

This scope type would capture each of `foo`, `bar`, `baz`, `baz(whatever)`, and `bongo`, and include the `.` in the removal range.

This scope type could work in many languages

AndreasArvidsson:
Do you mean that this would result in multiple selections?

pokey:
no

pokey:
but the "every" version of this would result in multiple selections

pokey:
Had a use case yesterday where I wanted to bring an access after a function call. Eg

```ts
aaa.bbb()
ccc()
```

I wanted to do "bring access bat after cap", but had to settle for "bring access bat after call cap"

pokey:
I just found myself having code `aaa.bbb()` and wanting to say `"chuck access"` to result in `bbb()`

PR #1519, opened 2023-06-08
Add `"access"` scope type

pokey:
- Implements #707 for Typescript
- Implements #707 for Python

## Questions

- [ ] Unfortunately, both the queries and tests are almost identically duplicated between Typescript and Python 😕.
  - [ ] I wonder if we should support some kind of templating language, eg handlebars, along with a meta-updater-style approach where we keep the generated files in source control as well
  - [ ] For the test cases, we could maybe support a list of language ids in the recorded test yaml 🤷‍♂️
- [ ] Also, should the `.` be considered part of the content range, or just the removal range?
- [ ] Should a function call on its own be considered `"access"`?  eg `foo()`.  What about an identifier on its own, eg `foo`?

## Checklist

- [ ] Make it so eg `bar` or `foo()` on its own is an access? That might cover a lot of things 🤔. Eg what happens with something in parens? That could end up as part of access, eg `(a + b).c`
- [ ] What does happen with `(a + b).c`?


queries/javascript.fieldAccess.scm
```diff
@@ -0,0 +1,85 @@
...
+  )
+  (#not-type? @fieldAccess call_expression member_expression subscript_expression)
+)
+
+;; foo().bar
+;; ^^^^^
+(
+  (member_expression
+    object: (call_expression
+      function: (_) @dummy
```
pokey, line 15: wasn't sure what to call nodes like this where we're just using them to allow us to write a predicate that rejects the pattern.  We should probably standardize on something that allows multiple per pattern, because in the future I'd like to be strict about allowable capture names to detect typos in capture names, which today just result in scopes silently not working

AndreasArvidsson:
I don't think the dot should be part of the content range. I view this as a delimiter.

I definitely understand what you mean about the test cases. On fixture for multiple languages might be reasonable.

pokey:
yeah I'm on the fence, because it is a bit like a delimiter, but not quite.  Comparing it to "arg", in that case, an arg can exist without the `,`, whereas something that is after the first element in the chain can't exist without the `.`. So the `.` really feels more like it is a part of the access.  That being said, I do think you'd probably expect "change" to leave the `.` 🤔

To make it a delimiter, we'd need to:

1. support insertion delimiters for our tree-sitter queries, and
2. Ensure that a single function call / identifier / access is considered an "access"

Which are not unreasonable anyway imo

pokey:
I guess one argument to treat it as a delimiter is that the decision to use `?.` instead of `.` is determined by the preceding identifier rather than the following, so arguably it's not wholly owned by the trailing access

pokey:
aligned on having it be delimiter, so need the extra things above

pokey:
Update from meet-up (see https://github.com/cursorless-dev/cursorless/issues/1923#issuecomment-1745342445 for more context):

- Make it private by not having it appear in csvs, but don't throw error if we manually add it to the csv. Also call it `private.access`
- Ship it as is where it includes the `.` in the content range

pokey:
Ok it's now a private scope type. You'll need to add a line like the following to your `modifier_scope_types.csv`:

```csv
access, private.fieldAccess
```

@AndreasArvidsson it's worth giving this a whirl to make sure that:

1. You can't use it without adding the above line, and there's no error if you don't
2. It works if you do add the above line

pokey:
Keep in mind that someone could still copy one of our settings csvs and get it that way, but hopefully the comment helps 🤷

AndreasArvidsson:
Tested and works as expected

pokey:
Yeah that would be nice. Could prob build something using the pair scopes once those are migrated

pokey:
@AndreasArvidsson I think this one is good to go wehen you get a minute

AndreasArvidsson:
There are a lot of unfinished tasks on this pr and, but you're still happy to merge?

AndreasArvidsson (approved):

pokey:
yeah I should prob check them, but I left them because this PR is really kind of unfinished but we want to merge it as a private feature to get some feedback from real use

PR #1520, opened 2023-06-08
Fix `"instance first three chars"`

pokey:
The problem was that constructing a character range inadvertently resulted in a target with `isToken = true`, because when we clone the anchor target to create the new range target, we forgot to include the `isToken = false`.

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/targets/PlainTarget.ts
+++ b/packages/cursorless-engine/src/processTargets/targets/PlainTarget.ts
@@ -20,5 +20,10 @@ export default class PlainTarget extends BaseTarget {
   getTrailingDelimiterTarget = () => undefined;
   getRemovalRange = () => this.contentRange;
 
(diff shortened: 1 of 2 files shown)


packages/cursorless-engine/src/processTargets/targets/PlainTarget.ts
```diff
@@ -20,5 +20,10 @@ export default class PlainTarget extends BaseTarget {
   getTrailingDelimiterTarget = () => undefined;
   getRemovalRange = () => this.contentRange;
 
-  protected getCloneParameters = () => this.state;
+  protected getCloneParameters() {
+    return {
+      ...this.state,
+      isToken: this.isToken,
```
pokey, line 26: We really gotta figure out a way to make the clone parameters have a stronger type.  Forgetting to include custom parameters is something that has hit us multiple times

Issue #259, opened 2021-09-05
Visualization for scope types


pokey:
- [x] add a command "show scopes <scopeType>" eg "show scopes funk"
- [x] will highlight all scopes on screen using semi-transparent background to make nesting visible
- [ ] could also do "show scope hierarchy" which would show hierarchy of all scope types containing given scope
- [x] note these are cursorless scope types, not raw tree-sitter scope types
- [ ] could use this one for keyboard version and do it as the user types
- [x] need to think of good api for scope types so that we can get this information easily
- [ ] could have "hint scopes <scopeType>" to make it disappear after a couple seconds, or after next command
- [x] might be useful to map to a noise?  not by default
- [ ] think about making it nonintrusive enough to leave on for a while
- [x] update as you scroll?

AndreasArvidsson:
The scope visualizer and the sidebar is good enough

PR #1523, opened 2023-06-12
Scope visualizer

pokey:
Partially addresses #259

## Checklist

- [ ] Make docs page for migrating scopes to new query setup. Be sure to link to contributor intro doc at the top as well.  Link to this page from error message
- [ ] File issue to migrate all our vscode references to use new thin wrapper
- [ ] File issue about reworking removal highlight range to return generalised range, and not using line range for line target content range
- [ ] Flatten settings?  Might make it easier to tweak one specific color https://github.com/cursorless-dev/cursorless/labels/to%20discuss
- [x] Add note to removal highlight ranges in target type and removal ranges mentioning example of document with only `\n`. In that case first and second line have identical removal range but want different removal highlight range
- [x] What to do when word wrapping is enabled?  It ends up looking like the scope ends when the line wraps.  Not sure there's much we can do about that unfortunately
- [x] Use these for normal highlights?
- [x] Use these for normal flashes?
- [x] Use property / random testing where we just walk through line char by char to ensure it is right: intersects with exactly one highlight and has the correct borders on all sides. Special case for zero-length line
- [x] Light theme
- [x] Iteration scopes
- [x] Try setting ranges to empty before disposing to fix strange artifacts
- [x] try dotted / dashed / transparent border on edges that continue
- [x] Figure out where hats are getting siphoned to.  That's causing spurious error message for "plaintext not implemented"
- [x] Figure out issue where visualizations disappear when you open up settings
- [x] Don't show errors when you're visualizing all editors.  We can add an error in the future if you run it targeted at a specific editor.  Maybe experiment with a code lens
  - [x] Maybe only show the error for the active editor at the moment you say the visualize command
- [x] Make sure visualisation style enum is part of api surface
- [x] Try [blending colours](https://gist.github.com/JordanDelcros/518396da1c13f75ee057) to avoid artifacts. Could try removing rounding and then always do blending. If that doesn't look good could at least have optimisation when domain equals range. Would prob need special api for visualisation because otherwise domain and range are independent
- [x] Have tests for cursorless-engine side of visualisation
- [x] Don't need to have initial 0 event for next line; can just start the line in the mode knowing that next line starts at 0
- [x] File issue for side bar for showing scopes for a given language
- [x] What to do about border colors?
- [x] filter to only ancestors of visible ranges
- [x] Investigate broken flashes
- [x] Investigate lack of dash (see image at bottom)
- [x] Make sure normal highlights still work
- [x] Figure out why I'm only seeing domain in installed version

## Questions

- [x] Should the visualizer run on all editors, or just one? _answer_: both, but let's start with all editors as that's already implemented

### `"funk"`
image

image

### `"funk name"`
image

### `"token"`
image

image

### `"char"`
image

### `"word"`
image

image

### `"line"`
image
image

### `"state"`
image

### `"block"`
image

image

### Bugs

image

image


cursorless-talon/src/cursorless.talon
```diff
@@ -20,3 +20,14 @@ tag: user.cursorless
...
 {user.cursorless_homophone} settings: user.cursorless_show_settings_in_ide()
+
+visualize <user.cursorless_scope_type>:
+    user.private_cursorless_run_rpc_command_and_wait("cursorless.showScopeVisualizer", cursorless_scope_type, "content")
+visualize <user.cursorless_scope_type> removal:
+    user.private_cursorless_run_rpc_command_and_wait("cursorless.showScopeVisualizer", cursorless_scope_type, "removal")
+visualize <user.cursorless_scope_type> iteration:
+    user.private_cursorless_run_rpc_command_and_wait("cursorless.showScopeVisualizer", cursorless_scope_type, "iteration")
+visualize every <user.cursorless_scope_type>:
+    user.private_cursorless_run_rpc_command_and_wait("cursorless.showScopeVisualizer", cursorless_scope_type, "every")
```
pokey, line 31: remove

cursorless-talon/src/cursorless.talon
```diff
@@ -20,3 +20,14 @@ tag: user.cursorless
...
+visualize nothing:
```
pokey, line 32: use `"nothing"` from mark csv

cursorless-talon/src/cursorless.talon
```diff
@@ -20,3 +20,14 @@ tag: user.cursorless
     user.cursorless_wrap(cursorless_wrap_action, cursorless_target, cursorless_wrapper)
 
 {user.cursorless_homophone} settings: user.cursorless_show_settings_in_ide()
+
+visualize <user.cursorless_scope_type>:
+    user.private_cursorless_run_rpc_command_and_wait("cursorless.showScopeVisualizer", cursorless_scope_type, "content")
+visualize <user.cursorless_scope_type> removal:
+    user.private_cursorless_run_rpc_command_and_wait("cursorless.showScopeVisualizer", cursorless_scope_type, "removal")
+visualize <user.cursorless_scope_type> iteration:
```
pokey, line 28: add `scope_visualizer.csv`

packages/cursorless-engine/src/ScopeVisualizer/ScopeVisualizerImpl.ts
```diff
@@ -0,0 +1,56 @@
+import { ScopeRenderer, showError } from "@cursorless/common";
+import { ModifierStageFactory } from "../processTargets/ModifierStageFactory";
+import { ScopeHandlerFactory } from "../processTargets/modifiers/scopeHandlers/ScopeHandlerFactory";
+import { ide } from "../singletons/ide.singleton";
+import { PerEditor } from "../util/PerEditor";
+import { EditorScopeVisualizer } from "./EditorScopeVisualizer";
+import { ScopeVisualizer } from "..";
+
+export class ScopeVisualizerImpl implements ScopeVisualizer {
```
pokey, line 9: call this `ScopeRangeProvider`?  Then probably we extract the scope extraction code and potentially use it for #1524

packages/cursorless-vscode/src/extension.ts
```diff
@@ -63,6 +64,7 @@ export async function activate(
     testCaseRecorder,
     storedTargets,
     hatTokenMap,
+    scopeVisualizer: engineScopeVisualizer,
```
pokey, line 67: some provider-based name

packages/cursorless-vscode/src/ScopeVisualizerCommandApi.ts
```diff
@@ -0,0 +1,8 @@
+import { ScopeType } from "@cursorless/common";
+
+export interface ScopeVisualizerCommandApi {
+  start(scopeType: ScopeType, visualizationType: VisualizationType): void;
+  stop(): void;
+}
+
+export type VisualizationType = "content" | "removal" | "iteration" | "every";
```
pokey, line 8: remove every

packages/common/src/ide/types/IdeScopeVisualizer.ts
```diff
@@ -0,0 +1,37 @@
...
+    targets?: TargetRanges[];
+  }[];
+}
+
+export interface ScopeVisualizerConfig {
+  scopeType: ScopeType;
+  includeScopes: boolean;
+  includeIterationScopes: boolean;
+  includeIterationNestedTargets: boolean;
+}
```
pokey, line 37: make this the argument to a function on the the engine api surface called something like `onScopeRangesChanged`

cursorless-talon/src/cursorless.talon
pokey: I opted to just make it its own entry

pokey:
Split into the following PRs:

- #1643
- #1644
- #1645
- #1646
- #1647
- #1648
- #1649
- #1650
- #1651
- #1652
- #1653

PR #1528, opened 2023-06-13
Fix snippet insertion indentation

### LINKED ISSUE #1527: Snippet with `insertionScopeType: branch` will inherit the target branch's indentation level (python)
Input:
```
if True:
    air()
```

Snippet definition (named `else if`)
```
{
  "elseIfStatement": {
    "definitions": [
      {
        "scope": {
          "langIds": [
            "python"
          ]
        },
        "body": [
          "elif $condition:",
          "\t$consequence"
        ]
      }
    ],
    "description": "Else if statement",
    "variables": {
      "consequence": {
        "wrapperScopeType": "statement"
      }
    },
    "insertionScopeTypes": [
      "branch"
    ]
  }
}
```

Saying `snippet else if after air` results in

```
if True:
    air()
    elif _:
        _
```

But we expect

```
if True:
    air()
elif _:
    _
```

pokey:
- Fixes #1527

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/targets/PositionTarget.ts
+++ b/packages/cursorless-engine/src/processTargets/targets/PositionTarget.ts
@@ -44,8 +44,15 @@ export default class PositionTarget extends BaseTarget {
     if (
       this.insertionDelimiter === "\n" &&
       this.position === "after" &&
-      this.thatTarget.contentRange.isSingleLine
+      this.state.thatTarget!.contentRange.isSingleLine
     ) {
+      // If the target that we're wrapping is not a single line, then we
+      // want to compute indentation based on the entire target.  Otherwise,
+      // we allow the editor to determine how to perform indentation.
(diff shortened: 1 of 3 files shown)


packages/cursorless-engine/src/processTargets/targets/PositionTarget.ts
```diff
@@ -44,8 +44,15 @@ export default class PositionTarget extends BaseTarget {
     if (
       this.insertionDelimiter === "\n" &&
       this.position === "after" &&
-      this.thatTarget.contentRange.isSingleLine
+      this.state.thatTarget!.contentRange.isSingleLine
```
AndreasArvidsson, line 47: Are you really sure that we always have a that target?

pokey: yeah because it's a position target, and we already use a `!` on the `thatTarget` in the constructor.  Moreover, this stuff will all get reworked in #803

PR #1534, opened 2023-06-15
Make BaseTarget parameters getters stricter

auscompgeek:
## Context

https://github.com/cursorless-dev/cursorless/pull/1520#discussion_r1222855994

## What

This adds a generic to `BaseTarget`, which is expected to receive the type of the subclass' parameters. This tries its best to enforce that is the case, and ensures implementations of `getCloneParameters` return that declared type.

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/targets/BaseTarget.ts
+++ b/packages/cursorless-engine/src/processTargets/targets/BaseTarget.ts
@@ -15,12 +15,16 @@ import {
   createContinuousRangeUntypedTarget,
 } from "../targetUtil/createContinuousRange";
 
-/** Parameters supported by most target classes */
-export interface CommonTargetParameters {
+/** Parameters supported by all target classes */
+export interface MinimumTargetParameters {
   readonly editor: TextEditor;
   readonly isReversed: boolean;
-  readonly contentRange: Range;
   readonly thatTarget?: Target;
+}
+
+/** Parameters supported by most target classes */
+export interface CommonTargetParameters extends MinimumTargetParameters {
+  readonly contentRange: Range;
 }
 
 export interface CloneWithParameters {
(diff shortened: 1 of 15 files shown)


packages/cursorless-engine/src/processTargets/targets/BaseTarget.ts
```diff
@@ -39,7 +51,7 @@ export default abstract class BaseTarget implements Target {
   isNotebookCell = false;
   isWord = false;
 
-  constructor(parameters: CommonTargetParameters) {
+  constructor(parameters: TParameters & CommonTargetParameters) {
```
auscompgeek, line 54: This ensures that whatever type is declared actually ends up existing in the subclass constructor somehow.

packages/cursorless-engine/src/processTargets/targets/SubTokenWordTarget.ts
```diff
@@ -43,11 +43,12 @@ export default class SubTokenWordTarget extends BaseTarget {
     return getDelimitedSequenceRemovalRange(this);
   }
 
-  protected getCloneParameters() {
+  protected getCloneParameters(): SubTokenTargetParameters {
     return {
       ...this.state,
       leadingDelimiterRange: this.leadingDelimiterRange_,
       trailingDelimiterRange: this.trailingDelimiterRange_,
+      insertionDelimiter: this.insertionDelimiter,
```
auscompgeek, line 51: All this ceremony wasn't for nothing, it seems! I added the return type here for autocomplete.

pokey (approved):
this is great.  left a couple questions inline mostly for my own education 😄

I'm not totally in love with needing to specify that generic parameter, when you'd think TS would infer it from some function arg or something, but it doesn't, so I don't think we have a choice

btw elsewhere, I've made it so that it extracts what it needs from the type itself, and you just pass the type in as the only generic.  we may consider going down that road if we have more things we need to infer.  See eg https://github.com/cursorless-dev/cursorless/blob/fcdf70c97e89ab773dd23aad3a9931fe4a1267b6/packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts#L55

packages/cursorless-engine/src/processTargets/targets/BaseTarget.ts
pokey: oh really?  so the type signature of the base class constructor influences the type signature of derived class constructors?  That's surprising to me

packages/cursorless-engine/src/processTargets/targets/SubTokenWordTarget.ts
pokey: So you don't get autocomplete for free from the generic at the top?

Issue #1524, opened 2023-06-12
New representation for scope handler tests


pokey:
# The problem

We need a better way to represent our test cases for scope handlers, because today we end up needing to record a bunch of commands to test edge cases, and the tests aren't easy to read.  In addition, once #1523 is merged, we expect that most scope type development will be done by viewing the output of the scope visualizer, so we'd like test cases that mirror the scope visualizer as closely as possible

# The solution

We'd like an ascii representation that is readable by humans and parsable by machine.  The idea is as follows:

- We intersperse comment lines among the source code that point out regions in the line above
- The comment lines begin with the language's comment token (eg `//`), followed by `!`
- Each source code line is indented by extra whitespace at the beginning to allow the comment lines to be able to point at the start of the line above if necessary.  If we didn't do this, then a source code line could begin at  the same spot as the `//` on the line below, so we couldn't point out the beginning of the line.
- The content range is marked by `{^^^}`
- The domain is marked by `[---]`
- The removal range is marked by `(xxx)`
- The iteration scope is marked by `<***>`
- We need to figure out `interior` / `leadingDelimiter` / `trailingDelimiter`
  - Would like to avoid unicode if possible
  - Here are suggestions from chatGPT:
    - The `interior` is marked by `/###\`
    - The `leadingDelimiter` is marked by `«@@@»`
    - The `trailingDelimiter` is marked by `§$$$¶`
- If any range is equal to `contentRange`, it does not get a line, as that is the implicit assumption
- In each of the above, the surrounding brackets are _outside_ of the range, so that we can handle empty ranges by using eg `{}`
- Each range type (eg content, domain, removal, etc) gets its own comment line
- If there are overlapping ranges, the range with later start or earlier end gets its own set of lines after the other scope, beginning with `//!2` (or `//!3`, etc).  Note that ranges are overlapping even if they are 1 character apart, because otherwise there's no room for the brackets
- The start of the file will contain metadata lines to indicate scope type being tested and language id, eg
  ```ts
   //!! scope: {type: namedFunction}
   //!! languageId: typescript, plaintext
   ```
- Note that multiple examples of the same scope type could be present in one file, so that we don't need a million test files like we have today
- Note that we don't handle indicating `isPreferredOver`; need special tests for that

## Components

- [ ] Automatically generate test case from source code.  Can probably steal some code from #1523 
[...]

PR #1535, opened 2023-06-16
Add almost testable comments to `.scm` files

pokey:
## Checklist


queries/javascript.core.scm
```diff
@@ -14,20 +17,33 @@
...
   (#not-type? @dummy variable_declarator)
 ) @_.domain
 
+;; Special cases for `(let | const | var) foo = ...;` because the full statement
+;; is actually a grandparent of the `name` node, so we want the domain to include
+;; this full grandparent statement.
 (
   [
+    ;;!! (const | let) foo = ...;
+    ;;!  --------------^^^-------
```
pokey, line 38: note that I merged into one line for compactness

queries/javascript.core.scm
```diff
@@ -14,20 +17,33 @@
...
+  ;; special cases from above because they can't be exported
   (#not-type? @dummy variable_declarator)
 ) @_.domain
 
+;; Special cases for `(let | const | var) foo = ...;` because the full statement
+;; is actually a grandparent of the `name` node, so we want the domain to include
+;; this full grandparent statement.
 (
   [
+    ;;!! (const | let) foo = ...;
```
pokey, line 37: Note that the `(const | let)` and `...` make this not strictly Javascript, so doesn't fully implement the idea from https://github.com/cursorless-dev/cursorless/issues/1524#issuecomment-1594788592.  On the fence about whether we want to go all out and make these testable or just use it as a convenient notation

I wonder if in cases where it's not fully valid syntax we leave off the `!!` so that future auto testing impl doesn't pick these up 🤔.  Or maybe we leave it so that we're forced to clean these up if we go down that road

We could also leave off the `!!` and `!` altogether, but in some cases the clarity is helpful; see below

queries/javascript.jsx.scm
```diff
@@ -1,56 +1,82 @@
...
 ) @_.domain
 
+;;!! <foo>bar</foo>
+;;!  --------^^^^^^
 (jsx_element
   (jsx_closing_element) @xmlEndTag @xmlBothTags
   (#allow-multiple! @xmlBothTags)
 ) @_.domain
 
+;;!! <foo/>
```
pokey, line 34: If there's no annotation below, it's implied that the whole thing is the content range and domain

queries/javascript.jsx.scm
```diff
@@ -1,56 +1,82 @@
...
   (jsx_fragment) @xmlEndTag @xmlBothTags @_.domain
   (#child-range! @xmlEndTag -3)
   (#child-range! @xmlBothTags -3)
   (#allow-multiple! @xmlBothTags)
 )
 
+;; Sets `name` to be empty range inside the fragment tag:
+;;!! <>foo</>
+;;!  {}    {}
+;;!  --   ---
```
pokey, line 79: This is the example that motivated me to use the `!!` and `!`, because in this case it might not be clear what's code and what's annotations

pokey: Also note that I'm including multiple scopes on one line here, as otherwise it gets pretty long

AndreasArvidsson (approved):

PR #1541, opened 2023-06-19
Make it easier to have unit tests in VSCode packages

pokey:
Introduces a new convention for tests that end with `.vscode.test.ts` must run within VSCode; other tests can be run as unit tests outside of VSCode.  This makes it easier to add unit tests within packages such as `cursorless-vscode`.  Just because a package is designed to be used with VSCode doesn't mean that all of its tests should have to be run within a VSCode context.  It might (and should!) have support code that consists of pure functions with no dependencies.  

- Required by #1523 

## Checklist


packages/test-harness/src/runners/endToEndOnly.ts
pokey, line 1: this wasn't used anywhere

AndreasArvidsson (approved):

Issue #1115, opened 2022-11-03
Use an output/log channel for Cursorless


AndreasArvidsson:
Instead of just spamming in the default channel create our own Cursorless channel. Ones this is enabled maybe we could even put debug setting to true by default

PR #1554, opened 2023-06-29
 Change debug log to represent tree sitter query syntax 

AndreasArvidsson:
from
```
> source_file
> source_file
>> declarations
>> declarations
>>> settings_declaration
>>> settings_declaration
>>>> left: settings()
>>>> left: settings()
>>>> "settings()"
>>>> "settings()"
```

to
```
(source_file
 (declarations
  (settings_declaration
   left: "settings()" @cursor
  )
 )
)
```

`console.debug` is the culprit in why we got to duplicated lines. Switching to `console.log` removes the duplicated lines. I think this is an improvement until we do #1115 

I'm not totally sure if I like the `@cursor`. 

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/core/Debug.ts
+++ b/packages/cursorless-engine/src/core/Debug.ts
@@ -93,27 +93,55 @@ export class Debug {
     }
 
     const cursor = node.tree.walk();
-    this.printCursorLocationInfo(cursor, 0);
-
-    for (let i = 1; i < ancestors.length; ++i) {
-      cursor.gotoFirstChild();
-      while (cursor.currentNode().id !== ancestors[i].id) {
-        if (!cursor.gotoNextSibling()) {
-          return;
-        }
-      }
-      this.printCursorLocationInfo(cursor, i);
-    }
-
-    const leafText = ancestors[ancestors.length - 1].text
-      .replace(/\s+/g, " ")
-      .substring(0, 100);
-    console.log(">".repeat(ancestors.length), `"${leafText}"`);
+    this.printCursorLocationInfo(ancestors, cursor, 0);
   }
 
-  private printCursorLocationInfo(cursor: TreeCursor, depth: number) {
+  private printCursorLocationInfo(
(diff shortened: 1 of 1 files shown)


packages/cursorless-engine/src/core/Debug.ts
```diff
@@ -93,27 +93,55 @@ export class Debug {
...
+  }
+
+  private printCursorLocationInfo(
+    nodes: SyntaxNode[],
+    cursor: TreeCursor,
+    index: number,
+  ) {
+    const field = cursor.currentFieldName();
+    const fieldText = field != null ? `${field}: ` : "";
+    const ident = " ".repeat(index);
```
pokey, line 106: Should be `indent`, no?

packages/cursorless-engine/src/core/Debug.ts
```diff
@@ -93,27 +93,55 @@ export class Debug {
...
+      if (nodeIsLast) {
+        text += ")";
       }
-      this.printCursorLocationInfo(cursor, i);
+    } else {
+      text += `"${cursor.nodeType}"`;
+    }
+
+    if (nodeIsLast) {
+      text += " @cursor";
```
pokey, line 121: Mixed feelings as well. Is it not obvious that this is where your cursor is due to the indentation? I think this tag might be a bit confusing if you don't expect it, but maybe it's helpful? Maybe `yourCursor` to make it more obvious? Idk. Could go either way

pokey (approved):
Nice. Left a couple minor comments; feel free to merge once addressed

packages/cursorless-engine/src/core/Debug.ts
AndreasArvidsson: I decided to just drop it

packages/cursorless-engine/src/core/Debug.ts
AndreasArvidsson: fixed

Issue #817, opened 2022-07-03
Add language support for `.talon` files


pokey:
Now that there is a [tree-sitter for Talon files](https://github.com/wenkokke/tree-sitter-talon), we should leverage this to add proper Cursorless support for Talon files

See https://www.cursorless.org/docs/contributing/adding-a-new-language/

phillco:
Just FYI, I've started working on this on a branch. I will push it when it gets a bit closer to completion :)

phillco:
Developing in this branch: https://github.com/phillco/cursorless/tree/kotlin-talon

Most of the functionality is working (with a couple of notes), just need to record the tests.

phillco:
Quirks with the grammar I noticed so far: 

- Some of the notes like `tag()` and `settings()` have parentheses in the names. Cursorless has difficulty matching them.
- The `-` node inside of the settings block is a sibling to the other rules, so we can't easily target just the rules without writing a custom matcher. It would be better if there was a parent node for all of the rules.
- @AndreasArvidsson suggested that if there is a general convention of using left/right parent blocks for the left hand side/right hand side of things we could reduce the number of matches needed, e.g. we could reduce this:

```
  value: [
    "assignment[right]",
    "match[pattern]",
    "include_tag[tag]",
]
```

pokey:
- For the type names with parentheses, that feels like a Cursorless-specific problem, so I'm not sure worth changing the parser.  Also, I believe @AndreasArvidsson mentioned he has a fix for those now, and the real problem was probably unnamed nodes?
- Filed https://github.com/wenkokke/tree-sitter-talon/issues/11 for second point
- Filed https://github.com/wenkokke/tree-sitter-talon/issues/12 for third point

phillco:
I think Andreas' point was that names just shouldn't have special symbols in them regardless. Also we updated the regex but they still didn't work in Cursorless, so if there's another fix I'm not aware of it

pokey:
Maybe cursorless patterns don't support unnamed children?

Fair point re special chars in type names; I'll file to see what makes sense

pokey:
On second thought, it's actually quite common for node types for unnamed nodes to have special chars.  For example, `"`, `,`, `(`, etc, so unless we want to argue for removing those everywhere, I'm not sure I see a good argument for removing them here?

AndreasArvidsson:
[...]

PR #1555, opened 2023-06-30
Talon language support

AndreasArvidsson:
Resolves https://github.com/cursorless-dev/cursorless/issues/817

**Condition**
- [x] context match
- [x] every context match

**Command**
- [x] command
- [x] every command

**Statement**
- [x] command line
- [x] command block
- [x] settings block
- [x] script action call
- [x] script key action call
- [x] script sleep action call
- [x] script assignment
- [x] key binding
- [x] tag binding
- [x] gamepad binding
- [x] parrot binding
- [x] face binding
- [x] every statement in file
- [x] every statement in command block
- [x] every statement in settings block

**Name / Collection key**
- [x] context match left
- [x] command left
- [x] settings block left
- [x] setting left
- [x] tag() left
- [x] key() left
- [x] gamepad() left
- [x] face() left
- [x] parrot() left
- [x] every name/key in context block
- [x] every name/key in file

**Value** 
- [x] context match right
- [x] command right
- [x] settings block right
- [x] setting right
- [x] tag() right
- [x] key() right
- [x] gamepad() right
- [x] face() right
- [x] parrot() right
- [x] every value in context block
- [x] every value in settings block
- [x] every value in file

**Interior**
- [x] command right
- [ ] settings block right
- [x] every command interior in file

**Function call**
- [x] action
- [x] key action
- [x] sleep action
- [x] every call in command block

**Function callee**
- [x] action
- [x] key action
- [x] sleep action
- [x] every callee in command block

**Argument or parameter**
- [x] action
- [x] key action
- [x] sleep action
- [x] every argument in action
 
**Update Tree sitter patterns**
- [x]  Exclude leading/trailing whitespace from `implicit_string`
     Waiting for release of  https://github.com/wenkokke/tree-sitter-talon/pull/23
- [x] Removed trailing parentheses `key(` and `sleep(` actions
    ~~Waiting for release of https://github.com/wenkokke/tree-sitter-talon/pull/26~~
    Instead solved with a pattern predicate

## Checklist

- [ ] Look into failures with new tree-sitter-talon.  Do they occur on shipped version?
  - [ ] "every value bat" in

          aaa: "bbb"
          ccc: "ddd"

  - [x] https://github.com/wenkokke/tree-sitter-talon/issues/29
  - [x] `"take key"` targeting `settings()`
  - [x] `"take key"` targeting `tag()`
  - [x] Rollback whitespace removal
  - [x] If the above are all fine on shipped tree-sitter-talon, let's just file follow-up to upgrade tree-sitter-talon later


pokey (commented):
One quick comment to have a look at; I'll give a more thorough review later

queries/talon.scm
```diff
@@ -0,0 +1,105 @@
...
+;;!  ^^^^^^^^^^^^^^^^^^^^
+(_
+  [
+    (expression_statement)
+    (assignment_statement)
+    (settings_declaration)
+    (tag_import_declaration)
+    (key_binding_declaration)
+  ] @statement
+) @_.iteration
```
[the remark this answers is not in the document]
AndreasArvidsson: I think I understand what you mean but I feel I'm a bit too new to this to grasp nuances. I would prefer if we had a proper discussion about this when you have the time.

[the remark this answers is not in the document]
AndreasArvidsson: No iteration scope is wildcard anymore.

AndreasArvidsson:
@pokey I have two things I need your help with
1. Pre commit is failing 
2. We have a few tests that are now failing and the resent being that there is empty nodes in the Tree sitter. Sow the iteration scope appears to be an empty node

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
```diff
@@ -19,6 +19,17 @@ class NotType extends QueryPredicateOperator<NotType> {
   }
 }
 
+/**
+ * A predicate operator that returns true if the nodes range is not empty.
+ */
+class NotEmpty extends QueryPredicateOperator<NotEmpty> {
```
AndreasArvidsson, line 25: That Talon Tree sitter can contain an empty matches node if there is no header. When this happens and the user has an empty cursor at the start of the document we get an empty range for the iteration scope for key/value/name/statement.

pokey:
https://github.com/cursorless-dev/cursorless/pull/1712 should fix pre-commit failure here

phillco (approved):

AndreasArvidsson:
@pokey Pre commit fixed. Except for that single update to the Tree sitter extension the rest is ready to go.

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/talon/clearArg.yml
pokey, line 1: What about "chuck arg"?

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/talon/clearEveryCall.yml
pokey, line 1: Not sure it makes sense to have an iteration scope for call, but I guess it's fine? What happens if you have nested calls? Not sure what iteration scope should be for that

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/talon/clearEveryCallee.yml
pokey, line 1: Same here; not sure if we want iteration scope and what happens for nested

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
```diff
@@ -116,6 +127,31 @@ class ChildRange extends QueryPredicateOperator<ChildRange> {
   }
 }
 
+class ShrinkToMatch extends QueryPredicateOperator<ShrinkToMatch> {
```
pokey, line 130: Docstring

pokey (commented):
Looks good with a few minor comments

pokey:
Also worth running through checkboxes on our language PR template. In particular I don't think you have "round" tests within comments / strings, unless I missed those

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/talon/clearArg.yml
AndreasArvidsson: I don't think we have the machinery for scm that we have in typescript definitions for delimiters and since item already works I'm not sure it's worth to make it work to be honest.

AndreasArvidsson:
Without adding `@textFragment` matching pairs in strings and comments just work? I have added tests.

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/talon/clearArg.yml
pokey: I think it should be doable. We're going to need to tackle it at some point; I don't want us to just keep punting it down the field.  Would you like me to take a swing?

AndreasArvidsson: I was more thinking along the lines that we should make it lazy. We should just tell it it's delimiter and the rest should be done in the target the way we do with the other targets?

pokey: Interesting idea 🤔.  How do we handle comments?

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
```diff
@@ -116,6 +127,36 @@ class ChildRange extends QueryPredicateOperator<ChildRange> {
...
+      node: { text },
+    } = nodeInfo;
+
+    const match = text.match(new RegExp(pattern));
+
+    if (match?.index == null) {
+      throw Error(`No match for pattern '${pattern}'`);
+    }
+
+    nodeInfo.range = new Range(
```
AndreasArvidsson, line 151: I just realize why I used to have an if statement that throw and error on multiple lines. Since we don't have access to the editor we can't convert in decease to positions. Just translating position character only works if it's on the same line.

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
```diff
@@ -116,6 +127,36 @@ class ChildRange extends QueryPredicateOperator<ChildRange> {
...
+  name = "shrink-to-match!" as const;
+  schema = z.tuple([q.node, q.string]);
+
+  run(nodeInfo: MutableQueryCapture, pattern: string) {
+    const {
+      range,
+      node: { text },
+    } = nodeInfo;
+
+    const match = text.match(new RegExp(pattern));
```
AndreasArvidsson, line 145: Is it a problem that we're using the text from the node even though the range might have changed?

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
pokey: Ah right. Should be fairly straightforward to add an editor field I'd think no? Would be a bit surprising if things fail when you break a line 😅

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
pokey: Huh. Yeah we should prob add editor and get text that way?

pokey: Does make me a bit nervous that we have two ways to get this stuff that can become out of sync 😬

AndreasArvidsson: I agree. Maybe we should make a simple node type. It basically only needs its type and its children and parent?

pokey: Yeah and then I guess it would return the same type from those. I'd start with just those fields

pokey: Agreed https://github.com/cursorless-dev/cursorless/issues/1725

pokey:
Hmm an unfortunate drawback of the direction I pushed you to go for iteration scopes is that we get some odd behaviour in situations like the following:

image

See how eg `"smile"` is its own iteration scope, because it is a node of type `body`.  Thus when you say `"every value sun"`, you get just ` "smile"`, which is obviously unexpected.

I believe your original approach would work, because it was requiring a statement with a `right` to actually appear as a child.  Idk, mixed feelings, as the new behaviour is "more correct" in some sense.  Eg in theory if you had cursors in a bunch of bodies, you might not want it to leapfrog to the top level if one of them were empty or had only one value 🤔

But in this case I'm leaning towards going with the possibly slightly wrong, but way more useful version, that only considers bodies that have something with a `right`.  WDYT?

AndreasArvidsson:
I guest this is technically correct. If the block had multiple lines you would expect it to be contained by the block. Here it just looks weird because it's on a single line. We could definitely add a predicate for that?

pokey:
I'm not sure it's a single/multiple line thing, because even if it were:

```talon
aaa: "bbb"
ccc: "ddd"
eee:
   "some really long text that should get inserted"
fff: "ggg"
```

I think I'd still expect `"every value"` to refer to all the values, even when I'm at the end of `inserted"`.  I think the criterion here is probably the one you were already using before our long conversation about iteration scopes 😅

But happy to discuss at meet-up

pokey:
I guess we could just leave it, but it does feel like a bit of a gotcha 🤷‍♂️

AndreasArvidsson:
When I said multiline I meant something more like this
```talon
aaa: "bbb"
ccc: "ddd"
eee:
   "some really long text that should get inserted"
   a = 5
fff: "ggg"
```

pokey:
Ok, then what about:

```talon
aaa:
   foo = "bar"
```

In that case I'd argue "every value" should just be `"bar"`, even tho it is single-line by your criterion.  Admittedly my example isn't particularly useful, but I think I'd lean towards a syntactic / semantic construct rather than one based on whitespace

AndreasArvidsson:
> Ok, then what about:
> 
> ```talon
> aaa:
>    foo = "bar"
> ```
> 
> In that case I'd argue "every value" should just be `"bar"`, even tho it is single-line by your criterion. Admittedly my example isn't particularly useful, but I think I'd lean towards a syntactic / semantic construct rather than one based on whitespace

And this is what's happening today I believe. So our current implementation is probably correct semantically?

pokey:
Yes, but I'm arguing that

```talon
aaa: "bbb"
```

and

```talon
aaa:
   bbb = "ccc"
```

should behave differently.  The main issue for me is that you get different behaviour for `"every value"` depending if you're on the left or right side of

```talon
aaa: "bbb"
```

AndreasArvidsson:
I'm fine with it behaving the same. I think the current implementation is reasonable. 

The same is true for `foo(): bar = 2` depending on if you are to the left or the right of the colon

pokey:
the problem for me is when it looks like a key-value map, and behaviour diverges based on whether you're on key or value.  your example doesn't look like a key-value map.  the right-hand side looks like its own key-value map, so I'd expect to just get `2` for `"every value"`

pokey:
let's discuss at meet-up, which at this point is in like 10 mins 😄

pokey:
update from meet-up: decided to revert to the wildcard parent containing a statement with a `right` field heuristic for `name`, `key`, and `value`, but use the existing code for `statement`

AndreasArvidsson:
Reverted to using wildcard for iteration scope

pokey (approved):
looks good to me! have a look at my changes and merge if happy

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
```diff
@@ -97,6 +97,15 @@ class ChildRange extends QueryPredicateOperator<ChildRange> {
  * A predicate operator that modifies the range of the match to shrink to regex
  * match.  For example, `(#shrink-to-match! @foo "\\S+")` will modify the range
  * of the `@foo` capture to exclude whitespace.
+ *
+ * If convenient, you can use a special capture group called `keep` to indicate
+ * the part of the match that should be kept.  For example,
+ *
+ * ```
+ * (#shrink-to-match! @foo "^\s+(?<keep>.*)$")
```
AndreasArvidsson, line 105: Why not just use the first group? <keep> just feels like unnecessary code.

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
```diff
@@ -105,14 +114,18 @@ class ShrinkToMatch extends QueryPredicateOperator<ShrinkToMatch> {
   run(nodeInfo: MutableQueryCapture, pattern: string) {
     const { document, range } = nodeInfo;
     const text = document.getText(range);
-    const match = text.match(new RegExp(pattern));
+    const match = text.match(new RegExp(pattern, "ds"));
```
AndreasArvidsson, line 117: Good catch about the d flag!

queries/talon.scm
```diff
@@ -96,7 +96,13 @@
 ;;!  ^^^^^^^^^^^^^
 (match) @condition
 
-(matches) @condition.iteration
+(
+  (source_file
+    (matches) @condition.iteration @condition @condition.trailing
```
AndreasArvidsson, line 116: I would probably skip the iteration scope and make the domain the entire file. Its just one condition per file?

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
pokey: I just worry that you'll use a `()` for another purpose and be surprised when all of a sudden it narrows to that. I wanted to make sure it was fully intentional

AndreasArvidsson: Not a fan of non capture groups? You want it more explicit?

pokey: Yeah I know non-capture groups exist, but my habit is to just use `()` without thinking. Idk maybe that's a bad habit? I don't feel super strongly, just trying to reduce surprise

queries/talon.scm
AndreasArvidsson: `(matches) @condition.iteration @condition @condition.trailing` looks like the entire matches node is the condition. Shouldn't it be `match` instead then?

pokey: That is also `condition`.  It is now hierarchical.  It is pretty much the exact same pattern we decided to use for `"name"` in Typescript's `const aaa = 0, bbb = 1;`, where

- each nested assignment (`aaa = 0` and `bbb = 1`) is its own domain
- there is a parent scope whose domain is the entire statement that has multiple content ranges, one per sub assignment
- the iteration scope is the entire statement

This situation is analogous, replacing "nested assignment" with "individual condition" and "statement" with "file"

AndreasArvidsson: Not sure if this is prefared, but its fine with me.

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
AndreasArvidsson: Heh maybe. A named group is fine though :)

queries/talon.scm
pokey: I do wonder if we should simplify things and not support the child conditions here 🤔.  Ie there would just be a single scope `"condition"` that refers to the whole condition block from anywhere in the file, and there is no `"every condition"`

pokey: Before we merge, any thoughts on my last two comments here?

AndreasArvidsson: I think a single condition is probably enough. You can use line for a single match.

pokey (approved):

PR #1556, opened 2023-07-02
Added compile clean script

AndreasArvidsson:
## Checklist


package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "pnpm -r build",
     "compile": "tsc --build",
+    "compile-clean": "tsc --build --clean",
```
pokey, line 9: Should we just call this `clean`?

Also, this doesn't fully clean in case you've deleted a file. So maybe we want a full clean command to do that either instead of or in addition to this one

AndreasArvidsson: Renaming it to us clean is reasonable.
What happens if you have deleted a file?

pokey: If you delete a TS file the JS file sticks around

AndreasArvidsson: Got it. Definitely not what we want. What is the most robust way to get rid of all the out directories? cli remove command?

pokey: Probably? Tbh I'd be tempted to defer to subdirectories and let them do their own clean. The top-level command would just call clean on subdirectories. We can use meta-updater to ensure some degree of uniformity. If you're not comfortable wiring all that up I can knock it out when I'm back

AndreasArvidsson: If I don't get to it you're well come to knock it out yourself

AndreasArvidsson: I have now used meta updated to distribute the script to all subdirectories.

pokey (approved):
👌

PR #1557, opened 2023-07-02
Better handling of escaped quotes in textual item

### LINKED ISSUE #1551: `item` in python dict with a value `"\\"` selects multiple items
Input:
```
symbol_key_words = {
    "air": "each",
    "bat": "\\",
    "cap": "gust",
    "drum": "harp",
}
```
`take item air` selects the first item as expected.

`take item drum` selects the three last items in the example.

It works as expected with input:
```
symbol_key_words = {
    "air": "each",
    "bat": "fine",
    "cap": "gust",
    "drum": "harp",
}
```

AndreasArvidsson:
Fixes #1551

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/ItemStage/tokenizeRange.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/ItemStage/tokenizeRange.ts
@@ -41,8 +41,8 @@ export function tokenizeRange(
   const lexemes = text
     // NB: Both the delimiters and the text between them are included because we
     // use a capture group in this split regex
-    .split(/([,(){}<>[\]"'`]|\\"|\\'|\\`)/g)
-    .filter((lexeme) => lexeme.length > 0);
+    .split(/([,(){}<>[\]"'`])|(?<!\\)(\\"|\\'|\\`)/g)
+    .filter(Boolean);
   const joinedLexemes = joinLexemesBySkippingMatchingPairs(lexemes);
   const tokens: Token[] = [];
   let offset = document.offsetAt(interior.start);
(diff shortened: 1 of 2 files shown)


packages/cursorless-engine/src/processTargets/modifiers/ItemStage/tokenizeRange.ts
```diff
@@ -41,8 +41,8 @@ export function tokenizeRange(
   const lexemes = text
     // NB: Both the delimiters and the text between them are included because we
     // use a capture group in this split regex
-    .split(/([,(){}<>[\]"'`]|\\"|\\'|\\`)/g)
-    .filter((lexeme) => lexeme.length > 0);
+    .split(/([,(){}<>[\]"'`])|(?<!\\)(\\"|\\'|\\`)/g)
+    .filter(Boolean);
```
pokey, line 45: Why the switch to boolean?

pokey (approved):
LGTM with minor question

packages/cursorless-engine/src/processTargets/modifiers/ItemStage/tokenizeRange.ts
AndreasArvidsson: Some of the items could be undefined. It's just a more terse way to do `.filter((lexeme) => lexeme != null && lexeme.length > 0);`

pokey: I generally lean toward verbose, and the conversion to boolean behaviour can be a bit unintuitive i believe. Aren't there strange things where the string "0" evaluates to false or something like that?

AndreasArvidsson: I don't believe so but I have no problem being more verbose if you prefer. Updated.

PR #1558, opened 2023-07-02
Surrounding pairs: Fall back to text based algorithm when error node is detected.

### LINKED ISSUE #712: Surrounding pair problem in error python code
`def funk(words:list<str>):`
`take pair` in `<str>` doesn't work

AndreasArvidsson:
Fixes #712

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/surroundingPair/index.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/surroundingPair/index.ts
@@ -99,6 +99,18 @@ function processSurroundingPairCore(
   try {
     node = languageDefinitions.getNodeAtLocation(document, range);
 
+    // Error nodes are unreliable and should be ignored. Fall back to text based
+    // algorithm.
+    if (nodeHasError(node)) {
+      return findSurroundingPairTextBased(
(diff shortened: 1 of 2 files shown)


packages/cursorless-engine/src/processTargets/modifiers/surroundingPair/index.ts
```diff
@@ -148,3 +160,22 @@ function processSurroundingPairCore(
     scopeType,
   );
 }
+
+function nodeHasError(node: SyntaxNode, includeChildren = false): boolean {
```
pokey, line 164: This seems quite conservative. Are there test cases that fail if you don't eg check grand-siblings?

AndreasArvidsson: The added test for example. The error node is a child to an ancestor of the selected node.

pokey: as discussed, let's use your proposed version for now and if it becomes bothersome we can revisit.  see also #1600

pokey (approved):

PR #1565, opened 2023-07-02
Added clear twin test

### LINKED ISSUE #328: Fix python surrounding pairs in new `tree-sitter-python` version
In newer versions of tree-sitter-python, it uses `string_start` and `string_end` as the type for string delimiter tokens. This will break all of our surrounding pairs in python.  

We should fix this by actually looking at the token text to rather than just its type. We to be a bit careful though because for some cases we actually do prefer to use the type, for example with a format string or triple quote.  We probably want to match the type, and then have a mapping from text to what kind of pair it is, eg `f"""` should map to `doubleQuotes`, etc.  Note that this will require us to take into account language id when determining surrounding pair

We should probably solve this one after #1061 

In addition, it uses `string_content` as the type for the content of the string, which appears to be different from today.

AndreasArvidsson:
## Checklist

### DIFF
--- a/packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/surroundingPair/parseTree/python/clearQuad.yml
+++ b/packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/surroundingPair/parseTree/python/clearQuad.yml
@@ -1,0 +1,22 @@
+languageId: python
+command:
+  version: 5
+  spokenForm: clear quad
+  action: {name: clearAndSetSelection}
+  targets:
+    - type: primitive
+      modifiers:
+        - type: containingScope
+          scopeType: {type: surroundingPair, delimiter: doubleQuotes}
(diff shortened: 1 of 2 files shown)


pokey:
wow super weird but this one doesn't work when I try it in development, but the tests pass locally 🤔

AndreasArvidsson:
Weird. It works fine for me in development. Maybe you need to clean out your out directories?

pokey:
Idk, but this is def not going to work with later versions of the tree-sitter-python parser, where the node type becomes `string_start`.  That breaks `"` strings as well fwiw.  So we'll still need to look at string text.  So I'd be inclined to

- Merge this PR as is; the tests are useful
- But leave https://github.com/cursorless-dev/cursorless/issues/328 open

wdyt?

AndreasArvidsson:
Sounds good but I think we should reword the issue a bit

pokey:
done

Issue #855, opened 2022-07-11
Automatically try to remove scopes as if they're token, line, block, file if they have no removal behaviour


pokey:
Most of our tree-sitter scopes don't remove newlines, and some of them don't remove spaces. We'd like a way to make them all automatically remove the appropriate whitespace 

I would argue that we should detect if something stops at a line boundary and use newline delimiter if so.

# Implementation

## Where
Let's start by putting it on `SyntacticScopeTarget`

## What

If target has leading or trailing range, we use today's behaviour.

Otherwise, we see if the "core removal range" corresponds to the entire file, then falling back to block, then line, then token.  If it corresponds to one of these, we forward calls to `removalRange` to the upgraded target

The core removal range will be the value passed in to `removalRange` in constructor, otherwise content range.

# Other thoughts

Alternately we could add this implementation to `BaseTarget`, so that things like `SyntacticScopeTarget` will get it, but not get `LineTarget`, `TokenTarget`, `UntypedTarget` (?) will override it, but let's start with `SyntacticScopeTarget` for now

pokey:
I think we can also consider detecting we're a block and set it to `\n\n`, though not quite as critical as the newline one.  That one gets me because it feels like a regression in situations like this one

pokey:
This one is also super annoying for "section"

AndreasArvidsson:
The current solution for these is of course that they have the wrong delimiter. Section should definitely not insert or remove as a token.

pokey:
Yeah maybe for now let's just add them to that switch statement, but I do worry it's going to start feeling like a game of whack-a-mole

AndreasArvidsson:
Agreed. We should definitely have a proper look at an alternative solution. But just for now I think we should just add them to the delimiter list.

PR #1568, opened 2023-07-03
Updated markdown item ranges

### LINKED ISSUE #1107: Properly determine markdown item insertion delimiter
Should inspect the leading item marker to see if they're using `- `, `* `, etc to delimit their lists.  Don't forget the newline

These seem to behave a bit like indentation.  The item content should actually go between the leading delimiter and the newline, just like with indentation.  So I don't think we can just treat it like a simple insertion delimiter.  We probably should try to find a way to reuse the pieces of `editNew` that we use for line targets, rather than simple delimiter insertion

A few tests we'll want:

- [ ] `"bring item air before bat"`
- [ ] `"bring item air after bat"`
- [ ] `"drink item air"`

### LINKED ISSUE #1199: Leading indentation of next item included in `"item"` content range
In the following markdown:

```markdown
- aaa
  - bbb
```

The scope `"item air"` includes from `aaa` until the `-` on the next line. For example, saying `"round wrap item air"` results in the following:

```markdown
- (aaa
  )- bbb
```

### LINKED ISSUE #1200: `"bring item"` broken for indented lists
In the following markdown:

```markdown
- aaa
  - bbb
  - ccc
  - ddd
  - eee
```

Saying `"bring item each after bat"` results in the following:

```markdown
- aaa
  - bbb
  - ccc
    eee
  - ddd
  - eee
```

It seems like this bug is in part because the leading indentation of `ccc` is included in item `- bbb` (#1199), but in addition, it seems that the range is being extended to the end of the line.

AndreasArvidsson:
Fixes #1200
Fixes #1199
Fixes #1107

## Checklist

### DIFF
--- a/packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/markdown/chuckItem2.yml
+++ b/packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/markdown/chuckItem2.yml
@@ -1,0 +1,28 @@
+languageId: markdown
+command:
+  version: 5
+  spokenForm: chuck item
+  action: {name: remove}
+  targets:
+    - type: primitive
+      modifiers:
+        - type: containingScope
+          scopeType: {type: collectionItem}
+  usePrePhraseSnapshot: true
+initialState:
+  documentContents: |-
+    - aaa
+      - bbb
+      - ddd
+  selections:
+    - anchor: {line: 1, character: 5}
+      active: {line: 1, character: 5}
+  marks: {}
+finalState:
+  documentContents: |-
+    - aaa
+
+      - ddd
+  selections:
+    - anchor: {line: 1, character: 0}
+      active: {line: 1, character: 0}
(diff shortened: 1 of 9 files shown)


packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/markdown/chuckItem2.yml
```diff
@@ -0,0 +1,28 @@
...
+      - bbb
+      - ddd
+  selections:
+    - anchor: {line: 1, character: 5}
+      active: {line: 1, character: 5}
+  marks: {}
+finalState:
+  documentContents: |-
+    - aaa
+
```
pokey, line 24: so is the idea to handle this one as part of [#855](https://github.com/cursorless-dev/cursorless/issues/855)?

pokey (approved):
Looks good with minor comment / question

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/markdown/chuckItem2.yml
AndreasArvidsson: Maybe? For now this is just an a improvement of what we had, but not perfect.

pokey: Seems fine

PR #1569, opened 2023-07-03
Added Cursorless output channel

### LINKED ISSUE #1115: Use an output/log channel for Cursorless

AndreasArvidsson:
Adds a Cursorless output channel that is used for Cursorless debug outputs when in production mode. For development we still use the old debug console.

Fixes #1115

## Checklist


packages/common/src/ide/types/ide.types.ts
```diff
@@ -213,6 +213,12 @@ export interface IDE {
     editor: TextEditor,
     ranges: GeneralizedRange[],
   ): Promise<void>;
+
+  /**
+   * Write message to the Cursorless log
+   * @param message The message to log
+   */
+  log(message: string): Promise<void>;
```
pokey, line 221: I'm hesitant to add more stuff to the top-level ide object.  As discussed, we'd like to break it up into smaller pieces.  Not strictly opposed to having it here, as it's just one function, but I'm not in love with this monolithic object that everyone needs

Also, do we want to encode some notion of log levels?  That way the places that we were using `console.debug` before could do something like one of the following

- `output.debug("foo")`, or
- `debug(output, "foo")`, or
- `output.log("foo", Log.debug)`

packages/cursorless-engine/src/core/Debug.ts
```diff
@@ -29,16 +30,21 @@ export class Debug {
...
 
-  log(...args: any[]) {
+  log(message: string) {
     if (this.active) {
-      console.log(...args);
+      if (this.isProduction) {
+        ide().log(message);
+      } else {
+        console.log(message);
+      }
```
pokey, line 47: why are we doing this here rather than defining the `log` function in `NormalizedIde` to use `console.log`?  Seems kind of vscode-specific that we need to use `console.log` vs output channel depending on how we're running

packages/cursorless-vscode/src/ide/vscode/VscodeIDE.ts
```diff
@@ -52,6 +53,9 @@ export class VscodeIDE implements IDE {
     this.flashHandler = new VscodeFlashHandler(this, this.highlights);
     this.capabilities = new VscodeCapabilities();
     this.editorMap = new WeakMap<vscode.TextEditor, VscodeTextEditorImpl>();
+    if (this.runMode === "production") {
```
pokey, line 56: if we're only creating channel in production, why don't we just use `console.log` below when it's null, and avoid needing checks in `NormalizedIde` or `Debug`

packages/common/src/ide/types/ide.types.ts
AndreasArvidsson: I don't think there are different ways of logging things to the output channel, but you want this only to have some kind of log level I'm not against it. We already have a settings for the debug log.

packages/cursorless-engine/src/core/Debug.ts
AndreasArvidsson: I guess we could.

packages/cursorless-vscode/src/ide/vscode/VscodeIDE.ts
AndreasArvidsson: We could do that

packages/common/src/ide/types/ide.types.ts
pokey: I think log level would be nice to have, and I would argue it's harder to add later, because we would need to go through and figure out what level everything should be

packages/cursorless-vscode/src/ide/vscode/VscodeIDE.ts
pokey: I would argue for this

AndreasArvidsson:
@pokey Ok updated. 
Right now there is only one log function on the ide and that is because there are no other ways to log in vscode. If you go directly to this it will always be logged, but if you instead go too `debug.log` we do have our debug setting that toggles this on and off.
Is that fine or do you want me to do something more elaborate?

pokey:
let's discuss.  some links for our discussion:

- https://code.visualstudio.com/api/references/vscode-api#LogOutputChannel
- https://stackoverflow.com/a/5646876/

AndreasArvidsson:
Log output channel is not available for the version of vscode that Cursorless is using. Do you know a way to find which version specific feature requires?

pokey:
binary search? 😅

AndreasArvidsson:
Okay so so until we have made Cursorless required on that version we can't implement the log outputs.

pokey:
We can do the same thing we're doing for notebooks, where we check the version. If it's old, we fall back to using regular output channel

pokey:
I don't think so. We can just ignore the log level for lower versions 🤷‍♂️

AndreasArvidsson:
Closing for now

PR #1571, opened 2023-07-03
Migrated notebook cell modifier to scope handler

### LINKED ISSUE #1053: Migrate `notebookCell` to use a scope handler

AndreasArvidsson:
Fixes #1053

## Checklist


pokey (commented):
I agree we should try a simple migration here before doing something fancy, but this feels pretty hacky to me.  Let's discuss

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/NotebookCellScopeHandler.ts
```diff
@@ -0,0 +1,25 @@
+import type { Direction } from "@cursorless/common";
+import { NestedScopeHandler } from ".";
+import { NotebookCellTarget } from "../../targets";
+import { TargetScope } from "./scope.types";
+
+export default class NotebookCellScopeHandler extends NestedScopeHandler {
+  public readonly iterationScopeType = { type: "token" } as const;
```
pokey, line 7: we should find a way to make "every cell" throw an exception

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/NotebookCellScopeHandler.ts
```diff
@@ -0,0 +1,25 @@
+import type { Direction } from "@cursorless/common";
+import { NestedScopeHandler } from ".";
+import { NotebookCellTarget } from "../../targets";
+import { TargetScope } from "./scope.types";
+
+export default class NotebookCellScopeHandler extends NestedScopeHandler {
```
pokey, line 6: should not be a nested scope handler

AndreasArvidsson: fixed

pokey:
Ah right forgot about that. There's code in our VSCode impl that deals with some of that fwiw

AndreasArvidsson:
Yeah but in a scope handler things should be quite generic so this is quite a pickle

AndreasArvidsson:
Nope this one was a bit devious. Definitely not a good first issue! :D

AndreasArvidsson:
To be honest I don't really have the time to do a lot of notebook stuffs at the moment. I think we should either just merge the simpler version I had initially, we can still throw an error on the iteration scope. Or just close this for now.

pokey:
Feel free to assign it to me and I'll take a look when I get a minute

AndreasArvidsson:
I don't think that this is a priority at the moment

PR #1578, opened 2023-07-05
Added explicit mark

### LINKED ISSUE #1577: Create new explicit range mark
For the side car we would require a new mark that instead of using the hat map gives an explicit range and editor id.

Since Cursorless at the moment don't have an identifier for editors we would either make this field optional or by default throw an exception. The side car can then create its own ide that actually defines this.

AndreasArvidsson:
Fixes #1577

## Checklist


packages/common/src/types/command/PartialTargetDescriptor.types.ts
```diff
@@ -43,14 +43,31 @@ export interface RangeMark {
...
+  readonly character: number;
+}
+
+interface PartialRange {
+  readonly start: PartialPosition;
+  readonly end: PartialPosition;
+}
+
+export interface ExplicitMark {
+  type: "explicit";
```
pokey, line 57: I think I'd be a bit more explicit (😊) and call this `explicitRange`.  I'd also change

- `ExplicitMark` to `ExplicitRangeMark` and
- `ExplicitMarkStage` to `ExplicitRangeMarkStage`

WDYT?

packages/common/src/types/command/PartialTargetDescriptor.types.ts
```diff
@@ -43,14 +43,31 @@ export interface RangeMark {
   excludeActive?: boolean;
 }
 
+interface PartialPosition {
+  readonly line: number;
+  readonly character: number;
+}
+
+interface PartialRange {
```
pokey, line 51: I think I'd call this `SimpleRange` and the other one `SimplePosition`.  WDYT?

pokey (commented):
Looks fine, but I might rename a couple things; see inline comments

packages/common/src/types/command/PartialTargetDescriptor.types.ts
AndreasArvidsson: agreed

packages/common/src/types/command/PartialTargetDescriptor.types.ts
AndreasArvidsson: agreed

AndreasArvidsson:
Updated with suggestions

packages/common/src/types/command/PartialTargetDescriptor.types.ts
```diff
@@ -43,14 +43,31 @@ export interface RangeMark {
...
 export type PartialMark =
   | CursorMark
   | ThatMark
   | SourceMark
   | DecoratedSymbolMark
   | NothingMark
   | LineNumberMark
-  | RangeMark;
+  | RangeMark
+  | ExplicitRangeMark;
```
pokey, line 70: oh sorry i forgot we have a `RangeMark` already.  A bit confusing to have both, esp since we have something called `hasExplicitRange`, and this name has nothing to do with that 😅.  Maybe let's discuss.  Sorry for the needless work

AndreasArvidsson: Good catch. We can have a talk about it.

pokey: update: go back to `ExplicitMark` and `"explicit"`, but add jsdoc

PR #1579, opened 2023-07-05
Removed reveal line for insert copy action

### LINKED ISSUE #1437: Disorienting scrolling when changes are made outside the viewport
(continued from Slack)

The editor sometimes scrolls in odd ways when changes are made outside the viewport.

Repro cases:

1. open a file larger than the screen. "pre file", "bottom file". "bring air" (any token on screen). the token is brought to line 1. vscode does not scroll to line 1 to show the change you just made (I'm sure it used to!)

2. same setup, "pre file", "bottom file". "clone air" (any token). now vscode does scroll to line 1, away from the change you just made. I'm sure it didn't use to do this.

Building on case 2, ctrl-z sometimes also scrolls away from the changes it's undoing/redoing, making it quite hard to follow where I am when I'm undoing more than one change. There's probably a shorter repro but this works at least:

3. "pre file", "bottom file". "clone air". (editor scrolls to top) "post file". "crown file". "clone bat" (anything on screen). (editor scrolls to bottom). "crown file". "undo that". (editor scrolls away from the change it just undid). "undo that" (again, editor scrolls away from the change)

I think this started happening after a vscode update about a month ago, but I'm not 100% sure on that.

AndreasArvidsson:
Fixes #1437

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/actions/InsertCopy.ts
+++ b/packages/cursorless-engine/src/actions/InsertCopy.ts
@@ -94,6 +94,7 @@ class InsertCopy implements Action {
         (selection) => selection.intersection(primarySelection) != null,
       )
     ) {
+        // If the original target contained the user's cursor, reveal it in case it got pushed off screen
       await editableEditor.revealRange(primarySelection);
     }
 
--- a/packages/cursorless-engine/src/actions/InsertCopy.ts
+++ b/packages/cursorless-engine/src/actions/InsertCopy.ts
@@ -87,7 +87,6 @@ class InsertCopy implements Action {
     );
 
     setSelectionsWithoutFocusingEditor(editableEditor, updatedEditorSelections);
-    await editableEditor.revealRange(editor.selections[0]);
 
     return {
       sourceMark: createThatMark(targets, insertionRanges),
--- a/packages/cursorless-engine/src/actions/InsertCopy.ts
+++ b/packages/cursorless-engine/src/actions/InsertCopy.ts
@@ -87,6 +87,15 @@ class InsertCopy implements Action {
     );
 
     setSelectionsWithoutFocusingEditor(editableEditor, updatedEditorSelections);
+    const primarySelection = editor.selections[0];
+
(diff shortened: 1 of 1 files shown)


pokey (commented):
I'm a bit confused how this one would address all the issues in #1437, when this change appears to only impact "clone" and "clone up"

AndreasArvidsson:
> I'm a bit confused how this one would address all the issues in #1437, when this change appears to only impact "clone" and "clone up"

The behavior of clone and bring is now consistent. ie we do not scroll on either. The behavior of ctrl-z I don't think we can really influence that much.

pokey (approved):

pokey (changes requested):
whoops; broken in case where new target contains cursor; that should reveal selection

packages/cursorless-engine/src/actions/InsertCopy.ts
```diff
@@ -87,7 +87,15 @@ class InsertCopy implements Action {
...
 
     setSelectionsWithoutFocusingEditor(editableEditor, updatedEditorSelections);
-    await editableEditor.revealRange(editor.selections[0]);
+    const primarySelection = editor.selections[0];
+
+    if (
+      updatedContentSelections.some(
+        (selection) => selection.intersection(primarySelection) != null,
+      )
+    ) {
```
pokey, line 96:
```suggestion
    ) {
        // If the original target contained the user's cursor, reveal it in case it got pushed off screen
```

packages/cursorless-engine/src/actions/InsertCopy.ts
```diff
@@ -87,7 +87,16 @@ class InsertCopy implements Action {
...
+        // If the original target contained the user's cursor, reveal it in case it got pushed off screen
```
pokey, line 97:
```suggestion
      // If the original target contained the user's cursor, reveal it in case it got pushed off screen
```

pokey (approved):

PR #1580, opened 2023-07-05
Remove use of constructor.name

### LINKED ISSUE #1300: Stop using `constructor.name` in `targetToPlainObject`
During minification, function / class names are changed, so the test case recorder / test case runner break when we minify the Cursorless VSCode extension, resulting in failures like https://github.com/cursorless-dev/cursorless/actions/runs/4378570568/jobs/7663476484.  Prior to #1281, we ran our tests in CI on the un-minified code, so the tests passed, even though technically the test case recorder is broken in production.  This is fine in practice, because we only ever run the test case recorder in extension development mode.

But I would argue that it is risky not to run our tests on the final minimised version of the code, as it can lead to behaviour skew like the above.  

The temporary fix was to enable [`--keep-names`](https://esbuild.github.io/api/#keep-names), but this results takes our bundle size from 788.3kb to 834.3kb, so would be good to roll back.  An alternative temporary fix would have been to continue running tests on un-minified code, but this seemed more dangerous, and not worth the 5.5% space saving.

The proper fix would probably be to have targets know how to serialise themselves, leveraging a `type` field.  We have been hesitant to add a `type` field because we don't want to encourage code that uses a `switch` statement on `type`.  We could mark the field `protected`, and then have `BaseTarget` define a `toPlainObject` function that leverages the derived `type` field

- [x] In addition, we should add a lint rule to prevent usage of `constructor.name` in the future.  Something like https://stackoverflow.com/a/72903871 should do the trick

AndreasArvidsson:
Fixes #1300

## Checklist


packages/cursorless-engine/src/processTargets/targets/BaseTarget.ts
```diff
@@ -194,6 +195,16 @@ export default abstract class BaseTarget<
...
+  toPlainObject(): TargetPlainObject {
+    return {
+      type: this.type,
+      contentRange: rangeToPlainObject(this.contentRange),
+      isReversed: this.isReversed,
+      hasExplicitRange: this.hasExplicitRange,
+    };
+  }
+
+  protected abstract get type(): string;
```
pokey, line 207: OUt of curiosity, why define this one as a getter in the base class rather than just a `readonly` property?

pokey (commented):
Looks great with one minor question

packages/cursorless-engine/src/processTargets/targets/BaseTarget.ts
AndreasArvidsson: I didn't know that you could have an abstract read only property. I have updated now.

pokey (approved):

PR #1595, opened 2023-07-07
Implemented sentence scope

### LINKED ISSUE #19: Create "sentence" transformation
It would be considered a delimited sequence, so that it would remove trailing periods as necessary.

AndreasArvidsson:
Fixes #19

While I was there I also sorted the textual scopes.

## Checklist


packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler.ts
```diff
@@ -0,0 +1,49 @@
...
+      const start = offset + segment.index + leadingWhitespace.length;
+      return new Range(
+        editor.document.positionAt(start),
+        editor.document.positionAt(
+          start +
+            segment.segment.length -
+            leadingWhitespace.length -
+            trailingWhitespace.length,
+        ),
+      );
```
pokey, line 31: maybe we should just have a `trimRange` function?  I think that would make it clearer what's happening here.  Plus we could use it for a `#trim!` query predicate in the future 😊

AndreasArvidsson: I kinda had the same thought. I can do that.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler.ts
```diff
@@ -0,0 +1,49 @@
+import { Direction, Range } from "@cursorless/common";
+import { imap } from "itertools";
+import { NestedScopeHandler } from ".";
+import type { TargetScope } from "./scope.types";
+import { TokenTarget } from "../../targets";
+
+export default class SentenceScopeHandler extends NestedScopeHandler {
+  public readonly scopeType = { type: "sentence" } as const;
+  public readonly iterationScopeType = { type: "paragraph" } as const;
```
pokey, line 9: 👍

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler.ts
```diff
@@ -0,0 +1,49 @@
...
+      const contentRange = segmentToRange(segment);
+      return {
+        editor,
+        domain: contentRange,
+        getTargets: (isReversed) => [
+          new TokenTarget({
+            editor,
+            contentRange,
+            isReversed,
+          }),
```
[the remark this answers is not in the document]
AndreasArvidsson: I think so

pokey (commented):
looks good with minor comment.  also

- [ ] would be good to add a test with something slightly more challenging, eg `Hello, Mr. Anderson.`

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler.ts
AndreasArvidsson: By the way why do we need line eight? The constructor of nested scope handler makes that public already?
https://github.com/cursorless-dev/cursorless/blob/1046ddac884bd411770c78bc5b592a2da662ca2b/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/NestedScopeHandler.ts#L55

AndreasArvidsson:
> * Hello, Mr. Anderson.

That test actually fails with the current implementation. Hmm...

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler.ts
AndreasArvidsson: Got it. The risk here is just that we now define scope handlers in two different places that could in theory be different.

pokey:
Huh.  So I guess the sentence segmenter is fairly primitive then. That's a fairly basic counterexample to pure punctuation-based splitting

AndreasArvidsson:
> Huh. So I guess the sentence segmenter is fairly primitive then. That's a fairly basic counterexample to pure punctuation-based splitting

Agreed. I was just counting on `Intl.Segmenter` to be smarter

AndreasArvidsson:
> A couple other thoughts:
> 
>     * Do we want an insertion delimiter of `". "`?
> 
>     * Do we want to file a follow-up for the capitalization checkbox on [Add "sentence" scope type #19](https://github.com/cursorless-dev/cursorless/issues/19)?  That kinda requires [Destinations as first-class objects #803](https://github.com/cursorless-dev/cursorless/issues/803)

I don't think so. There's a lot of delimiters for a sentence. Some languages like spanish even have them at the beginning of the sentence. I would be conservative here and just add a whitespace.

I wasn't sure what you actually wanted here. When would the capitalization feature come in to play?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler.ts
AndreasArvidsson: done

pokey:
> I don't think so. There's a lot of delimiters for a sentence. Some languages like spanish even have them at the beginning of the sentence. I would be conservative here and just add a whitespace.
> 
> I wasn't sure what you actually wanted here. When would the capitalization feature come in to play?

I think both come into play when you say eg "bring air past bat after sentence".  In that case, you probably want "air" to get capitalized and you want a `. ` after it.  The former is extra credit.  The latter seems easy / useful.  But yeah I guess a decent amount of the time you wouldn't want a `.`

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler.ts
pokey: sorry what do you mean "we now define scope handlers in two different places".  Which places are those?

AndreasArvidsson: One scope gets passed into the constructor in `ScopeHandlerFactoryImpl.ts` and then another scope could be defined in the actual scope handler. I just thought this pattern was a bit peculiar.

AndreasArvidsson:
I would be perfectly fine with running with this and then filing a follow up for a smarter implementation. than this was a really quick implementation and I think it gives us a lot compared to nothing.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export default class SentenceScopeHandler extends NestedScopeHandler {
+  public readonly scopeType = { type: "sentence" } as const;
+  public readonly iterationScopeType = { type: "paragraph" } as const;
+
+  protected generateScopesInSearchScope(
+    direction: Direction,
+    { editor, domain }: TargetScope,
+  ): Iterable<TargetScope> {
+    const segmenter = new Intl.Segmenter("en", { granularity: "sentence" });
```
auscompgeek, line 16: Should we use the user locale instead of hardcoding `en`?

AndreasArvidsson: We could just pass undefined or `best fit`

AndreasArvidsson:
I just want to lift the concern that if we implement support for other languages when we later try to switch to a better library we kinda need to pick something that supports multiple languages. Most of the ones I found or very language specific.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler.ts
AndreasArvidsson: fixed

AndreasArvidsson:
We're now using the default locale

pokey:
> I just want to lift the concern that if we implement support for other languages when we later try to switch to a better library we kinda need to pick something that supports multiple languages. Most of the ones I found or very language specific.

I think that's fine. We can try to detect language and upgrade to better lib if it's English, falling back to intl segmenter if not

AndreasArvidsson:
@pokey @auscompgeek Okay I couldn't sleep and put some more thought into this. I looked into a few different libraries but I didn't think any of them was a good fit for Cursorless. For example we need to support sentences with new line breaks in them since we have a maximum line with in our jsdoc comments for example. Anyhow it I ended up implementing my own sentence segmenter complete with unit tests :D

pokey:
Ha ok looks like you had some fun 😄

I'm a bit confused why you didn't go for https://github.com/Tessmore/sbd tho?  Looks very lightweight, has an option to disable newline-based splitting, and in my testing handles most cases we care about, including the comment example if you strip unwanted punctuation in postprocessing.  You can play with it [here](http://tessmore.github.io/sbd/).  If you look at their code you can see they handle a bunch of cases that you miss; not sure we want to maintain our own sentence boundary segmentation library.  It's a deceptively hard problem

AndreasArvidsson:
I didn't think that sbd fulfilled our needs. 

1. If you disabled new line splitting this example will be interpreted as one sentence
```
Foo.
Bar.
```

2. It does not trim leading symbols `* Foo bar.` we'll get interpreted just like that with the asterisk at the beginning.

3. I wanted an implementation that used a generator.

pokey:
> 1. If you disable new line splitting this example will be interpreted as one sentence
> 
> ```
> 
> Foo.
> 
> Bar.
> 
> ```
> 
> This is also one sentence
> 
> ```
> 
> Foo
> 
> Bar.
> 
> 
> 
> Baz bongo.
> 
> ```

Neither of those are English.  The library is designed for English. Do you have similar examples that are real English sentences?  Also, you split on blocks, so none of those examples would actually make it to the segmenter without already being split

> 2. It does not trim leading symbols `* Foo bar.` we'll get interpreted just like that with the asterisk at the beginning.

You can easily reuse your postprocessing code to strip those

> 3. I wanted an implementation that used a generator. 

A list can easily be used as a generator. I don't think we should be sacrificing quality for efficiency until we've established we actually have a performance problem. You're only handling one block at a time anyway

Fundamentally, the problem is that your code is missing lots of cases, and I don't think we want to invest in adding them.  See [the implementation of sbd](https://github.com/Tessmore/sbd/blob/master/lib/Match.js) if you're not sure what I mean. 
Here are some test cases, courtesy of chatgpt.  Not that sbd gets a perfect score, but just to give you some examples of why it's such a hard problem:

```ts
[
  {
    input: "Visit www.example.com. Best site ever!",
    expectedOutput: ["Visit www.example.com.", "Best site ever!"],
  },
  {
    input: "I scored 3.5. What about you?",
    expectedOutput: ["I scored 3.5.", "What about you?"],
  },
  {
    input: "And so...! He exclaimed.",
    expectedOutput: ["And so...!", "He exclaimed."],
  },
  {
    input: "I am watching E.T. It's a good movie.",
    expectedOutput: ["I am watching E.T.", "It's a good movie."],
  },
  {
    input: "The temperature was -5.2. Quite cold.",
    expectedOutput: ["The temperature was -5.2.", "Quite cold."],
  },
  {
    input: "She shouted, 'No!' He ran away.",
    expectedOutput: ["She shouted, 'No!'", "He ran away."],
  },
  {
    input: "They live at No. 42. Best place!",
    expectedOutput: ["They live at No. 42.", "Best place!"],
  },
  {
    input: "Version 2.0.1. was released.",
    expectedOutput: ["Version 2.0.1. was released."],
  },
  {
    input: "This is www.example.com/site. Visit it.",
    expectedOutput: ["This is www.example.com/site.", "Visit it."],
  },
  {
    input: "He ran 10.5 miles. Very fast.",
    expectedOutput: ["He ran 10.5 miles.", "Very fast."],
  },
  {
    input: "She loves dogs!He loves cats.",
    expectedOutput: ["She loves dogs!", "He loves cats."],
  },
  {
    input: "But then... she arrived.",
    expectedOutput: ["But then... she arrived."],
  },
  {
    input: "John has a Ph.D. He is smart.",
    expectedOutput: ["John has a Ph.D.", "He is smart."],
  },
  {
    input: "It is great etc. Isn't it?",
    expectedOutput: ["It is great etc.", "Isn't it?"],
  },
  {
    input: "He said 'Bye!'She didn't reply.",
    expectedOutput: ["He said 'Bye!'", "She didn't reply."],
  },
  {
    input: "I was born in Mar. I love spring.",
    expectedOutput: ["I was born in Mar.", "I love spring."],
  },
  {
    input: "He exclaimed, 'What!?' She was confused.",
    expectedOutput: ["He exclaimed, 'What!?'", "She was confused."],
  },
  {
    input: "Finished at 5 p.m.We went home.",
    expectedOutput: ["Finished at 5 p.m.", "We went home."],
  },
  {
    input: "The location is (42.3601° N, 71.0589° W).",
    expectedOutput: ["The location is (42.3601° N, 71.0589° W)."],
  },
  {
    input: "The equation is 2x - 3y = 4.We solved it.",
    expectedOutput: ["The equation is 2x - 3y = 4.", "We solved it."],
  }
]
```

Happy to discuss at meetup if you'd like

AndreasArvidsson:
Let's discuss it

AndreasArvidsson:
I did some more testing and with the proper post processing sbd appears to be working fine. Still kind of annoying api where just get an array of strings instead of actual indices, but we can work with that.

pokey:
Agreed that's kind of an annoying api, but glad to hear it's working well. Should we still discuss or is this one ready for review?

AndreasArvidsson:
I think it's ready for review. We might take a quick discussion any how but that shouldn't stop you from reviewing it I think.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler/SentenceSegmenter.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  return {
+    text: sentence.trimEnd(),
+    index,
+  };
+}
+
+function isInvalidLine(text: string): boolean {
+  return invalidLineRegex.test(text);
```
pokey, line 59: I think this might hit the weird JS regex reuse issue; we should always use our own functions

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler/SentenceSegmenter.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+}
+
+function createSegment(
+  sentence: string,
+  index: number,
+): MatchedText | undefined {
+  const leadingOffset =
+    sentence.match(leadingOffsetRegex)?.index ?? sentence.length;
```
pokey, line 41: should prob use our own here, even tho the regex has no `g`, just for consistency

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler/SentenceSegmenter.ts
```diff
@@ -0,0 +1,60 @@
...
+const invalidLineRegex = /(\n[^\p{L}]*\n)/gu;
+
+const options: sbd.Options = {
+  ["newline_boundaries"]: false,
+  ["preserve_whitespace"]: true,
+};
+
+export default class SentenceSegmenter {
+  *segment(text: string): Iterable<MatchedText> {
+    const sentences = sbd.sentences(text, options);
```
pokey, line 16: Is there any way we can fall back to intl segmenter if the language isn't English?  Not sure we want to roll a language detector into our code as well 🤔

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler/SentenceSegmenter.ts
```diff
@@ -0,0 +1,60 @@
...
+    let index = 0;
+
+    for (const sentence of sentences) {
+      const parts = sentence.split(invalidLineRegex);
+
+      for (const part of parts) {
+        if (!isInvalidLine(part)) {
```
pokey, line 23: if we're using invalidLineRegex `invalidLineRegex` to split, then I don't think we'd get any invalid lines, right?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler/SentenceSegmenter.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+function createSegment(
+  sentence: string,
+  index: number,
+): MatchedText | undefined {
+  const leadingOffset =
+    sentence.match(leadingOffsetRegex)?.index ?? sentence.length;
+
+  if (sentence.length === leadingOffset) {
```
pokey, line 43: I'm nitpicking here, but I'd prob just drop the `?? sentence.length` above and check if `leadingOffset` is `null` here.  I'd usually just make this change myself but I don't want to lose my other comment on that line 😅

pokey (commented):
Looking good; left a few comments

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler/SentenceSegmenter.ts
AndreasArvidsson: I don't think we should do that. Many languages including Swedish has very similar structure to english. intl segmenter was truly atrocious in many aspects and would require special post processing before I think it would be useful.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler/SentenceSegmenter.ts
AndreasArvidsson: The regex has a capture group. If you split on a regex with the capture group that content will be in the final result. We need that since we need to calculate a correct index for the start of each sentence.

AndreasArvidsson:
updated with suggestions

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SentenceScopeHandler/SentenceSegmenter.ts
pokey: Ok let's leave it for now.  If people file issues for supporting other languages, we can look for better segmenters for those languages and deal with language detection then.

AndreasArvidsson: Sounds good

pokey:
Ran this with the scope visualizer.  For the most part it seems to work quite well.  There were some interesting rough edges, but good enough to ship I think.  Once scope visualizer is shipped out I'd recommend giving that a whirl on "sentence".  

## Good

image
image
image


## Odd

One thing we might consider doing in the future is restricting "sentence" to comments / strings when we're within a source file, as it is a bit confusing when applied to code that's not in a comment or string:

image

Also a bit strange to apply it to frontmatter, but I don't think it really matters:

image

## Bad

Just to be clear, none of these issues are blockers for me; I'm happy to merge as is, but thought it was interesting and maybe we address in a follow-up at some point if they bother us in practice:

The following really shouldn't be one sentence, but I'm not sure "sentence" is what we should be targeting them with; probably more like "item" or something

image

Note how it drops the leading quotes on the following items, when they really are part of the sentences.  The postprocessing might be a bit too aggressive, ie maybe we should include quotes in addition to letters.  

image

No idea what's going on with the break before `is false` here:

image

The following is a pretty tough case, but these really shouldn't be one sentence:

image

I do wonder if we want to have some kind of tree-sitter-aware splitting, so that eg markdown items are automatically broken up.  Prob also should break up markdown table cells if we do that:

image

pokey (approved):

PR #1602, opened 2023-07-09
Migrate a couple go scopes

pokey:
## Checklist


pokey:
hmm looks like I may need to migrate text fragment scope type for pairs.  Interesting if that breaks when you add one scope type for the language.  not sure that's a bad thing tbh, as we don't want people to forget to add those

queries/go.scm
```diff
@@ -0,0 +1,46 @@
...
+  (return_statement)
+  (select_statement)
+  (type_declaration)
+  (type_switch_statement)
+  (var_declaration)
+] @statement
+
+(composite_literal
+  type: [
+    (map_type)
```
josharian, line 30: On further reflection, there's more subtlety here. More details later, but we probably shouldn't merge this as-is.

pokey: That's fine.  Do you want to take this one over, or should we pair up again?  Fine either way by me

josharian: Probably easiest for me to take it over. I have a design question here, though.

The problem is that the type of a Go composite literal isn't a purely syntactic property. Examples:

```go
type T map[int]string
type U [2]string
type V struct{ a, b string }

var t T{1: "a"} // a map literal
var u U{1: "a"} // an array literal

var u2 U{"a", "b"}  // an array literal
var v V{"a", "b"} // a slice literal
```

There are heuristics we could use that would probably be 90-99% accurate.

* If there's an explicit syntactic type, use it. (That's what this PR does.)
* If there's an index present, call it a map. (Fails on `u` above.)
* If there's no index present, call it a list. (Fails on `v` above.)

Or we could give up and declare all composite literals to be both `map` and `list`, the current behavior.

The question is: Is it better to be (a) more precise, more complicated, and occasionally wrong, or (b) less precise, and simpler.

pokey: how about if it has `:`s in it, it's a map, otherwise, it's a list?

josharian: yep, that's the heuristic i alluded to above. just wanted to confirm that heuristic is better than simple. sounds like it is; i will start with that.

josharian:
In case I don’t make the meet-up tomorrow (kids still sick): I think I understand the path forward here, and how to do it. I just need to play with it a bit more, document it, and add tests. Thanks for the running start!

pokey:
For sure! One thing we didn't cover fwiw is how to migrate

```
foo: "bar[baz]"
```

In that case, you migrate to

```
(bar
   baz: (_) @foo
) @_.domain
```

This allows you to match the node at field `baz` from anywhere in the node with type `bar`, which is consistent with how our legacy pattern language handles field names. If you don't want this behaviour for the desired scope type, then you don't have to use the domain there, but often you will. Eg for "funk name", you'll likely want the domain to be the entire function so you can say "name" from anywhere in the function

josharian:
This has been superseded and can probably be closed.

PR #1604, opened 2023-07-10
Delete ts cache on clean

pokey:
This is necessary to reset the ts cache; otherwise it will still think that the generated files are there and won't build them after we clean

## Checklist


AndreasArvidsson (approved):

AndreasArvidsson:
So dumb that there isn't a compile flag that actually does this properly

Issue #1351, opened 2023-03-23
Don't show warning for `"past end of"`


pokey:
This behaviour could actually be useful in rare cases.

- [x] Remove warning
- [ ] Add note to old videos that use "past end of" to use "tail"

AndreasArvidsson:
@pokey I was just about to implement this one, but I can't do the second task so it's probably better if you take ownership? Or do you want me to go ahead and do the first task and then hand it over to you?


This is where in the code the warning is defined
https://github.com/cursorless-dev/cursorless/blob/0e4c9d903829b9852da4327a31065cf43ffc5410/packages/cursorless-engine/src/core/commandVersionUpgrades/canonicalizeAndValidateCommand.ts#L122-L151

pokey:
I don't feel strongly either way. I think either is fine

AndreasArvidsson:
@pokey The warning is now removed and I have assigned it to us you can update your videos.

PR #1605, opened 2023-07-10
Destinations as first-class objects

### LINKED ISSUE #803: Destinations as first-class objects

AndreasArvidsson:
- Fixes https://github.com/cursorless-dev/cursorless/issues/1630
- Partially #1351

## Checklist


packages/cursorless-engine/src/actions/actions.types.ts
```diff
@@ -65,4 +65,20 @@ export interface Action {
 /**
  * Keeps a map from action names to objects that implement the given action
  */
-export type ActionRecord = Record<ActionType, Action>;
+export interface ActionRecord extends Record<SimpleActionName, Action> {
+  replaceWithTarget: {
+    run(
+      sources: Target[],
+      destinations: Destination[],
+    ): Promise<ActionReturnValue>;
```
pokey, line 73: Not that we need to solve the world while we're here, but we could prob make these return types stronger now too. But happy to call that out of scope as well

packages/cursorless-engine/src/actions/BringMoveSwap.ts
```diff
@@ -270,80 +253,157 @@ class BringMoveSwap implements Action {
...
+    await this.decorateThatMark(markEntries);
+
+    return { thatSelections: markEntries };
+  }
+
+  private getEditsSwap(targets1: Target[], targets2: Target[]): ExtendedEdit[] {
+    const results: ExtendedEdit[] = [];
+
+    targets1.forEach((source, i) => {
+      const destination = targets2[i];
```
pokey, line 383: I was imagining you'd convert the target to a destination on the fly here when you use it as a destination

pokey (commented):
Coming along nicely! Left a couple super quick comments from a very cursory look

packages/cursorless-engine/src/actions/actions.types.ts
AndreasArvidsson: I agree. But I think we should do that in the follow up

packages/common/src/types/command/ActionDescriptor.ts
```diff
@@ -0,0 +1,107 @@
...
+  "generateSnippet",
+  "getText",
+  "highlight",
+  "indentLine",
+  "insertCopyAfter",
+  "insertCopyBefore",
+  "insertEmptyLineAfter",
+  "insertEmptyLineBefore",
+  "insertEmptyLinesAround",
+  "insertSnippet",
```
pokey, line 29: Ideally this would accept a destination, but maybe it's simpler to start with a target for now? Or maybe you just haven't implemented that part yet 😅

packages/cursorless-engine/src/actions/Call.ts
```diff
@@ -1,27 +1,27 @@
...
     this.run = this.run.bind(this);
   }
 
-  async run([sources, destinations]: [
-    Target[],
-    Target[],
-  ]): Promise<ActionReturnValue> {
+  async run(
+    sources: Target[],
+    destinations: Target[],
```
pokey, line 13: Hesitant to call something a destination if it's a target. Maybe we go semantically richer and say "callees" and "arguments"?

packages/cursorless-engine/src/actions/CallbackAction.ts
```diff
@@ -32,13 +32,13 @@ interface CallbackOptions {
...
  */
-export class CallbackAction implements Action {
+export class CallbackAction {
   constructor(private rangeUpdater: RangeUpdater) {
     this.run = this.run.bind(this);
   }
 
   async run(
-    [targets]: [Target[]],
+    targets: Target[],
```
pokey, line 41: Nice

packages/cursorless-engine/src/core/commandRunner/CommandRunnerImpl.ts
```diff
@@ -35,36 +41,15 @@ export class CommandRunnerImpl implements CommandRunner {
...
-
-    const prePositionStages =
-      action.getPrePositionStages?.(...actionArgs) ?? [];
-    const finalStages = action.getFinalStages?.(...actionArgs) ?? [];
-
-    const targets = targetDescriptors.map((targetDescriptor) =>
-      this.pipelineRunner.run(targetDescriptor, prePositionStages, finalStages),
-    );
-
+  async run({ action }: CommandComplete): Promise<unknown> {
```
pokey, line 44: Huh if this is all you need to run a command I wonder if we should rename it 🤔. Or inline it. Let's discuss

packages/cursorless-engine/src/core/commandRunner/CommandRunnerImpl.ts
```diff
@@ -78,6 +63,214 @@ export class CommandRunnerImpl implements CommandRunner {
...
+        return action.run(targets);
+      }
+    }
+  }
+
+  private getTargets(
+    inferenceContext: InferenceContext,
+    partialTargetsDescriptor: PartialTargetDescriptor,
+    actionPrePositionStages?: ModifierStage[],
+    actionFinalStages?: ModifierStage[],
```
pokey, line 215: I'm curious how this distinction works now that positions aren't modifiers, other than start of and end of

packages/cursorless-engine/src/core/commandRunner/CommandRunnerImpl.ts
```diff
@@ -78,6 +63,214 @@ export class CommandRunnerImpl implements CommandRunner {
...
+  ) {
+    const sourceTargetDescriptor = inferenceContext.run(
```
pokey, line 217: Nit: should be `targetDescriptor`

packages/cursorless-engine/src/actions/InsertSnippet.ts
```diff
@@ -121,12 +109,12 @@ export default class InsertSnippet implements Action {
   }
 
   async run(
-    [targets]: [Target[]],
+    targets: Target[],
```
pokey, line 112: I think this has to be a destination, because otherwise how would "snip funk after bat" work? But that should come out of the tests

pokey (commented):
Getting there! Left a few sporadic comments

packages/common/src/types/command/PartialTargetDescriptor.types.ts
```diff
@@ -371,3 +381,25 @@ export type PartialTargetDescriptor =
...
+export interface PartialListDestinationDescriptor {
+  type: "destinationList";
+  destinations: {
+    type: "destination";
+    insertionMode: InsertionMode;
+    target:
+      | PartialPrimitiveTargetDescriptor
+      | PartialRangeTargetDescriptor
+      | PartialListTargetDescriptor;
+  }[];
```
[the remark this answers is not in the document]
AndreasArvidsson: A list can't contain implicit targets. I try to make the types as specific as possible.

packages/cursorless-engine/src/actions/Call.ts
AndreasArvidsson: I had the same thought.

packages/cursorless-engine/src/core/commandRunner/CommandRunnerImpl.ts
AndreasArvidsson: I kinda like the separation between the running the action and the rest of the command.

packages/common/src/types/command/PartialTargetDescriptor.types.ts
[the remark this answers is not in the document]
pokey: Can a primitive destination contain an implicit target?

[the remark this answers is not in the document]
AndreasArvidsson: Yes. `bring air` for example would create an primitive destination containing an implicit target

packages/common/src/types/command/ActionDescriptor.ts
```diff
@@ -0,0 +1,194 @@
...
+  name: "executeCommand";
+  commandId: string;
+  options?: ExecuteCommandOptions;
+  target: PartialTargetDescriptor;
+}
+
+export type ReplaceWith = string[] | { start: number };
+
+export interface PartialReplaceActionDescriptor {
+  name: "replace";
```
pokey, line 171: Holy crap we still support this? 😂

I've actually been thinking to use this for case-aware insertion when you use case-insensitive "instance" so I guess I'm glad it's stayed with us all this time. Tests ftw

packages/cursorless-engine/src/actions/BringMoveSwap.ts
```diff
@@ -133,7 +116,9 @@ class BringMoveSwap implements Action {
         usedSources.push(source);
         if (this.type !== "move") {
           results.push({
-            edit: source.constructChangeEdit(destination.contentText),
+            edit: source
+              .toDestination("to")
+              .constructChangeEdit(destination.target.contentText),
             editor: source.editor,
             originalTarget: source,
             isSource: true,
```
pokey, line 124: pretty sure this case is only for swap, which is now handled separately, so you should be able to remove this `if` statement altogether.  Then you can also remove the filtering here https://github.com/cursorless-dev/cursorless/blob/6c95c6d7da735a3f12e0976a17326e86a44109fa/packages/cursorless-engine/src/actions/BringMoveSwap.ts#L154-L157

pokey (commented):
couple more quick comments

packages/cursorless-engine/src/actions/BringMoveSwap.ts
pokey, line 1: Good call to separate these implementations; I think the unification actually made things harder to read.  I think we can go even further, but let's leave that for a follow-up PR

packages/cursorless-engine/src/actions/BringMoveSwap.ts
pokey: Oh actually this will break our handling of source marks right?  That all feels pretty hacky now.  I'm ok with leaving this for now and I can do a quick follow-up in #1608 to clean things up

packages/cursorless-engine/src/core/commandRunner/CommandRunnerImpl.ts
AndreasArvidsson: Yeah we don't need preposition stages anymore

packages/common/src/types/command/ActionDescriptor.ts
AndreasArvidsson: We do! :D

packages/cursorless-engine/src/actions/BringMoveSwap.ts
AndreasArvidsson: Correct

packages/cursorless-engine/src/core/commandVersionUpgrades/canonicalizeAndValidateCommand.ts
```diff
@@ -32,26 +33,21 @@ export function canonicalizeAndValidateCommand(
...
   return {
     version: LATEST_VERSION,
     spokenForm,
     action: {
+      ...action,
       name: actionName,
-      args: action.args ?? [],
-    },
-    targets: partialTargets,
+    } as PartialActionDescriptor,
```
AndreasArvidsson, line 50: Not a fan of having to cast like this, but you to the typed nature of our actions it will be hard to get around.

packages/common/src/types/command/command.types.ts
```diff
@@ -1,14 +1,19 @@
...
 import type { CommandV4 } from "./legacy/CommandV4.types";
+import type { CommandV5 } from "./legacy/CommandV5.types";
 
-export type CommandComplete = Required<Omit<CommandLatest, "spokenForm">> &
-  Pick<CommandLatest, "spokenForm"> & { action: Required<ActionCommand> };
+// TODO:
+// export type CommandComplete = Required<Omit<CommandLatest, "spokenForm">> &
+//   Pick<CommandLatest, "spokenForm"> & {
+//     action: Required<PartialActionDescriptor>;
+//   };
```
AndreasArvidsson, line 12: We can't make all fields in the action descriptor required since we have optional parameters where the default value is set in the action.

packages/common/src/types/command/ActionDescriptor.ts
pokey, line 1: Should we get rid of all the `Partial`s in this file?  We don't have corresponding full types, and these names are quite a mouthful 😅

packages/cursorless-engine/src/core/commandRunner/CommandRunnerImpl.ts
pokey: I guess I'm not opposed to that, tho it breaks our nomenclature a bit, because "action" usually refers to the action without its targets, whereas "command" is the full command including targets.  Eg in the cheatsheet we have a table of "actions".  Let's discuss at meet-up.  I think just a better name here is fine, because inlining them is slightly more awkward because then all the different versions need to extend a base

packages/common/src/types/command/PartialTargetDescriptor.types.ts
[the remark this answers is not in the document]
pokey: Oh weird yeah I guess you're right.  Hadn't thought about that.  I'm tempted to have something called `ImplicitDestination`.  Seems semantically richer than a primitive destination wrapping an implicit target

packages/common/src/types/command/ActionDescriptor.ts
AndreasArvidsson: Okay with me

packages/common/src/types/command/PartialTargetDescriptor.types.ts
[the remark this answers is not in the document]
AndreasArvidsson: Sounds good

packages/cursorless-engine/src/actions/actions.types.ts
```diff
@@ -47,22 +54,93 @@ export interface ActionReturnValue {
 }
 
 export interface Action {
```
pokey, line 56: Should this be called `SimpleAction`?

packages/cursorless-engine/src/actions/ToggleBreakpoint.ts
```diff
@@ -15,7 +15,7 @@ export default class ToggleBreakpoint implements Action {
     this.run = this.run.bind(this);
   }
 
-  async run([targets]: [Target[], Target[]]): Promise<ActionReturnValue> {
+  async run(targets: Target[]): Promise<ActionReturnValue> {
```
pokey, line 18: why did the previous type def have two arg lists?

packages/common/src/types/command/PartialTargetDescriptor.types.ts
```diff
@@ -371,3 +381,25 @@ export type PartialTargetDescriptor =
   | PartialRangeTargetDescriptor
   | PartialListTargetDescriptor
   | ImplicitTargetDescriptor;
+
+export interface PartialPrimitiveDestinationDescriptor {
+  type: "primitiveDestination";
```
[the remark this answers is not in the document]
AndreasArvidsson: I guess we could do that. I was just thinking that when you have a serialized log it might be a bit more obvious if we have different names between destinations and targets

packages/cursorless-engine/src/processTargets/modifiers/PositionStage.ts
```diff
@@ -1,11 +1,46 @@
...
+    const parameters: CommonTargetParameters = {
+      editor: target.editor,
+      isReversed: target.isReversed,
+      contentRange: this.constructContentRange(target.contentRange),
+    };
+
+    return [
+      target.isRaw
+        ? new RawSelectionTarget(parameters)
+        : new PlainTarget({ ...parameters, isToken: false }),
```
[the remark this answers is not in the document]
AndreasArvidsson: It is so that we don't lose raw/just. `end of just air` vs `end of air`.
This test for example fails if I remove that logic:
https://github.com/cursorless-dev/cursorless/blob/9c3366a2e4f71d7cb96e69fb3a342dd18618e185/packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/positions/bringArgMadeAndGustToEndOfJustThis.yml

packages/cursorless-engine/src/typings/target.types.ts
```diff
@@ -164,3 +161,17 @@ export interface Target {
    */
   toPlainObject(): TargetPlainObject;
 }
+
+export interface Destination {
+  readonly insertionMode: InsertionMode;
+  readonly editor: TextEditor;
```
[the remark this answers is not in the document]
AndreasArvidsson: Yeah those are just for convenience

packages/cursorless-engine/src/actions/actions.types.ts
AndreasArvidsson: Agreed

packages/cursorless-engine/src/typings/target.types.ts
```diff
@@ -164,3 +161,17 @@ export interface Target {
...
+  readonly target: Target;
+  readonly contentRange: Range;
+  readonly contentSelection: Selection;
+  readonly isRaw: boolean;
+  readonly insertionDelimiter: string;
```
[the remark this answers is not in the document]
pokey: out of scope.  let's discuss in a follow-up issue

packages/common/src/types/command/PartialTargetDescriptor.types.ts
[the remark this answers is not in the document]
pokey: update: let's do it

packages/cursorless-engine/src/processTargets/modifiers/PositionStage.ts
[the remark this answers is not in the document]
pokey: maybe try `new RawSelectionTarget(parameters)` always and see if tests pass

packages/cursorless-engine/src/core/commandVersionUpgrades/canonicalizeAndValidateCommand.ts
pokey: I'll take a crack at it

packages/cursorless-engine/src/processTargets/modifiers/PositionStage.ts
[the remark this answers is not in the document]
AndreasArvidsson: worked

packages/common/src/types/command/ActionDescriptor.ts
pokey: This file still has a ton of  `Partial`s in it (eg `PartialPasteActionDescriptor`, `PartialCallActionDescriptor`, etc

AndreasArvidsson: Ah sorry missed that.

AndreasArvidsson: Updated now

pokey: Cool.  You missed one; I took care of it (0f5818c746c64c12d7fcefee2f6b98855bac93a6)

AndreasArvidsson: Thanks. Long day... :)

packages/cursorless-engine/src/core/commandVersionUpgrades/upgradeV5ToV6/upgradeV5ToV6.ts
```diff
@@ -0,0 +1,359 @@
...
+}
+
+/**
+ * Converts a list target to a destination. This is a bit tricky because we need
+ * to split the list into multiple destinations if there is more than one insertion
+ * mode.
+ * @param target The target to convert
+ * @returns The converted destination
+ */
+function listTargetToDestination(
```
pokey, line 222: Your implementation was creating one destination per element of a list target.  E.g. it was mapping `"to end of air and end of bat"` to `"to end of air and to end of bat"` (note the second `"to"`).  I changed it so it only creates a list of there are multiple insertion modes, which is what the Talon side will do in practice.  For the most part, there is no difference, except for one small fact: if the targets refer to the same range, they get deduplicated in my implementation, because they will go through the target pipeline together, whereas in your version they would not.  This actually affected a couple unit tests:

- It turns out I had to roll back 5a95c6517d27e8b590285986c7c9cea3f4e31c8f because that did actually have an effect, it's just that without deduplication the insertion delimiter joining wasn't actually being exercised for a couple tests because they weren't getting deduplicated into a single target
- I was able to revert your change to `bringPointAndHarpToEndOfThisAndEndOfWhaleTakeWhale`.  The lack of deduplication was what was causing your test case failure there.

I wonder if we should have a final deduplication step for destinations where we deduplicate if they are the same target and insertion mode 🤔.  I would be tempted to handle that in follow-up work, but wouldn't be too bad to tackle in this PR; wdyt?

packages/cursorless-engine/src/processTargets/TargetPipelineRunner.ts
```diff
@@ -347,17 +347,14 @@ function targetsToVerticalTarget(
...
-      results.push(
-        new PlainTarget({
-          editor: anchorTarget.editor,
-          isReversed: anchorTarget.isReversed,
-          contentRange,
-        }),
-      );
-    }
+    results.push(
+      new VerticalRangeTarget({
```
pokey, line 351: How come you need a new target type here?

AndreasArvidsson: It behaves like a plan target with no special removal range with the exception of the insertion delimiter being given as an argument. We had no existing target like that.

AndreasArvidsson: Just didn't want to make the definition of plain target looser. If you prefer that go ahead.

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/updateSelections/bringFineAfterThis.yml
```diff
@@ -19,7 +19,7 @@ initialState:
       start: {line: 0, character: 0}
       end: {line: 0, character: 3}
 finalState:
-  documentContents: foo  fooworld
+  documentContents: foo world foo
```
pokey, line 22: Oh wow how did we miss this? Definitely seems like an improvement 👍

pokey: ...but no longer really testing what we wanted it to.  I'll move this test somewhere else to indicate it is about "this" expansion and re-record these tests with "just"

pokey: One thing that is a bit confusing now is that "bring air to end of this" won't expand to token, but "bring air after this" will 🤔.  I'm happy to call it out of scope for this PR, but maybe I should file a follow up to discuss?

PR #1613, opened 2023-07-11
Require at least two targets with sort action

### LINKED ISSUE #1610: warn when sort/shuffle/reverse provided only a single target
Easy mistake: `sort three lines air` instead of `sort every line three lines air`. It then appears not to work. For these actions that are fundamentally multi-target, perhaps we should warn if they are provided with only a single target. Ideally the warning should be more than just "needs multiple targets", but also provide a hint or two about how to accomplish that.

[Slack thread](https://talonvoice.slack.com/archives/C026KPTJE6T/p1689019734860659)

AndreasArvidsson:
Fixes #1610

## Checklist


packages/cursorless-engine/src/actions/Sort.ts
```diff
@@ -11,6 +11,10 @@ abstract class SortBase implements Action {
   protected abstract sortTexts(texts: string[]): string[];
 
   async run([targets]: [Target[]]): Promise<ActionReturnValue> {
+    if (targets.length < 2) {
+      throw new Error("Require more than one targets with this action");
```
pokey, line 15:
```suggestion
      throw new Error('This action requires than one target. For example, instead of "sort block", you might say "sort every line block" to sort the lines in a block.  Or you might say "reverse every key" to reverse the order of the keys in a map.');
```

@josharian is that the kind of error message you had in mind?

AndreasArvidsson: I don't think we should put verbose error messages on just this action. A lot of our different actions already have error messages for to few targets and they are as terse as the above one. I would argue for uniformity.

pokey: Maybe link to page in docs that elaborates then? Just because we have terse messages elsewhere doesn't make it a good idea 😄

AndreasArvidsson: I'm not against verbose error messages, but I am very much against treating some features like they are more important than others. I most definitely think that we should have a uniform error message system. We probably shouldn't use just the base error either and have more custom and verbose messages overall. Linking to the docs is not bad idea.

pokey: I also wonder if we want this to be a warning rather than an error.  I could see for example something like "scout air" then "sort every key scout next", and then popping to repeat a few times, to sort the keys in a bunch of maps which have a particular token in them.  I know it's a bit far fetched, but in that case, I might end up sorting a map with one element along the way, and I don't want the command chain to break.  Technically it is completely valid to sort a single target, it just so happens to be a no-op.  I'm happy to ignore the warning in that case.  Tbh it would be nice to be able to just check a box to suppress it in the future, but that's extra credit.

AndreasArvidsson: I agree with that.

josharian: My 2c, generally:

Warning > error

The best messages are both concise and helpful. But if I had to pick one, I'd choose helpful.

Maybe something like:

> ACTION works on multiple targets, e.g. "sort every line block" instead of "sort block".

Where ACTION is the actual action used, if we have ready access to it.

AndreasArvidsson: I think we need to discuss this

AndreasArvidsson: done

josharian (approved):

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/shuffleThis.yml
```diff
@@ -12,16 +12,5 @@ initialState:
...
-      active: {line: 0, character: 5}
-  thatMark:
-    - type: UntypedTarget
-      contentRange:
-        start: {line: 0, character: 0}
-        end: {line: 0, character: 5}
-      isReversed: false
-      hasExplicitRange: true
 fullTargets: [{type: primitive, mark: {type: cursor}, selectionType: token, position: contents, insideOutsideType: inside, modifier: {type: identity}, isImplicit: false}]
+thrownError: {name: Error}
```
pokey, line 16: I was leaning towards a warning for reasons mentioned in other comment. Do you prefer error?

AndreasArvidsson: updated

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/shuffleThis.yml
```diff
@@ -12,16 +12,12 @@ initialState:
...
   selections:
     - anchor: {line: 0, character: 0}
       active: {line: 0, character: 5}
-  thatMark:
-    - type: UntypedTarget
-      contentRange:
-        start: {line: 0, character: 0}
-        end: {line: 0, character: 5}
-      isReversed: false
-      hasExplicitRange: true
```
pokey, line 26: Not that we care for this test, but I wonder why it dropped the `that` mark. Did you just run update fixtures?

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/shuffleThis.yml
```diff
@@ -12,16 +12,12 @@ initialState:
     - anchor: {line: 0, character: 0}
       active: {line: 0, character: 5}
   marks: {}
+fullTargets: [{type: primitive, mark: {type: cursor}, selectionType: token, position: contents, insideOutsideType: inside, modifier: {type: identity}, isImplicit: false}]
```
pokey, line 15: Huh strange our update fixtures moves this up here. I think it used to put it at the end, but I believe now we totally ignore it in code so it just kind of ends up where it ends up? Maybe we should just remove it here. I think recorder doesn't capture it anymore

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/shuffleThis.yml
```diff
@@ -12,16 +12,12 @@ initialState:
...
-    - type: UntypedTarget
-      contentRange:
-        start: {line: 0, character: 0}
-        end: {line: 0, character: 5}
-      isReversed: false
-      hasExplicitRange: true
-fullTargets: [{type: primitive, mark: {type: cursor}, selectionType: token, position: contents, insideOutsideType: inside, modifier: {type: identity}, isImplicit: false}]
+ide:
+  messages:
+    - {type: warning, id: toFewTargets}
```
pokey, line 23:
```suggestion
    - {type: warning, id: tooFewTargets}
```

packages/cursorless-engine/src/actions/Sort.ts
```diff
@@ -11,6 +13,14 @@ abstract class SortBase implements SimpleAction {
   protected abstract sortTexts(texts: string[]): string[];
 
   async run(targets: Target[]): Promise<ActionReturnValue> {
+    if (targets.length < 2) {
+      showWarning(
+        ide().messages,
+        "toFewTargets",
```
pokey, line 19:
```suggestion
        "tooFewTargets",
```

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/shuffleThis.yml
AndreasArvidsson: Correct

AndreasArvidsson: But in two stages. Didnt revert in between.

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/shuffleThis.yml
AndreasArvidsson: Ok yeah then we should have it bee removed. Can do that in a separate pr.

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/shuffleThis.yml
AndreasArvidsson: Reverting back to the original and running update fixtures once kept the that mark.

pokey (approved):
Looks good!

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/shuffleThis.yml
pokey: Huh interesting. Yeah I guess that makes sense. Once it became an error the fact that it should preserve that mark was lost. Not sure how to solve that one but prob not a big deal

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/shuffleThis.yml
pokey: Yeah we should prob just remove them to reduce noise. Pretty sure we removed all references to it in code

PR #1623, opened 2023-07-13
implement "visible" scope

### LINKED ISSUE #1607: Add "visible" scope
Defined to be the visible content range in the editor. (Open question: Include partially visible lines?)

The main use case is with `from`. That would help in the case in which there are matching identifiers far away in the file that I don't care about. You can approximate `visible` using other scopes, but `visible` is a very simple scope to say and to reason about.

Sample usage: `from visible take every instance air`

Filed for discussion. If there's interest, I might attempt to implement this myself.

josharian:
Some progress! But definitely not working yet.

Very much :confused_dog: here. Would love some help,
ideally live at the next community meeting that I can attend.
No need to look at this in advance.

Things to discuss (notes to self):

* seems like containment = null is correct for "visible" 
  scope, and it seems like that's what the code says, 
  but...vscode begs to differ. halp?
  - we yield multiple ranges but only one gets used
  - if primary cursor is offscreen, we get an error 
    indicating that the range is not found
* definition of visible: i propose we leave it up to 
  VSCode, which mean does not include folded or partially 
  visible lines. the value of visible is that it does 
  exactly what it says on the tin.
* testing strategy: use lots of newlines?
* did I find all the places that need updating?

Fixes #1607

### DIFF
--- a/cursorless-talon/src/modifiers/scopes.py
+++ b/cursorless-talon/src/modifiers/scopes.py
@@ -78,6 +78,7 @@ scope_types = {
     "short paint": "boundedNonWhitespaceSequence",
     "link": "url",
     "cell": "notebookCell",
+    "visible": "visible",
 }
 
 
--- a/docs/user/experimental/keyboard/modal.md
+++ b/docs/user/experimental/keyboard/modal.md
@@ -59,6 +59,7 @@ To bind keys that do not have modifiers (eg just pressing `a`), add entries like
     "ss": "boundedNonWhitespaceSequence",
     "sa": "argumentOrParameter",
     "sl": "url",
+    "vz": "visible",
   },
   "cursorless.experimental.keyboard.modal.keybindings.actions": {
     "t": "setSelection",
--- a/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/VisibleScopeHandler.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/VisibleScopeHandler.ts
@@ -1,0 +1,35 @@
+import { TextEditor, Position } from "@cursorless/common";
+import { Direction, ScopeType } from "@cursorless/common";
+import { PlainTarget } from "../../targets";
+import BaseScopeHandler from "./BaseScopeHandler";
+import { TargetScope } from "./scope.types";
+
+export default class VisibleScopeHandler extends BaseScopeHandler {
+  public readonly scopeType = { type: "visible" } as const;
+  public readonly iterationScopeType = { type: "document" } as const;
+  protected readonly isHierarchical = false;
+
+  constructor(_scopeType: ScopeType, _languageId: string) {
+    super();
+  }
+
+  protected *generateScopeCandidates(
+    editor: TextEditor,
+    _position: Position,
+    _direction: Direction,
+  ): Iterable<TargetScope> {
+    yield {
+      editor,
+      domain: editor.document.range,
+      getTargets: (isReversed) =>
+        editor.visibleRanges.map(
+          (range) =>
+            new PlainTarget({
+              editor,
+              isReversed,
+              contentRange: range,
+            }),
+        ),
+    };
+  }
+}
--- a/docs/user/README.md
+++ b/docs/user/README.md
@@ -337,6 +337,7 @@ We have experimental support for prefixing a command with `"from <target>"` to n
 
 - `"from funk take every instance air"`: selects all instances of the token with a hat over the letter `a` in the current function
 - `"from air take next instance bat"`: selects the next instance of the token with a hat over the letter `b` starting from the token with a hat over the letter `a`
+- `"from visible take every instance drum"`: selects all instances of the token with a hat over the letter `d` that are currently visible in the editor
 
 Note that the `"from"` modifier is not enabled by default; you must remove the `-` at the start of the line starting with `-from` in your `experimental/experimental_actions.csv` [settings csv](./customization.md). Note also that this feature is considered experimental and may change in the future.
 
--- a/packages/cheatsheet/src/lib/sampleSpokenFormInfos/defaults.json
+++ b/packages/cheatsheet/src/lib/sampleSpokenFormInfos/defaults.json
@@ -1465,6 +1465,16 @@
           ]
         },
         {
+          "id": "visible",
+          "type": "scopeType",
+          "variations": [
+            {
+              "spokenForm": "visible",
+              "description": "Visible content in editor"
+            }
+          ]
+        },
+        {
           "id": "word",
           "type": "scopeType",
           "variations": [
--- a/packages/common/src/types/command/PartialTargetDescriptor.types.ts
+++ b/packages/common/src/types/command/PartialTargetDescriptor.types.ts
@@ -131,7 +131,8 @@ export type SimpleScopeTypeType =
   | "nonWhitespaceSequence"
   | "boundedNonWhitespaceSequence"
   | "url"
-  | "notebookCell";
+  | "notebookCell"
+  | "visible";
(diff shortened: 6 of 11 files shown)


packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/VisibleScopeHandler.ts
```diff
@@ -0,0 +1,35 @@
...
+  constructor(_scopeType: ScopeType, _languageId: string) {
+    super();
+  }
+
+  protected *generateScopeCandidates(
+    editor: TextEditor,
+    _position: Position,
+    _direction: Direction,
+  ): Iterable<TargetScope> {
+    for (const range of editor.visibleRanges) {
```
pokey, line 21: This will interact with folding in an interesting way. If you say "take visible" from one side of a fold, you'll only get up until the fold. If you say "take next visible" you'll get the code between the next fold until the following. To refer to all sections between folds, you'd say "take every visible".

I'd be tempted to either

- yield a single scope whose range is the union all the visible ranges, or
- yield a single scope that returns all the visible regions as a single list in its `getTargets`. I'd consider making the domain of that scope just be the document.

pokey (commented):
Wow great start! Tbh it's almost mergeable as is. But see comment below

pokey:
> * seems like containment = null is correct for "visible" 
>   scope, and it seems like that's what the code says, 
>   but...vscode begs to differ. halp?

What do you mean by `containment = null`?

>   - we yield multiple ranges but only one gets used
>   - if primary cursor is offscreen, we get an error 
>     indicating that the range is not found
> * definition of visible: i propose we leave it up to 
>   VSCode, which mean does not include folded or partially 
>   visible lines. the value of visible is that it does 
>   exactly what it says on the tin.

This would be my second proposal in my inline review comment

> * testing strategy: use lots of newlines?

That's a tough one. We struggle to test visible range stuff because we don't know how big the window is. Probably the best way to test this would be to leverage the fact that the editor is actually an abstract interface, but we haven't done anything like that yet. We will in the future, but I'd be tempted to punt on that. I'd probably just have a test with and without folds and declare victory

> * did I find all the places that need updating?

I think so. If the test passes you're probably good. In fact you were more thorough than we usually are 😅

pokey:
Come to think of it, this probably shouldn't be a scope type at all. I think it probably makes more sense as a mark

@AndreasArvidsson wdyt?

josharian:
> What do you mean by containment = null?

You can ignore. I mistakenly thought that the way to return a disjoint/union range was by yielding each of its pieces. Then I noticed that it behaved wrong, tried to figure out why, in my flailing found docs indicating that the wrong `containment` scope iterator setting would potentially explain the behavior I saw, and assumed that that was wrong.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/VisibleScopeHandler.ts
josharian: There's a third option, which is to reject `visible` as an allowed scope when folds are present. That sidesteps the question of what the correct behavior is, and punts the decision to the future, where we could loosen that constraint.

I lean towards your second option, and will implement that to play with, but am open to guidance. :)

pokey:
> You can ignore. I mistakenly thought that the way to return a disjoint/union range was by yielding each of its pieces. Then I noticed that it behaved wrong, tried to figure out why, in my flailing found docs indicating that the wrong `containment` scope iterator setting would potentially explain the behavior I saw, and assumed that that was wrong.

Sorry, I didn't understand anything you said after "You can ignore". But I'm happy to leave it there 😄

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/VisibleScopeHandler.ts
josharian: Implemented, played with second option. I'm happy with it. Another use case: fixing variable shadowing. Fold all shadowed inner scopes, work on every instance in what remains visible. (You can do that with take and give and this and that, but folding is more stateful, which can be useful.)

josharian: Although that makes me wonder whether visible should be a modifier.

`from visible file take every instance air`
`from visible func fine take every instance air`

Nah, probably not.

josharian:
Thanks for all the help. PTAL. Note the new commit message.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/VisibleScopeHandler.ts
pokey: If you revert to your original approach, then you could say "from every visible funk" to target all visible ranges that touch the current function. Not necessarily suggesting you do that, but just fwiw

josharian: That's true! But on balance, I think I prefer the simpler semantics. I'm happy with what is implemented now.

pokey: Brain dump:

- This is the first example of a scope that's not purely a function of the document itself
- Being a scope type here doesn't buy us anything useful over just making it a modifier (eg we can't leverage "every visible", "next visible", etc even tho the grammar now allows them). However that's also true or document scope ("file")
- Having it be a mark would preclude ability to say "visible air" to target visible in another split, so we'd prob want it to at least be a modifier
- Having it be modifier with intersection as you proposed is interesting but might be unintuitive when you have a nonempty selection and say "take visible": that would then intersect your selection with visible
- it might actually be useful to say "next visible" / "first visible" etc. I think the latter does work with your impl fwiw
- There is an interesting parallel here to a hypothetical scope for referring to your selections, eg "first selection", or "selection air" to refer to the selection in the split containing air

docs/user/experimental/keyboard/modal.md
```diff
@@ -59,6 +59,7 @@ To bind keys that do not have modifiers (eg just pressing `a`), add entries like
     "ss": "boundedNonWhitespaceSequence",
     "sa": "argumentOrParameter",
     "sl": "url",
+    "vz": "visible",
```
auscompgeek, line 62: The `v` key is already used for the `curve` shape below, so this will need to use a different key sequence.

cursorless-talon/src/modifiers/scopes.py
```diff
@@ -78,6 +78,7 @@
     "short paint": "boundedNonWhitespaceSequence",
     "link": "url",
     "cell": "notebookCell",
+    "visible": "visible",
```
auscompgeek, line 81: `visible` is being added next to `document` in other places. The same should be done here for consistency.

pokey:
If I recall correctly, the decision was to implement this one as a modifier for now? Or just close until we understand it better? I'm fine either way

josharian:
yep, we're going to try it as a modifier. Still in my queue!

AndreasArvidsson:
@josharian You want me to take this one? Implementing this as a modifier should be easy.

josharian:
@AndreasArvidsson great! That’d be lovely.

AndreasArvidsson:
@josharian Your wish is my command ;)
https://github.com/cursorless-dev/cursorless/pull/2094

josharian:
Breaking news, ChatGPT has no sense of humor.
IMG_9876

josharian:
Closing this. Thanks again, Andreas.

AndreasArvidsson:
You're very welcome :)

PR #1624, opened 2023-07-13
minor docs improvements

josharian:
- expand contributing docs slightly
- fix typo in comment

### DIFF
--- a/docs/contributing/CONTRIBUTING.md
+++ b/docs/contributing/CONTRIBUTING.md
@@ -52,9 +52,15 @@ extension](#running--testing-extension-locally), you may want to check out the
 ## Running / testing extension locally
 
 In order to test out your local version of the extension or to run unit tests
-locally, you need to run the extension in debug mode. To do so you need to run
-the `workbench.action.debug.selectandstart` command in VSCode and then select either "Run
-Extension" or "Extension Tests".
+locally, you need to run the extension in debug mode. To do so, open the root directory:
+
+```bash
+code .
+```
+
+and then, in that window, run the `workbench.action.debug.selectandstart` command in VSCode.
+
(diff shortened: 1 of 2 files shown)


docs/contributing/CONTRIBUTING.md
```diff
@@ -52,9 +52,15 @@ extension](#running--testing-extension-locally), you may want to check out the
 ## Running / testing extension locally
 
 In order to test out your local version of the extension or to run unit tests
-locally, you need to run the extension in debug mode. To do so you need to run
-the `workbench.action.debug.selectandstart` command in VSCode and then select either "Run
-Extension" or "Extension Tests".
+locally, you need to run the extension in debug mode. To do so, open the root directory:
+
+```bash
+code .
```
pokey, line 58: If the developer has followed step 2 above (open directory in VSCode) then this step shouldn't be necessary. Did you run into some issue where it was necessary to run this command? Is it a no-op if they're already in the VSCode window for the extension?

pokey (commented):
One question; see below

docs/contributing/CONTRIBUTING.md
josharian: Oh, I apparently mentally combined that with step 3, which included the step

> and then close the window that opens

maybe a better fix would be to move step 2 to be the last step, since that's the moment you're ready to do some editing.

pokey: That's actually a completely separate window. Is that not what happened for you?

josharian: I didn't run step 2, because i mentally merged it w/ step 3. (That's what I Get for reading the instructions before running them!) Anyway, happy to abandon this change; it's probably not needed.

PR #1632, opened 2023-07-13
Update Talon side for new destination types

### LINKED ISSUE #725: Use data classes on talon side in order to get better type hints

### LINKED ISSUE #803: Destinations as first-class objects

### LINKED ISSUE #1440: Rework Talon side to use destinations
- [x] Update Talon side
- [x] Add deprecation warning if user says "to after"?  I believe we support this today, and some users are still saying that, as indicated by slack questions from the past month.  We could probably have this as an alternate spoken form for "after" and show a warning if they use it
- [x] Add destination box to the cheatsheet with "before target", "to target" and "after target"
- [x] Update "paste", "bring", "move" and others in cheatsheet to accept a destination and link to that box
- [x] Add destination to public api, and use it for snippets
- [x] #1319

## Talon side

On the Talon side, we would like to do the following:

### 1. Make `"before"`, `"after"` and `"to"` into "insertion modes"

We'll have a new list called `cursorless_insertion_mode`, which will consist of `"before"`, `"after"` and `"to"`.

These terms should also be removed from both `cursorless_position` and `cursorless_source_destination_connective`.

Note that `"end of"` and `"start of"` remain as position modifiers.

### 2. Add new capture for destinations

Replace the [`<cursorless_positional_target>` capture](https://github.com/cursorless-dev/cursorless/blob/2ee7ca137e801b922ec020b59c15350818f2d2d0/cursorless-talon/src/positional_target.py#L8-L20) with the following:

```python
@mod.capture(
    rule=(
        "{user.cursorless_insertion_mode} <user.cursorless_target> "
        "({user.cursorless_list_connective} {user.cursorless_insertion_mode} <user.cursorless_target>)*"
    )
)
def cursorless_destination(m) -> list[dict]:
    return {
        "type": "destinationList",
        "destinations": [
            {
                "type": "destination",
                "insertionMode": insertion_mode,
                "target": target,
            }
            for insertion_mode, target in zip(m.cursorless_insertion_mode, m.cursorless_target)
        ]
    }
```

Note that we are using the `cursorless_target` capture as part of this capture, which can contain its own `"and"`s. That's how we support `"after air and bat and before cap"`.

We would then use this new `cursorless_destination` capture wherever we use `cursorless_positional_target` today.

Note that we might instead want to do something like we do with targets, where the primitive destination can appear at the top level when there is no "and"

## Problems

- We already have `"before"` and `"after"` in `positions.csv`, and `"to"` in `connectives.csv`, but ideally we'd move them into the same file as each other

AndreasArvidsson:
Since I had to basically update every capture and action with the new formats I took the liberty of doing some general cleanup and maintenance in this pr. 

Ponderings:
* Should we create a single python file with Cursorless public interface? Today I believe some of them are in actions and some in snippets and so on.
* It's quite hard to know where the default spoken forms are located. I'm leaning more and more towards just having a single folder with son files that the fine all our spoken forms.

Fixes #1440
Fixes #725 
Fixes #803

## Checklist

- [ ] Handled migration of csv files
- [ ] Update cheat sheet
- [ ] Update documentation
- [ ] Migrate modifiers to data classes?
- [ ] Figure out how to handle fallback of Cursorless commands when the editor is not in focus in a more standardized way


AndreasArvidsson:
@pokey I think this is ready for a preliminary review. There are still a few tasks left to do, but I think discussing the current direction could be useful.
I've done some rudimentary testing and everything appears to be working. My plan is to start working on the Talon grammar testing you using this branch of Cursorless to get some more manual testing done.

pokey:
Nice!

Re testing: you mean you'll base the testing branch off main, but will install this PR locally to drive while you're developing?

AndreasArvidsson:
Exactly like that

AndreasArvidsson:
Closed in favor of https://github.com/cursorless-dev/cursorless/pull/1637

PR #1637, opened 2023-07-16
Talon side spoken form tests

### LINKED ISSUE #449: Add regression tests for voice commands

### LINKED ISSUE #1675: Change "take vest past after air" test to "take vest past end of air"
Interesting. We should prob change this test to use "end of". Or just drop it 🤷‍♂️

_Originally posted by @pokey in https://github.com/cursorless-dev/cursorless/pull/1674#discussion_r1271428440_

AndreasArvidsson:
- Fixes #449
- Fixes #1675

## Potential manual tests to add

For these, just bake payload and spoken form into `const`s in the test itself

- [x] "slice past"
- [x] "two tokens forward"
- [x] "one token"
- [x] "take inside pair"

## Checklist

- [x] Handle if Talon was in sleep mode initially
- [x] Add deprecation warning if user says "to after"?  I believe we support this today, and some users are still saying that, as indicated by slack questions from the past month.  We could probably have this as an alternate spoken form for "after" and show a warning if they use it
- [x] Stop using `generateSpokenForms` to check whether to run a test; instead read `spokenFormError` field on recorded test
- [x] How do we actually want to launch this? Launch config in vscode and then?
- [x] Add notifications when test start and stop
- [/] I have updated the [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [/] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs)


packages/cursorless-engine/src/core/commandVersionUpgrades/canonicalizeSpokenFormTestCommand.ts
pokey, line 1: This is epic 😍

packages/common/src/definitions/modifiers.ts
```diff
@@ -0,0 +1,126 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+
+import { objectFromKeys, swapKeyAndValue } from "../util/objectUtils";
+
+const modifiers = {
+  bounds: "excludeInterior",
```
pokey, line 6: I would have mapped id to spoken form, as the former is the primary key for that entity. How come you went this way?

pokey (commented):
Just left a couple very quick comments

packages/common/src/definitions/modifiers.ts
AndreasArvidsson: This is the way we are used to seeing them in Talon. I'm also thinking that we're going to use the same source to actually generate the spoken forms as well as cheatsheet. In my mind this just feels "right".

packages/cursorless-engine/src/core/commandVersionUpgrades/canonicalizeSpokenFormTestCommand.ts
AndreasArvidsson: Glad you like it :)

packages/common/src/definitions/modifiers.ts
AndreasArvidsson: All true.

pokey:
Worth running through the checkboxes in these two issues to see if you've addressed everything.  If not, I'm not saying you should do it in this PR, but we should prob not close those issues:

- https://github.com/cursorless-dev/cursorless/issues/449
- https://github.com/cursorless-dev/cursorless/issues/1440

AndreasArvidsson:
@pokey Updated and ready for review

AndreasArvidsson:
> Ok this looks good to me, but for some reason is still failing for me locally, so we'll need to debug. Also worth having a look at the changes I made

The changes just looks fine. I just discover that after Talon has restarted some of the tests are failing for me as well. Touching a single Talon user file fixes this for some reason.

AndreasArvidsson:
Its the Cursorless simple actions list that don't get updated to default unless you touch a file. I will investigate further

pokey (approved):

PR #1648, opened 2023-07-17
Add `vscodeApi` for mocking

pokey:
- Required by https://github.com/cursorless-dev/cursorless/pull/1649

## Checklist

### DIFF
--- a/packages/cursorless-vscode/src/constructTestHelpers.ts
+++ b/packages/cursorless-vscode/src/constructTestHelpers.ts
@@ -17,9 +17,10 @@ import {
   takeSnapshot,
 } from "@cursorless/cursorless-engine";
 import { TestHelpers } from "@cursorless/vscode-common";
-import * as vscode from "vscode";
+import type { TextEditor as VscodeTextEditor } from "vscode";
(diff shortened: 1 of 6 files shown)


packages/cursorless-vscode/src/constructTestHelpers.ts
```diff
@@ -61,7 +62,7 @@ export function constructTestHelpers(
     },
 
     setStoredTarget(
-      editor: vscode.TextEditor,
```
AndreasArvidsson, line 64: Is this really an improvement?

AndreasArvidsson (approved):

packages/cursorless-vscode/src/constructTestHelpers.ts
pokey: I made these changes when I thought we'd be using `vscode` as the name of our wrapper, so was planning to start deprecating our use of `import * as vscode from "vscode"` . After renaming it to `vscodeApi`, we prob don't need to deprecate that. Happy to roll these back

AndreasArvidsson: Ah got it. Please do that :)

PR #1652, opened 2023-07-17
Add `ScopeSupportChecker`

pokey:
- Required by #1653 
- Depends on https://github.com/cursorless-dev/cursorless/pull/1651

## Checklist

- [ ] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)
- [ ] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [ ] I have not broken the cheatsheet

PR #1649, opened 2023-07-17
Add `VscodeFancyRangeHighlighter`

pokey:
- Depends on https://github.com/cursorless-dev/cursorless/pull/1648
- Depends on https://github.com/cursorless-dev/cursorless/pull/1645
- Required by https://github.com/cursorless-dev/cursorless/pull/1653

## Checklist

- [ ] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)
- [ ] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [ ] I have not broken the cheatsheet

PR #1650, opened 2023-07-17
Add `ScopeRangeProvider`

pokey:
- Required by #1653 
- Depends on #1647 

## Checklist

- [ ] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)
- [ ] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [ ] I have not broken the cheatsheet

PR #1653, opened 2023-07-17
Add the scope visualizer

pokey:
- Depends on https://github.com/cursorless-dev/cursorless/pull/1651
- Depends on https://github.com/cursorless-dev/cursorless/pull/1652
- Depends on https://github.com/cursorless-dev/cursorless/pull/1649
- Depends on https://github.com/cursorless-dev/cursorless/pull/1644
## Checklist

- [x] Add image to doc page
- [x] File issue to migrate all our vscode references to use new thin wrapper
- [x] File issue about reworking removal highlight range to return generalised range, and not using line range for line target content range
- [x] Investigate cheatsheet performance issue
- [x] Migrate todos from #1523 
- [x] Compute border colors from background color?


packages/cursorless-vscode/package.json
```diff
@@ -399,6 +411,198 @@
           },
           "additionalProperties": false
         },
+        "cursorless.scopeVisualizer.colors.dark": {
```
pokey, line 414: Should we flatten these?  Might make it easier to change them

cursorless-talon/src/scope_visualizer.py
```diff
@@ -0,0 +1,49 @@
...
+        """Hides scope visualizer"""
+        run_rpc_command_no_wait("cursorless.hideScopeVisualizer")
+
+
+def on_ready():
+    init_csv_and_watch_changes(
+        "scope_visualizer",
+        {
+            "show_scope_visualizer": {"visualize": "showScopeVisualizer"},
+            "hide_scope_visualizer": {"visualize nothing": "hideScopeVisualizer"},
```
pokey, line 43: I decided to just make `"visualize nothing"` have its own entry in the csv, rather than reusing `"visualize"` and `"nothing"`.  This allows the user more flexibility to use whatever command they want

cursorless-talon/src/scope_visualizer.py
```diff
@@ -0,0 +1,49 @@
...
+        )
+
+    def private_cursorless_hide_scope_visualizer():
+        """Hides scope visualizer"""
+        run_rpc_command_no_wait("cursorless.hideScopeVisualizer")
+
+
+def on_ready():
+    init_csv_and_watch_changes(
+        "scope_visualizer",
```
pokey, line 40: We're introducing yet another `.csv` file 😅.  I think that's ok, but after our discussion I wonder if we want to keep adding more

packages/cursorless-vscode-e2e/src/suite/scopeVisualizer/runBasicMultilineContentTest.ts
```diff
@@ -0,0 +1,84 @@
+import { openNewEditor } from "@cursorless/vscode-common";
+import * as vscode from "vscode";
+import { checkAndResetFakes } from "./checkAndResetFakes";
+import { injectFakes } from "./injectFakes";
+import { ExpectedArgs } from "./scopeVisualizerTest.types";
+
+/**
+ * Tests that the scope visualizer works with multiline content, by
+ * ensuring that the correct decorations are applied so that it looks
+ * like `./runBasicMultilineContentTest.png`.
```
pokey, line 10: Annoyingly, we can't embed this image or link to it because then our docs build breaks 🙄.  See https://github.com/cursorless-dev/cursorless/issues/1254

packages/cursorless-vscode-e2e/src/suite/scopeVisualizer/runBasicMultilineContentTest.ts
```diff
@@ -0,0 +1,84 @@
...
+  checkAndResetFakes(fakes, expectedArgs);
+}
+
+const contents = `
+function helloWorld() {
+
+}
+`;
+
+const expectedArgs: ExpectedArgs = {
```
pokey, line 36: Admittedly, these are a bit tough to read.  Hopefully the image helps, and for the complex line decoration drawing aspect, we have unit tests

packages/cursorless-vscode/src/extension.ts
```diff
@@ -63,6 +74,7 @@ export async function activate(
     testCaseRecorder,
     storedTargets,
     hatTokenMap,
+    scopeProvider,
```
pokey, line 77: from #1650 / #1651 / #1652

cursorless-talon/src/scope_visualizer.py
AndreasArvidsson: Yeah a bit tentative to do that at the moment

cursorless-talon/src/scope_visualizer.py
AndreasArvidsson: Could you expand on this? What else would you combine with visualize nothing?

packages/cursorless-vscode-e2e/src/suite/scopeVisualizer/runBasicMultilineContentTest.ts
AndreasArvidsson: Yeah really annoying

packages/cursorless-vscode/package.json
AndreasArvidsson: I would recommend as many settings as possible to be changeable in the gui

packages/cursorless-vscode-e2e/src/suite/scopeVisualizer/runBasicMultilineContentTest.ts
AndreasArvidsson: Images definitely helps

AndreasArvidsson (approved):

cursorless-talon/src/scope_visualizer.py
pokey: Any alternative suggestions?

cursorless-talon/src/scope_visualizer.py
pokey: I mean that they could change "visualise nothing" to something like "disappear". They can't do that if we tie it to the spoken forms for "visualise" and "nothing"

AndreasArvidsson: Got it. I guess that's reasonable.

pokey: Possibly.  Tho I worry that would be harder to find

packages/cursorless-vscode/package.json
pokey: Hmm because these aren't proper color theme colors (because we need to access them for programmatic blending), they are a bit annoying to edit with settings gui:

image

If I make them one nested object, then when you click the settings button it dumps all the defaults to json and shows you rendered color and color picker

image

wdyt?  I don't feel super strongly either way, but it almost seems like ux is better without settings gui here

PR #1662, opened 2023-07-19
migrate Go scopes statement, comment, string, and textFragment

### LINKED ISSUE #1592: Incorrect statement definition
Looks like a shortcoming in the definition of statement, specifically around [expression statements](https://go.dev/ref/spec#Expression_statements). `take state air` correctly picks out this statement:

```s = lit```
but not this one:

```scan.Init(file, buf, nil, 0)```

[Slack Message](https://talonvoice.slack.com/archives/C026KPTJE6T/p1688671785119089?thread_ts=1688663398.532399&cid=C026KPTJE6T)

### PR DESCRIPTION
**Pull request by josharian**
This adds statement scopes for block and expression statement.

Fixes #1592

Co-authored-by: Pokey Rule <755842+pokey@users.noreply.github.com>

## Checklist

### CODE REVIEW

**Comment by josharian** [comment]
Decided to do scope migrations very incrementally.

The scope visualizer is AMAZING for this.

**Comment by josharian** [comment]
this should be pretty straightforward, but one thing to look out for while reviewing, is whether the block statement should be included. It is a statement, strictly speaking, but it only contains other statements. I don't know whether this will be expected to users, or even helpful. There is a test for it, so you can look at what the behavior is. It might make more sense to remove it. Input requested.

**Comment by josharian** [comment]
I have a new test. Yes. I have not broken the existing test? No comment.

**Comment by josharian** [comment]
I need to learn about how `item` is implemented before I can make more progress here. Please hold. :)

File: queries/go.scm [hunk 1]
```diff
@@ -0,0 +1,109 @@
...
+;; This guarantees that a user always knows how to refer to any composite literal.
+;; There are cases in which being overgenerous in matching is not ideal,
+;; but they are rarer, so let's optimize for the common case.
+;; Mixed keyed and non-keyed elements are also rare in practice.
+;; The main ambiguity is with {}, but there's little we can do about that.
+;;
+;; Go users may also expect that the map and list scopes will include the type definition.
+;; However, there is a Cursorless type scope, and defining the map
+;; and list scope to include the type if present has proved to be extremely difficult.
+;; The sticking point is the "if present". It is easy to end up with duplicates.
```
pokey, line 75: This should actually be quite easy using our custom `@foo.start` / `@foo.end` construct. It constructs a range between the two nodes and calls it `@foo`, and if one end is missing, it just uses the other without constructing a range. You can even apply query predicates to the ends. Have a look at our other `.scm` files for examples of this and our custom predicates

queries/go.scm
```diff
@@ -0,0 +1,109 @@
...
+;;
+;; Combining all these constraints suggests the following simple rules:
+;;
+;;   * If there is a keyed element present, then it is a map.
+;;   * If there is a non-keyed element present, then it is a list.
+;;   * If there are both or neither, then it is both a map and a list.
+;; Conveniently, this is also simple to implement.
+;;
+;; This guarantees that a user always knows how to refer to any composite literal.
+;; There are cases in which being overgenerous in matching is not ideal,
```
pokey, line 67: Better to be overgenerous. I'd go for your rule, or even simpler and just have map and list be synonymous. But I think your rule makes sense

pokey (commented):
Just left a couple quick comments

pokey:
Also, `item` is going to be hard fwiw. Not sure I'd block on that one

josharian:
I’m making steady progress, including on item.

pokey:
Item is often best left to the generic item scope handler. If you just remove `collectionItem` from the definitions, it should fall back to that. If that doesn't work well in certain circumstances, then you can define it using scope queries where appropriate, but remember that the removal range should include a comma, and that comma may be before or after the item, or not exist at all, depending on the item and the list

josharian:
> Item is often best left to the generic item scope handler. If you just remove collectionItem from the definitions, it should fall back to that.

With no collectionItem in either the legacy or the new style definitions, items doesn't work. it is something about the partial migration, or about missing `.domain` or `.iteration` annotations.

I'll keep playing around with it for a little bit, and then maybe we can discuss some details over voice/video.

josharian:
A bit of experimentation suggests leading/trailing for key/value/item is currently not working with Go either. :)

josharian:
Ok, I've made some progress. I'm down to three failing tests. One of them is mysterious--I don't know how my changes could have caused it to fail; it is an entirely unrelated part of the code. The two comprehensible failures are going from a key to its associated value and vice versa. Also, leading and trailing do not work at all, although that is existing behavior and there are no tests .

The code as written is mildly horrifying. There's a lot of duplication and complication.

 I secretly hope for a more elegant way. But if not, that's okay.

The next step here is probably going to involve non-trivial discussion and teaching. Let's both set this down and not look at it until we do so together over voice.

Thanks in advance for all your help with this.  It has been involved and I know I'veasked a lot.

josharian:
Hard to call this ready for review, but the tests all pass!

Let's discuss leading and tailing over voice. For reals this time.

And then I'll rebase my mess away. :)

josharian:
https://github.com/cursorless-dev/cursorless/assets/67496/2359a03f-f224-4cf0-8755-2c92abfc96d4

josharian:
Don't know why the pre-commit is failing. If I run locally:

```
find packages/cursorless-vscode-e2e/src/suite -name "*.yml" | xargs pnpm transform-recorded-tests
```

I get no changes. 🤷🏽

pokey:
When you mean "item doesn't work", you mean it doesn't work with scope visualizer? That's because it's still legacy impl. It should still function as a scope type. I would be tempted to use it rather than defining your own. For example, I don't think your "chuck item" is working, unless I've missed something

Re CI, I can fix. It has to do with the way pre-commit is splitting up args

pokey:
Also, it looks like you're treating the condition of an if statement as a statement?

josharian:
> Also, it looks like you're treating the condition of an if statement as a statement?

Go [if statements can contain simple statements](https://go.dev/ref/spec#If_statements). Useful for reducing scope of variables.

josharian:
> Re CI, I can fix.

I'd prefer to learn to fish, if that's OK.

josharian:
> When you mean "item doesn't work", you mean it doesn't work with scope visualizer? That's because it's still legacy impl. It should still function as a scope type. I would be tempted to use it rather than defining your own. For example, I don't think your "chuck item" is working, unless I've missed something

I removed every reference to `collectionItem` from `go.scm`, and items behave...incorrectly. I didn't delete anything related to items from `go.ts`. So I'm not quite sure how to proceed here.

With this PR as written, items work now in the sense that they get correct ranges, but yeah, they have no delimiter support, so `chuck item` doesn't clean up commas. `key` and `value` also lack delimiter support. And I'm not convinced iteration is working exactly as intended.

None of this is urgent. Let's look together Sunday.

josharian:
DO NOT REVIEW

pushing code to have it available for discussion only

josharian:
closing this mess in favor of #1685, which is much smaller in scope.

once that and a few other outstanding pull requests of mine are merged, I will revive the relevant code from this and start new smaller pull requests.

PR #1663, opened 2023-07-20
Scope support tree view

pokey:
image

## Checklist

- [x] File follow-up issue to use the language icon for language-specific scopes
- [x] Add tests for talon side. Use context to set csv overrides dir setting to tmp dir, then override spokenforms json output location, then call the `update` function. Check output, then tweak csv, then check output again
- [x] Add buttons to visualize scope, removal range, and iteration scope (if it's supported and present)
- [x] Indicate which ones contain your selections in their domain?
- [x] Rename view to just `cursorless.scopes`, and change title
- [x] Change section titles to be "present", "supported", "legacy", and "unsupported"
- [x] What to do when we add a new scope type? See TODO comment. update: let's detect that it's not in the json, and report that they need to update
- [x] Click on visualized scope to deactivate scope visualizer
- [x] Indicate which scope is being visualized
- [x] Other things from tree provider / tree item api that we want to support?
- [x] Voice command to show tree view
- [x] Where to put the tree view? Currently in explorer sidebar
- [x] Click on scope to show scope visualizer
- [x] Add hover (alternate spoken forms?)
- [x] Move `-` entries to the end (maybe in own section?)
- [x] Add tooltip with alternate spoken forms
- [x] Fix custom regex
- [x] Add test for custom regex generate spoken form
- [x] Place language-specific scopes at the top.  ~~Have sub-levels for lang-specific vs global?  Collapse global by default?~~ Use icon to indicate whether global or lang-specific. Just use the icon for the current language. For global, either use globe or something, or just no icon at all
- [x] Add entries for other complex scope types? (eg custom regex?)
- [x] Add icons
- [x] Use spoken forms from `~/.cursorless/spoken.json`, falling back to defaults
- [x] Add entries for surrounding pairs


packages/cursorless-vscode/src/ScopeSupportTreeProvider.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  private registerScopeSupportListener() {
+    this.onDidChangeScopeSupportDisposable =
+      this.scopeProvider.onDidChangeScopeSupport((supportLevels) => {
+        this.supportLevels = supportLevels;
+        this._onDidChangeTreeData.fire();
+      });
+  }
+
+  getTreeItem(element: TreeItem): TreeItem {
```
AndreasArvidsson, line 78: Maybe rename your own type to make it clear that you're not just passing around a vscode tree item

pokey:
update from meet-up:

- Have our own cursorless sidebar which shows icon
- spoken form is "bar cursorless", which focuses cursorless sidebar
- use `code` icon for now

packages/cursorless-vscode/src/ScopeSupportTreeProvider.ts
```diff
@@ -0,0 +1,198 @@
...
+  > = new vscode.EventEmitter<MyTreeItem | undefined | null | void>();
+  readonly onDidChangeTreeData: vscode.Event<
+    MyTreeItem | undefined | null | void
+  > = this._onDidChangeTreeData.event;
+
+  constructor(
+    private context: vscode.ExtensionContext,
+    private scopeProvider: ScopeProvider,
+  ) {
+    this.treeView = vscode.window.createTreeView("cursorlessScopeSupport", {
```
auscompgeek, line 29: Since we get a command for free, should we use dots in this so all the commands Cursorless adds are in the same namespace?
```suggestion
    this.treeView = vscode.window.createTreeView("cursorless.scopeSupport", {
```

pokey: huh yeah that makes sense to me

packages/common/src/types/command/PartialTargetDescriptor.types.ts
```diff
@@ -75,88 +75,107 @@ export type PartialMark =
...
+  "notebookCell",
   // Talon
-  | "command";
+  "command",
+] as const;
+
+export function isSimpleScopeType(
+  scopeType: ScopeType,
+): scopeType is SimpleScopeType {
+  return simpleScopeTypeTypes.includes(scopeType.type as any);
```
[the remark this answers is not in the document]
pokey: huh ok that works; did it inline tho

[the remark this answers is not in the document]
pokey: Btw would you recommend using ts-reset? Out of scope for this PR I think but I wonder if we should add it in a follow-up?

packages/cursorless-engine/src/CustomSpokenForms.ts
```diff
@@ -0,0 +1,169 @@
...
+    for (const entryType of ENTRY_TYPES) {
+      // FIXME: How to avoid the type assertion?
+      const entry = Object.fromEntries(
+        entries
+          .filter((entry) => entry.type === entryType)
+          .map(({ id, spokenForms }) => [id, spokenForms]),
+      );
+
+      this[entryType] = Object.fromEntries(
+        Object.entries(defaultSpokenFormInfo[entryType]).map(
```
pokey, line 122: Fix custom regexes

PR #1671, opened 2023-07-22
Added generation of spoken form

AndreasArvidsson:
## Checklist

- [x] Use spoken form generation when recording tests. Update message with the generated spoken form.
- [x] Add tests for list targets with 0 and 1 element, ensuring there's no spoken form
- [x] Add tests for list destination with 0 and 1 element, ensuring there's no spoken form


packages/common/src/spokenForms/actions.ts
```diff
@@ -0,0 +1,74 @@
...
+  ["experimental.setInstanceReference"]: "from",
+
+  editNew: null,
+  executeCommand: null,
+  getText: null,
+  replace: null,
+
+  // applyFormatter: "format",
+  // findInDocument: "scout",
+  // nextHomophone: "phones",
```
pokey, line 62: What's up with these?

packages/common/src/spokenForms/actions.ts
```diff
@@ -0,0 +1,74 @@
...
+};
+
+export function actionToSpokenForm(name: ActionType): string {
+  if (name === "wrapWithSnippet") {
+    name = "wrapWithPairedDelimiter";
+  }
```
pokey, line 68: Why is this necessary?  I see both entries above

packages/common/src/spokenForms/characters.ts
```diff
@@ -0,0 +1,73 @@
...
+// https://github.com/talonhub/community/blob/9acb6c9659bb0c9b794a7b7126d025603b4ed726/core/keys/keys.py
+
+const alphabet =
+  "air bat cap drum each fine gust harp sit jury crunch look made near odd pit quench red sun trap urge vest whale plex yank zip"
+    .split(" ")
+    .reduce<Record<string, string>>((result, phrase, index) => {
+      const letter = String.fromCharCode("a".charCodeAt(0) + index);
+      result[letter] = phrase;
+      return result;
+    }, {});
```
pokey, line 14: Do you prefer this to using `Object.fromEntries`?

packages/common/src/spokenForms/modifiers.ts
```diff
@@ -0,0 +1,202 @@
...
+export const modifiersExtra = {
+  first: "first",
+  last: "last",
+  previous: "previous",
+  next: "next",
+  forward: "forward",
+  backward: "backward",
+};
+
+const scopes: Record<ScopeTypeType, string | null> = {
```
pokey, line 40: why not `SimpleScopeTypeType`?

pokey (commented):
partial review

packages/common/src/spokenForms/marks.ts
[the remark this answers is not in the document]
pokey: Oh right I'm guessing we can't because those are VSCode-specific?  Hmm.  Not sure I see an obvious way to get them here

packages/common/src/spokenForms/marks.ts
```diff
@@ -0,0 +1,61 @@
...
+  hole: "hole",
+  frame: "frame",
+  curve: "curve",
+  eye: "eye",
+  play: "play",
+  crosshairs: "cross",
+  bolt: "bolt",
+};
+
+const marks: Record<string, string | null> = {
```
pokey, line 24: But we could prob have something stronger here, no?

packages/common/src/types/command/PartialTargetDescriptor.types.ts
```diff
@@ -189,6 +189,8 @@ export type ScopeType =
   | CustomRegexScopeType
   | OneOfScopeType;
 
+export type ScopeTypeType = ScopeType["type"];
```
pokey, line 192: Do we still need this if we migrate the places I suggest to `SimpleScopeTypeType`?

packages/common/src/types/command/PartialTargetDescriptor.types.ts
```diff
@@ -344,8 +346,8 @@ export interface RangeModifier {
   type: "range";
   anchor: Modifier;
   active: Modifier;
-  excludeAnchor?: boolean;
-  excludeActive?: boolean;
+  excludeAnchor: boolean;
+  excludeActive: boolean;
```
pokey, line 350: oh are these always sent by Talon these days?

packages/common/src/types/command/hatStyles.types.ts
pokey, line 1: Oh you moved this here?  So then you should be able to use these types in your spoken forms map.  Mixed feelings bout moving them here tho; maybe let's discuss https://github.com/cursorless-dev/cursorless/labels/to%20discuss

packages/common/src/spokenForms/modifiers.ts
```diff
@@ -0,0 +1,202 @@
...
+  map: "map",
+  name: "name",
+  regularExpression: "regex",
+  section: "section",
+  sectionLevelOne: "one section",
+  sectionLevelTwo: "two section",
+  sectionLevelThree: "three section",
+  sectionLevelFour: "four section",
+  sectionLevelFive: "five section",
+  sectionLevelSix: "six section",
```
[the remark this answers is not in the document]
AndreasArvidsson: They are, but we want to be able to do spoke in form tests because we have fixtures for these. What should break? These spoken forms are only used for the Talon grammar test and nothing else.

packages/cursorless-engine/src/core/commandVersionUpgrades/upgradeV1ToV2/upgradeV1ToV2.ts
```diff
@@ -55,12 +55,14 @@ function upgradeModifier(modifier: ModifierV0V1): ModifierV2[] {
...
-
-      if (delimiterInclusion === "excludeInterior") {
-        return [{ type: "excludeInterior" }, surroundingPairModifier];
+      if (
+        delimiterInclusion === "interiorOnly" ||
+        delimiterInclusion === "excludeInterior"
+      ) {
+        if (surroundingPairModifier.scopeType.delimiter === "any") {
+          return [{ type: delimiterInclusion }];
+        }
```
pokey, line 64: How come we need to drop "any"?

packages/cursorless-engine/src/core/commandVersionUpgrades/upgradeV2ToV3/upgradeV2ToV3.ts
pokey, line 1: Not sure I follow the changes in this file, but not sure how much it matters, as long as the tests are still working. I can't imagine we have too many users on V2 anyway.  Maybe worth a very quick walkthrough https://github.com/cursorless-dev/cursorless/labels/to%20discuss

packages/cursorless-engine/src/core/commandVersionUpgrades/upgradeV4ToV5/upgradeV4ToV5.ts
pokey, line 1: Also could use just a quick walkthrough here

packages/cursorless-engine/src/core/commandVersionUpgrades/canonicalizeSpokenFormTestCommand.ts
```diff
@@ -0,0 +1,593 @@
...
+import { upgradeV1ToV2 } from "./upgradeV1ToV2";
+import { upgradeV2ToV3 } from "./upgradeV2ToV3";
+import { upgradeV3ToV4 } from "./upgradeV3ToV4";
+import { upgradeV4ToV5 } from "./upgradeV4ToV5";
+
+/**
+ * Temporarily here to support spoken form tests for current(v5) Talon side grammar
+ * TODO: remove
+ */
+export function canonicalizeSpokenFormTestCommand(
```
pokey, line 52: how come we need this again?  I thought it was for the Talon tests

packages/cursorless-engine/src/core/commandVersionUpgrades/canonicalizeSpokenFormTestCommand.ts
```diff
@@ -0,0 +1,593 @@
...
+        result.mark = target.mark;
+      }
+      return result;
+    }
+    case "implicit":
+      return target;
+  }
+}
+
+function generateSpokenForm(command: CommandComplete): string | null {
```
pokey, line 148: how come this function is in this file, and not in its own file?  I'd think this is kind of the punchline of the PR

packages/cursorless-engine/src/test/generateSpokenForm.test.ts
pokey, line 1: just directly use generateSpokenForm

packages/cursorless-engine/src/core/commandVersionUpgrades/canonicalizeSpokenFormTestCommand.ts
pokey: update: move to talon test PR

packages/common/src/spokenForms/actions.ts
AndreasArvidsson: Those are spoken forms where we don't have a corresponding action extension side. These are implemented Talon side.
Should I add a comment or just remove them?

packages/common/src/spokenForms/modifiers.ts
```diff
@@ -0,0 +1,202 @@
...
+  escapedSquareBrackets: "escaped box",
+  doubleQuotes: "quad",
+  parentheses: "round",
+  backtickQuotes: "skis",
+  squareBrackets: "box",
+  singleQuotes: "twin",
+  any: "pair",
+  string: "string",
+
+  collectionBoundary: null,
```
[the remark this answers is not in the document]
AndreasArvidsson: Used here
https://github.com/cursorless-dev/cursorless/blob/92fa046c06496a2f444706ef1b0286bed603b153/packages/cursorless-engine/src/processTargets/modifiers/surroundingPair/delimiterMaps.ts#L42-L47

packages/common/src/spokenForms/modifiers.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  // whitespace: "void",
```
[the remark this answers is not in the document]
AndreasArvidsson: It's just there to capture that we do have a default spoken form for this type. Do you want me to remove it?

packages/common/src/types/command/PartialTargetDescriptor.types.ts
AndreasArvidsson: Correct

packages/common/src/types/command/hatStyles.types.ts
AndreasArvidsson: Already discussed

AndreasArvidsson:
@pokey Updated wit requested changes from our meetup

pokey:
@AndreasArvidsson did we want to use the spoken form generator for recording test cases in this PR? Tbh, I think I'd almost prefer if update fixtures launch config updated spoken forms.  Makes me a bit less nervous than just automatically generating new test cases for the spoken form generator that are based on its behaviour at that moment and requiring us to carefully review them.  

But I'd lean towards having one or the other in this PR, as otherwise it's going to be pretty inconvenient the next time someone records a test

packages/common/src/spokenForms/actions.ts
```diff
@@ -61,7 +61,7 @@ const actions: Record<ActionType, string | null> = {
   getText: null,
   replace: null,
 
-  // These actions are implemented talent-side, usually using `getText` followed
+  // These actions are implemented talon-side, usually using `getText` followed
```
pokey, line 64: Looks like new Whisper model doesn't like the word Talon 😄

packages/common/src/spokenForms/actions.ts
```diff
@@ -0,0 +1,77 @@
...
+  // nextHomophone: "phones",
+};
+
+export function actionNameToSpokenForm(name: ActionType): string {
+  const result = actions[name];
+  if (result == null) {
+    throw Error(`Unknown action '${name}'`);
+  }
+  return result;
+}
```
pokey, line 77: Do we need this function?  Couldn't we just export the map like we do for connectives?  It's strongly typed, so we shouldn't be able to index with something that's not an action without typescript complaining

packages/common/src/spokenForms/marks.ts
```diff
@@ -0,0 +1,73 @@
...
+const hatColors: Record<HatColor, string | null> = {
+  blue: "blue",
+  green: "green",
+  red: "red",
+  pink: "pink",
+  yellow: "yellow",
+  userColor1: "navy",
+  userColor2: "apricot",
+
+  default: null,
```
pokey, line 13: This is a bit different from the other situations where we use a new value, right? In those situations, it means that there is no spoken form. So we'll raise a no spoken form error. Whereas in this situation, it just means that the spoken form is omitted. I wonder if we just want to map this to an empty string? I don't have a strong feeling either way.

pokey: nvm I think this is consistent

packages/common/src/spokenForms/marks.ts
```diff
@@ -0,0 +1,73 @@
...
+  return result;
+}
+
+export function markTypeToSpokenForm(mark: MarkType): string {
+  const result = marks[mark];
+  if (result == null) {
+    throw Error(`Unknown mark '${mark}'`);
+  }
+  return result;
+}
```
pokey, line 73: I wonder if we also want to just export a map here or at least remove the error throw because TypeScript should be protecting us.

packages/common/src/spokenForms/modifiers.ts
```diff
@@ -0,0 +1,202 @@
...
+  if (scope === "switchStatementSubject") {
+    throw new NoSpokenFormError(`No spoken form for scope '${scope}'`);
+  }
+
+  const result = scopes[scope];
+  if (result == null) {
+    throw Error(`Unknown scope '${scope}'`);
+  }
+  return result;
+}
```
pokey, line 181: It's a bit unfortunate that we can't distinguish between whether something has no spoken form or is an error just from the map. I'd be tempted to have something a bit stronger typed in our map, but I think we can revisit this when we start to actually get the spoken form from the user's customizations as per my other PR. In my PR we represent spoken forms as a list, and always have an entry for everything. Things that don't have a spoken form just have an empty list. More verbose and annoying here were we just want our one canonical default. But we may possibly use these to generate something closer to my format. We could also use a sentinel value to indicate there is no default spoken form, to distinguish from the other case where we generate the spoken form via other mechanisms (eg ordinal scope)

packages/common/src/spokenForms/modifiers.ts
```diff
@@ -0,0 +1,202 @@
...
+  squareBrackets: "box",
+  singleQuotes: "twin",
+  any: "pair",
+  string: "string",
+
+  // Used internally by the "item" scope type
+  collectionBoundary: null,
+
+  // This one can't be used as a scope type; only for wrapping
+  // whitespace: "void",
```
pokey, line 124: How do we generate spoken form for "void wrap"?

packages/common/src/spokenForms/actions.ts
AndreasArvidsson: No not anymore

packages/common/src/spokenForms/marks.ts
AndreasArvidsson: Agreed

packages/common/src/spokenForms/modifiers.ts
AndreasArvidsson: Sounds reasonable to revisit this

packages/common/src/spokenForms/modifiers.ts
pokey: yep this broke when I tried to record "void wrap air".  Fixed and added test

packages/common/src/spokenForms/modifiers.ts
pokey: On second thought, I don't see why `switchStatementSubject` is any different from the section level scope types. They're not activated by default, but we still have them mapped in the spoken forms here. So why don't we just do the same thing with `switchStatementSubject`?

AndreasArvidsson: `switchStatementSubject` has no default spoken form in our python files and does not exist in the users csv.

pokey: sure but I'm not seeing why we shouldn't have a spoken form here?

packages/common/src/types/command/hatStyles.types.ts
pokey, line 1: I'm a bit nervous about having this file in common.  I don't want us to inadvertently start assuming that these are the only hats that will be supported, when we made an effort to factor that out so that other editors can support arbitrary styles.  Any objection to rolling this back?  I'm happy to do so.  Then we'd just have weak typing for hat shapes and colors, which is kind of correct if other editors do start supporting other shapes

Alternately, we could decide that we want to restrict to these shapes and editors are only allowed to support a subset of these.  But I think that's a bigger conversation; not sure I want to tie us to that as part of this PR

Lmk if you have objections to me rolling this back; should be trivial I think because those are the only changes to cursorless-vscode

packages/cursorless-engine/src/test/generateSpokenForm.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  const spokenForm = await generateSpokenFormWithHatTokenMapSuffix(
+    file,
+    fixture.command,
+  );
+
+  if (spokenForm == null) {
+    return null;
+  }
+
+  if (shouldUpdateFixtures()) {
```
pokey, line 39: Do we have a launch config to actually run this?

packages/common/src/spokenForms/modifiers.ts
pokey: I would be inclined to have a spoken form for everything where it makes any sense to have a spoken form, as you've done for section headings

pokey: if no objections i'll do that

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
```diff
@@ -0,0 +1,486 @@
...
+      }
+
+      const number = numberToSpokenForm(modifier.length);
+
+      if (modifier.start === 0) {
+        return [modifiersExtra.first, number, pluralize(scope)];
+      }
+      if (modifier.start === -modifier.length) {
+        return [modifiersExtra.last, number, pluralize(scope)];
+      }
```
pokey, line 300: how come we don't throw NoSpokenForm here?

packages/cursorless-engine/src/test/generateSpokenForm.test.ts
AndreasArvidsson: Same one as before.

packages/common/src/types/command/hatStyles.types.ts
AndreasArvidsson: I was thinking that we should only allow a subset of this, but if you prefer please revert this.

packages/common/src/spokenForms/modifiers.ts
AndreasArvidsson: Just because we don't have one and I didn't think making up new spoken forms was part of this pr, but if you have something please add it. You also need to add it Talon side or our spoken form tests will fail.

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
```diff
@@ -0,0 +1,477 @@
...
+
+      throw Error(`Modifier '${modifier.type}' is not fully implemented`);
```
pokey, line 293: How come we don't throw `NoSpokenForm` here?

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
```diff
@@ -0,0 +1,477 @@
...
+            : ordinalToSpokenForm(modifier.anchor.start + 1);
+        const active = modifierToSpokenForm(modifier.active);
+        const connective = getRangeConnective(
+          modifier.excludeAnchor ?? false,
+          modifier.excludeActive ?? false,
+        );
+        return [anchor, connective, active];
+      }
+
+      throw Error(`Modifier '${modifier.type}' is not fully implemented`);
```
pokey, line 316: and this one, how come not `NoSpokenFormError`?

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
AndreasArvidsson: This is more of an "this shouldn't happen and if so something in my logic has gone wrong and need to be looked over" compared to "this is expected, but we don't have a spoken form"

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
AndreasArvidsson: See above

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
```diff
@@ -0,0 +1,478 @@
...
+    return [direction, number, scopePlural];
+  }
+
+  if (modifier.length === 1) {
+    const ordinal = ordinalToSpokenForm(modifier.offset);
+    // second next/previous token
+    return [ordinal, direction, scope];
+  }
+
+  throw Error(`Modifier '${modifier.type}' is not fully implemented`);
```
pokey, line 379: here as well, I'd think we'd throw `NoSpokenFormError`

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
```diff
@@ -0,0 +1,478 @@
...
+}
+
+function scopeTypeToSpokenForm(scopeType: ScopeType): string {
+  switch (scopeType.type) {
+    case "oneOf":
+    case "customRegex":
+    case "switchStatementSubject":
+      throw new NoSpokenFormError(`Scope type '${scopeType.type}'`);
+    case "string":
+      throw new AmbiguousSpokenFormError(`Scope type '${scopeType.type}'`);
```
pokey, line 389: Why not just throw `NoSpokenFormError` here?  I'm not sure I see the distinction.  There is no spoken form for the `string` scope, so I'd argue that `NoSpokenFormError` applies.  It used to be ambiguous, but after #1661 it is not

AndreasArvidsson: Yes no it's totally fine to change that. Just wanted to be more specific in the past.

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
```diff
@@ -0,0 +1,478 @@
...
+
+    default:
+      return scopeSpokenForms[scopeType.type];
+  }
+}
+
+function markToSpokenForm(mark: PartialMark): RecursiveArray<string> {
+  switch (mark.type) {
+    case "decoratedSymbol": {
+      const [color, shape] = mark.symbolColor.split("-");
```
pokey, line 410: This is VSCode-specific. I'm inclined to get this stuff merged in, but we should definitely revisit the decorated symbol aspect of all this. My understanding was that the decorated symbols were to be editor-specific, so really VSCode should be injecting a function to convert decorated symbol to spoken form. But then obv we can't use this in CLI. Either way, let's leave it and file an issue to revisit it. Prob can do that when we get further along with other editors

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
```diff
@@ -0,0 +1,478 @@
...
+          const connective = getRangeConnective(
+            mark.excludeAnchor ?? false,
+            mark.excludeActive ?? false,
+          );
+          // Row five past seven
+          return [typeAnchor, numberAnchor, connective, numberActive];
+        }
+      }
+
+      throw Error(`Mark '${mark.type}' is not fully implemented`);
```
pokey, line 443: Once again, I think I'd throw `NoSpokenFormError`, no?

packages/cursorless-engine/src/test/generateSpokenForm.test.ts
pokey: Ah right.  Was looking for way to update just unit tests.  Prob fine to just use full test update

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
pokey: But there's nothing inherently wrong with having `length > 1` and `start > 0`? And we do support it in the extension I'm pretty sure.  And we could definitely introduce a spoken form for it in the future, eg `"second three words`".  So I'd argue that the only reason we're throwing an exception here is that there is currently no spoken form for it.

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
pokey: This one I'm slightly more on the fence about, but again I think mostly the problem is just that we don't currently have a spoken form for it. I don't think there's anything inherently wrong with this case, and we do support it extension-side. But again this one's a bit different because spoken form would prob be ambiguous and we may remove this construct in the future. But I'm still not sure I see the benefit of not just throwing `NoSpokenFormError`. I think I'd reserve `Error` for when there's truly something wrong / unexpected. But if you want to push the other direction for this specific case I could be convinced

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
pokey: This one I'm slightly more on the fence about, but again I think mostly the problem is just that we don't currently have a spoken form for it. I don't think there's anything inherently wrong with this case, and we do support it extension-side. But again this one's a bit different because spoken form would prob be ambiguous and we may remove this construct in the future. But I'm still not sure I see the benefit of not just throwing `NoSpokenFormError`. I think I'd reserve `Error` for when there's truly something wrong / unexpected. But if you want to push the other direction for this specific case I could be convinced

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
AndreasArvidsson: If you prefer please change it

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
pokey: Again, in this case we could absolutely support `"second next three funks`, so I'd argue that this should be `NoSpokenFormError`.  As mentioned above, if we support it extension side, and it is logically valid, I would argue that `NoSpokenFormError` is appropriate

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
pokey: I guess the main reason to keep all these cases as errors is that we don't have tests for them, so if we added spoken forms for them in the future, we wouldn't notice that generation doesn't work and they just get skipped in CI.  And their talon spoken form tests would get skipped too.  So I think I'd argue to either keep them as errors or add tests for them.  I think maybe for "second two funks" and "second next two funks", I'll add tests and switch them to no spoken form, and for range modifier / mark, let's leave as error, because it's unclear we'll ever support them, and when we do, we'll notice quickly when the error gets thrown.  Make sense?

pokey: Tbh I'm kinda tempted to capture which tests don't have proper spoken forms in their fixtures.  I'm not in love with the silent / possibly unpredictable skipping. Should be easy for us to automatically add those fields now.  Would be good to notice if they change.  wdyt?

pokey: Fwiw that would further decouple this PR from talon test PR, because then the field is just in the test itself, so it doesn't even need to call `generateSpokenForms`.  All in all, I think the idea of using some complex logic to determine whether to skip a test feels like an anti-pattern the more I think about it

AndreasArvidsson: That was kind of the purpose of keeping them as errors. If something in the future changes we will not silently as drop them.

pokey: Ok but to be clear this is what I'm proposing:

1. In your `generateSpokenForms`, keep everything as it is except that we change `"second two funks"` and `"second next two funks"` to throw `NoSpokenFormError`
2. Record tests for`"second two funks"` and `"second next two funks"`
3. Add a field to all recorded tests that currently have no spoken form
4. Use that field to decide whether we're ok with spoken form being `null`
5. In your PR, use that field to decide whether to run Talon test 

Otherwise, we will not notice if something starts throwing `NoSpokenFormError` in the future. It will just silently be dropped.  Correct?

AndreasArvidsson: I think that is fine.

pokey: Cool I'm happy to take a crack at it. I'd also be tempted to let `NoSpokenFormError` bubble out of `generateSpokenForms` fwiw because today there's no way for the caller to know why there's no spoken form. But it's a bit unfortunate that the type system gives them no indication that they should be expecting that error. In fact, that case is not really exceptional at all, because the user may have just disabled a spoken form. So maybe let's leave it as null for now, but consider making `generateSpokenForms` return an error object that includes message if we need that message at some point in the future

AndreasArvidsson: Please do
I think it's probably better to just use null for now so we get that help from the type system. Returning in an object is of course also an alternative.

packages/cursorless-engine/src/generateSpokenForm/defaultSpokenForms/characters.ts
```diff
@@ -0,0 +1,74 @@
...
+  "(": "paren",
+  ")": "right paren",
+  "{": "brace",
+  "}": "right brace",
+  "[": "square",
+  "]": "right square",
+  "<": "angle",
+  ">": "right angle",
+
+  "\uFFFD": "special",
```
pokey, line 59: added this to make a couple more spoken forms work

packages/cursorless-engine/src/generateSpokenForm/defaultSpokenForms/marks.ts
```diff
@@ -0,0 +1,64 @@
+import { PartialMark } from "@cursorless/common";
+
+const hatColors: Record<string, string | null> = {
```
pokey, line 3: switched this back to `string`, to keep this PR more self-contained.  We can revisit decoration spoken form generation in follow-up

packages/cursorless-engine/src/generateSpokenForm/generateSpokenForm.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    assert(hatMapSpokenForm.type === "success");
+    generatedSpokenForm.value += " " + hatMapSpokenForm.value;
+  }
+
+  if (shouldUpdateFixtures()) {
+    if (generatedSpokenForm.type === "success") {
+      fixture.command.spokenForm = generatedSpokenForm.value;
+      fixture.spokenFormError = undefined;
+    } else {
+      fixture.spokenFormError = generatedSpokenForm.reason;
```
pokey, line 51: note how we now capture / check the error in the test

packages/cursorless-engine/src/testCaseRecorder/TestCaseRecorder.ts
```diff
@@ -318,6 +327,7 @@ export class TestCaseRecorder {
         this.startTimestamp!,
         this.captureFinalThatMark,
         this.extraSnapshotFields,
+        spokenForm.type === "error" ? spokenForm.reason : undefined,
```
pokey, line 330: Capture the error here if there was one

packages/common/src/spokenForms/modifiers.ts
pokey: Let's leave it and revisit when we start supporting custom spoken forms

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
pokey: done.  last checkbox is for your talon test PR so I think we can resolve this issue here

packages/cursorless-engine/src/generateSpokenForm/destinationToSpokenForm.ts
```diff
@@ -0,0 +1,43 @@
...
+import { connectives } from "./defaultSpokenForms/connectives";
+import { targetToSpokenForm } from "./targetToSpokenForm";
+
+export function destinationToSpokenForm(
+  destination: DestinationDescriptor): RecursiveArray<string> {
+  switch (destination.type) {
+    case "list":
+      if (destination.destinations.length < 2) {
+        throw new NoSpokenFormError("List destination with < 2 elements");
+      }
```
pokey, line 16: Notice we now throw an error for this one

packages/cursorless-engine/src/generateSpokenForm/primitiveTargetToSpokenForm.ts
pokey, line 1: Could have broken this one even further but I think this was a good place to stop

packages/cursorless-engine/src/generateSpokenForm/generateSpokenForm.ts
```diff
@@ -0,0 +1,203 @@
...
+}
+
+function destinationToSpokenForm(
+  destination: DestinationDescriptor,
+): RecursiveArray<string> {
+  switch (destination.type) {
+    case "list":
+      if (destination.destinations.length < 2) {
+        throw new NoSpokenFormError("List destination with < 2 elements");
+      }
```
pokey, line 175: No way to speak this one

packages/cursorless-engine/src/generateSpokenForm/generateSpokenForm.test.ts
pokey, line 1: Did fairly extensive rewrite of this one; worth a look.  Also note how I moved it right next to the file it's testing.  I think I prefer this pattern; as you've argued, keep related stuff close together 😊

packages/cursorless-engine/src/generateSpokenForm/NoSpokenFormError.ts
pokey, line 1: Moved this closer to where it's used.  More inline with the trademark @AndreasArvidsson "move stuff close to where it's needed" philosophy I think 😊

packages/common/src/types/command/hatStyles.types.ts
pokey: See https://github.com/cursorless-dev/cursorless/issues/1709

packages/cursorless-engine/src/core/commandVersionUpgrades/generateSpokenForm.ts
pokey: See https://github.com/cursorless-dev/cursorless/issues/1709

pokey (approved):
I'm happy, but @AndreasArvidsson have a look through my comments to make sure you're happy with my changes

packages/cursorless-engine/src/generateSpokenForm/generateSpokenForm.ts
```diff
@@ -0,0 +1,165 @@
...
+}
+
+export function targetToSpokenForm(
+  target: PartialTargetDescriptor,
+): RecursiveArray<string> {
+  switch (target.type) {
+    case "list":
+      if (target.elements.length < 2) {
+        throw new NoSpokenFormError("List target with < 2 elements");
+      }
```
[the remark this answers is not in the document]
AndreasArvidsson: I'm wondering if this should not be a normal error. If we end up here we have failed somewhere in our upgrade path probably. This isn't just the case where there is no spoken form its actually incorrect state to my mind.

packages/cursorless-engine/src/generateSpokenForm/generateSpokenForm.ts
AndreasArvidsson: Same here. I would argue that this should be a normal error.

packages/cursorless-engine/src/generateSpokenForm/generateSpokenForm.ts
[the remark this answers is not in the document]
pokey: I'd lean towards allowing it, as it could happen if someone is doing sophisticated things using our api directly, but happy to discuss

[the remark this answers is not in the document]
pokey: Also fwiw this was actually happening in our upgrade path in a couple places; I added fixes in this PR

pokey:
aggh looks like it's time to fix that E2BIG thing.  wrote up a fix; going to test that it works here then cherry pick it to new PR

PR #1680, opened 2023-07-23
don't override file.exclude and search.exclude

josharian:
This is more likely to be a user preference than a Cursorless preference.

This is why out files were showing up during the live coding during the meetup.

## Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,11 +1,5 @@
-// Place your settings in this file to overwrite default and user settings.
+// Place settings in this file to overwrite default end user settings.
 {
-  "files.exclude": {
-    "out": false // set this to true to hide the "out" folder with the compiled JS files
-  },
-  "search.exclude": {
-    "out": true // set this to false to include "out" folder in search results
-  },
   "python.formatting.provider": "black",
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   // Turn off tsc task auto detection since we have the necessary tasks as npm scripts


AndreasArvidsson:
I don't really mind, but for me user settings will take precedence and by default the out directory is not excluded so it doesn't really make a lot of difference. I know that official documentation says that workspace settings should override user settings, but that doesn't happen in my own testing. I'm fine with this change.

AndreasArvidsson:
Sounds reasonable

pokey:
Not sure I understand why these settings caused you to see these files in search

AndreasArvidsson:
Workspace settings take precedence and here we have explicitly said that `out` should not be excluded.

auscompgeek:
The pattern that exists here says to exclude `/out` from search

AndreasArvidsson:
> The pattern that exists here says to exclude `/out` from search

Yes but not from files. And with them mono repository structure in Cursorless we have quite a few out directories.

auscompgeek:
Yeah that's my point, none of the actual `out` subdirectories are being excluded.

pokey:
But what's confusing to me here is that @josharian was seeing search results from the out directories but I don't

josharian:
I guess the key point here is that it only makes sense to have things in the cursorless VSCode settings that are directly related to hacking on cursorless. Things like which code formatter to use. Other things should be left to the user to decide.

auscompgeek:
> But what's confusing to me here is that @josharian was seeing search results from the out directories but I don't

Do either of you have `search.useIgnoreFiles` disabled or enabled respectively?

josharian:
> Do either of you have search.useIgnoreFiles disabled or enabled respectively?

it looks like I have it explicitly disabled. based on where it is in the settings file, that was a long time ago. I have no idea why it is set that way. :)

pokey:
But the fact that `out` is generated is a property of the repo, not the user, right? Or maybe there's a way to mark it as generated and let the user decide whether to show it in search results / sidebar? Eg via gitignore or some other ignore?

josharian:
> But the fact that out is generated is a property of the repo, not the user, right? 

fair point. I will close this for now, and if it becomes painful, I will dig around deeper for other solutions.

pokey:
Fine by me.  So removing `"search.useIgnoreFiles": false` from your personal settings fixed the issue for you for now?

josharian:
I'm not sure, I had this commit in a working branch. I'll remove it now and mess around my settings and find out what works.

josharian:
assuming I understand correctly, I have sent #1720

PR #1682, opened 2023-07-23
replace treesitter #start-position with .startof

josharian:
This particular query pattern is going to be very common.
It's worth having syntactic sugar for.
In particular, this reduces the level of indentation
required to express that a node's associated scope
(iteration, trailing, leading, etc.)
should start or end at the start or end of a different node.

While we're here, make a related error message more helpful.

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
+++ b/packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
@@ -8,6 +8,7 @@ import { parsePredicates } from "./parsePredicates";
 import { predicateToString } from "./predicateToString";
 import { groupBy, uniq } from "lodash";
 import { checkCaptureStartEnd } from "./checkCaptureStartEnd";
+import { rewriteStartOfEndOf } from "./rewriteStartOfEndOf";
 
 /**
  * Wrapper around a tree-sitter query that provides a more convenient API, and
@@ -95,6 +96,7 @@ export class TreeSitterQuery {
         const captures: QueryCapture[] = Object.entries(
           groupBy(match.captures, ({ name }) => normalizeCaptureName(name)),
         ).map(([name, captures]) => {
+          captures = rewriteStartOfEndOf(captures);
           const capturesAreValid = checkCaptureStartEnd(
             captures,
             ide().messages,
@@ -123,7 +125,7 @@ export class TreeSitterQuery {
 }
 
 function normalizeCaptureName(name: string): string {
-  return name.replace(/\.(start|end)$/, "");
+  return name.replace(/(\.(start|end))?(\.(startOf|endOf))?$/, "");
 }
 
 function positionToPoint(start: Position): Point {
--- a/packages/cursorless-engine/src/languages/TreeSitterQuery/rewriteStartOfEndOf.test.ts
+++ b/packages/cursorless-engine/src/languages/TreeSitterQuery/rewriteStartOfEndOf.test.ts
@@ -1,0 +1,73 @@
+import { Range } from "@cursorless/common";
+import { MutableQueryCapture } from "./QueryCapture";
+import { SyntaxNode } from "web-tree-sitter";
+import { rewriteStartOfEndOf } from "./rewriteStartOfEndOf";
+import assert = require("assert");
+
+type NameRange = Pick<MutableQueryCapture, "name" | "range">;
+
+interface TestCase {
+  name: string;
+  captures: NameRange[];
+  expected: NameRange[];
+}
+
+const testCases: TestCase[] = [
+  {
+    name: "should rewrite startOf to start of range",
+    captures: [
+      { name: "@value.iteration.start.startOf", range: new Range(1, 2, 1, 3) },
+      { name: "@collectionKey.startOf", range: new Range(1, 2, 1, 3) },
+    ],
+    expected: [
+      { name: "@value.iteration.start", range: new Range(1, 2, 1, 2) },
+      { name: "@collectionKey", range: new Range(1, 2, 1, 2) },
+    ],
+  },
+
+  {
+    name: "should rewrite endOf to start of range",
+    captures: [
+      { name: "@value.iteration.start.endOf", range: new Range(1, 2, 1, 3) },
+      { name: "@collectionKey.endOf", range: new Range(1, 2, 1, 3) },
+    ],
+    expected: [
+      { name: "@value.iteration.start", range: new Range(1, 3, 1, 3) },
+      { name: "@collectionKey", range: new Range(1, 3, 1, 3) },
+    ],
+  },
+
+  {
+    name: "should leave other captures alone",
+    captures: [
+      { name: "@value.iteration.start", range: new Range(1, 2, 1, 3) },
+      { name: "@collectionKey", range: new Range(1, 2, 1, 3) },
+    ],
+    expected: [
+      { name: "@value.iteration.start", range: new Range(1, 2, 1, 3) },
+      { name: "@collectionKey", range: new Range(1, 2, 1, 3) },
+    ],
+  },
+];
+
+suite("rewriteStartOfEndOf", () => {
+  for (const testCase of testCases) {
+    test(testCase.name, () => {
+      const actual = rewriteStartOfEndOf(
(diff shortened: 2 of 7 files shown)


josharian:
The code of the end of the test is 😬 . Looking forward to knowing the right way to do it. And anything else. :) Be brutal.

josharian:
I guess alternative approach would be to make the rewrite a method on MutableQueryCapture, and in line the forEach loop that calls it. Then the test would mostly go away. :)

packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
```diff
@@ -123,7 +125,7 @@ export class TreeSitterQuery {
 }
 
 function normalizeCaptureName(name: string): string {
-  return name.replace(/\.(start|end)$/, "");
+  return name.replace(/\.(start|end)(\.(startof|endof))?$/, "");
```
pokey, line 128: I don't think this works with `@foo.startOf` (without the `.start`)

queries/javascript.jsx.scm
```diff
@@ -77,10 +77,7 @@
...
 ;;!  --   ---
-(
-  (jsx_fragment
-    "<" @_.domain.start
-    ">" @name @_.domain.end
-  )
-  (#start-position! @name)
+(jsx_fragment
+  "<" @_.domain.start
+  ">" @name.start.startof @_.domain.end
```
pokey, line 82: why is this not just `@name.startOf`?

packages/cursorless-engine/src/languages/TreeSitterQuery/rewriteStartOfEndOf.test.ts
```diff
@@ -0,0 +1,61 @@
...
+      { name: "@value.iteration.start", range: new Range(1, 2, 1, 3) },
+    ],
+    want: [{ name: "@value.iteration.start", range: new Range(1, 2, 1, 3) }],
+  },
+];
+
+suite("rewriteStartOfEndOf", () => {
+  for (const testCase of testCases) {
+    test(testCase.name, () => {
+      const got = rewriteStartOfEndOf(
```
pokey, line 44: we usually use `actual` / `expected` rather than `got` / `want`.  I don't feel strongly if you want to push for the latter, but we should def be consistent

packages/cursorless-engine/src/languages/TreeSitterQuery/rewriteStartOfEndOf.test.ts
```diff
@@ -0,0 +1,61 @@
+import { Range } from "@cursorless/common";
+import { MutableQueryCapture } from "./QueryCapture";
+import { SyntaxNode } from "web-tree-sitter";
+import { rewriteStartOfEndOf } from "./rewriteStartOfEndOf";
+import assert = require("assert");
+
+type NameRange = Omit<MutableQueryCapture, "allowMultiple" | "node">;
```
pokey, line 7: the name `NameRange` implies to me that you care that you have `name` and `range`, so I'd argue to use `Pick` rather than `Omit`, eg
```suggestion
type NameRange = Pick<MutableQueryCapture, "name" | "range">;
```

packages/cursorless-engine/src/languages/TreeSitterQuery/rewriteStartOfEndOf.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  },
+
+  {
+    name: "should leave other captures alone",
+    captures: [
+      { name: "@value.iteration.start", range: new Range(1, 2, 1, 3) },
+    ],
+    want: [{ name: "@value.iteration.start", range: new Range(1, 2, 1, 3) }],
+  },
+];
```
pokey, line 39: i'd add tests for `@value.startOf` (without `.start`)

packages/cursorless-engine/src/languages/TreeSitterQuery/rewriteStartOfEndOf.ts
```diff
@@ -0,0 +1,31 @@
...
+export function rewriteStartOfEndOf(
+  captures: MutableQueryCapture[],
+): MutableQueryCapture[] {
+  // TODO: mutate instead of copy/create/return?
+  return captures.map((capture) => {
+    // Remove trailing .startof and .endof, adjusting ranges.
+    if (capture.name.endsWith(".startof")) {
+      return {
+        ...capture,
+        name: capture.name.replace(/\.startof$/, ""),
```
pokey, line 18: camel-case please (`startOf`) 😊

pokey (commented):
mostly looks fine, with some minor comments.  still slightly on the fence here but seems you and @AndreasArvidsson are leaning towards it so I'll defer because I don't feel too strongly.  I do wonder if we want to use `startPosition` rather than `startOf`.  But again I'll defer to you and @AndreasArvidsson as I don't feel too strongly bout that either

AndreasArvidsson:
I think this is an improvement. I agree with all pokeys comments.

pokey:
But @AndreasArvidsson you're leaning towards `startOf` over `startPosition`?

AndreasArvidsson:
> But @AndreasArvidsson you're leaning towards `startOf` over `startPosition`?

Yes I think so

packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
josharian: I think that would only be used in cases in which you want a zero width range for an entire scope. And I can't think of any such cases offhand; are there any?

josharian: oh, haha, the very next comment gives an example of one. ok. :)

packages/cursorless-engine/src/languages/TreeSitterQuery/rewriteStartOfEndOf.test.ts
josharian: got and want are idiomatic go. I will use idiomatic typescript for typescript. :)  thanks for pointing it out.

packages/cursorless-engine/src/languages/TreeSitterQuery/rewriteStartOfEndOf.test.ts
josharian: TIL about pick. Thanks. done

packages/cursorless-engine/src/languages/TreeSitterQuery/rewriteStartOfEndOf.test.ts
josharian: done

packages/cursorless-engine/src/languages/TreeSitterQuery/rewriteStartOfEndOf.ts
josharian: done

queries/javascript.jsx.scm
josharian: it is now.

packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
josharian: done

josharian:
thanks! Please take another look.

packages/cursorless-engine/src/languages/TreeSitterQuery/rewriteStartOfEndOf.ts
```diff
@@ -0,0 +1,30 @@
+import { MutableQueryCapture } from "./QueryCapture";
+
+/**
+ * Rewrite captures, absorbing .startOf and .endOf into ranges.
```
pokey, line 4:
```suggestion
 * Modifies captures by applying any `.startOf` or `.endOf` suffixes. For
 * example, if we have a capture `@value.startOf`, we would rename it to
 * `@value` and adjust the range to be the start of the original range.
```

pokey (commented):
Left one tweak to make jsdoc clearer, but otherwise looks good!

pokey (approved):

PR #1683, opened 2023-07-24
add .projections.json

josharian:
I have been going back and forth about whether
this belongs in my global git ignores.
Open to either opinion. What do you think?

## Checklist

### DIFF
--- a/.projections.json
+++ b/.projections.json
@@ -1,0 +1,5 @@
+{
+  "*.ts": {
+    "alternate": "{}.test.ts"
+  }
+}


AndreasArvidsson:
Is this for the alternate file extension? In that case I don't think it belongs in the Cursorless repository. First because it's an extension that not every developer is going to have and also we have very few cases where the test file is named the same as a typescript file.

josharian:
Yeah. I’m convinced. :)

PR #1685, opened 2023-07-24
migrate Go scopes statement, comment, string, and textFragment

### LINKED ISSUE #1592: Incorrect statement definition

josharian:
This begins the Go migration to treesitter.

In addition to migrating scopes, this commit includes two bug fixes.

(1)

It adds a statement scope for expression statements.
This fixes #1592.

Note that block statements are omitted.
There are many block statements in a normal Go program,
most of which are not of interest.

For example, in:

if x() {
    // nada
}

there are two statements, one that is the entire if statement,
and one block statement that runs from brace to brace.
As the typical block statement includes other statements,
and is usually part of other named statements,
it will not typically be helpful during editing.

(2)

It fixes an existing bug in the handling of paired delimiters
when used with treesitter-based scopes.

When looking for paired delimiters from inside a string,
we first search inside the string only,
then outside the string only.

This lets us find the parens in both

"(hello)"

and

("hello")

given command "take round odd".

However, that fallback logic did not get replicated
to the treesitter-based scope flow.
As a result, we never searched outside the string.

Co-authored-by: Pokey Rule <755842+pokey@users.noreply.github.com>

## Checklist


packages/cursorless-engine/src/processTargets/modifiers/surroundingPair/index.ts
```diff
@@ -86,13 +86,16 @@ function processSurroundingPairCore(
...
+      const surroundingRange = findSurroundingPairTextBased(
         editor,
         range,
         containingScope[0].contentRange,
         delimiters,
         scopeType,
       );
+      if (surroundingRange != null) {
+        return surroundingRange;
+      }
```
pokey, line 98: excellent catch 👍.  I would be tempted to do 

```
return findSurroundingPairTextBased(
        editor,
        range,
        null,
        delimiters,
        scopeType,
      );
```

directly when we fallback rather than falling through to rest of function as I'm not sure we want to hit that next try/catch block

pokey (commented):
looks good, but see command and

- [ ] please add test for your `"change round"` example on `("aaa")`

packages/cursorless-engine/src/processTargets/modifiers/surroundingPair/index.ts
josharian: done

josharian:
> please add test for your "change round" example on ("aaa")

I added a test for both forms. and this also has existing coverage in the change items tests for Go, which were previously failing.

pokey:
https://github.com/cursorless-dev/cursorless/pull/1712 should fix pre-commit failure here

josharian:
> https://github.com/cursorless-dev/cursorless/pull/1712 should fix pre-commit failure here

it does, hurrah

pokey (approved):
looks good!

PR #1687, opened 2023-07-25
Make svg narrower if it will clip adjacent svgs

pokey:
If you increase your line height and increase Cursorless hat size, you may end up with hats that are wide enough to clip their neighboring hats.  This PR detects that situation and horizontally shrinks the hats so that they have a 1px margin from one another.  This does alter the aspect ratio of the hats, but if you've made them so big that they start clipping their nieghbors, I think this is the best solution.  If you don't want them to start deforming, make them smaller.  I don't really see any other option 🤔

Before:

image

After:

image

## Checklist

### DIFF
--- a/packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
+++ b/packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
@@ -247,6 +247,11 @@ export default class VscodeHatRenderer {
     const hatHeightPx = defaultHatHeightPx * scaleFactor;
     const hatWidthPx = defaultHatWidthPx * scaleFactor;
 
+    // If svg is too wide, scale it down. This will change the aspect ratio
+    const maxHatWidthPx = characterWidth - 1;
+    const widthFactor =
+      hatWidthPx > maxHatWidthPx ? maxHatWidthPx / hatWidthPx : 1;
+
     const hatVerticalOffsetPx =
       (DEFAULT_VERTICAL_OFFSET_EM + hatVerticalOffsetEm) * fontSize -
       hatHeightPx / 2;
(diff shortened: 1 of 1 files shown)


AndreasArvidsson (approved):

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
```diff
@@ -257,22 +262,24 @@ export default class VscodeHatRenderer {
...
+      .replace(/width="[^"]+"/, ``)
+      .replace(/height="[^"]+"/, ``)
+      .replace(/viewBox="([^"]+)"/, `style="overflow:visible"`);
+
+    const svg =
+      `<svg xmlns="http://www.w3.org/2000/svg" ` +
+      `viewBox="${newViewBoxString}" ` +
+      `width="${svgWidthPx}px" ` +
+      `height="${svgHeightPx}px">` +
+      `<g transform="scale(${widthFactor}, 1)">${innerSvg}</g></svg>`;
```
MartinRykfors, line 282: Had a problem here: If the `rawSvg` string includes an xml-header like
```
<?xml version="1.0" encoding="UTF-8"?><svg ...
```
Then the hat will not be displayed. I also had problems if the svg has a trailing newline or if it is formatted with indentation, but I could not pin down why that would cause problems.

MartinRykfors: I can make a PR for adding xml-header stripping to the preprocess script if you would like

pokey: Interesting.  Yeah feel free to add to preprocessor.  We also bundle an xml parser for use in cheatsheet so should prob use that to upgrade to something more robust.  Feel free to take a crack at it or even just file an issue

PR #1690, opened 2023-07-27
move subset regular expressions to a separate file

josharian:
This enables us to ask git not to [track changes to that file.

And while we're here, support multiple regular expressions.

## Checklist


packages/common/src/testUtil/testSubsetGrep.txt
```diff
@@ -0,0 +1,12 @@
+# This file contains the grep strings to pass to Mocha when running a subset of tests.
+# These grep strings will be used with the "Run Single Extension Test" launch configuration.
```
AndreasArvidsson, line 2: "Run test subset"

packages/common/src/testUtil/runTestSubset.ts
```diff
@@ -1,10 +1,26 @@
...
+  const inFile = path.join(
+    getCursorlessRepoRoot(),
+    "packages",
+    "common",
+    "src",
+    "testUtil",
+    "testSubsetGrep.txt",
+  );
+  const raw = fs.readFileSync(inFile, "utf-8");
+  const withoutComments = raw.replace(/#.*/g, "");
```
AndreasArvidsson, line 19: Wouldn't all of this been simpler to do with split.map.filter instead of all the regexes?

packages/common/src/testUtil/testSubsetGrep.txt
```diff
@@ -0,0 +1,12 @@
...
+# See https://mochajs.org/#-grep-regexp-g-regexp for supported syntax.
+#
+# One regular expression per line.
+# Tests matching any of the regular expressions will be run.
+#
+# To ask git not to track changes to this file, run:
+#   git update-index --skip-worktree packages/common/src/testUtil/testSubsetGrep.txt
```
AndreasArvidsson, line 9: Isn't there a way we can automate this so that each developer doesn't have to run this command?
Otherwise we can just use that command on the current `runTestSubset.ts` file?

packages/common/src/testUtil/runTestSubset.ts
josharian: will change. I'm new to typescript, and poking around the internet suggested that  regular expressions were de rigeur. I'd be delighted to switch to a functional style.

packages/common/src/testUtil/testSubsetGrep.txt
josharian: Unfortunately, you can't git ignore an existing file. I thought you could ( Maybe in the past you could?)  But it doesn't work. in classic git style, this mouthful is the recommendedapproach.

 an alternative is to git ignore the path and then create the file the first time any tests get run.   Assuming that the user runs the tests before they ever try to read the documentation and change this file, it will work like magic.

yet another alternative is to just document this git command. Although in that case it is still worth splitting out the easily editable parts into a separate file so you can just ignore that part.

I'm open to any approach as long as it leaves me with a clean tree. :)

packages/common/src/testUtil/testSubsetGrep.txt
josharian: Thanks. silly of me to copy paste stale doks without updating them. I will change this once there is clarity about what overall approach we want to take.

packages/common/src/testUtil/testSubsetGrep.txt
pokey: I'd be inclined to not have the file in source control at all.  I'd argue to either

- generate the file, or
- just show them a command in the docs that they can use to generate it, eg `echo "fooBar" > the/path/to/the/file.txt`, or
- we could map a voice command in our dev talon files

packages/common/src/testUtil/runTestSubset.ts
AndreasArvidsson: Absolutely nothing wrong with regular expressions per se, but I think we do prefer to use them only when necessary. This kind of patron is very common in javascript and something we use a lot in Cursorless.
```
raw
    .split(/\r?\n/)
    .map(line => line.trim())
    .filter(line => line.length > 0 && !line.startsWith("#"))
    .join("|")
```

packages/common/src/testUtil/testSubsetGrep.txt
AndreasArvidsson: I would prefer for the file to be generated. Adding additional voice commands for something that's gonna be done once feels unnecessary.

josharian: OK, unless I hear otherwise in the next day or two, I will generate the file whenever a test command is run.

pokey: Sure I'm fine with generating it, as long as that workflow is documented

josharian:
all comments addressed. I wouldn't be surprised if there's one more round of design feedback here, but at least it's ready for another look.

auscompgeek:
Instead of shoving this in a file, why not read an environment variable? Would be a lot simpler, and can be specified on the CLI (or prompted in the VSCode launch config) without surprises.

pokey:
The problem with prompting is that it's common that you will use the same subset repeatedly, so a bit tedious to have to keep providing the same subset

pokey:
to summarize from meet-up:

- fail with message if the regex file wasn't found
- ideally, would open the regex file if it wasn't found
- could possibly use a task for that, but if this or previous step too hard, ok to just show error message with path, which should be easy to ctrl-/cmd-click
- would be cool if there were another task that opens the regex file whether or not it exists, because I'd personally map that to a voice command because I frequently tweak that file

josharian:
OK, I've hooked this up to tasks. It's all pretty slick except for one problem: When we need to create the file on the very first run, we have to decide whether to exit 0 or exit 1. If we exit 0, it keeps on running. If we exit 1, it pops up an alarming error message saying that the build failed. I don't really like either behavior. We could make the default grep subset be something that matches nothing at all, so running "all the subtests" is superfast, but then you don't have nice useful templates to work off of. I've pushed up that change so you can play with it and see for yourself.

I'm running out of steam working on what is fundamentally a relatively minor improvement.

pokey:
the force push broke my review workflow pretty badly here 😕.  wdyt of https://github.com/cursorless-dev/cursorless/discussions/1752?

pokey:
made some tweaks; wdyt?

.vscode/tasks.json
```diff
@@ -46,6 +46,37 @@
...
+      "options": {
+        "cwd": "packages/common/src/testUtil"
+      },
+      "presentation": {
+        "reveal": "silent"
+      },
+      "group": "test"
+    },
+    {
+      "label": "Edit test subset file",
```
pokey, line 62: added this because I'll map it to a voice command; I do this all the time

josharian:
@AndreasArvidsson this needs a fare a bit of work still but if you had a few minutes to test it on your machine as is, that would be helpful. the important question I'd like to know the answer to is whether powershell/windows can handle the tiny bit of shell scripting that I shoved into the tasks.json.

AndreasArvidsson:
` *  The terminal process "C:\Program Files\Git\bin\bash.exe '--login', '-i', '-c', '[ -f testSubsetGrep.txt ] || {cp testSubsetGrep.{tmpl,txt}; code testSubsetGrep.txt; echo Please edit testSubsetGrep.txt and re-run.; exit 1}'" terminated with exit code: 2.`
` *  The terminal process "C:\Windows\System32\cmd.exe /d /c [ -f testSubsetGrep.txt ] || {cp testSubsetGrep.{tmpl,txt}; code testSubsetGrep.txt; echo Please edit testSubsetGrep.txt and re-run.; exit 1}" terminated with exit code: 1.`

pokey:
We can't use VSCode apis unless we want the file to open in the developer instance, unless I've missed something

AndreasArvidsson:
If that's the case we could at least use node to create the file instead of having to rely on terminal commands?

pokey:
I would probably prefer a node script to inline bash here as well; was just trying to avoid dragging this PR out even longer, and wasn't sure it was worth the effort. But I won't get in your way if you wanna push for it; prob worth @josharian learning how we write our TS node scripts as they're generally easier to read / maintain

pokey:
Or if you feel like just rolling up your sleeves (hmm that saying doesn't quite work with voice coding 🤔) go for it; I'm guessing it would take you ~5 mins 😄. I'm sure @josharian would learn to fish by either reading or writing the script

AndreasArvidsson:
I have faith in Josh abilities to figure it out :D

pokey:
@josharian here are links to the components for an example of how we do that.  These files together define the `populate-dist` script:

- https://github.com/cursorless-dev/cursorless/blob/93918bd75e78aef9bf5a0e1411df5f493c999fa9/.vscode/tasks.json#L39-L48
- https://github.com/cursorless-dev/cursorless/blob/93918bd75e78aef9bf5a0e1411df5f493c999fa9/packages/cursorless-vscode/src/scripts/populateDist/populateDist.ts
- https://github.com/cursorless-dev/cursorless/blob/93918bd75e78aef9bf5a0e1411df5f493c999fa9/packages/cursorless-vscode/src/scripts/populateDist/index.ts
- https://github.com/cursorless-dev/cursorless/blob/93918bd75e78aef9bf5a0e1411df5f493c999fa9/packages/cursorless-vscode/package.json#L1051

Noting that:

- you should prob put the script in the `cursorless-engine` package rather than `cursorless-vscode`, and
- you prob don't need separate `index.ts`; for a simple script just having one `.ts` file is fine

pokey:
Also, I forgot we already have some voice commands mapped to tasks, so it should be an easy one-liner to add a new one https://github.com/cursorless-dev/cursorless/blob/main/cursorless-talon-dev/src/cursorless_dev_vscode.talon

josharian:
OK, I took another crack at this. I still don't particularly like the behavior when someone runs the test subset before ever creating/editing the file. But because of how it's documented, it's unlikely to ever happen much. 🤷🏽 

I had to switch from .txt to .properties to get good commenting behavior.

packages/cursorless-engine/src/scripts/generateTestSubsetFile/generateTestSubsetFile.ts
```diff
@@ -0,0 +1,43 @@
...
+
+snippets
+languages/go/
+`;
+
+function run() {
+  // Two possible modes:
+  //   * user explicitly asked us to open; create as needed, always open
+  //   * run as a side effect of "test subset"; create as needed, and open and fail if so
+  const testSubsetGrepPath = path.join(
```
josharian, line 21: Should I dedup this by exporting a function in testUtil?

cursorless-talon-dev/src/cursorless_dev_vscode.talon
```diff
@@ -11,3 +11,8 @@ app: vscode
     user.vscode("workbench.action.tasks.terminate")
     insert("Serve")
     key(enter)
+
+^debug generate subset$:
+    user.vscode_with_plugin("workbench.action.tasks.runTask", "Generate test subset file")
+^debug edit subset$:
+    user.vscode_with_plugin("commands.openFolder", "./packages/common/src/testUtil/testSubsetGrep.properties")
```
pokey, line 18: I split this into two commands because the version that runs as a task was painfully slow on my computer. Which is fine for the initial generation, but too slow for something I do many times a day. This version is ~instant

Would be great if we could get away with one command that was fast, but I couldn't figure out how to do it 🤷‍♂️

packages/cursorless-engine/package.json
```diff
@@ -6,7 +6,9 @@
   "scripts": {
     "compile": "tsc --build",
     "watch": "tsc --build --watch",
-    "clean": "rm -rf ./out tsconfig.tsbuildinfo"
+    "clean": "rm -rf ./out tsconfig.tsbuildinfo",
+    "generate-test-subset-file": "tsx --conditions=cursorless:bundler ./src/scripts/generateTestSubsetFile/generateTestSubsetFile.ts",
```
pokey, line 10: Looks like you dropped the bundle arg we always use with `tsx`. That will cause unpredictable behaviour if you import something from another package

See
https://youtu.be/Od7BZHc7w6c for more than you probably ever wanted to know on the subject 😄

packages/cursorless-engine/src/scripts/generateTestSubsetFile/generateTestSubsetFile.ts
pokey: I'll do that in a follow-up; we should migrate all this stuff into one place; no reason to have it spread over three packages when it's all consumed in one package

packages/cursorless-engine/src/scripts/generateTestSubsetFile/generateTestSubsetFile.ts
pokey, line 1: this file simplified a bit now that it's no longer responsible for pure editing

packages/cursorless-engine/src/scripts/generateTestSubsetFile/generateTestSubsetFile.ts
```diff
@@ -0,0 +1,44 @@
...
+    "common",
+    "src",
+    "testUtil",
+    "testSubsetGrep.properties",
+  );
+
+  if (!fs.existsSync(testSubsetGrepPath)) {
+    fs.writeFileSync(testSubsetGrepPath, TEMPLATE);
+
+    child.execSync(`code ${testSubsetGrepPath}`);
```
pokey, line 35: had to make this sync in order for it to reliably open for me

pokey (approved):
Made some tweaks in d084268594bba4c98e88356bfff11cd80e309fb5.  I'm happy if you are

pokey (changes requested):
actually we should prob move stuff to correct place now

packages/cursorless-engine/src/scripts/generateTestSubsetFile/generateTestSubsetFile.ts
josharian: oops. i'm so used to things being sync by default.

packages/cursorless-engine/src/scripts/generateTestSubsetFile/generateTestSubsetFile.ts
josharian: thanks :)

packages/cursorless-engine/package.json
josharian: thanks! will watch that video after the meetup

cursorless-talon-dev/src/cursorless_dev_vscode.talon
josharian: ack

docs/contributing/CONTRIBUTING.md
```diff
@@ -60,9 +60,11 @@ Extension" or "Extension Tests".
 
 The entire test suite takes a little while to run (1-2 mins), so you may want to run just a subset of the tests.
 
-To specify the set of tests to run, run the VSCode task "Edit test subset file". To do this, choose "Tasks: Run Task" from the command pallete. Or if you have the cursorless-talon-dev files installed, say "edit test subset".
+To specify the set of tests to run, say `"debug generate subset"`, or if you haven't installed the cursorless-talon-dev files, run the VSCode task "Generate test subset file". To do this, choose "Tasks: Run Task" from the command pallete.
 
-To run that subset of tests, use the "Run test subset" launch config instead of the usual "Extension Tests".
+To run the given subset of tests, use the "Run test subset" launch config instead of the usual "Extension Tests".
```
josharian, line 63: s/given/specified/ maybe

josharian (commented):
thank you so much for getting this to the finish line for me

josharian:
ugh, github didn't thread my comments nicely.

josharian:
> if that's not in the next ~2 hours I may do it because I can no longer live without this feature now that it's so close 

i have lots of errands today so thrilled to leave this to you. or i'll do it monday if not.

pokey (approved):
Ok I think this is good to go.  @AndreasArvidsson I'll give you a chance to have a look / make sure it works as expected for you and then let's ship it!

AndreasArvidsson:
Works fine but I have two things I would like to fix in this pr if possible
1. Rename and reorder launch options `Runs tests subset` => `Extension tests subset` and move it directly below `Extension tests`.
2. Update the parsing of the property's file to replace `/` with `\\` on windows. The way it is right now you have to think about which platform you are on and I so often forget this and wonder why the tests are not running.

pokey:
> 2. Update the parsing of the property's file to replace `/` with `\\` on windows. The way it is right now you have to think about which platform you are on and I so often forget this and wonder why the tests are not running.

I'm not sure parsing the properties file is the right place to do that. These are just regexes on test names, so in principle the `/` could mean something other than a path.  Maybe we should normalize those [when we construct the test name](https://github.com/cursorless-dev/cursorless/blob/63ea8e9076828b72ce198b58fcb060bd49f06ddc/packages/common/src/testUtil/getFixturePaths.ts#L31)?

AndreasArvidsson:
> > 2. Update the parsing of the property's file to replace `/` with `\\` on windows. The way it is right now you have to think about which platform you are on and I so often forget this and wonder why the tests are not running.
> 
> I'm not sure parsing the properties file is the right place to do that. These are just regexes on test names, so in principle the `/` could mean something other than a path. Maybe we should normalize those [when we construct the test name](https://github.com/cursorless-dev/cursorless/blob/63ea8e9076828b72ce198b58fcb060bd49f06ddc/packages/common/src/testUtil/getFixturePaths.ts#L31)?

That sounds reasonable

pokey:
So maybe that should be a follow-up? It's not really anywhere near anything this PR touches

pokey:
> 1. Rename and reorder launch options `Runs tests subset` => `Extension tests subset` and move it directly below `Extension tests`.

On it...

AndreasArvidsson:
> So maybe that should be a follow-up? It's not really anywhere near anything this PR touches, and it's not a problem introduced by this PR, right?

I can live with that

pokey:
queuing up for merge...

PR #1692, opened 2023-07-28
stop early when there are no candidate hats for a grapheme

josharian:
No functional changes.

However it prevents a bunch of pointless runs through the
scoring functions with an empty set of candidates.

(All the early stopping conditions check for a
unique winner, not the empty set.
Those aren't worth changing, when we can trivially
avoid calling the function at all.)

The main benefit is reduced noise when logging
inside the scoring functions when debugging.

## Checklist


pokey (commented):
I'd be tempted to short circuit at the start of [`maxByFirstDiffering`](https://github.com/cursorless-dev/cursorless/blob/5ec99720b8a3e10973d8c263d5ed06d3c10ad730/packages/cursorless-engine/src/util/allocateHats/maxByFirstDiffering.ts#L19) instead to avoid added complexity to this function.  The docstring of that function already says it returns `undefined` if list is empty, it just does so today in a roundabout way

josharian:
> I'd be tempted to short circuit at the start of [maxByFirstDiffering](https://github.com/cursorless-dev/cursorless/blob/5ec99720b8a3e10973d8c263d5ed06d3c10ad730/packages/cursorless-engine/src/util/allocateHats/maxByFirstDiffering.ts#L19)

yeah, I wavered back and forth on that exact point. done.

pokey (approved):
👍

PR #1693, opened 2023-07-28
fix inverted score in hatOldTokenRank

josharian:
hatOldTokenRank decides which token, if any, to steal a hat from.
The goal is to steal a hat from the token that is least important.
As written, the code steals the hat from the token that is most important,
but only when there are no completely free hats available.
Fix that.

This does not happen often in practice.
This commit will have minimal impact on user experience.
The main benefit is avoiding future confusion over the code,
and being more robust if surrounding parts of the code change.

## Checklist


josharian:
See https://youtu.be/czsNj4lLagU for a little demo

pokey (approved):
sold!  very nice catch 🙌.  youtube video was super helpful

Issue #1549, opened 2023-06-22
"short paint" (boundedNonWhitespaceSequence?) is undocumented


jmegner:
"short paint" (boundedNonWhitespaceSequence?) is undocumented; even a very short blurb in the reference would be great

AndreasArvidsson:
Duplicate of #839

Issue #839, opened 2022-07-07
"paint" and "short paint" targets are missing from documentation


paul-schaaf:

AndrewDant:
@pokey you can assign this one to me. I won't be taking on any more big ones for a few weeks but some stuff like this I can handle.

PR #1698, opened 2023-07-31
document short paint

jmegner:
For #839 and #1549.

Added a section about "short paint" in main reference page.  Improved explanation in cheatsheet.

## Checklist

- [X] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [X] I have not broken the cheatsheet

### DIFF
--- a/docs/user/README.md
+++ b/docs/user/README.md
@@ -311,6 +311,25 @@ foo.bar baz|bongo
 
 Saying `"every paint"` would select `foo.bar` and `baz|bongo`.
 
+##### `"short paint"`
+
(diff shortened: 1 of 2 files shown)


pokey (approved):
Looks good.  Made a couple tweaks; merge if happy

We should revisit the way we talk about scopes in the docs at some point, because the language describing these examples is very much rooted in the old world where scopes and containing scope modifiers are one and the same.  Probably a visualization would be the easiest, but we can wait  until `"short paint"` has been upgraded to next-gen scope handler

jmegner:
Looks good to me.  I don't have to do anything to officially accept your changes, right?

PR #1712, opened 2023-08-01
Fix E2BIG error in pre-commit

pokey:
## Checklist


AndreasArvidsson (approved):

auscompgeek:
> annoying and duplicative

You can avoid the duplication by making the package.json script call `pre-commit run` instead

pokey:
I don't think that works because we want to support additional args and no args ets

PR #1724, opened 2023-08-03
migrate Go scopes map and list to treesitter

josharian:
## Checklist

### DIFF
--- a/queries/go.scm
+++ b/queries/go.scm
@@ -35,3 +35,150 @@
 ] @string @textFragment
 
 (comment) @comment @textFragment
+
+;; What should map and list refer to in Go programs?
+;;
+;; The obvious answer is that map should refer to map and struct composite literals,
+;; and that list should refer to slice and array composite literals.
+;;
+;; There are two problems with this answer.
+;;
+;;   * The type of a composite literal is a semantic, not a syntactic property of a program.
+;;       - What is the type of T{1: 2}? It could be array, map, or slice.
+;;       - What about T{a: 1}? It could be map or struct.
+;;       - What about T{1, 2}? It could be struct, array, or slice.
+;;     Cursorless only has syntactic information available to it.
+;;
+;;   * The user might not know the type either. With a named type, the type definition might be far away.
+;;     Or it might just be offscreen. Either way, the user needs to be able to make a decision about
+;;     what scope to use using only locally available, syntactic information.
+;;     Note that this also means that has-a predicates work better than has-no predicates.
+;;     The user can locally confirm that there is a keyed element.
+;;     She cannot confirm locally that there is no keyed element; it might just not be visible.
+;;
+;; Combining all these constraints suggests the following simple rules:
+;;
+;;   * If there is a keyed element present, then it is a map.
+;;   * If there is a non-keyed element present, then it is a list.
+;;   * If there are both or neither, then it is both a map and a list.
+;;
+;; Conveniently, this is also simple to implement.
+;;
+;; This guarantees that a user always knows how to refer to any composite literal.
+;; There are cases in which being overgenerous in matching is not ideal,
+;; but they are rarer, so let's optimize for the common case.
+;; Mixed keyed and non-keyed elements are also rare in practice.
+;; The main ambiguity is with {}, but there's little we can do about that.
+;;
+;; Go users also expect that the map and list scopes will include the type definition,
+;; as well as any & before the type. (Strictly speaking it is not part of the literal,
+;; but that's not how most humans think about it.)
+;;
+;; If you are considering changing the map and list scopes, take a look at the examples in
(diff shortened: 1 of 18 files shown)


josharian:
The reason that I put a Go subdirectory in the playground is that the toolchain is not happy with them sitting next to c++ files without explicitly declaring it.

However, it also gives a nice opportunity to keep things organized. And when the new test format rolls around, some of these playground files can be used to generate scope tests right away.

queries/go.scm
```diff
@@ -35,3 +35,104 @@
...
+;;
+;; If you are considering changing the map and list scopes, take a look at the examples in
+;; data/playground/go/maps_and_lists.go, which cover a fairly wide variety of cases.
+
+;; maps
+
+;; &T{a: 1}
+(unary_expression
+  operator: "&"
+  (composite_literal body: (literal_value (keyed_element)))
```
pokey, line 85: I would love to have an autoformatter for our `.scm` files, but unfortunately we don't yet.  In the meantime, we will unfortunately have to operate via PR comments 😖.  Have a look at our other files for examples of the convention, which is that any node with children should break onto multiple lines.  Eg
```suggestion
  (composite_literal
    body: (literal_value
      (keyed_element)
    )
  )
```

queries/go.scm
```diff
@@ -35,3 +35,104 @@
...
+) @map
+
+;; T{a: 1}
+(
+  (composite_literal body: (literal_value (keyed_element))) @_comp_lit
```
pokey, line 90: as above re formatting

Also, I would just replace `@_comp_lit` with `@map`, so you can remove the outermost annotation

queries/go.scm
```diff
@@ -35,3 +35,104 @@
...
+  (#not-parent-type? @_comp_lit unary_expression)
```
pokey, line 91: what happens if the `composite_literal` node has a parent of type `unary_expression` whose operator is not `&`?  then neither this pattern nor the other will match. Is that desirable?  If the operator is always `&` for a `unary_expression` containing a `composite_literal`, then why specify it above?

queries/go.scm
```diff
@@ -35,3 +35,104 @@
...
+) @map
+
+;; {a: 1}
+(
+  (literal_value (keyed_element)) @_lit_val
+  (#not-parent-type? @_lit_val composite_literal)
+) @map
```
pokey, line 98: same comments above re removing  `@_lit_val` and reformatting

also, optional parents would really make things more compact 😅.  I wish tree-sitter patterns supported that

pokey (commented):
looking good! unfortunate there needs to be so much duplication, but I don't see a way around it either.  In addition to the comments inline,

- [ ] please add a few more tests, eg with no type annotation, combining keyed with non-keyed, etc

queries/go.scm
josharian: no worries. Done.

queries/go.scm
josharian: oh nice. Done.

queries/go.scm
josharian: the only valid operator is `&`. I specified it for two reasons: a form of documentation, and as future proofing in case the treesitter grammar decides that it is happy to spit out invalid unary expressions there. if you feel strongly I am happy to remove it, but I think it improves readability.

queries/go.scm
josharian: would it ever. I wish a lot of things from treesitter patterns. 😆

josharian:
thanks for the review! PTAL.

queries/go.scm
```diff
@@ -35,3 +35,146 @@
...
+  (composite_literal
+    body: (literal_value
+      .
+      "{"
+      .
+      "}"
+      .
+    )
+  )
+) @list @map
```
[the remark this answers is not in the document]
josharian: Nice catch, and definitely something to think about in the future as I go. In this particular case I think the problem is going to be extremely rare, because  I would wager that comments inside `{}`  Are all but nonexistent. Given that the fix is very verbose, I think I'd just like to let leave this one be.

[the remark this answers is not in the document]
josharian: I would certainly be happy to add a comment to that effect if you'd like. ( Or fix it if you care deeply.)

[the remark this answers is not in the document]
josharian: I guess one kind of hacky implementation would be `(#count-non-comment-children 2)`.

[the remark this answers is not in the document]
pokey: I'm fine with a one-line comment indicating it's not handled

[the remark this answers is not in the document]
josharian: It's three short lines. You found my kryptonite :)

pokey (approved):
🙌

PR #1732, opened 2023-08-06
Use simple syntax node for mutable query captures

AndreasArvidsson:
## Checklist

### DIFF
--- a/packages/cursorless-engine/src/languages/TreeSitterQuery/QueryCapture.ts
+++ b/packages/cursorless-engine/src/languages/TreeSitterQuery/QueryCapture.ts
@@ -1,5 +1,19 @@
 import { Range } from "@cursorless/common";
-import { SyntaxNode } from "web-tree-sitter";
+import { Point } from "web-tree-sitter";
+
+/**
+ * Simple representation of the tree sitter syntax node. Used by
+ * {@link MutableQueryCapture} to avoid using range/text and other mutable
+ * parameters directly from the node.
+ */
+export interface SimpleSyntaxNode {
+  readonly id: number;
+  readonly type: string;
+  readonly startPosition: Point;
+  readonly endPosition: Point;
+  readonly parent: SimpleSyntaxNode | null;
+  readonly children: Array<SimpleSyntaxNode>;
+}
 
 /**
  * A capture of a query pattern against a syntax tree.  Often corresponds to a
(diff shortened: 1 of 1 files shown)


packages/cursorless-engine/src/languages/TreeSitterQuery/QueryCapture.ts
```diff
@@ -1,5 +1,19 @@
...
+/**
+ * Simple representation of the tree sitter syntax node. Used by
+ * {@link MutableQueryCapture} to avoid using range/text and other mutable
+ * parameters directly from the node.
+ */
+export interface SimpleSyntaxNode {
+  readonly id: number;
+  readonly type: string;
+  readonly startPosition: Point;
+  readonly endPosition: Point;
```
pokey, line 13: This seems like a loophole. Do we need these?

pokey (commented):
Looks good with one comment

packages/cursorless-engine/src/languages/TreeSitterQuery/QueryCapture.ts
AndreasArvidsson: Unfortunately we do for the child range predicate. You could use it as a loophole but in the only case we actually use it we use it on the children and not the mutated node itself.
https://github.com/cursorless-dev/cursorless/blob/f84512dcc9a3fd8573c544ccef7bc7b72e472503/packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts#L87-L90

pokey: Hmm. I wonder if we could remove that predicate entirely by using a syntax like `.child0`. Then combined with `.start` / `.startOf` you can recover all the same functionality

We'd need to figure out what to do for `-1` tho. Eg `.lastChild1` or something 🤔

Alternately we could exclude `startPoint` and `endPoint` from the node itself and only include it on parent / children. There'd still be a bit of a loophole because you could say `.children[0].parent` but at that point it would be pretty obvious in a PR review. I guess we could throw an exception from the `startPoint` accessor of the main node. Idk

I guess the main thing we're trying to prevent is doing it without thinking, as you and I have both done at this point I believe 😅

AndreasArvidsson: I think omitting them from the current mutable node is the simplest solution and should prevent us from accidentally doing stuff like that. If you on purpose go to the parent and then down to the child again you have not accidentally done anything.

Updated
https://github.com/cursorless-dev/cursorless/blob/6b8c3221edb7ee507db999d6935df4111868e89c/packages/cursorless-engine/src/languages/TreeSitterQuery/QueryCapture.ts#L58

pokey (commented):
looks good!

pokey (approved):
looks good!

PR #1738, opened 2023-08-07
Move default spoken forms to json file

### LINKED ISSUE #1737: Move default spoken forms to json single file
Today all the default spoken forms are scattered across our Talon python implementation. Moving all of these to a single json file would give us a single source of truth. A next step is to actually generate this file using meta updater from our typescript definitions.

AndreasArvidsson:
` 2713 passing (6s)`

Fixes #1737

## Checklist


cursorless-talon/src/spoken_forms.json
```diff
@@ -0,0 +1,260 @@
+{
+  "actions": {
+    "simple_action": {
```
pokey, line 3: Huh. A bit odd that Typescript needs to know about the list names, in snake case, that we expect to use Talon side 🤔

pokey: I guess maybe this is a reasonable first step, and we can revisit when we actually gen from typescript?

cursorless-talon/src/spoken_forms.json
```diff
@@ -0,0 +1,260 @@
...
+      "bottom": "scrollToBottom",
```
pokey, line 4: I'd be tempted to map the other direction (id to spoken form), as that's how we prefer to represent things typescript-side, but again, maybe a reasonable first step to keep things closer to how they are now?

cursorless-talon/src/spoken_forms.json
```diff
@@ -0,0 +1,260 @@
...
+    "swap_action": { "swap": "swapTargets" },
+    "wrap_action": { "wrap": "wrapWithPairedDelimiter", "repack": "rewrap" },
+    "insert_snippet_action": { "snippet": "insertSnippet" },
+    "reformat_action": { "format": "applyFormatter" }
+  },
+  "targetConnectives": {
+    "range_connective": {
+      "between": "rangeExclusive",
+      "past": "rangeInclusive",
+      "-": "rangeExcludingStart",
```
pokey, line 63: here's an example where it would be better to map the other direction, so that we can have no default value for multiple spoken forms

cursorless-talon/src/spoken_forms.json
pokey: I'd also be tempted to map to a list of spoken forms, and to allow richer data, such as a default spoken form that is disabled by default, without using shorthand such as `-foo`, but again, maybe for further work

cursorless-talon/src/spoken_forms.json
```diff
@@ -0,0 +1,260 @@
...
+  },
+  "scopeVisualizer": {
+    "show_scope_visualizer": { "visualize": "showScopeVisualizer" },
+    "hide_scope_visualizer": { "visualize nothing": "hideScopeVisualizer" },
+    "visualization_type": {
+      "removal": "removal",
+      "iteration": "iteration"
+    }
+  },
+  "experimental.actions": {
```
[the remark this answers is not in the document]
AndreasArvidsson: Correct. As I said on Slack I tried to change as little as possible. This just moved things from python to json.

cursorless-talon/src/spoken_forms.json
AndreasArvidsson: Kiss for now :D. Format change could be reasonable when we do meta updater.

cursorless-talon/src/spoken_forms.json
AndreasArvidsson: No just iterate keys and ignore the value. Python controls the format and ts just updates what's already there was my idea.

cursorless-talon/src/spoken_forms.json
```diff
@@ -0,0 +1,260 @@
...
+    "escaped box": "escapedSquareBrackets",
+    "quad": "doubleQuotes",
+    "round": "parentheses",
+    "skis": "backtickQuotes",
+    "void": "whitespace",
+    "box": "squareBrackets",
+    "twin": "singleQuotes",
+    "pair": "any"
+  },
+  "hatColors": {
```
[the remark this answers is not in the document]
AndreasArvidsson: Because they need special treatment in python. Same with matching pairs.

cursorless-talon/src/spoken_forms.py
```diff
@@ -0,0 +1,97 @@
...
+
+from .csv_overrides import SPOKEN_FORM_HEADER, init_csv_and_watch_changes
+from .marks.mark import init_marks
+from .paired_delimiter import paired_delimiter_spoken_form_defaults
+
+JSON_FILE = Path(__file__).parent / "spoken_forms.json"
+
+
+def on_ready() -> None:
+    with open(JSON_FILE) as file:
```
pokey, line 14: I wonder if we want to use talon resource tracker or something. During development of our cursorless-talon files, we won't trigger a reload of this file if it changes, so not ideal DX.  I remember there being some gotchas on resource tracker that were fixed in knausj in last 8 months tho so may be worth an issue search if we decide it's worth the effort here

pokey (commented):
Looks good with some minor comments; nice incremental change

cursorless-talon/src/spoken_forms.py
AndreasArvidsson: Good call. I will look into that. It requires 0.4 I believe, but I think its reasonable to expect people to update talon and cursorless.

cursorless-talon/src/spoken_forms.json
AndreasArvidsson: If we dont want a spoken form we could just remove the entire item?

AndreasArvidsson:
> A couple more high-level thoughts:
> 
>     * this setup makes things even more magical, because you can't even see where the spoken forms are coming from in any given file.  I guess that's kind of a natural trade-off of moving them all to one place 🤷‍♂️
> 
>     * we have no way of putting a warning at the top of our json file not to modify the spoken forms

The json file has the list name, so not to hard to connect. We still have captures that contains captures and so on so a very clear name should be our goal. Just because something was in the same file didnt always help before.

Do we really need that? How many people modify our cursorless files and they should find the csv first hopefully.

cursorless-talon/src/spoken_forms.py
pokey: Might be worth a look at the community 0.4 update PR to see if they're trying to detect old talon version / holding off on merge. I guess we could just detect old talon and fall back; prob not too hard

pokey:
Possible solutions for both problems:

- add a dummy field to top of the json like `"NOTE FOR USERS": "Please don't edit this json file; see https://www.cursorless.org/docs/user/customization/`
- add a comment to the top of all of our implementation files, pointing maintainers to json file and users to https://www.cursorless.org/docs/user/customization/

I don't feel strongly, but might be nice UX / DX

cursorless-talon/src/spoken_forms.json
pokey: Then it won't show in csv

cursorless-talon/src/spoken_forms.json
pokey: Huh interesting idea. We can discuss in follow-up to this PR

cursorless-talon/src/spoken_forms.py
AndreasArvidsson: Pretty sure the community pr doesn't do that but we could of course.

AndreasArvidsson: File watcher added. I decided to go with the old one I'm let community the merge first. Once community require the latest version of Talon I think Cursorless is free to do the same.

cursorless-talon/src/spoken_forms.json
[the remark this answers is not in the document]
AndreasArvidsson: Changed my mind and updated to the same formats

cursorless-talon/src/spoken_forms.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+    init_marks(
+        spoken_forms["hats"]["hat_color"],
+        spoken_forms["hats"]["hat_shape"],
+    )
+
+
+def on_watch(path, flags):
+    if JSON_FILE.match(path):
+        update()
```
pokey, line 99: Should probably un watch all the csv files before we rerun the csv watch functions, no? Our csv watcher function returns an unsubscribe function so should be fairly straightforward to just use that

pokey: I would assume that `fs.watch` automatically de-registers when you touch a python file, no?

pokey: I agree it may be overkill, but it's also a bit weird that we're careful to return unsubscribe functions from our csv watcher and then completely ignore them 🤔

cursorless-talon/src/spoken_forms.json
```diff
@@ -0,0 +1,263 @@
...
+    }
+  },
+  "insertionSnippets": {
+    "insertion_snippet_no_phrase": {
+      "if": "ifStatement",
+      "if else": "ifElseStatement",
+      "try": "tryCatchStatement"
+    }
+  },
+  "insertionSnippetsSinglePhrase": {
```
pokey, line 216: if these are just names of csvs, why not make that even clearer and just use the csv name as the key, eg `insertion_snippet_single_phrase.csv`, and even for the experimental ones just go eg `experimental/miscellaneous.csv`

pokey: I would argue to keep it simple.  This should either cater completely to Python or not at all.  The in-between setup means it's neither intuitive coming from the TS or the Python side.  But let's discuss at meet up in next half hour

AndreasArvidsson:
Fixes applied from today's meet up
1. Added top item with warning message about changing the file
2. Change top level key to be full filename

pokey (approved):
looks good to me!

PR #1745, opened 2023-08-08
Bump docusaurus to `3.0.0-alpha.0`

pokey:
We have some security warnings on a couple packages that are dependencies of docusaurus, so I decided to upgrade to the latest to remove those dependencies. Docusaurus uses their nightly version on their own docs, and things seem to work for us on this version with no issue, so I think it's ok. In addition, we were violating their peer dependencies by using react 18; we now no longer need to do that.  Also, this version allows us to use mdx v2, which has a lot of niceties that we'll probably want as we do our big docs refactor (https://github.com/cursorless-dev/cursorless/issues/867)

- depends on https://github.com/cursorless-dev/cursorless/pull/1750 to fix typedoc failure

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,14 +41,11 @@
   },
   "pnpm": {
     "patchedDependencies": {
-      "@docusaurus/theme-search-algolia@2.3.1": "patches/@docusaurus__theme-search-algolia@2.3.1.patch"
+      "@docusaurus/theme-search-algolia@3.0.0-alpha.0": "patches/@docusaurus__theme-search-algolia@2.3.1.patch"
     },
     "peerDependencyRules": {
       "ignoreMissing": [
-        "@docusaurus/core",
-        "@docusaurus/mdx-loader",
-        "@docusaurus/plugin-content-blog",
-        "@docusaurus/preset-classic"
+        "react-json-view"
       ],
       "allowedVersions": {
         "react": "18",
--- a/packages/meta-updater/src/updateTSConfig.ts
+++ b/packages/meta-updater/src/updateTSConfig.ts
@@ -5,6 +5,7 @@ import exists from "path-exists";
 import { TsConfigJson } from "type-fest";
 import { toPosixPath } from "./toPosixPath";
 import { Context } from "./Context";
+import { uniq } from "lodash";
 
 /**
  * Given a tsconfig.json, update it to match our conventions.  This function is
@@ -76,9 +77,7 @@ export async function updateTSConfig(
 
   return {
     ...input,
-    extends: toPosixPath(
-      path.join(pathFromPackageToRoot, "tsconfig.base.json"),
-    ),
+    extends: getExtends(pathFromPackageToRoot, input.extends),
     compilerOptions: {
       ...(input.compilerOptions ?? {}),
       rootDir: "src",
@@ -98,4 +97,24 @@ export async function updateTSConfig(
       toPosixPath(path.join(pathFromPackageToRoot, "typings", "**/*.d.ts")),
     ],
   };
+}
+
+function getExtends(
+  pathFromPackageToRoot: string,
+  inputExtends: string | string[] | undefined,
+) {
+  let extendsList =
+    inputExtends == null
+      ? []
+      : Array.isArray(inputExtends)
+      ? [...inputExtends]
+      : [inputExtends];
+
+  extendsList.push(
+    toPosixPath(path.join(pathFromPackageToRoot, "tsconfig.base.json")),
+  );
+
+  extendsList = uniq(extendsList);
+
+  return extendsList.length === 1 ? extendsList[0] : extendsList;
 }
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -41,14 +41,11 @@
...
-      "@docusaurus/theme-search-algolia@2.3.1": "patches/@docusaurus__theme-search-algolia@2.3.1.patch"
+      "@docusaurus/theme-search-algolia@3.0.0-alpha.0": "patches/@docusaurus__theme-search-algolia@2.3.1.patch"
     },
     "peerDependencyRules": {
       "ignoreMissing": [
-        "@docusaurus/core",
-        "@docusaurus/mdx-loader",
-        "@docusaurus/plugin-content-blog",
-        "@docusaurus/preset-classic"
+        "react-json-view"
```
pokey, line 48: We still need this peer dependency override until https://github.com/facebook/docusaurus/pull/9116 is released

package.json
```diff
@@ -41,14 +41,11 @@
   },
   "pnpm": {
     "patchedDependencies": {
-      "@docusaurus/theme-search-algolia@2.3.1": "patches/@docusaurus__theme-search-algolia@2.3.1.patch"
+      "@docusaurus/theme-search-algolia@3.0.0-alpha.0": "patches/@docusaurus__theme-search-algolia@2.3.1.patch"
```
pokey, line 44: this patch seemed to apply cleanly, and still is having the intended effect in local testing, so that's a relief 😅

pnpm-lock.yaml
pokey, line 1: @auscompgeek anything jump out at you here?

packages/meta-updater/src/updateTSConfig.ts
```diff
@@ -76,9 +77,7 @@ export async function updateTSConfig(
 
   return {
     ...input,
-    extends: toPosixPath(
-      path.join(pathFromPackageToRoot, "tsconfig.base.json"),
-    ),
+    extends: getExtends(pathFromPackageToRoot, input.extends),
```
pokey, line 80: need this because we are now extending docusaurus tsconfig in our docs site

AndreasArvidsson (approved):

pokey:
ok @auscompgeek I've merged your dedupe. Fwiw I just ran a dedupe on this branch and did a soft reset over your dedupe; didn't feel like fighting with merge conflicts in pnpm lockfile 😅

auscompgeek:
Yep makes sense, I think the end result should be the same however you slice it 😄

pnpm-lock.yaml
```diff
@@ -11692,67 +11947,413 @@ packages:
...
+
+  /mem@6.1.1:
+    resolution: {integrity: sha512-Ci6bIfq/UgcxPTYa8dQQ5FY3BzKkT894bwXWXxC/zqs0XgMO2cT20CGkOqda7gZNkmK5VP4x89IGZ6K7hfbn3Q==}
+    engines: {node: '>=8'}
+    dependencies:
+      map-age-cleaner: 0.1.3
+      mimic-fn: 3.1.0
+    dev: true
+
+  /mem@8.1.1:
```
auscompgeek, line 11963: wat

... oh, github diff didn't notice these already existed?

auscompgeek (approved):
Nothing that leapt out as majorly alarming to me. Left some comments on the more curious new deps coming in.

A few cases where there's -1 +1 versions of a dep, largely different semver major versions (judging by the node engine constraints on them, I'm guessing ES6 conversions or something). A few cases where we have one fewer version of an existing dep though, which is good to see.

pnpm-lock.yaml
```diff
@@ -7115,6 +7100,11 @@ packages:
     resolution: {integrity: sha512-Gmy6FhYlCY7uOElZUSbxo2UCDH8owEk996gkbrpsgGtrJLM3J7jGxl9Ic7Qwwj4ivOE5AWZWRMecDdF7hqGjFA==}
     engines: {node: '>=10'}
 
+  /camelcase@7.0.1:
```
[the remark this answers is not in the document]
pokey: If true, then the fact that we pull in 3 major versions of a one-line function is pretty much the npm ecosystem in a nutshell 😂

pnpm-lock.yaml
```diff
@@ -7716,6 +7713,14 @@ packages:
   /crypto-random-string@2.0.0:
     resolution: {integrity: sha512-v1plID3y9r/lPhviJ1wrXpLeyUIGAZ2SHNYTEapm7/8A9nLPoyvVp3RK/EPFqn5kEznyWgYZNsRtYYIWbuG8KA==}
     engines: {node: '>=8'}
+    dev: true
+
+  /crypto-random-string@4.0.0:
+    resolution: {integrity: sha512-x8dy3RnvYdlUcPOjkEHqozhiwzKNSq7GcPuXFbnyMOCHxX8V3OgIg/pYuabl2sbUPfIJaeAQB7PMOK8DFIdoRA==}
+    engines: {node: '>=12'}
+    dependencies:
+      type-fest: 1.4.0
```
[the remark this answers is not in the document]
pokey: Ha yeah that's...interesting. Should type-fest every be anything other than dev dependency?

PR #1746, opened 2023-08-08
Run `pnpm dedupe`

auscompgeek:
#1745 is almost impossible to review :sweat_smile:

I'd suggest running `pnpm dedupe` on the PR but it turns out `main` has a whole bunch of duplicate transitive deps too, so running it against `main` first.

## Checklist

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -270,7 +270,7 @@ importers:
         version: 13.2.3(eslint@8.38.0)(typescript@5.0.3)
       next:
         specifier: 13.2.3
-        version: 13.2.3(@babel/core@7.21.0)(react-dom@18.2.0)(react@18.2.0)
+        version: 13.2.3(@babel/core@7.21.4)(react-dom@18.2.0)(react@18.2.0)
       react:
         specifier: ^18.2.0
         version: 18.2.0
@@ -456,7 +456,7 @@ importers:
         version: 11.1.0
       glob:
         specifier: ^7.1.7
-        version: 7.2.3
+        version: 7.1.7
       mocha:
         specifier: ^8.1.3
         version: 8.4.0
@@ -548,7 +548,7 @@ importers:
         version: link:../common
       glob:
         specifier: ^7.1.7
-        version: 7.2.3
+        version: 7.1.7
     devDependencies:
       '@types/glob':
         specifier: ^7.1.3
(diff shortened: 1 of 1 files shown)


pokey (approved):

pnpm-lock.yaml
```diff
@@ -8117,6 +7991,7 @@ packages:
 
   /debuglog@1.0.1:
     resolution: {integrity: sha512-syBZ+rnAK3EgMsH2aYEOLUW7mZSY9Gb+0wUMCFsZvcmiz+HigA0LOcq/HoQqVuGG+EKykunc7QG2bzrponfaSw==}
+    deprecated: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
```
auscompgeek, line 7994: Looks like this is a dependency of [`readdir-scoped-modules@1.1.0`](https://www.npmjs.com/package/readdir-scoped-modules) which in itself is also deprecated.

> This functionality has been moved to @npmcli/fs

pnpm-lock.yaml
```diff
@@ -15932,6 +15786,7 @@ packages:
   /typedoc-plugin-resolve-crossmodule-references@0.3.3(typedoc@0.23.28):
     resolution: {integrity: sha512-ZWWBy2WR8z9a6iXYGlyB3KrpV+JDdZv1mndYU6Eh6mInrfMCrQJi3Y5K9ihMBfuaBGB//le1nEmQLgzU3IO+dw==}
     engines: {node: '>=14'}
+    deprecated: Upgrade to typedoc >= 0.24 and remove typedoc-plugin-resolve-crossmodule-references from your dependencies
```
auscompgeek, line 15789: New deprecation for a direct dependency :eyes:

pokey: yeah typedoc has been a bit of a nightmare tbh.  i'm tempted to switch off our api docs, as I don't think anyone is using them anyways 🤔

auscompgeek:
> I thought pnpm deduped by default nowadays?

:shrug: I don't think it'll try to update existing locked transitive dependencies, but will reuse existing resolutions. Both pnpm and [yarn behave this way](https://yarnpkg.com/cli/dedupe/) I think.

pnpm-lock.yaml
pokey: appears to be mainly coming from webpack, so maybe we bump webpack and see if that fixes?

auscompgeek: So it seems! Looks like webpack is a dependency of docusaurus too, not just a direct `devDependency` for the cheatsheet.

pokey: https://github.com/cursorless-dev/cursorless/issues/1748

pnpm-lock.yaml
pokey: https://github.com/cursorless-dev/cursorless/issues/1749

PR #1750, opened 2023-08-08
Stop publishing API docs

### LINKED ISSUE #1749: Remove our api docs
Our [Cursorless api docs](https://www.cursorless.org/docs/contributing/api/) have been a bit of a nightmare to maintain, and don't seem to be used by any maintainers.  I believe the fundamental problem is that [typedoc](https://typedoc.org/) is designed to be used for public-facing apis, but we're using it to document internal APIs.  That leads to us having to use some strange flags, so we're constantly fighting with the typedoc library as it gets updated

At the point where Cursorless publishes npm packages, such as the engine, we should revisit, but we'd probably use a more standard typedoc setup where we just document our public-facing api.

- [x] Close https://github.com/cursorless-dev/cursorless/issues/1304 as not planned
- [x] Close https://github.com/cursorless-dev/cursorless/issues/601 as not planned

pokey:
- Fixes #1749
- Will do a `pnpm remove` on all our typedoc dependencies once #1745 is merged so we don't get massive merge conflicts on our `pnpm.lock` 😅
- Our website deploy has gone from more than 12 minutes to 1 min  20 sec 🙌

## Checklist


packages/cursorless-engine/src/testCaseRecorder/RecordTestCaseCommandArg.ts
pokey, line 1: Moved this into its own file to make it easier to link to

packages/cursorless-engine/src/testCaseRecorder/RecordTestCaseCommandArg.ts
```diff
@@ -0,0 +1,52 @@
+import { ExtraSnapshotField } from "@cursorless/common";
+
+/**
+ * This is the type of the `recordTestCase` command's argument. It is used to
+ * specify the options for recording a test case.
+ */
+export interface RecordTestCaseCommandArg {
```
auscompgeek, line 7: I've seen these referred to as "`options` objects", which I typically shorten to `opts`. https://stackoverflow.com/questions/32948475/argument-vs-options-in-typescript

Perhaps options would be a better name than `arg` if it's being exposed?

AndreasArvidsson (approved):

PR #1757, opened 2023-08-09
Try to be fancy with csv

pokey:
## Checklist

### DIFF
--- a/cursorless-talon/src/spoken_forms.py
+++ b/cursorless-talon/src/spoken_forms.py
@@ -24,7 +24,7 @@ R = TypeVar("R")
 
 def get_defaults_from_file(
     f: Callable[Concatenate[str, dict[str, dict[str, str]], P], R],
-) -> Callable[Concatenate[str, P], R]:
+):
     with open(JSON_FILE) as file:
         spoken_forms = json.load(file)
 
--- a/cursorless-talon/src/spoken_forms.py
+++ b/cursorless-talon/src/spoken_forms.py
@@ -1,6 +1,6 @@
 import json
 from pathlib import Path
-from typing import Callable
+from typing import Callable, Concatenate, ParamSpec, TypeVar
 
 from talon import app, fs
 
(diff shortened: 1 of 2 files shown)


pokey:
@AndreasArvidsson I was mostly just curious if this was possible with Python's type system.  Feel free to reject if you feel it's too magical 😄.  I think it makes it a bit clearer which custom flags each csv is using, but the wrapper function is a bit of a doozy 😅

cursorless-talon/src/spoken_forms.py
```diff
@@ -18,58 +18,68 @@ def watch_file(spoken_forms: dict, filename: str) -> Callable:
     )
 
 
+P = ParamSpec("P")
+R = TypeVar("R")
+
+
+def get_defaults_from_file(
```
AndreasArvidsson, line 25: Why not just call this function `watch_csv` and just internally use `init_csv_and_watch_changes` instead of the callable argument?

pokey: I wanted the `watch_csv` function to mirror the kwargs from `init_csv_and_watch_changes`, excluding `default_values`.  If you know of a way to do that without this wrapper function then I am all for it 😅

pokey: I just fixed a bug I introduced during some refactoring so pull if you're messing with it

AndreasArvidsson: That part is fine. I'm not quite sure what passing the argument `init_csv_and_watch_changes` in gives us?

AndreasArvidsson: If it has to be this complicated I'm still fine with it.

pokey: The only reason to have the argument is that I can destructure the type in order to strip off the first two args from `init_csv_and_watch_changes`. I don't know any other way to do that

AndreasArvidsson: I would probably just have copied the arguments personally instead of all this type stuff, but I think the result is really nice.

pokey: Ha yeah thought about that 😄. Maybe that's a better solution? We still should get a type error if we mess something up so I don't think the duplication has many drawbacks. I'm worried this code is quite hard to read

pokey: The advantage of this approach I guess is it mirrors the docstring and we see all the kwargs. Maybe I just add a docstring to the wrapper

AndreasArvidsson: Go with either direction. I'm happy

pokey:
ok added docstring; merging into your PR

cursorless-talon/src/spoken_forms.py
```diff
@@ -1,6 +1,6 @@
 import json
 from pathlib import Path
-from typing import Callable
+from typing import Callable, Concatenate, ParamSpec, TypeVar
```
auscompgeek, line 3: ParamSpec and Concatenate were new in Python 3.10. Did we intentionally drop support for Talon 0.3?

See [PEP 612](https://peps.python.org/pep-0612/)

auscompgeek: (I see our [pyproject.toml] still targets py39, so I'm guessing not 😬)

[pyproject.toml]: https://github.com/cursorless-dev/cursorless/blob/ae5304ddf3bd602ff2eda089a2fb1d5cddcf4e96/pyproject.toml

pokey: Oh huh right good point. @AndreasArvidsson any thoughts on this one?

AndreasArvidsson: Talon 0.3 was python 3.10 I believe? Anyhow community is going to move to Talon list so I don't think we need to support older Talon versions

auscompgeek: I still have Talon 0.3.1 on my laptops; it was Python 3.9.

pokey: We were originally planning to wait for community to move to 0.4 before we did, but I don't think it's the end of the world to be slightly ahead of community here. From what I can tell, the delay on community side has more to do with ironing out details of upgrade PR than intentionally waiting for users to upgrade their talon version

That being said, we should prob bump our explicit requirement for Python to 3.10 if we're relying on 3.10 features

PR #1762, opened 2023-08-09
Refactored decorated marks into separate file

AndreasArvidsson:
## Checklist

### DIFF
--- a/cursorless-talon/src/marks/decorated_mark.py
+++ b/cursorless-talon/src/marks/decorated_mark.py
@@ -1,0 +1,180 @@
+from typing import Any
+from pathlib import Path
+
+from talon import Module, actions, cron, fs
+
+from ..csv_overrides import init_csv_and_watch_changes
+
+
+mod = Module()
+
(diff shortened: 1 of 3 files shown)


cursorless-talon/src/marks/special_mark.py
pokey, line 1: we should probably stop using the term "special" for these marks, because the default spoken form for `unknownSymbol` is `"special"`, so I got really confused 😅.  Maybe we use `simple_mark`, as `simple` is the term we use in lots of places for an object that only has a `type` field

AndreasArvidsson: Agreed

pokey (commented):
Left minor comment but otherwise looks good!

AndreasArvidsson:
updated

pokey (approved):
Looks good!  I'd be tempted to merge `simple_mark.py` into `mark.py`, like we do with our simple modifiers, but I could very much go either way on that one

PR #1765, opened 2023-08-10
Ran formatter on scm files

AndreasArvidsson:
First pass of the new scm formatter. Looks reasonable to me.

## Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,11 @@
 // Place your settings in this file to overwrite default and user settings.
 {
-  "python.formatting.provider": "black",
+  "[python]": {
+    "editor.defaultFormatter": "ms-python.black-formatter"
+  },
+  "[scm]": {
+    "editor.defaultFormatter": "AndreasArvidsson.andreas-talon"
+  },
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
   "typescript.tsc.autoDetect": "off",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,6 +6,9 @@
   "[scm]": {
     "editor.defaultFormatter": "AndreasArvidsson.andreas-talon"
   },
+  "[talon]": {
+    "editor.defaultFormatter": "wenkokke.tree-sitter-talon"
+  },
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
   "typescript.tsc.autoDetect": "off",
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,5 +1,11 @@
 {
   // See http://go.microsoft.com/fwlink/?LinkId=827846
   // for the documentation about the extensions.json format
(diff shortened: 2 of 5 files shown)


.vscode/settings.json
```diff
@@ -1,6 +1,11 @@
 // Place your settings in this file to overwrite default and user settings.
 {
-  "python.formatting.provider": "black",
+  "[python]": {
+    "editor.defaultFormatter": "ms-python.black-formatter"
+  },
+  "[scm]": {
+    "editor.defaultFormatter": "AndreasArvidsson.andreas-talon"
```
pokey, line 7: I wonder if we want to require all Cursorless developers to use the `andreas-talon` extension.  Probably not a big deal, and we can probably revisit fairly easily, but feels a bit opinionated.  Eg your talon formatter, tho that's off by default, right?

I wonder if the scm formatter should be its own extension. We could also add cool stuff like function signature hints for our query predicates, hover / go to definition for those, etc.  Tho given they're custom to us, I wonder if separate extension makes sense or if it should really be part of cursorless developer tools

Another advantage of having your scm formatter be separate extension is you can give it a pre-commit config so we can easily use it in CI

But I'm not opposed to merging as is just to get the ball rolling and revisit later

pokey (approved):
looks good with one comment!

.vscode/settings.json
AndreasArvidsson: I definitely feel that this is a topic we should revisit. I like many of those suggestions.

pokey: So when people install your extension it's going to reformat the talon files? That's not great, because we use wen's formatter in pre-commit

AndreasArvidsson: Only if they run the formatter. Format on save is not default in vscode. The readme describes how to deactivate format on save for Talon files specifically. We could put that setting in the repository vscode file?

pokey: Yeah I'm pretty sure I had to manually deactivate the formatter when I first installed your extension

Maybe we could just have a config that sets wen's extension as default formatter for talon files like we're having yours for scm?

Also, maybe we should make the scm syntax colouring extension a recommended extension while we're here?

AndreasArvidsson: We can definitely do that setting in Cursorless.
Maybe.

AndreasArvidsson: Wens formatter is now the default for Talon files.

pokey: I mean do we want to add `jrieken.vscode-tree-sitter-query` as a recommended extension

AndreasArvidsson: I guess we could. Do we have other extension recommendations? if yes then where are they? and if know why this one in particular?

pokey: https://github.com/cursorless-dev/cursorless/blob/main/.vscode/extensions.json

looks like we recommend our autoformatters; maybe we should add the new autoformatters as well?

AndreasArvidsson: Sounds reasonable. What exactly are we missing?

pokey: - `AndreasArvidsson.andreas-talon`
- `wenkokke.tree-sitter-talon`
- `jrieken.vscode-tree-sitter-query`

AndreasArvidsson: done

pokey (approved):
Looks good!

Issue #1767, opened 2023-08-11
confusing hat in fixture moveFirstWordZipBeforeFineTakeZip.yml


josharian:
Consider `packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/scss/moveFirstWordZipBeforeFineTakeZip.yml`:

```yaml
languageId: scss
command:
  spokenForm: move first word zip before fine take zip
  version: 2
  targets:
    - type: primitive
      mark: {type: decoratedSymbol, symbolColor: default, character: z}
      modifiers:
        - type: ordinalRange
          scopeType: {type: word}
          anchor: 0
          active: 0
          excludeAnchor: false
          excludeActive: false
    - type: primitive
      mark: {type: decoratedSymbol, symbolColor: default, character: f}
      modifiers:
        - {type: position, position: before}
  usePrePhraseSnapshot: true
  action: {name: moveToTarget}
marksToCheck: [default.z]
initialState:
  documentContents: foo-bar baz-bongo
  selections:
    - anchor: {line: 0, character: 17}
      active: {line: 0, character: 17}
  marks:
    default.z:
      start: {line: 0, character: 8}
      end: {line: 0, character: 17}
    default.f:
      start: {line: 0, character: 0}
      end: {line: 0, character: 7}
finalState:
  documentContents: baz-foo-bar bongo
  selections:
    - anchor: {line: 0, character: 17}
      active: {line: 0, character: 17}
  marks:
    default.z:
      start: {line: 0, character: 12}
      end: {line: 0, character: 17}
    default.f:
      start: {line: 0, character: 0}
      end: {line: 0, character: 11}
```

In the finalState, the document is: `baz-foo-bar bongo`, and there's supposed to be a `z` hat in range 12-17, which is `bongo`. But there's no "z" in bongo.

Would it be OK to manually delete the `default.z` entry in `finalState` there? Or is it testing something non-obvious?

josharian:
This occurs in a handful of fixtures. I can generate a list as appropriate.

pokey:
Ha yep that is intended behaviour. Because it is a single command phrase, the "zip" hat should still refer to the same place it did at the start of the phrase, but with its range updated to reflect how the token has changed over the course of the phrase. That's actually one of the the main purposes of the test case. See https://www.cursorless.org/docs/contributing/test-case-recorder/#testing-the-hat-map

josharian:
> Is it causing issues for something you're trying to do?

[...]

PR #1768, opened 2023-08-11
experiment with hatdocs / txtar

josharian:
For voice discussion. Lots of important caveats and explaining required.

interesting fixtures:

* packages/cursorless-vscode-e2e/src/suite/fixtures/recorded//everyRange/clearEveryFunkNameFinePastBlueDrum.yml
* packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/scss/moveFirstWordZipBeforeFineTakeFine.yml
* packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/giveEqualsPastColon.yml

```go
type RangeDoc struct {
	Content string
	Ranges  []NamedRangeSet // "hats", "selections", "that", etc.
}

type NamedRangeSet struct {
	Name   string
	Ranges []AnnotatedRange // ranges along with annotations
}

type AnnotatedRange struct {
	Annotation    string
	AnnotationEnd int // how many chars into the range should the annotatin end (e.g. for hats); zero means "don't care", which we will interpret for now as "center"
	Range         Range
	Direction     Direction
}

type Range struct {
	Start Position `yaml:"start,flow"`
	End   Position `yaml:"end,flow"`
}

type Direction int

const (
	DirectionNone Direction = iota
	DirectionStart
	DirectionEnd
)
```

```
range
!
#
[]
[-]
[--]

directed range
!
>
<
[>
<]
[->
<-]
```

```
--- raw notes ---
- Q: primary cursor: how? range? use a hat ref? can't quite cram that last single bool in there :/
- Q: visible vs invisible contents using > vs <?
- Q: how to sync with scope tests, subtle differences due to different needs
- Q: unicode? tabs?

txtar: way to combine hatdoc + metadata for other purposes
- hat tests: hats, tokens (set of ranges), split editors, multi-stage assignment (either file naming conventinos or config at top)
- test fixtures: extract initial/final docs and some ranges (maybe now can run some of these editor-free?!)
- txtar makes it easier to update just part, do diffs, do multi-step things, do multi-editor things, etc.

# problems

currently losing some thatMark (etc) metadata: reversed, explicit, etc.
ranges need lots more metadata (isReversed, explicit, etc) / annotations :(


# stats test

given doc + other metadata:

number of 0, 1, 2 penalty hats
nubmer of hats on first letter of word
percent of tokens with no hat
distance to nearest token with no hat

```


josharian:
https://github.com/cursorless-dev/cursorless/pull/1768/commits/1ac8931cf0c6274873c61f2f97a73a8a1c4ecd86

pokey:
Fwiw if you mark a PR as draft, as you've done, we won't merge it, and will assume it's WIP, so you don't need the big all-caps caveat in the title 😄. We also have a `to discuss` tag you can put on it which indicates it should be discussed in a meet-up

pokey:
Also, I'm guessing you've seen https://github.com/cursorless-dev/cursorless/issues/1524, but figured I'd mention it just in case

josharian:
> I'm guessing you've seen https://github.com/cursorless-dev/cursorless/issues/1524

This was very much inspired by #1524. Unfortunately due to different needs, they diverged in minor ways, and continue to do so. I was designing for hat tests and then got distracted by attempting to re-use for other fixtures. https://github.com/cursorless-dev/cursorless/issues/1767 is an added wrinkle I'm still pondering.

josharian:
this is actually ready to discuss a bit now. will need voice intro, though, for sure.

josharian:
outcome from meetup: as anticipated, the advantages of having a single uniform data format are outweighed by the disadvantages of it not being a perfect fit for any particular use case.

plan is that @pokey (and maybe @AndreasArvidsson?) are going to look through a few of the interesting fixtures in case there are any ideas they want to steal, and we will put this on ice. :)

PR #1783, opened 2023-08-14
Update indentation for snippet body

### LINKED ISSUE #1780: Make insert snippet action do indentation normalization
Today when we insert a snippet the indentation is left unchanged. This means that the actual indentation settings in the current editor are ignored. The solution for this is to replace indentation spaces with tabs. If you use tabs in a snippet vscode will replace them with your preferred indentation setting.

* add this logic to the action 
* do for both insertion and wrapper snippets
* if the snippet already contains tabs don't change anything

AndreasArvidsson:
Fixes #1780

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/snippets/updateSnippetBody.test.ts
+++ b/packages/cursorless-engine/src/snippets/updateSnippetBody.test.ts
@@ -1,0 +1,17 @@
+import * as assert from "assert";
+import { updateSnippetBody } from "./updateSnippetBody";
+
+function testUpdate(title: string, body: string, expected: string) {
+  test(title, () => {
+    const actual = updateSnippetBody(body);
+    assert.equal(actual, expected);
+  });
+}
+
+suite("updateSnippetBody", () => {
+  testUpdate("if", "if($1) {\n    $0\n}", "if($1) {\n\t$0\n}");
+  testUpdate("increasing", "foo\n    bar\n        baz", "foo\n\tbar\n\t\tbaz");
+  testUpdate("decreasing", "        foo\n    bar\nbaz", "\t\tfoo\n\tbar\nbaz");
+  testUpdate("pyramid", "foo\n    bar\nbaz", "foo\n\tbar\nbaz");
+  testUpdate("uneven", "  foo\n   bar\n    baz", "\tfoo\n\t bar\n\t\tbaz");
+});
--- a/packages/cursorless-engine/src/snippets/updateSnippetBody.ts
+++ b/packages/cursorless-engine/src/snippets/updateSnippetBody.ts
@@ -1,0 +1,41 @@
+import { getLeadingWhitespace } from "../util/regex";
+
+export function updateSnippetBody(body: string): string {
+  // If snippet body containins tabs don't touch anything.
+  if (body.includes("\t")) {
+    return body;
+  }
+
+  const lines: { indentation: string; rest: string }[] = [];
+  let smallestIndentation: string | undefined = undefined;
+
+  body.split(/\r?\n/).forEach((line) => {
+    const indentation = getLeadingWhitespace(line);
+
+    // Keep track of smallest non empty indentation
+    if (
+      indentation.length > 0 &&
+      (smallestIndentation == null ||
+        indentation.length < smallestIndentation.length)
+    ) {
+      smallestIndentation = indentation;
+    }
+
+    lines.push({ indentation, rest: line.slice(indentation.length) });
+  });
+
+  // No indentation found in snippet body. No change.
+  if (smallestIndentation == null) {
+    return body;
+  }
+
+  return lines
+    .map((line) => {
+      const indentation = line.indentation.replaceAll(
+        smallestIndentation!,
+        "\t",
+      );
+      return `${indentation}${line.rest}`;
+    })
+    .join("\n");
+}
--- a/packages/cursorless-engine/src/actions/InsertSnippet.ts
+++ b/packages/cursorless-engine/src/actions/InsertSnippet.ts
@@ -20,6 +20,7 @@ import {
   findMatchingSnippetDefinitionStrict,
   transformSnippetVariables,
 } from "../snippets/snippet";
+import { updateSnippetBody } from "../snippets/updateSnippetBody";
 import { SnippetParser } from "../snippets/vendor/vscodeSnippet/snippetParser";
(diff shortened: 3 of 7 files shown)


packages/cursorless-engine/src/snippets/updateSnippetBody.ts
```diff
@@ -0,0 +1,41 @@
+import { getLeadingWhitespace } from "../util/regex";
+
+export function updateSnippetBody(body: string): string {
+  // If snippet body containins tabs don't touch anything.
+  if (body.includes("\t")) {
+    return body;
```
pokey, line 6: do we instead want to throw an error if they mix them? I think that's where we landed yesterday, but feels a bit aggressive on second thought. This is not a new api, so we could be breaking things that are working in the wild. Maybe a warning? Which would be super annoying, but as discussed, if they are mixing tabs and spaces it's probably by accident so they might appreciate knowing

packages/cursorless-engine/src/snippets/updateSnippetBody.ts
```diff
@@ -0,0 +1,41 @@
...
+  if (smallestIndentation == null) {
+    return body;
+  }
+
+  return lines
+    .map((line) => {
+      const indentation = line.indentation.replaceAll(
+        smallestIndentation!,
+        "\t",
+      );
```
pokey, line 37: This is assuming perfect multiples of `smallestIndentation`. If that assumption is broken, we will mix spaces and tabs. Is that what we want?

packages/cursorless-engine/src/snippets/updateSnippetBody.test.ts
```diff
@@ -0,0 +1,17 @@
...
+  });
+}
+
+suite("updateSnippetBody", () => {
+  testUpdate("if", "if($1) {\n    $0\n}", "if($1) {\n\t$0\n}");
+  testUpdate("increasing", "foo\n    bar\n        baz", "foo\n\tbar\n\t\tbaz");
+  testUpdate("decreasing", "        foo\n    bar\nbaz", "\t\tfoo\n\tbar\nbaz");
+  testUpdate("pyramid", "foo\n    bar\nbaz", "foo\n\tbar\nbaz");
+  testUpdate("uneven", "  foo\n   bar\n    baz", "\tfoo\n\t bar\n\t\tbaz");
+});
```
pokey, line 17: maybe add a test with `\r\n`?

packages/cursorless-engine/src/snippets/updateSnippetBody.ts
```diff
@@ -0,0 +1,41 @@
+import { getLeadingWhitespace } from "../util/regex";
+
+export function updateSnippetBody(body: string): string {
```
pokey, line 3: I think I might call this `normalizeSnippetBodyTabs` to make it super clear what's happening here. And a quick jsdoc would be nice

packages/cursorless-engine/src/snippets/updateSnippetBody.test.ts
pokey, line 1: this is nice, tho I think maybe one end-to-end test would be good too just to connect the dots?

pokey (commented):
Looks good with a few comments

packages/cursorless-engine/src/snippets/updateSnippetBody.ts
AndreasArvidsson: I don't want to break anything existing. If they were mixing tabs and spaces today they probably would already have problems with indentation?

packages/cursorless-engine/src/snippets/updateSnippetBody.ts
AndreasArvidsson: Yes. I have added a test for that use case.

AndreasArvidsson:
Updated

packages/cursorless-engine/src/snippets/updateSnippetBody.ts
pokey: Why is that desirable?

packages/cursorless-engine/src/snippets/updateSnippetBody.ts
pokey: I don't feel super strongly.  If this were a new api I think I'd argue for an error, but given it's an existing api, there is an argument for maintaining the status quo when we don't know what to do.  @josharian any thoughts here?

packages/cursorless-engine/src/snippets/updateSnippetBody.ts
AndreasArvidsson: Just because there really is no correct answer. We can round down, round up or leave the extra spaces untouched. I think the latter is the least wrong thing to do.

pokey: Fwiw in my generate snippet code I just consider whether it's an increase or decrease relative to the previous line, regardless of how much. I believe that's what Python does, but I could be wrong about that

pokey: Not sure I agree. Mixed tabs and spaces is a very bad failure mode imo. Maybe let's discuss?

AndreasArvidsson: Lets

packages/cursorless-engine/src/snippets/updateSnippetBody.ts
josharian: Blah. I'd prefer an error, but to avoid actively breaking anyone, let's make it a warning. Maybe even with a request to file an issue if see the warning, we have some idea whether this is going to be a recurring problem...

AndreasArvidsson: We can't just warn if the snippet contains a tab we have to check if they are mixing; a snippet containing tab indentation is valid.

pokey: Yep that's right. I was thinking warn if they're mixed

You should be able to detect that fairly easily when you're computing min indentation below right?

AndreasArvidsson: Definitely doable

AndreasArvidsson:
We decided in today's meet up that this should move to the parser of the snippet format instead

Issue #492, opened 2022-01-18
Expand cursorless talon command api


pokey:
We'd like for users to be able to define more complex custom grammars.  To do so, they'll need full access to the types of commands supported by cursorless, whereas [today](https://github.com/pokey/cursorless-talon/blob/main/src/actions/actions.py#L37) they can just do a single target, with no extra arguments

- [ ] Support [multiple arguments](https://github.com/pokey/cursorless-talon/blob/main/src/command.py#L85)
- [ ] Support [`extraArgs`](https://github.com/pokey/cursorless-talon/blob/main/src/command.py#L58)

tararoys:
Allow custom grammars for reformat action is what I vote!

josharian:
Sample use case:


```
hug <phrase>$:
    mimic("chuck leading " + phrase)
kiss <phrase>$:
    mimic("chuck trailing " + phrase)
(cosy|cozy) <phrase>$:
    m = "chuck leading " + phrase
    m = m + " and trailing " 
    m = m + phrase
    mimic(m)
```

josharian:
Since it often helps to have concrete, real world use cases written down, here is a running list of things I am currently implementing using `mimic`. (Will update it over time.)

```
pre line <x>
chuck leading <x>
chuck trailing <x>
```

josharian:
GitHub failed to backlink, so here it is manually: https://github.com/cursorless-dev/cursorless/discussions/1655#discussioncomment-6570805

pokey:
update from meet-up: there are a couple approaches

## Use csvs

We could have custom modifier and action csvs which map from a spoken form to some representation of compound modifiers / actions.  Here are a couple possibilities for representing `"chuck previous char"` as a custom action:

- `remove previous character`
- `remove relative(-1; character)`
- `remove relative(-1) character`
- `remove <~ relative(-1; character)`

Benefit of using identifiers rather than spoken form is that they can change spoken form without breaking these, at the expense of being a bit harder to read / write

In a csv this would be something like

`experimental/custom_actions.csv`:
```csv
Spoken form, value
shave, remove <~ relative(-1; character)
```

For custom modifiers, we could have another csv where it's the same syntax but without the leading action.  Eg

`experimental/custom_modifiers.csv`:
```csv
Spoken form, value
duet, nScopes(2; token)
```

We could use this syntax in other places, eg

- To pass in to an action eg `user.cursorless_command("remove previous character", cursorless_target)`
[...]

PR #1784, opened 2023-08-14
Experimental private Cursorless Talon api

pokey:
- Private experimental api; one version of #492
- Currently used by [`wax_talon`](https://github.com/pokey/wax_talon); see https://github.com/pokey/wax_talon/commit/4bd7d9b9112a4172d4fc46fe5644e33b8b1340b7 for example usage
- Depends on https://github.com/cursorless-dev/cursorless/pull/1880 for Python 3.10 `match` statements

## Checklist


cursorless-talon/src/public_api/public_api_v1.py
```diff
@@ -0,0 +1,75 @@
...
+    RangeTargetType,
+)
+from .extract_decorated_marks import extract_decorated_marks
+
+mod = Module()
+
+
+@mod.action_class
+class MiscActions:
+    def cursorless_v1_extract_decorated_marks(capture: Any) -> list[dict]:
```
pokey, line 19: This is currently used as follows: https://github.com/pokey/wax_talon/blob/4bd7d9b9112a4172d4fc46fe5644e33b8b1340b7/recorders/cursorless_recorder.py#L151-L159.  The `parsed` argument is expected to be `list(j["parsed"])`, where `j` is the argument to an `on_phrase` callback.  I wonder if we want to expose something that operates directly on the phrase object rather than requiring the pre-processing 🤔.  This function does work on the output of any of our spoken form captures, so I think may still be a reasonable api

pokey:
update from meet-up:

- Keep this "internal" for now as we don't want to commit to maintaining this expansive of an api yet

pokey:
ok @AndreasArvidsson as discussed I made this private for now so that I can use it for wax_talon without needing to expose it to the world. Ready for review

pokey:
one sec I'll add a couple quick talon api tests so this doesn't break

cursorless-talon/src/private_api/private_api.py
```diff
@@ -0,0 +1,66 @@
...
+class TargetActions:
+    def cursorless_private_target_nothing() -> PrimitiveTarget:
+        """Cursorless private api: Creates the "nothing" target"""
+        return PrimitiveTarget({"type": "nothing"}, [])
+
+
+@mod.action_class
+class ActionActions:
+    def cursorless_private_action_highlight(
+        target: CursorlessTarget, highlightId: Optional[str] = None
```
AndreasArvidsson, line 53: I assume you actually use the highlight id? Otherwise the highlight action should just work with our existing public api.

cursorless-talon/src/private_api/extract_decorated_marks.py
```diff
@@ -0,0 +1,47 @@
...
+    ListTarget,
+    PrimitiveDestination,
+    PrimitiveTarget,
+    RangeTarget,
+)
+
+
+def extract_decorated_marks(capture: Any) -> list[Any]:
+    match capture:
+        case PrimitiveTarget(mark=mark):
```
AndreasArvidsson, line 18: This syntax is interesting.

pokey: Yeah kinda nice I think

AndreasArvidsson: I'm not sure how to read this. I totally follow the examples `case MyClass(value=5)` checks if the instance is of the correct type and the value field is the correct value. `mark=mark` I'm unsure how to interpret.

pokey: It creates a local variable `mark` bound to whatever `mark` was in that dataclass. It's analogous to `const {x: x} = foo", where it constructs a local variable `x` bound to whatever was in the `x` field of `foo`

AndreasArvidsson: So that syntax can be made for both comparisons and assignments. hmm. Sometimes python is a little bit to clever for my taste.

pokey: Ha yeah I hadn't thought about that you can use it with constant. Yes maybe a bit clever 😅

AndreasArvidsson (approved):

PR #1789, opened 2023-08-15
Added insertion delimiter to Talon command

### LINKED ISSUE #1785: talon command scope delimiters not quite correct
```
foo:
    insert("foo\n")
```

run `clone command fine`. result:

```
foo:
    insert("foo\n") foo:
    insert("foo\n")
```

cc @AndreasArvidsson

AndreasArvidsson:
Fixes #1785

## Checklist

### DIFF
--- a/queries/talon.scm
+++ b/queries/talon.scm
@@ -119,9 +119,12 @@
 
 ;;!! slap: key(enter)
 ;;!  ^^^^^^^^^^^^^^^^
-(command_declaration
-  right: (_) @_.interior
-) @command
+(
+  (command_declaration
+    right: (_) @_.interior
+  ) @command
+  (#insertion-delimiter! @command "\n")
+)
 
 (source_file
   (declarations) @command.iteration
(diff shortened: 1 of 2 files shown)


queries/talon.scm
```diff
@@ -119,9 +119,12 @@
...
 ;;!! slap: key(enter)
 ;;!  ^^^^^^^^^^^^^^^^
-(command_declaration
-  right: (_) @_.interior
-) @command
+(
+  (command_declaration
+    right: (_) @_.interior
+  ) @command
+  (#insertion-delimiter! @command "\n")
```
pokey, line 126: This isn't where we usually do this. Are you proposing that we always do this in our query file rather than in the usual spot?

AndreasArvidsson: I think having the power to control this per language is an improvement? If you prefer I can put this in typescript.

pokey: No I don't have a strong feeling, just wondering why we didn't do this before, as we had a way to define insertion delimiters per-language before. I guess it's slightly less awkward now in this new syntax

pokey (approved):

queries/talon.scm
AndreasArvidsson: I think things like functions and statements are similar enough that they probably have the same delimiter in most languages. For now this scope type is only used for a single language. Yeah and our new syntax makes it a bit easier to do this kind of thing.

PR #1794, opened 2023-08-15
Release 0.27.0

pokey:
## Checklist

- [ ] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)
- [x] Add link to josh videos
- [x] Highlight some issues form shortlist, and link to it
- [x] Highlight some issues from longlist, and link to it (esp cursorless everywhere stuff; maybe draw from what's next stuff on 2022 summary)
- [x] Add images
- [ ] Write some destination docs?
- [ ] Write docs for other things I had to handwave?
- [ ] Record scope visualizer video?
- [ ] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [ ] I have not broken the cheatsheet

PR #1801, opened 2023-08-17
Add release notes pop-up

pokey:
- Depends on #1794
- Partially addresses #491 

## Checklist


AndreasArvidsson (approved):

pokey:
I just realised one possible problem with this approach is that users will see this when they first install Cursorless 🤔. I guess that's not the end of the world but probably not particularly helpful. I wonder what is the best way to avoid that. Would be fine if we had set a date of some sort on install but at this point not sure how to detect new install vs update

pokey:
I realise now that what we should have done was:

- Modify this PR to not show the pop up if they have no version number set in global storage. Just set it if it's not there and show nothing
- Merge this PR
- Then merge release 0.27.0

That way we just track their initial install and show for subsequent updates

Not sure what we should do now though. That ship has kind of sailed for 0.27.0 😅

pokey:
I guess we might need a hack of some sort. Eg a one-off key just for release 0.27.0 that we set in a pre-PR and then check in another one that merges a day or two later so everyone has the first one

pokey:
I also feel like `seenReleaseNotesVersion` is maybe not quite right key name with this mechanism, as after first install it's not technically accurate

AndreasArvidsson:
Agreed that this should not pop up on first install.

pokey:
Ok I have modified this PR as follows:

- It doesn't show the popup on first install
- I added tests

I propose we do the following:

1. Merge this PR
2. Wait 1 week
3. Merge #1818

I don't think anyone has likely noticed that we released 0.27.0, so it seems fine to just rebrand that release as 0.28.0.  Over the next week, existing users will have their stored `version` field set, so that next week when we release 0.28.0, existing users will get a popup

Make sense?

AndreasArvidsson (approved):

PR #1810, opened 2023-08-18
support hat strokes as a hidden feature

josharian:
To use, add e.g. `-#aa10ff` after any color in settings to add `#aa10ff` as the stroke color.

For example, my `userColor2` is `#000000-#ffffff`, and my green is `#14ff07-#00a336`.

This is intentionally undocumented: It is a hidden feature for folks to play around with, but only if they are paying a lot of attention, at which point they will know that is not officially supported. :)

If we decide we like it, we can eventually see about documenting and testing and generally "unhiding" it.

## Checklist

### DIFF
--- a/packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
+++ b/packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
@@ -50,6 +50,9 @@ interface SvgInfo {
   svg: string;
   svgHeightPx: number;
   svgWidthPx: number;
+  originalViewBoxWidth: number;
+  originalViewBoxHeight: number;
+  strokeWidth: number;
 }
 
 /**
@@ -230,6 +233,7 @@ export default class VscodeHatRenderer {
             this.fontMeasurements,
             shape,
             scaleFactor,
+            defaultShapeAdjustments[shape].strokeFactor ?? 1,
             finalVerticalOffsetEm,
           ),
         ];
@@ -248,7 +252,7 @@ export default class VscodeHatRenderer {
             ];
           }
 
-          const { svg, svgWidthPx, svgHeightPx } = svgInfo;
+          const { svgWidthPx, svgHeightPx } = svgInfo;
 
           const { light, dark } = getHatThemeColors(color);
 
@@ -258,12 +262,18 @@ export default class VscodeHatRenderer {
               rangeBehavior: vscode.DecorationRangeBehavior.ClosedClosed,
               light: {
                 before: {
-                  contentIconPath: this.constructColoredSvgDataUri(svg, light),
+                  contentIconPath: this.constructColoredSvgDataUri(
+                    svgInfo,
+                    light,
+                  ),
                 },
               },
               dark: {
                 before: {
-                  contentIconPath: this.constructColoredSvgDataUri(svg, dark),
+                  contentIconPath: this.constructColoredSvgDataUri(
+                    svgInfo,
+                    dark,
+                  ),
                 },
               },
               before: {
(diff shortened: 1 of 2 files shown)


josharian:
I've been using this for several weeks now, and I really like it. Not only do I get to use the color black, but I have a bright stroke around my green, and I no longer confuse it with blue.

Screenshot 2023-09-05 at 5 49 56 PM

Requires thicker hats.

One minor problem is that the stroke is around the outside of the shape, which can cause clipping. It's very worth it. 

So let's talk about what it would take to get this in. :)

pokey:
update:

- [ ] wait until thicker hats ship
- [ ] add comment saying to subscribe to this PR if they use it

josharian:
marking as draft until thick hats (of some variety or another) are easily available

pokey:
thicker hats are here! (#1868)

josharian:
> thicker hats are here!

Yay! Rebased, ready for y'all to play with.

FWIW, my colors (dark mode only) using this are:

```json
  "cursorless.colors.dark": {
    "blue": "#388aff",
    "green": "#64ba57-#66ff00",
    "red": "#ff0000",
    "pink": "#ffa8b5",
    "yellow": "#efef0b",
    "userColor1": "#ffffff",
    "userColor2": "#000000-#ffffff"
  },
```

Screenshot 2023-10-31 at 4 18 07 PM

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
```diff
@@ -307,10 +315,25 @@ export default class VscodeHatRenderer {
...
+  private constructColoredSvgDataUri(svgInfo: SvgInfo, color: string) {
+    const { svg: originalSvg, svgWidthPx, originalViewBoxWidth } = svgInfo;
+    let replacementEquals, replacementColon: string;
+    // If color contains a dash, the second part is a stroke.
+    // If you are code spelunking and have found this undocumented (and thus potentially transient) feature,
+    // please subscribe to https://github.com/cursorless-dev/cursorless/pull/1810
+    // so that you can be notified if/when it changes or is removed.
+    if (color.includes("-")) {
+      const [fill, stroke] = color.split("-");
+      const strokeScaleFactor = (0.7 * originalViewBoxWidth) / svgWidthPx;
```
AndreasArvidsson, line 327: This should probably be called `strokeWidth`

AndreasArvidsson:
I think that this is really interesting. Unfortunately some of our hats have negatives internal spaces that gets squeezed to almost nothing when you add the stroke.

This is just adding a stroke to the green color
image
image

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
```diff
@@ -307,10 +315,25 @@ export default class VscodeHatRenderer {
...
+      replacementEquals = `fill="${fill}" stroke="${stroke}" stroke-width="${strokeScaleFactor}"`;
+      replacementColon = `fill: ${fill}; stroke: ${stroke} stroke-width: {strokeScaleFactor};`;
```
AndreasArvidsson, line 329: Missing a `$` before `{strokeScaleFactor}`

josharian:
> some of our hats have negatives internal spaces that gets squeezed to almost nothing when you add the stroke.

Oh, I see that. I don't use any of them.

Not sure how to fix that without this turning into a large project.

> I tried with stroke-alignment="inner" but that doesn't seem to do anything in vscode.

Alas, SVG stroke controls are super limited. See: https://stackoverflow.com/questions/7241393/can-you-control-how-an-svgs-stroke-width-is-drawn

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
josharian: indeed, done

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
josharian: eek done thanks

AndreasArvidsson:
@pokey Any thoughts?

pokey:
update from meetup:

- @josharian will update the code to rely on the fact that all shapes only contain a single path element, and use that path element to perform clipping on the stroke so that it doesn't fill anything that is currently negative space
- the changes should be mostly contained in a single function that is only called if the color contains a `-` (modulo some refactoring 😊)
- we should try to fail at least somewhat gracefully if the shape isn't a single path, if nothing else because we may forget about this requirement in the future when shipping new hat shapes 😅

josharian:
> @josharian will update the code to rely on the fact that all shapes only contain a single path element, and use that path element to perform clipping on the stroke so that it doesn't fill anything that is currently negative space

done, although hole is such an edge-heavy shape that the fills gets lost. to compensate, i dialed down the stroke width considerably from a factor of 1.4 to a factor of 1), which makes it less useful for other shapes. i don't really want to start tweaking the stroke width per shape, but it may be necessary.

i also don't want to overinvest in this.

> the changes should be mostly contained in a single function that is only called if the color contains a - (modulo some refactoring 😊)

done

> we should try to fail at least somewhat gracefully if the shape isn't a single path, if nothing else because we may forget about this requirement in the future when shipping new hat shapes 😅

done enough, i think. although see the usual comments about idiot-proofing.

josharian:
Screenshot 2024-01-11 at 1 02 23 PM

screenshot, zoomed in, showing black with white stroke and green with light green stroke.

josharian:
> i don't really want to start tweaking the stroke width per shape, but it may be necessary.

never mind, made it possible to tweak stroke width per shape, and restored all other shapes to original stroke width.

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
```diff
@@ -50,6 +50,9 @@ interface SvgInfo {
   svg: string;
   svgHeightPx: number;
   svgWidthPx: number;
+  originalViewBoxWidth: number;
+  originalViewBoxHeight: number;
```
pokey, line 54: from what I can tell, these are unused?

pokey (commented):
looks good with minor comments!

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
```diff
@@ -307,17 +317,51 @@ export default class VscodeHatRenderer {
...
+  private addInnerStrokeToSvg(
+    svgInfo: SvgInfo,
+    svg: string,
+    stroke: string,
+  ): string {
+    // All hat svgs have exactly one path element. Extract it.
+    const pathRegex = /<path[^>]*d="([^"]+)"[^>]*\/>/;
+    const pathMatch = pathRegex.exec(svg);
+    if (!pathMatch) {
+      return svg;
```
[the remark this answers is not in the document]
josharian: Added a log.  Per previous discussion ("fail at least somewhat gracefully if the shape isn't a single path"), we don't want to throw here.

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
josharian: oh yeah, refactoring made those obsolete. thanks, removed.

pokey (approved):
🚀

PR #1821, opened 2023-08-22
Added get targets action

### LINKED ISSUE #1820: Extension-level API to resolve targets to ranges
This will be useful for the sidecar to implement actions without having to add extra layers of RPC. We can ask cursorless to resolve the given target and return those up to Talon as content ranges on the document, who can then pass them to the editor to do custom commands [1], or implement the builtin actions such as "follow".

Proposal is to add the command to the extension, but it will be internal, so only include the Talon action definition in the pull request description so that people don't call it accidentally.

[1] asking the sidecar to resolve the ranges, and then acting on them, could also be done entirely within the exterior editor, without going up to Talon

AndreasArvidsson:
Fixes #1820

```talon
info <user.cursorless_target>:
    target_info = user.private_cursorless_get_targets(cursorless_target)
    print(target_info)
```

```python
def private_cursorless_get_targets(target: Any) -> list[dict]:
        """Get information about Cursorless target"""
        return actions.user.private_cursorless_command_get(
            {
                "name": "private.getTargets",
                "target": target,
            }
        )
```

```json
[
    {"contentRange": {"start": {"line": 71, "character": 23}, "end": {"line": 71, "character": 53}}}
]
```

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/actions/GetTargets.ts
+++ b/packages/cursorless-engine/src/actions/GetTargets.ts
@@ -1,0 +1,17 @@
+import { Target } from "../typings/target.types";
+import { ActionReturnValue, SimpleAction } from "./actions.types";
+
+export default class GetTargets implements SimpleAction {
+  constructor() {
+    this.run = this.run.bind(this);
+  }
+
+  async run(targets: Target[]): Promise<ActionReturnValue> {
+    return {
+      returnValue: targets.map(({ contentRange }) => ({
+        contentRange,
+      })),
+      thatTargets: targets,
+    };
+  }
+}
--- a/packages/common/src/types/command/ActionDescriptor.ts
+++ b/packages/common/src/types/command/ActionDescriptor.ts
@@ -19,7 +19,6 @@ const simpleActionNames = [
   "findInWorkspace",
   "foldRegion",
   "followLink",
-  "private.getTargets",
   "indentLine",
   "insertCopyAfter",
   "insertCopyBefore",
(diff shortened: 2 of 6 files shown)


AndreasArvidsson:
@phillco

phillco:
Thank you! ❤️

packages/cursorless-engine/src/actions/GetTargets.ts
```diff
@@ -0,0 +1,17 @@
+import { Target } from "../typings/target.types";
+import { ActionReturnValue, SimpleAction } from "./actions.types";
+
+export default class GetTargets implements SimpleAction {
+  constructor() {
+    this.run = this.run.bind(this);
+  }
+
+  async run(targets: Target[]): Promise<ActionReturnValue> {
```
pokey, line 9: I think a flash would be nice. WDYT @phillco @dyamito

pokey (commented):
Looks good but would be nice to record a quick test. Should be trivial, no?

packages/cursorless-engine/src/actions/GetTargets.ts
phillco: You mean a visual flash?

AndreasArvidsson: I left it out since the sidecar wouldnt benefit by flashes in vscode.

phillco: I would agree. I also don't know if flashing is synchronous but we wouldn't want to have any delay here.

pokey: Good call

AndreasArvidsson:
@pokey test added

pokey (approved):
Good stuff!

PR #1824, opened 2023-08-22
do a single pass in maxByFirstDiffering

josharian:
This cuts the runtime for hat allocation
in my tests by about 20%.

No functional changes.

I have not added any tests, but I have confirmed
that there are no functional changes
using my not-yet-submitted hat golden tests.

Part of the reason that the hat golden tests are not
yet ready for review is that they are way too slow.

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/util/allocateHats/maxByFirstDiffering.test.ts
+++ b/packages/cursorless-engine/src/util/allocateHats/maxByFirstDiffering.test.ts
@@ -1,0 +1,40 @@
+import * as assert from "assert";
+import { maxByAllowingTies } from "./maxByFirstDiffering";
+
+// known good but slow
+function goldenMaxByAllowingTies<T>(arr: T[], fn: (item: T) => number): T[] {
+  const max = Math.max(...arr.map(fn));
+  return arr.filter((item) => fn(item) === max);
+}
+
+suite("maxByFirstDiffering", () => {
+  test("maxByAllowingTies", () => {
+    const testCases: number[][] = [
+      [],
+      [0],
+      [1],
+      [-Infinity],
+      [+Infinity],
+      [0, 0],
+      [0, 1],
+      [1, 0],
+      [-Infinity, -Infinity],
+      [-Infinity, +Infinity],
+      [+Infinity, -Infinity],
+      [+Infinity, 0],
+      [0, +Infinity],
+      [-Infinity, 0],
+      [0, -Infinity],
+      [0, 1, 0, 1],
+      [1, 0, 1, 0],
+      [0, 1, 1, 0],
+      [1, 0, 0, 1],
+    ];
+
+    testCases.forEach((testCase) => {
+      const actual = maxByAllowingTies(testCase, (x) => x);
+      const expected = goldenMaxByAllowingTies(testCase, (x) => x);
+      assert.deepStrictEqual(actual, expected);
+    });
+  });
+});
--- a/packages/cursorless-engine/src/util/allocateHats/maxByFirstDiffering.ts
+++ b/packages/cursorless-engine/src/util/allocateHats/maxByFirstDiffering.ts
@@ -30,8 +30,22 @@ export function maxByFirstDiffering<T>(
       return remainingValues[0];
     }
 
-    const max = Math.max(...remainingValues.map(fn));
-    remainingValues = remainingValues.filter((item) => fn(item) === max);
+    // Take a single pass through, accumulating all
+    // items with the single highest value.
+    let best: number = -Infinity;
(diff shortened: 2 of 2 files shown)


josharian:
marking as draft because I think there may be a deeper fix...

josharian:
this appears to be cumulative with my other fixes, ready for review

packages/cursorless-engine/src/util/allocateHats/maxByFirstDiffering.ts
```diff
@@ -30,8 +30,22 @@ export function maxByFirstDiffering<T>(
...
+      if (value < best) {
+        continue;
+      }
+      if (value > best) {
+        best = value;
+        keep.length = 0;
+      }
+      keep.push(item);
+    }
+    remainingValues = keep;
```
pokey, line 48: I would be tempted to make this its own function for clarity, but I don't feel strongly. I guess there may be performance implications if this code is on the hot path

pokey (approved):
Looks good with one minor comment!

packages/cursorless-engine/src/util/allocateHats/maxByFirstDiffering.ts
josharian: Done. I'm still finding my feet for typescript performance, but I'm pretty confident that the JIT will figure out how to inline this. And empirically it is just as fast.

josharian:
What is the convention for PRs that are approved with comments? wait for another round of review to confirm you like the fixes?

packages/cursorless-engine/src/util/allocateHats/maxByFirstDiffering.ts
```diff
@@ -24,15 +24,45 @@ export function maxByFirstDiffering<T>(
...
 }
+
+/**
+ * Given an array of items and a function that returns a number for each item,
+ * return all items that share the maximum value according to that function.
+ * @param arr The array to find the max values of
+ * @param fn A function that returns a number for each item in the array
+ * @returns All items in the array that share the maximum value
+ **/
+function highest<T>(arr: T[], fn: (item: T) => number): T[] {
```
pokey, line 43: Maybe `maxByAllowingTies` instead of `highets`?

pokey (approved):
Minor quibble; otherwise good.  I would be very tempted to add unit tests for this function, but I'm hesitant to make you invest more in this PR, which was supposed to be a quick win

If you think not worth it to add tests, and you're happy with my rename in comment below, feel free to merge

pokey:
> What is the convention for PRs that are approved with comments? wait for another round of review to confirm you like the fixes?

It's a good question. I think if the comments resulted in changes that don't really have any room for disagreement, then feel free to merge. In this case, I think there's room for disagreement over the function name (see my comment 😄), so maybe better to leave it?

josharian:
> I think if the comments resulted in changes that don't really have any room for disagreement, then feel free to merge.

OK. Although given the naturally widely divergent range of opinions, I would have a weak preference for only approving once you are sold that the only remaining changes are trivial.

Will rename and will probably add tests. Thanks for the speedy review.

pokey:
> OK. Although given the naturally widely divergent range of opinions, I would have a weak preference for only approving once you are sold that the only remaining changes are trivial.

yeah fwiw that's what I usually do. in this case I approved because I wasn't sure whether it was worth making the changes at all

josharian:
since I'm new to the project and to typescript, I'm going to be more annoying about asking for re-reviews rather than just merging. so: PTAL. :)

packages/cursorless-engine/src/util/allocateHats/maxByFirstDiffering.test.ts
```diff
@@ -0,0 +1,40 @@
...
+      [0, -Infinity],
+      [0, 1, 0, 1],
+      [1, 0, 1, 0],
+      [0, 1, 1, 0],
+      [1, 0, 0, 1],
+    ];
+
+    testCases.forEach((testCase) => {
+      const actual = maxByAllowingTies(testCase, (x) => x);
+      const expected = goldenMaxByAllowingTies(testCase, (x) => x);
```
pokey, line 36: Cute. I like it

pokey (approved):
Looks good!

packages/cursorless-engine/src/util/allocateHats/maxByFirstDiffering.test.ts
josharian: An old favorite approach from fuzz testing. Particularly well suited to performance optimization. :)

https://en.wikipedia.org/wiki/Differential_testing

PR #1829, opened 2023-08-24
tsconfig: Enable isolatedModules

auscompgeek:
esbuild recommends enabling this to avoid miscompilation. https://esbuild.github.io/content-types/#isolated-modules

## Checklist

- [ ] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)
- [ ] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [ ] I have not broken the cheatsheet

PR #1832, opened 2023-08-25
eslint: Enable consistent-type-imports

auscompgeek:
Revives #872.

It might be a good idea to enable [`import/no-duplicates`] first though.

[`import/no-duplicates`]: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md

https://typescript-eslint.io/rules/consistent-type-imports/

## 🥞

- #1829

## Checklist


packages/cheatsheet-local/src/app/app.tsx
```diff
@@ -1,4 +1,5 @@
-import { CheatsheetPage, CheatsheetInfo } from "@cursorless/cheatsheet";
+import type { CheatsheetInfo } from "@cursorless/cheatsheet";
+import { CheatsheetPage } from "@cursorless/cheatsheet";
```
pokey, line 2: It's possible to do this in one line

```suggestion
import type { type CheatsheetInfo, CheatsheetPage } from "@cursorless/cheatsheet";
```

Is there a way to get it to do that?

pokey (commented):
Had one question

Also, re running deduplicate, do we not want to run deduplicate after this one rather than before, due to interaction with https://github.com/import-js/eslint-plugin-import/pull/334. Not totally sure tbh 🤔

packages/cheatsheet-local/src/app/app.tsx
pokey: @auscompgeek maybe let's aim to discuss this inline `type` thing at a meet-up? https://github.com/cursorless-dev/cursorless/labels/to%20discuss

pokey: After discussion, we want to:

1. enable `"@typescript-eslint/consistent-type-imports": "error"`
2. enable [`--verbatimModuleSyntax` TS flag](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax)
3. enable `"import/consistent-type-specifier-style": ["error", "prefer-top-level"]`
4. enable `"import/no-duplicates": "error"` (note we don't need `prefer-inline` because we don't use inline)

Each step should be its own PR, and they should be in the above order. Note that 1) must come before 2) because without 2), 1) should have no side effects, and then the effects of 2) should become more clear once 1) is in

Note that we don't need [`no-import-type-side-effects` lint rule](https://typescript-eslint.io/rules/no-import-type-side-effects/) because 2 and 3 won't allow us to have inline `type` anywhere

The reasoning is that we want `import type` to be kind of like it's own type of statement that we know always gets dropped, and in addition, if we were to have a rule where `import {type foo}` => `import type {foo}` only when there are no non-type imports in the statement, it would feel a bit inconsistent that it doesn't happen when there are non-type imports. This rule is simpler: types always get imported in their own statement, and never have side effects

pokey: We want to test the following:

- [ ] Auto-import when you use something as a type but not runtime and it's currently imported as runtime
- [ ] Auto-import when you use something as a type that is not yet imported
- [ ] Auto-import when you use something as runtime that is currently imported as a type
- [ ] Auto-fix in each of the above cases
- [ ] Organize imports in each of the above cases

auscompgeek: I still never got a chance to test that matrix, but up until TypeScript 5.3 whether auto-import used `import type` depends on the `tsconfig`. TypeScript 5.3 adds an IDE setting to prefer auto-importing with `import type` https://devblogs.microsoft.com/typescript/announcing-typescript-5-3/#settings-to-prefer-type-auto-imports

The presence of this lint rule will obviously not affect auto-import behaviour since eslint doesn't change the tsconfig. The auto-import behaviour probably isn't too big of an issue anyway since we all have eslint autofix on save, I believe.

pokey:
Note from meet-up:

- For ease of git-fu, just open a new PR with contents of first commit
- Then test the matrix in https://github.com/cursorless-dev/cursorless/issues/2618

AndreasArvidsson:
Replaced by https://github.com/cursorless-dev/cursorless/pull/2619

PR #1834, opened 2023-08-26
implement non-quadratic version of uniqWith

### LINKED ISSUE #1830: unusably slow with a few thousand lines
I was trying to do some bulk editing with cursorless today and found that it becomes unusably slow when it's operating on a bunch of lines (in this case, about 3,236 of them).

Here's a sample ([file contents](https://gist.github.com/josharian/9df01b0380be79855e2ea2e8eabc3277)) of the UX:

https://github.com/cursorless-dev/cursorless/assets/67496/c9c5d9c5-8dfb-40bd-8340-fa8364dcf2f0

josharian:
When given a comparator, uniqWith is quadratic.
Add uniqWithHash that uses an optional hash function
to return to approximately linear behavior,
assuming that the hash function is good.
For targets and selections,
Using the contentRange/selection as the hash value
seems to be pretty good in practice.

I've added "concise" printing versions of positions,
ranges, and selections. These are handy as hash functions,
but I've also made heavy use of them while investigating
hat allocation. They are generally useful.

"take every line" on a file containing 3000 lines
used to take about 15s on my computer.
Now it takes about 1s.
That's still not great but it's fast enough
that it doesn't appear that Cursorless has hung.

I've half a mind to try upstreaming this,
or at least sending a documentation fix,
because a quadratic API is a big footgun.

Fixes #1830 enough for now

## Checklist

### DIFF
--- a/packages/common/src/types/Selection.ts
+++ b/packages/common/src/types/Selection.ts
@@ -72,4 +72,12 @@ export class Selection extends Range {
       this.anchor.isEqual(other.anchor) && this.active.isEqual(other.active)
     );
   }
+
+  /**
+   * Return a concise string representation of the selection
+   * @returns concise representation
+   **/
+  public concise(): string {
+    return `${this.anchor.concise()}->${this.active.concise()}`;
+  }
 }
--- a/packages/common/src/types/Selection.ts
+++ b/packages/common/src/types/Selection.ts
@@ -80,4 +80,8 @@ export class Selection extends Range {
   public concise(): string {
     return `${this.anchor.concise()}->${this.active.concise()}`;
   }
+
+  public toString(): string {
+    return this.concise();
+  }
 }
--- a/packages/cursorless-engine/package.json
+++ b/packages/cursorless-engine/package.json
@@ -27,6 +27,7 @@
     "@types/mocha": "^8.0.4",
     "@types/sbd": "^1.0.3",
     "@types/sinon": "^10.0.2",
+    "fast-check": "3.12.0",
     "js-yaml": "^4.1.0",
     "mocha": "^10.2.0",
     "sinon": "^11.1.1"
--- a/packages/cursorless-engine/src/util/uniqWithHash.ts
+++ b/packages/cursorless-engine/src/util/uniqWithHash.ts
@@ -3,13 +3,13 @@ import { uniqWith, isEqual } from "lodash";
 /**
  * Like lodash.uniqWith, but uses a hash function to (mostly) avoid quadratic runtime.
  * @param array The array to uniq
- * @param fn The equality function
+ * @param isEqual The equality function
  * @param hash The hash function. It must be a valid hash function, insofar as it must return the same value for equal items. A hash function that returns a constant value is equivalent to lodash.uniqWith.
  * @returns The uniq array
  */
 export function uniqWithHash<T>(
   array: T[],
-  fn: (a: T, b: T) => boolean,
+  isEqual: (a: T, b: T) => boolean,
   hash: (t: T) => string,
 ): T[] {
   // Handle the common, tiny cases without allocating anything extra.
@@ -17,15 +17,16 @@ export function uniqWithHash<T>(
     return [...array];
   }
   if (array.length === 2) {
-    if (fn(array[0]!, array[1]!)) {
-      return [array[0]!];
+    if (isEqual(array[0], array[1])) {
+      return [array[0]];
     }
     return [...array];
   }
   // First, split up the array using the hash function.
   // This keeps the sets of items passed to uniqWith small,
   // so that the quadratic runtime of uniqWith less of a problem.
-  // Keep track of which sets have multiple items, so that we can uniq them.
+
+  /* Keep track of which sets have multiple items, so that we can uniq them. */
   const needsUniq: string[] = [];
   const hashToItems: Map<string, T[]> = array.reduce((acc, item) => {
     const key = hash(item);
@@ -34,17 +35,18 @@ export function uniqWithHash<T>(
       acc.set(key, [item]);
       return acc;
     }
-    if (items.length === 1) {
+
+    acc.get(key)!.push(item);
+    if (items.length === 2) {
       needsUniq.push(key);
     }
-    acc.get(key)!.push(item);
     return acc;
   }, new Map<string, T[]>());
 
   // For hash collisions, uniq the items,
   // letting uniqWith provide correct semantics.
   needsUniq.forEach((key) => {
-    hashToItems.set(key, uniqWith(hashToItems.get(key)!, fn));
+    hashToItems.set(key, uniqWith(hashToItems.get(key)!, isEqual));
   });
 
   // Another common case: Everything is unique.
@@ -54,6 +56,10 @@ export function uniqWithHash<T>(
 
   // To preserve order, step through the original items
   // one at a time, returning it as appropriate.
+  // We need to do this because uniqWith preserves order,
+  // and we are mimicking its semantics.
+  // Note that items were added in order,
+  // and uniqWith preserved that order.
   return array.flatMap((item) => {
     const key = hash(item);
     const items = hashToItems.get(key)!;
@@ -68,7 +74,6 @@ export function uniqWithHash<T>(
     }
     // Emit item.
     items.shift();
-    hashToItems.set(key, items);
     return first;
   });
 }
--- a/packages/cursorless-engine/src/util/uniqWithHash.ts
+++ b/packages/cursorless-engine/src/util/uniqWithHash.ts
@@ -47,6 +47,11 @@ export function uniqWithHash<T>(
     hashToItems.set(key, uniqWith(hashToItems.get(key)!, fn));
   });
 
+  // Another common case: Everything is unique.
+  if (needsUniq.length === 0) {
+    return array;
+  }
+
   // To preserve order, step through the original items
   // one at a time, returning it as appropriate.
   return array.flatMap((item) => {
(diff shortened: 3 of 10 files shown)


josharian:
https://github.com/cursorless-dev/cursorless/assets/67496/b3f74beb-e9c5-42bd-bd72-adaa300e40ae

josharian:
There is one small behavior difference worth calling out here. The new implementation can return an aliased version of the input; I don’t think the original ever does that. I don’t think it matters in practice here, but I could (a) document it (b) change it at a small performance cost or (c) leave it as is.

auscompgeek:
We could throw some `readonly` around to see if there's anything actually depending on getting a copy.

josharian:
I am still :confuseddog: with typescript, but readonly appears to work on properties, whereas this gets used with local variables. But happy to try anything with a slightly more detailed pointer as to how. Or to just put in a defensive copy. Anything is better than continuing quadratic. 🤷🏽

auscompgeek:
Arrays can be `readonly` - it's shorthand for `ReadonlyArray`.

josharian:
Ah, thanks. No dice, because that breaks an interface satisfaction:

```
Interface 'EditableTextEditor' incorrectly extends interface 'TextEditor'.
  Types of property 'selections' are incompatible.
    The type 'readonly Selection[]' is 'readonly' and cannot be assigned to the mutable type 'Selection[]'.ts(2430)
```

Not necessarily a meaningful problem. This particular input to this function ultimately traces back to a `.map` call, so it's a new copy each time to begin with, so I'm pretty sure this is safe. The question is whether we should guarantee it.

josharian:
Meh. The entire codebase allocates everywhere. Added copies so we don't have to think about it.

auscompgeek:
Ah. I've been considering throwing `readonly` on a bunch of signatures to loosen what they accept. Maybe I should actually do it.

josharian:
update from meetup: josh to add a simple regression test not only of correctness but to be sure that performance issues don't get reintroduced.

pokey:
> update from meetup: josh to add a simple regression test not only of correctness but to be sure that performance issues don't get reintroduced.

What's your approach to detecting perform regressions?

packages/cursorless-engine/src/util/uniqWithHash.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Like lodash.uniqWith, but uses a hash function to (mostly) avoid quadratic runtime.
+ * @param array The array to uniq
+ * @param fn The equality function
+ * @param hash The hash function. It must be a valid hash function, insofar as it must return the same value for equal items. A hash function that returns a constant value is equivalent to lodash.uniqWith.
+ * @returns The uniq array
+ */
+export function uniqWithHash<T>(
+  array: T[],
+  fn: (a: T, b: T) => boolean,
```
pokey, line 12: I might rename this `isEqual` just to make it super clear what it does?

packages/cursorless-engine/src/util/uniqWithHash.ts
```diff
@@ -0,0 +1,74 @@
...
+  hash: (t: T) => string,
+): T[] {
+  // Handle the common, tiny cases without allocating anything extra.
+  if (array.length < 2) {
+    return [...array];
+  }
+  if (array.length === 2) {
+    if (fn(array[0]!, array[1]!)) {
```
pokey, line 20: do you need the `!` here?

packages/cursorless-engine/src/util/uniqWithHash.ts
```diff
@@ -0,0 +1,74 @@
...
+      return [array[0]!];
+    }
+    return [...array];
+  }
+  // First, split up the array using the hash function.
+  // This keeps the sets of items passed to uniqWith small,
+  // so that the quadratic runtime of uniqWith less of a problem.
+  // Keep track of which sets have multiple items, so that we can uniq them.
```
pokey, line 28: I think this line should be a JSDoc so we get hover hints on `needsUniq`
```suggestion
  /** Keep track of which sets have multiple items, so that we can uniq them. */
```

packages/cursorless-engine/src/util/uniqWithHash.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+    acc.get(key)!.push(item);
+    return acc;
+  }, new Map<string, T[]>());
+
+  // For hash collisions, uniq the items,
+  // letting uniqWith provide correct semantics.
+  needsUniq.forEach((key) => {
+    hashToItems.set(key, uniqWith(hashToItems.get(key)!, fn));
+  });
```
pokey, line 48: very much nitpicking here, but I think clearer to move this one after the next if statement

packages/cursorless-engine/src/util/uniqWithHash.ts
```diff
@@ -0,0 +1,74 @@
...
+      return [];
+    }
+    const first = items[0]!;
+    if (!isEqual(first, item)) {
+      // Removed by uniqWith.
+      return [];
+    }
+    // Emit item.
+    items.shift();
+    hashToItems.set(key, items);
```
pokey, line 71: Is this necessary?

packages/cursorless-engine/src/util/uniqWithHash.ts
```diff
@@ -0,0 +1,74 @@
...
+  return array.flatMap((item) => {
+    const key = hash(item);
+    const items = hashToItems.get(key)!;
+    if (items == null || items.length === 0) {
+      // Removed by uniqWith.
+      return [];
+    }
+    const first = items[0]!;
+    if (!isEqual(first, item)) {
+      // Removed by uniqWith.
```
pokey, line 66: So this is relying on the facts that:

1. We add them to `items` in the same order they originally appear in `array`, and
2. Lodash's `uniqWith` preserves order?

I believe if either of those assumptions were violated, you wouldn't be able to return `[]` here, right? Might be worth pointing this fact out if I'm correct, so we don't accidentally violate those assumptions later

packages/common/src/types/Selection.ts
```diff
@@ -72,4 +72,12 @@ export class Selection extends Range {
       this.anchor.isEqual(other.anchor) && this.active.isEqual(other.active)
     );
   }
+
+  /**
+   * Return a concise string representation of the selection
+   * @returns concise representation
+   **/
+  public concise(): string {
```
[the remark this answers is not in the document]
pokey: Ok, I see now that you're using this function not just for debugging, so I'm not sure I'd just rely on `toString()` as the hash function.

I'd still be tempted to add a `toString` function that just calls this one tho, so we get it during debugging without thinking

pokey (commented):
Very cool! Left some minor comments

pokey:
> > What's your approach to detecting performance regressions?
> 
> 
> 
> I don't know the preferred way in typescript. As a hack, I was going to call uniqWithHash on a large array. (Maybe ten thousand elements? Need to experiment.) If the performance goes quadratic again, the test will start timing out. I'm very open to better ideas, or to skipping it.

Hmm not sure whether it's worth it just for this function, as I don't think it's likely this function would go quadratic. Seems more likely we'd get a quadratic somewhere else we're not expecting it. Would be tempted to have a "take every char" on a large-ish file, tho might want to run it only in CI and only on Ubuntu if it's slow

packages/common/src/types/Selection.ts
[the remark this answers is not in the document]
josharian: indeed it does. done.

packages/cursorless-engine/src/util/uniqWithHash.ts
josharian: done

packages/cursorless-engine/src/util/uniqWithHash.ts
josharian: nope! done

packages/cursorless-engine/src/util/uniqWithHash.ts
josharian: done

packages/cursorless-engine/src/util/uniqWithHash.ts
josharian: done (now it is === 2)

packages/cursorless-engine/src/util/uniqWithHash.ts
josharian: in go, yes...

done

packages/cursorless-engine/src/util/uniqWithHash.ts
josharian: those assumptions are necessary, precisely becausewe need to preserve Lodash's behavior that uniqWith preserves order.

the tests cover this quite well, because it was one of the annoying parts of implementing this. I'll add a comment.

josharian:
thanks for the review, please take another look

pokey (approved):
I made a few minor changes in dca325a29639ccdc72da30e31d23a9008d1548f5; merge if happy 👍

packages/cursorless-engine/package.json
```diff
@@ -27,6 +27,7 @@
     "@types/mocha": "^8.0.4",
     "@types/sbd": "^1.0.3",
     "@types/sinon": "^10.0.2",
+    "fast-check": "3.12.0",
```
auscompgeek, line 30: Curious why fast-check is being pinned?

josharian: Because that's what `pnpm install` did. 🤷🏽 Happy to change to anything you prefer.

pokey: Huh yeah we should look into that. I think that has to do with a setting I set set so that pnpm would install local dependencies using `*`

I also think once we activate renovate bot we may want to pin anyway, tho there's obv a discussion to be had there

PR #1836, opened 2023-08-26
WIP: Add more hats

MartinRykfors:
This pull request adds a bunch of new hats. I am pretty happy with how they are now, though I think there are some issues that need further refinement but I am opening this PR so others can take part and give their feedback.

My goal for the hats has been:

* Chunky designs so that their colors can be distinguished even at small scales.

* Easy, intuitive spoken forms.

* Avoiding conflicts (visually and phonetically) with other hats, old as new.

A preview of the new hats can be found here: https://github.com/MartinRykfors/cursorless_hat_sandbox/blob/master/preview.md

The preview includes a list of alternative spoken forms that one can add to their `hat_styles.csv`.

To start things off, here is an unorganized list of what I think of the current state of the new hats:

* There is a collision between `meeple` and `star` because they have similar shapes. One should probably be deleted. `meeple` is easier to see when it is small, but it could be deleted if `star` is made chunkier. `star` is a more intuitive name for the hat, but it could cause ambiguities with the star character (`post star star` becomes ambiguous, like the `red red` problem).

* Talon will match `star` much more strongly than `stair` even when `stair` is enabled. The default spoken form for `stair` could be made to be `step`.

* I had a hard time making `leaf` be leaf-like without being reduced to a formless blob and I think it shows.

* `singer` has a spoken form with two syllables and it refers to a brand of sewing machines, so the connection between the shape and spoken form is kind of tenuous. I like the shape though because it is very distinct and chunky.

* `stupa` has a problem with being blobby and losing its distinct shape at small scales.

* `mosque` has small details that become fuzzy at small scales.

* `pail` and `cone` become very similar with the new hat rendering, when the hats start growing only vertically once they reach their maximum width. Also, I found that having a hat that is another hat but reflected would cause me to start mixing up the two, even if they are visually distinct. One should probably be deleted.

To test out the new hats you will need to checkout this branch and symlink your `.talon/user/cursorless-talon` to the one in the repo, then bulid the extension locally.

## Checklist

- [X] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)


pokey:
I think we're going to close this, as it's unlikely we will be accepting new hats, but we do have the custom svg setting, as I believe you may be aware, and @AndreasArvidsson and I had an idea for how we might be able to support extra hat shapes in a simple way. Stay tuned on that if you're interested

MartinRykfors:
Yes, this should be closed. I'd be interested in hearing about your idea. :+1:

pokey:
at a very high level, the idea is as follows:

- if any svgs appear in custom svg directory (currently a hidden setting), render those as hats with id `user.xxx` where `xxx` gets replaced with the filename of the svg
- change the talon-side `hat_styles.csv` to allow the user to put lines of the form `user.xxx` (where `xxx` can be anything)

In theory, this is an easy way to allow as many shapes as you want without any adverse impact on users who don't know about / need this functionality. You won't be able to configure them, but that is probably ok for a start. We could probably also secretly allow you to add extra keys to the various shape configuration dicts, ie don't change our schema but just read them if they appear

PR #1838, opened 2023-08-31
optimize getTokenRemainingHatCandidates

josharian:
Object.entries + map allocates a copy of the list twice.
And it ends up calling a tiny function over and over.

Using a boring nested for loop and an array
cuts hat allocation time by about 35%.

Removing items from an array one at a time is quadratic (triangular).
However, N is capped at the number of hats a user has (per grapheme),
which is fundamentally capped at a fairly small number.
And copying an array is typically a highly optimized operation.

This bit of hot code showed up in a profile from user saidelike on Slack.

I have tested that this does not modify behavior
on a branch with golden hat tests.

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
+++ b/packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
@@ -69,7 +69,7 @@ export function allocateHats(
   );
 
   /* All initially enabled hat styles. */
-  const allEnabledHatStyles = Object.keys(enabledHatStyles);
+  const enabledHatStyleNames = Object.keys(enabledHatStyles);
 
   /**
    * A map from graphemes to the remaining hat styles that have not yet been
@@ -80,7 +80,7 @@ export function allocateHats(
    * and this is a hot path.
    */
   const graphemeRemainingHatCandidates = new DefaultMap<string, HatStyleName[]>(
-    () => [...allEnabledHatStyles],
+    () => [...enabledHatStyleNames],
   );
 
   // Iterate through tokens in order of decreasing rank, assigning each one a
--- a/packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
+++ b/packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
@@ -9,7 +9,6 @@ import {
   Token,
   TokenHat,
 } from "@cursorless/common";
-import { clone } from "lodash";
 import { Grapheme, TokenGraphemeSplitter } from "../../tokenGraphemeSplitter";
 import { chooseTokenHat } from "./chooseTokenHat";
 import { getHatRankingContext } from "./getHatRankingContext";
@@ -69,13 +68,19 @@ export function allocateHats(
     tokenGraphemeSplitter,
   );
 
+  /* All initially enabled hat styles. */
+  const allEnabledHatStyles = Object.keys(enabledHatStyles);
+
   /**
    * A map from graphemes to the remaining hat styles that have not yet been
    * used for that grapheme.  As we assign hats to tokens, we remove them from
-   * these lists so that they don't get used again in this pass.
+   * these arrays so that they don't get used again in this pass.
+   * We use an array rather than a map to make iteration cheaper;
+   * we only remove elements once, but we iterate many times,
+   * and this is a hot path.
    */
-  const graphemeRemainingHatCandidates = new DefaultMap<string, HatStyleMap>(
-    () => clone(enabledHatStyles),
+  const graphemeRemainingHatCandidates = new DefaultMap<string, HatStyleName[]>(
+    () => [...allEnabledHatStyles],
   );
 
   // Iterate through tokens in order of decreasing rank, assigning each one a
@@ -90,6 +95,7 @@ export function allocateHats(
         tokenGraphemeSplitter,
         token,
         graphemeRemainingHatCandidates,
+        enabledHatStyles,
       );
 
       const chosenHat = chooseTokenHat(
@@ -107,10 +113,12 @@ export function allocateHats(
       }
 
       // Remove the hat we chose from consideration for lower ranked tokens
-      delete graphemeRemainingHatCandidates.get(chosenHat.grapheme.text)[
-        chosenHat.style
-      ];
-
+      graphemeRemainingHatCandidates.set(
+        chosenHat.grapheme.text,
+        graphemeRemainingHatCandidates
+          .get(chosenHat.grapheme.text)
+          .filter((style) => style !== chosenHat.style),
+      );
       return constructHatRangeDescriptor(token, chosenHat);
     })
     .filter((value): value is TokenHat => value != null);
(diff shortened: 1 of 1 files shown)


josharian:
don't worry, I'm not gonna continue on my current trajectory of replacing all your pretty functional code with boring go-like imperative code. :)

my current goal is to get the golden hat test branch fast enough that I doesn't time out in CI. I believe that with this and the other outstanding pull requests, this should be the end of it. 🤞🏽

packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
```diff
@@ -107,10 +113,12 @@ export function allocateHats(
...
-      delete graphemeRemainingHatCandidates.get(chosenHat.grapheme.text)[
-        chosenHat.style
-      ];
-
+      graphemeRemainingHatCandidates.set(
+        chosenHat.grapheme.text,
+        graphemeRemainingHatCandidates
+          .get(chosenHat.grapheme.text)
+          .filter((style) => style !== chosenHat.style),
+      );
```
pokey, line 121: Wow is this really faster than what I had? Performance tuning can be quite counterintuitive 😅

packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
```diff
@@ -135,19 +143,27 @@ function getTokenOldHatMap(oldTokenHats: readonly TokenHat[]) {
 function getTokenRemainingHatCandidates(
   tokenGraphemeSplitter: TokenGraphemeSplitter,
   token: Token,
-  availableGraphemeStyles: DefaultMap<string, HatStyleMap>,
+  availableGraphemeStyles: DefaultMap<string, HatStyleName[]>,
```
pokey, line 146: why not call this one `graphemeRemainingHatCandidates`, as we do in the calling function?

packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
```diff
@@ -135,19 +143,27 @@ function getTokenOldHatMap(oldTokenHats: readonly TokenHat[]) {
...
+  allHatStyles: HatStyleMap,
```
pokey, line 147: why not call this `enabledHatStyles` like we do in the calling function?

packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
```diff
@@ -135,19 +143,27 @@ function getTokenOldHatMap(oldTokenHats: readonly TokenHat[]) {
...
-        }),
-      ),
-    );
+  // Use iteration here instead of functional constructs,
+  // because this is a hot path and we want to avoid allocating arrays
+  // and calling tiny functions lots of times.
+  const candidates: HatCandidate[] = [];
+  const graphemes = tokenGraphemeSplitter.getTokenGraphemes(token.text);
+  for (const grapheme of graphemes) {
+    for (const style of availableGraphemeStyles.get(grapheme.text)!) {
```
pokey, line 155: do we need the `!`?

pokey (commented):
Good stuff! Left a few minor comments

packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
auscompgeek: Ah, here's a reference for you: https://v8.dev/blog/fast-properties

pokey: yikes 😅. thanks for the reference 🙏

packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
josharian: deeply ingrained habits from go. these are already extraordinarily long by my usual standards. :P

see https://google.github.io/styleguide/go/decisions#variable-names

anyway, I'm in rome, I have fixed them :)

packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
josharian: done

packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
josharian: nope! I did at some point... removed. :)

josharian:
ptal :)

packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
josharian: @auscompgeek fabulous reference. Thank you!

The slowest part now by far is this line:

```ts
      candidates.push({
        grapheme,
        style,
        penalty: enabledHatStyles[style].penalty,
      });
```

Any tips for that? I can't tell whether it is object allocation, array growth, or what. Nor would I know what to do with that. (In Go, I would pre-allocate the slice and use an object pool.)

Any ideas? :)

packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
pokey: Definitely some interesting heuristics in there. I still err strongly on the side of verbosity to avoid ambiguity, and I also like the visual echoing of incorporating a full name into a bigger name

My biggest gripe with the linked naming convention is that I think omitting words due to context forces you to keep more of the context in your head. I would argue that the cognitive load is higher when one tries to be clever with naming and omit words.

For example, I could see shortening the variable to `graphemeRemainingStyles`, but `hatStyle` is a term of art, and we try to be consistent about using it everywhere. If we sometimes use `style`, and sometimes `hatStyle`, I'd argue that makes the code harder to read

That being said, I do like some of the ideas in there, esp the notion of taking into account length / complexity of variable scope, and the idea that "the name of a local variable should reflect what it contains and how it is being used in the current context, rather than where the value originated"

Cc/ @AndreasArvidsson

pokey (approved):
Looks good!

packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
pokey: Maybe split it into 3 lines so profiling can tell you which piece is the problem? (Ie first get penalty, then construct object, then push)

josharian: Oh right, now I remember I did that. it is all in constructing the object.

packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
pokey: Wow yeah very different-looking 😅

I def didn't take it as proselytising, no worries 😊

packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
pokey: Oh interesting. I wonder if there's a way to make that object construction faster. Any ideas @auscompgeek

auscompgeek: I can't think of any good ideas other than turning this into WebAssembly instead, sorry 😅

pokey: Ha yeah fair. Rewrite in rust / zig / c into wasm would prob be much faster. But would hurt ability to easily experiment with this code I think

As a less dramatic change, it appears switching to a function prototype might speed things up. See eg https://www.measurethat.net/Benchmarks/Show/19556/0/es6-class-vs-prototype-vs-object-literal

We could also potentially switch to an array, tho that would make all the code that consumes these objects a lot less readable

josharian: > I can't think of any good ideas other than turning this into WebAssembly instead, sorry 😅
> [...] Rewrite in rust [edit: or Go] [...]

omg don't tempt me

> it appears switching to a function prototype might speed things up

Using a class cuts about 5%, with fairly minimal readability impact (diff below). But it's hard to get excited about that. Using a function prototype shaves off a few more percent, but makes typescript grouchy. I think this is probably not worth the code.

> We could also potentially switch to an array

At some point I experimented with an array. The diff was prohibitive--we'd be better off with webasm.

---

```diff
diff --git a/packages/cursorless-engine/src/util/allocateHats/allocateHats.ts b/packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
index 09bb49fc..ab3c5f92 100644
--- a/packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
+++ b/packages/cursorless-engine/src/util/allocateHats/allocateHats.ts
@@ -140,6 +140,18 @@ function getTokenOldHatMap(oldTokenHats: readonly TokenHat[]) {
   return tokenOldHatMap;
 }
 
+class ConcreteHatCandidate implements HatCandidate {
+  grapheme: Grapheme;
+  style: HatStyleName;
+  penalty: number;
+
+  constructor(grapheme: Grapheme, style: HatStyleName, penalty: number) {
+    this.grapheme = grapheme;
+    this.style = style;
+    this.penalty = penalty;
+  }
+}
+
 function getTokenRemainingHatCandidates(
   tokenGraphemeSplitter: TokenGraphemeSplitter,
   token: Token,
@@ -156,11 +168,13 @@ function getTokenRemainingHatCandidates(
       // Allocating and pushing all of these objects is
       // the single most expensive thing in hat allocation.
       // Please pay attention to performance when modifying this code.
-      candidates.push({
-        grapheme,
-        style,
-        penalty: enabledHatStyles[style].penalty,
-      });
+      candidates.push(
+        new ConcreteHatCandidate(
+          grapheme,
+          style,
+          enabledHatStyles[style].penalty,
+        ),
+      );
     }
   }
   return candidates;
```

PR #1839, opened 2023-09-01
migrate Go keys and values to treesitter

josharian:
This doesn't handle all uses of key, value, and name.
Missing still are migrate Go key and value scope to treesitter

This doesn't handle all uses of key and value.

Missing still are assignment statements (name, value),
function signatures (name, value), and possibly other
places (perhaps type declarations?).

It does handle all existing uses of key, value, and name,
though, so we can start here.

The handling of multiple return values is complicated,
particularly around the handling of comments and removal ranges.
It would be nice to find a simpler approach.

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
+++ b/packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
@@ -73,6 +73,22 @@ class HasMultipleChildrenOfType extends QueryPredicateOperator<HasMultipleChildr
   }
 }
 
+/**
+ * A predicate operator that returns true if the node has more than 1 child not of
+ * type {@link type} (inclusive).  For example, `(has-multiple-children-not-of-type?
+ * @foo bar)` will accept the match if the `@foo` capture has 2 or more children
+ * not of type `bar`.
+ */
+class HasMultipleChildrenNotOfType extends QueryPredicateOperator<HasMultipleChildrenNotOfType> {
+  name = "has-multiple-children-not-of-type?" as const;
+  schema = z.tuple([q.node, q.string]);
+
+  run({ node }: MutableQueryCapture, type: string) {
+    const count = node.children.filter((n) => n.type !== type).length;
+    return count > 1;
+  }
+}
+
 class ChildRange extends QueryPredicateOperator<ChildRange> {
   name = "child-range!" as const;
   schema = z.union([
@@ -187,4 +203,5 @@ export const queryPredicateOperators = [
   new AllowMultiple(),
   new InsertionDelimiter(),
   new HasMultipleChildrenOfType(),
+  new HasMultipleChildrenNotOfType(),
 ];
--- a/queries/go.scm
+++ b/queries/go.scm
@@ -222,3 +222,87 @@
     .
   )
 ) @anonymousFunction @namedFunction
+
+;; keys in maps
+(literal_value
+  "{" @collectionKey.iteration.start.endOf
+  (keyed_element
+    (_) @collectionKey @collectionKey.trailing.start.endOf
+    ":"
+    (_) @collectionKey.trailing.end.startOf
+  ) @collectionKey.domain
+  "}" @collectionKey.iteration.end.startOf
+)
+
+;; values in maps
+(literal_value
+  "{" @value.iteration.start.endOf
+  (keyed_element
+    (_) @value.leading.start.endOf
+    ":"
+    (_) @value @value.leading.end.startOf
+  ) @value.domain
+  "}" @value.iteration.end.startOf
+)
+
+;; values in return statements
+
+;; one value within a return statement
+(return_statement
+  (expression_list
+    .
+    (_)
+    .
+  ) @value
+  (#insertion-delimiter! @value ", ")
+) @value.domain @value.iteration
+
+;; multiple values within a return statement
+
+;; NB: gofmt puts the comma after block comments in lists of things
+;;
+;; Like this:
+;;   return "lorem" /* comment */, "ipsum"
+;; Not like this:
+;;   return "lorem", /* comment */ "ipsum"
+;;
+;; It's really hard to deal with commas both before and after,
+;; and they're rare anyway, so assume gofmt for now.
+;; Non-gofmt commas will mess up removal ranges.
+
+;; the first value
+
+;; BUG: in this code:
+;;   return "lorem" /* comment */ , "ipsum"
+;; the comment is included in the removal range of "lorem".
+;; This is too hard to fix now, because it would require
+;; disjoint removal ranges. And it is rare anyway.
+
+;; the first of many return values...
+(return_statement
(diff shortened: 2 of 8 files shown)


josharian:
I haven't added any tests yet. Before I did that, I wanted to get around of feedback about the behavior, so that I don't waste time adding tests with the wrong behavior.

There are _a lot_ of tests to add.

To make it easier to get initial feedback I have included two videos of the scope visualizer.

josharian:
return values:

https://github.com/cursorless-dev/cursorless/assets/67496/e15d9ba4-7063-4526-bc0c-4c215cb779ca

josharian:
keys and values in maps and lists:


https://github.com/cursorless-dev/cursorless/assets/67496/7b7675ef-1fd2-4af1-90bb-46d6668f4413

pokey:
> keys and values in maps and lists:
> 
>  visualize_key_value.mov

this looks great!

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
```diff
@@ -73,6 +73,22 @@ class HasMultipleChildrenOfType extends QueryPredicateOperator<HasMultipleChildr
   }
 }
 
+/**
+ * A predicate operator that returns true if the node has more than 1 child not of
+ * type {@link type} (inclusive).  For example, `(has-multiple-children-not-of-type?
+ * @foo bar)` will accept the match if the `@foo` capture has 2 or more children
+ * not of type `bar`.
+ */
+class HasMultipleChildrenNotOfType extends QueryPredicateOperator<HasMultipleChildrenNotOfType> {
```
pokey, line 82: Wow this really feels like a slippery slope 😅. Getting to the point where I wonder if we want to create our own mini-language

queries/go.scm
```diff
@@ -222,3 +222,87 @@
...
+
+;; values in return statements
+
+;; one value within a return statement
+(return_statement
+  (expression_list
+    .
+    (_)
+    .
+  ) @value
```
pokey, line 256: what happens with comments here?

queries/go.scm
```diff
@@ -222,3 +222,87 @@
...
+;; NB: gofmt puts the comma after block comments in lists of things
+;;
+;; Like this:
+;;   return "lorem" /* comment */, "ipsum"
+;; Not like this:
+;;   return "lorem", /* comment */ "ipsum"
+;;
+;; It's really hard to deal with commas both before and after,
+;; and they're rare anyway, so assume gofmt for now.
+;; Non-gofmt commas will mess up removal ranges.
```
pokey, line 271: I generally lean slightly towards being conservative wrt removing comments, but this behaviour might be nice if it's almost always what you want. I'll defer to you on this one

queries/go.scm
```diff
@@ -222,3 +222,87 @@
...
+
+;; the first value
+
+;; BUG: in this code:
+;;   return "lorem" /* comment */ , "ipsum"
+;; the comment is included in the removal range of "lorem".
```
pokey, line 277: is this a bug? I might argue the comment is part of `"lorem"`, tho as mentioned above, I'd lean towards being conservative about deleting people's comments

queries/go.scm
```diff
@@ -222,3 +222,87 @@
...
+;; This is too hard to fix now, because it would require
+;; disjoint removal ranges. And it is rare anyway.
+
+;; the first of many return values...
+(return_statement
+  (expression_list
+    .
+    (_) @value @value.trailing.start.endOf
+    (#insertion-delimiter! @value ", ")
```
pokey, line 286: interesting to put predicate right after the capture itself

queries/go.scm
```diff
@@ -222,3 +222,87 @@
...
+    .
+    (comment)* @value.trailing.omit
+    .
+    ","
+    .
+    (_) @value.trailing.end.startOf
+  ) @_exprlist @value.leading.start.startOf
```
pokey, line 293: the `_` prefix is your convention for things that are just dummy captures used for predicates? We've been using `dummy`, but I kinda like this `_` prefix convention. wdyt @AndreasArvidsson

queries/go.scm
```diff
@@ -222,3 +222,87 @@
...
+  (#not-type? @value comment)
+  (#has-multiple-children-not-of-type? @_exprlist comment)
```
pokey, line 295: it's a bit annoying we need to use negative specifier here rather than specifying the type itself, but I'm assuming the individual expressions have no wrapper nodes of uniform type, so I'm not sure what else we can do 🤷‍♂️

queries/go.scm
```diff
@@ -222,3 +222,87 @@
...
+) @value.iteration
+
+;; ...and the rest of the values
```
pokey, line 298: would be nice to have a comment explaining why you need to handle first / rest separately

pokey (commented):
Solid direction; left a bunch of minor inline comments, and see also my responses to your videos above 👍

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
AndreasArvidsson: dsl ftw! :D

queries/go.scm
AndreasArvidsson: Sounds good to me

queries/go.scm
```diff
@@ -222,3 +222,87 @@
...
+;; keys in maps
+(literal_value
+  "{" @collectionKey.iteration.start.endOf
+  (keyed_element
+    (_) @collectionKey @collectionKey.trailing.start.endOf
+    ":"
+    (_) @collectionKey.trailing.end.startOf
+  ) @collectionKey.domain
+  "}" @collectionKey.iteration.end.startOf
+)
```
pokey, line 235: I have mixed feelings about defining a scope’s pattern and its iteration scope’s pattern in one go. I think it's often clearer to define them separately. You could arguably define `collectionKey.iteration` and `@value.iteration` in one pattern. I think that might be clearer

That being said, it's more of a rule of thumb than a hard rule; if you feel defining `collectionKey` and `collectionKey.iteration` in the same pattern is clearer in this case, I don't have a strong objection

There will potentially be a slight performance implication later fwiw, because this pattern will yield the iteration scope once per scope instance in the iteration scope, and we silently merge them internally. Defining the pattern separately will only yield the iteration scope once

josharian:
I need to take another pass over this, marking as draft in the interim.

pokey:
update: I'll have a look to see if there's stuff we can easily pick up to take this one home

josharian:
> update: I'll have a look to see if there's stuff we can easily pick up to take this one home

Sounds good, thanks. I'm eyeball deep in other stuff, so I don't anticipate much cursorless bandwidth for a fair while.

pokey:
Removing myself as assignee so that someone else can pick it up

PR #1844, opened 2023-09-03
Elixir support

r-tae:
## What

Adds support for the `elixir` programming language

## Checklist

- [ ] Added a few specific tests that use `"chuck"` instead of `"change"` to test removal behaviour when it's interesting, especially:
  - [ ] `"chuck arg"` with single argument in list
  - [ ] `"chuck arg"` with multiple arguments in list
  - [ ] `"chuck item"` with single argument in list
  - [ ] `"chuck item"` with multiple arguments in list
- [/] Supported` "type"` both for type annotations (eg `foo: string`) and declarations (eg `interface Foo {}`) (and added tests for this behaviour 😊)
- [ ] Supported` "item"` both for map pairs and list entries (with tests of course)
- [ ] Add support for `"name"`


queries/elixir.scm
```diff
@@ -0,0 +1,41 @@
+(comment) @comment
+(anonymous_function) @anonymousFunction
+(arguments) @arguments_list
```
pokey, line 3: I don't think this is a real scope 🤔. We'd like to support it, though. But if we did, you'd probably want to be able to target it from anywhere in the function.

queries/elixir.scm
```diff
@@ -0,0 +1,41 @@
+(comment) @comment
+(anonymous_function) @anonymousFunction
```
pokey, line 2: would be good to have tests for these

queries/elixir.scm
```diff
@@ -0,0 +1,41 @@
...
+(arguments) @arguments_list
+(call
+  target: (_) @functionCallee) @functionCall @functionCallee.domain
```
pokey, line 5: Please install [Andreas's VSCode extension](https://marketplace.visualstudio.com/items?itemName=AndreasArvidsson.andreas-talon) to get correct formatting here. We will eventually format in CI but we're not there yet for this file type

Also, would be good to get tests for this scope

queries/elixir.scm
```diff
@@ -0,0 +1,41 @@
...
+
+;;(keywords
+;;  (pair
+;;    key: (_) @collectionKey @collectionItem.trailing.end.startOf
+;;    value: (_) @value) @collectionItem @collectionItem.trailing.start.endOf
+;;    (",")? @collectionItem.trailing.start.startOf
+;;    (#allow-multiple! @collectionItem))
```
pokey, line 12: did you want to discuss this one?

queries/elixir.scm
```diff
@@ -0,0 +1,41 @@
...
+
+;; How does surrounding whitespace work for this
+(map_content
+  (binary_operator
+      left: (_) @collectionKey @collectionKey.trailing.start.endOf
+      right: (_) @value @collectionKey.trailing.end.startOf) @collectionItem
```
pokey, line 18: you want the `binary_operator` to be the domain for key and value, and we'll want a test for that

queries/elixir.scm
```diff
@@ -0,0 +1,41 @@
...
+      ","? @collectionItem.trailing
```
pokey, line 19: this isn't quite right; you're missing the an optional `.leading` as well. And the whitespace is a bit off.  Have a look at https://github.com/cursorless-dev/cursorless/blob/c5ef8fc559302c721dff8dbf25b3501a5f83fbdd/queries/talon.scm#L168-L179. Tho keep in mind that one might behave strangely when there are comments mixed between args. That's still kind of uncharted territory as all this stuff is fairly new. Worth wiring it up and testing it with comments in between the items to see what happens.

queries/elixir.scm
```diff
@@ -0,0 +1,41 @@
...
+      (#shrink-to-match! @collectionItem "([^,]+)"))
```
pokey, line 20: What's this doing?

queries/elixir.scm
```diff
@@ -0,0 +1,41 @@
...
+
+(call
+  target: (identifier) @ignore
```
pokey, line 23: We tend to use the term `dummy` for things that aren't scope types, and are just used for predicates. At some point that may be enforced

queries/elixir.scm
```diff
@@ -0,0 +1,41 @@
...
+        left: (call target: (identifier) @functionName)
+        operator: "when")
+    ])
+    (do_block
+      .
+      "do" @namedFunction.interior.start.endOf
+      .
+      "end" @namedFunction.interior.end.startOf
+      .)?
+  (#match? @ignore "^(def|defp|defdelegate|defguard|defguardp|defmacro|defmacrop|defn|defnp)$"))  @namedFunction @functionName.domain
```
pokey, line 41: Does this predicate do anything? I've never used it before, but I think you're right that it should work out of the box

queries/elixir.scm
```diff
@@ -0,0 +1,41 @@
...
+      ; function clause with a guard clause
+      (binary_operator
+        left: (call target: (identifier) @functionName)
+        operator: "when")
+    ])
+    (do_block
+      .
+      "do" @namedFunction.interior.start.endOf
+      .
+      "end" @namedFunction.interior.end.startOf
```
pokey, line 39: worth having some "inside funk" tests for this one

queries/elixir.scm
```diff
@@ -0,0 +1,41 @@
...
+      right: (_) @value @collectionKey.trailing.end.startOf) @collectionItem
+      ","? @collectionItem.trailing
+      (#shrink-to-match! @collectionItem "([^,]+)"))
+
+(call
+  target: (identifier) @ignore
+  (arguments
+    [
+      ; zero-arity functions with no parentheses
+      (identifier) @functionName
```
pokey, line 27: worth having some `"funk name"` tests

pokey (commented):
wow off to a fantastic start 😍! I left a bunch of small comments. Feel free to drop into a meet-up as well; we have them 3 times a week on various timezones

queries/elixir.scm
r-tae: I was trying to fix `item removal` being greedy, it's just from me still learning how the queries work. I see that a period does roughly what I was looking for, am I more on the right track here?:

```
(map_content
  (binary_operator
    left: (_) @collectionKey
    right: (_) @value
  ) @collectionItem
  .
  ","? @collectionItem.trailing
)
```

queries/elixir.scm
r-tae: I pulled it from the [`ts-elixir` repo](https://github.com/elixir-lang/tree-sitter-elixir/blob/a2861e88a730287a60c11ea9299c033c7d076e30/queries/tags.scm#L10), it seems to be working

queries/elixir.scm
r-tae: Is this looking better?

```
;; %{:a => "lorem", "b" => "ipsum", 3 => "dolor"},
(map_content
  (_)? @_.leading.start.endOf
  .
  (binary_operator
    left: (_) @collectionKey
    right: (_) @value
  ) @collectionItem @_.leading.end.startOf @_.trailing.start.endOf
  .
  (_)? @_.trailing.end.startOf
  (#insertion-delimiter! @collectionItem ", ")
)
```

queries/elixir.scm
pokey: ah yeah a `.` should fix that 👍

queries/elixir.scm
pokey: although see https://github.com/cursorless-dev/cursorless/pull/1839#discussion_r1314817213

queries/elixir.scm
pokey: nice. Would be worth a negative test using "cursorless record error", where you try saying "change funk" from something that is a function call that uses a function other than eg `def` to ensure it isn't treated as `"funk"`, if that makes sense

queries/elixir.scm
pokey: not quite, because you only want the outer leading / trailing to be for `collectionItem`, so eg you want `collectionItem.leading.start.endOf` rather than `_.leading.start.endOf`, because `_` is a wildcard

I'd be tempted to give `collectionItem` its own pattern, even if slightly duplicative, to make things easier to read

queries/elixir.scm
r-tae: So this is mostly working at the moment (after removing that extra period), but I need commands like `chuck inside funk` to leave `end` on a separate line, or at least to leave a space between `do end`. Is that possible?

r-tae: Ideally
```
def elixir(foo) do
  bar = 5
end
```
- "chuck inside funk" ->
```
def elixir(foo) do
end
```

Right now it gives `def elixir(foo) doend`, which is a syntax error

queries/elixir.scm
r-tae: I've switched to `_target`, also I liked putting the predicate immediately after the match

queries/elixir.scm
pokey: ooh huh interesting one 😅. as a hack, maybe we could shift the endpoint of the range to the left one?

queries/elixir.scm
```diff
@@ -0,0 +1,60 @@
...
+) @functionCall @functionCallee.domain
+
+(keywords
+  (pair
+    key: (_) @collectionKey @collectionItem.trailing.end.startOf
+    value: (_) @value
+  ) @collectionItem @collectionItem.trailing.start.endOf
+  (
+    ","
+  )
```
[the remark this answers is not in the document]
r-tae: Ah thanks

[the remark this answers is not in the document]
r-tae: This whole `(keywords)` matcher is messed up anyway, I'll rewrite it based on the other one

queries/elixir.scm
```diff
@@ -0,0 +1,60 @@
...
+)
+
+;; %{:a => "lorem", "b" => "ipsum", 3 => "dolor"},
+(map_content
+  (_)? @_.leading.start.endOf
+  .
+  (binary_operator
+    left: (_) @collectionKey
+    right: (_) @value
+  ) @collectionItem @collectionKey.domain @value.domain @_.leading.end.startOf @_.trailing.start.endOf
```
pokey, line 28: unfortunately you can't have more than 3 captures on a single node 😕. They get silently dropped by the tree-sitter parser 😅. Another argument for splitting `collectionItem` handling into its own pattern

r-tae: Good to know 😅

queries/elixir.scm
pokey: Cool. As you can see, this stuff is fairly new for us, so we're still finding the conventions that we like 😊

queries/elixir.scm
r-tae: Thanks! Much nicer with the extension

queries/elixir.scm
r-tae: Interesting, is that something I'd do with a predicate?

pokey: Yes, although I'm not sure the predicate exists yet, now that you mention it. You could maybe use the regex shrink predicate?

queries/elixir.scm
r-tae: Good plan, I've added one

.gitignore
```diff
@@ -42,3 +42,8 @@ next-env.d.ts
 
 # test subset config
 packages/test-harness/testSubsetGrep.properties
+
+data/playground/.metals
+data/playground/.vscode
```
pokey, line 47: what is this one for?

.gitignore
```diff
@@ -42,3 +42,8 @@ next-env.d.ts
...
+data/playground/.scala-build
+.talon
```
pokey, line 49: what about this one?

packages/common/src/extensionDependencies.ts
```diff
@@ -5,7 +5,7 @@ export const extensionDependencies = [
   "scalameta.metals",
   "mrob95.vscode-talonscript",
   "jrieken.vscode-tree-sitter-query",
-
   // Necessary for the `drink cell` and `pour cell` tests
   "ms-toolsai.jupyter",
+  "jakebecker.elixir-ls",
```
pokey, line 10: after you update this one to latest main, you'll want to move this extension a couple lines up and add a comment

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/elixir/changeKey.yml
pokey, line 1: we might want to support "chuck key" / "chuck value" removing the `=>`; might make it easier to convert a map to a list. We support that in other languagse

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/elixir/changeLambda.yml
pokey, line 1: is it common to define a named function like this `add = fn x, y -> x + y end`? Ie would you want to support `"funk"` targeting that whole statement? We do that in typescript, but if it's much more common to define a named function using the other syntax then you might not bother

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/elixir/changePreviousComment.yml
pokey, line 1: I would remove this test. "next" / "previous" are defined independently of your code, so I don't think it's worth testing them here. We usually just use containing scope. But having the "next" instead of usual containing is fine

queries/elixir.scm
```diff
@@ -0,0 +1,121 @@
...
+
+;; Map
+;; %{:a => "lorem", "b" => "ipsum", 3 => "dolor"}
+(map_content
+  (_)? @_.leading.start.endOf
+  .
+  (binary_operator) @collectionItem @_.leading.end.startOf @_.trailing.start.endOf
+  .
+  (_)? @_.trailing.end.startOf
+  (#insertion-delimiter! @collectionItem ", ")
```
pokey, line 15: this won't behave nicely in a vertical list, eg if you have newlines between elements. I'm hesitant to introduce poor clone / pour behaviour for a new language. Maybe you can drop into a meet-up and we can discuss how to do this one right? It hasn't yet been done for any next-gen language

queries/elixir.scm
```diff
@@ -0,0 +1,121 @@
...
+    "end" @class.interior.end.startOf
+    .
+  )
+) @class @className.domain
+
+(source) @className.iteration @class.iteration
+(source) @statement.iteration
+
+(source) @namedFunction.iteration @functionName.iteration
+;; Is it better to have the class as the function iteration scope?
```
pokey, line 116: you want both the file and the class to be function iteration scope

pokey (commented):
Looking good! Left some minor comments. Also

- [ ] There are a few empty checkboxes on PR description
- [ ] It's worth thinking bout whether there are other scopes you want to implement. Eg "state", "map", "list", etc
- [ ] What about items in lists? Seems like you've only done maps, but maybe I missed it
- [ ] I see you checked box for `textFragment`, but I don't see that in the code
- [ ] Looks like you're still missing a few tests, eg `"change item"` (I only see `"chuck"`)

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/elixir/changeLambda.yml
r-tae: It wouldn't feel idiomatic to call this a `namedFunk` imo.

Unlike TS, in Elixir there's a clear split between named functions and lambdas, there's never a case where you could replace one with the other and they're even called differently:

```elixir
# has to be called as `add.(1, 2)`
add = fn x, y -> x + y end

# has to be called as `add(1, 2)`
def add(x, y) do
  x + y
end
```

All that's the say, I think keeping them distinct is valuable and intuitive enough

queries/elixir.scm
r-tae: Yeah sounds good, I'll be at the next Thursday one

.gitignore
r-tae: My mistake, I meant to ignore them in vscode

.gitignore
r-tae: Same for this

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/elixir/changePreviousComment.yml
r-tae: Done, cheers

queries/elixir.scm
r-tae: I believe I've done it, I don't use iteration stuff myself yet and I don't even know how to test if I've done it right 😅

queries/elixir.scm
josharian: I would like to discuss drink and pour as well. I can’t make Thursdays, but hopefully you all can tolerate having multiple conversations about it.

jaresty:
This is great! I'm about to be doing some elixir work. What's left to this? Can I help?

r-tae:
@jaresty help would be great! I haven't had a chance to do any more on this since Sept. The biggest outstanding thing is drink/pour in lists/maps (see Pokey's last comment, I haven't made it to a meetup since).

I don't remember what else is outstanding, but I think its mergable if we get just that working

jaresty:
@r-tae I'd love to help! Could we maybe find some time where you could introduce me to the work?

pokey:
# `"call"`

image

⬆️ I'm a bit surprised that a module definition is considered a call

image

⬆️ It's also surprising that a function definition is considered a call

pokey:
# `"funk"`

image

Are these not function definitions?

pokey:
# `"key"`

image

doesn't look right to me; should just be `a` and `b`

pokey:
# `"value"`

image

Looks like nested values are missing here (ie `"Jane"` and `"jane@example.com"`)

pokey:
Ok I kicked the ball down the field a bit; see [my changes](https://github.com/cursorless-dev/cursorless/pull/1844/files/9f7ad925a151c8df7fad5e61f32621c2471057d7..2471282ea193c63b9a958fc55dad5b908b7d8101) @jaresty I'm happy to get you up to speed on this PR in a meet-up if you want to carry it to the finish line; not too much left

jaresty:
Thank you for that @pokey - I'm unexpectedly in a job search so it may be some time before I can get back to this.

pokey:
ok @r-tae if you wanna bring this home go for it. Otherwise I may take a pass again in the future

jaresty:
@r-tae wanted to let you know that community now has an elixir language mode 😄

PR #1850, opened 2023-09-05
Type hint Talon cheatsheet get_list

auscompgeek:
## Checklist

### DIFF
--- a/cursorless-talon/src/cheatsheet/get_list.py
+++ b/cursorless-talon/src/cheatsheet/get_list.py
@@ -1,33 +1,48 @@
 import re
+from collections.abc import Mapping, Sequence
+from typing import Optional, TypedDict
 
 from talon import registry
 
 from ..conventions import get_cursorless_list_name
 
 
-def get_list(name, type, descriptions=None):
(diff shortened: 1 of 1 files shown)


cursorless-talon/src/cheatsheet/get_list.py
```diff
@@ -1,33 +1,48 @@
...
+
+
+def get_list(
+    name: str, type: str, descriptions: Optional[Mapping[str, str]] = None
+) -> list[ListItemDescriptor]:
     if descriptions is None:
         descriptions = {}
 
     items = get_raw_list(name)
-    item_dict = items if isinstance(items, dict) else {item: item for item in items}
```
auscompgeek, line 13: AFAICT (from the Talon type stubs) we're guaranteed to get out a `dict` when reading the registry like we do here, so we never get a `list` (which is what I presume this is attempting to handle).

pokey: as long as the cheatsheet works then I'm fine with it 👍

pokey (approved):
Looks good!

PR #1851, opened 2023-09-05
csv_overrides: Add stronger type hints

auscompgeek:
## Checklist

### DIFF
--- a/packages/cursorless-engine/src/test/fixtures/communitySnippets.fixture.ts
+++ b/packages/cursorless-engine/src/test/fixtures/communitySnippets.fixture.ts
@@ -22,7 +22,13 @@ const snippetAfterAction: ActionDescriptor = {
     snippets: [
       {
         type: "custom",
-        body: "```\n$0\n```",
+        languages: [
+          "javascript",
+          "typescript",
+          "javascriptreact",
+          "typescriptreact",
+        ],
+        body: 'import * as $0 from "$0";',
       },
     ],
   },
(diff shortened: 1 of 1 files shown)


cursorless-talon/src/csv_overrides.py
```diff
@@ -126,35 +124,44 @@ def on_watch(path, flags):
...
+    key: str
+    value: str
+    #: The list name.
+    list: str
+
+
+def check_for_duplicates(
+    filename: str, default_values: Mapping[str, Mapping[str, str]]
+):
     results_map = {}
```
auscompgeek, line 149: I can't see what type this is supposed to be... because I don't think any items are ever assigned to this :confused:

I recall this bit was in one of the other functions previously. Think this ended up being broken after a recent refactor.

cursorless-talon/src/csv_overrides.py
```diff
@@ -60,27 +62,25 @@ def init_csv_and_watch_changes(
...
+        no_update_file: Set this to `TRUE` to indicate that we should
         not update the csv. This is used generally in case there was an issue coming up with the default set of values so we don't want to persist those to disk
         pluralize_lists: Create plural version of given lists
     """
-    if extra_ignored_values is None:
-        extra_ignored_values = []
-
     file_path = get_full_path(filename)
     super_default_values = get_super_values(default_values)
+    assert allow_unknown_values == (default_list_name is not None)
```
auscompgeek, line 76: Turns out we have dependent types here! It would've been too annoying to check both `allow_unknown_values` and `default_list_name` in `update_dicts` to appease the type checker, but I'm also too lazy to update the callers of `init_csv_and_watch_changes` right now.

pokey:
I see this is still draft—did you want me to have a look or wanted to make some changes first?

auscompgeek:
It'd probably make sense to address the above issue first, since as it stands there are no types that are coherent in `check_for_duplicates`.

pokey:
Ah right yeah makes sense

AndreasArvidsson:
@codex review

AndreasArvidsson (approved):

PR #1805, opened 2023-08-17
add doc contents and row/col as yaml comments

josharian:
experimental, to discuss over voice

interesting diffs:

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded//languages/go/takeType.yml
packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/hatTokenMap/sendEmoji.yml
packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/snippets/snipMakeFunk.yml
packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/snippets/snipSpaghetti.yml
packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/cpp/tryCatchWrapThis2.yml

Issue #1631, opened 2023-07-13
Support destination pseudo-scopes


pokey:
# The problem

Consider the following code

```ts
const aaa = 0;
```

In this case, there is no type. However, we want the following to work:

- `"change type"` from within the statement should insert a `: ` after the `aaa` and leave your cursor after the `: `
- `"bring bat to type air"` should insert a `: ` after the `aaa`, followed by "bat"

# The solution

## Query language update

- In our query language, support `@myScopeType.destination`
- Could be used as follows:

  ```scm
  (
    (lexical_declaration
       (variable_declarator
          name: (_) @type.destination
          !type
       )
    )
    (#end-position! @type.destination)
    (#insertion-delimiter! @type.destination ": ")
  )
  ```

## How to handle these queries

The naive approach is to create a stub target that only has a `getDestination` method and throws on everything else.  Unfortunately the problem is that this would break the following:

```ts
function aaa(): void {
    const bbb = 0;
}
```

If you say `"copy type bat"`, you'll get an error, because it will get trapped in the pseudo-scope.  Maybe that's fine because it could be confusing to have `"copy type bat"` and `"change type bat"` refer to different scopes (the latter is legitimately a destination pseudo-scope that allows adding a type to `bbb`)

pokey:
update from meet-up:

- "drink", "pour", and "change" all do the same thing when you have no type
- "drink" and "pour" don't work when you have a type
- If you say "take type" from within something that has no type but could (ie has a destination pseudo-scope), it is an error.  This is useful when combined with https://github.com/cursorless-dev/cursorless/issues/1015
- for something that can have more than one (eg "name" in scm)
- Add `(#prefix-insertion-delimiter! ": ")` predicate
- Add `(#postfix-insertion-delimiter! ": ")` predicate
- Rename `(#insertion-delimiter!)` to `#infix-insertion-delimiter`?
- if no infix, then we don't support "pour" when there is already 1

pokey:
Note that the interaction with #1015 is a bit nontrivial because for eg "pour soft type" or "bring air to soft type", we actually want "soft" to apply to the conversion to a destination rather than to the modifier after it. Not sure how to distinguish that from wanting it to apply to the modifier after it 🤔

pokey:
[...]

PR #1854, opened 2023-09-05
add Go branch scope support

josharian:
Not all the behavior is perfect (I'm looking at you, pour and drink), but it's better than not existing.

The perfect is the enemy of the good.

## Checklist


josharian:
outstanding:

* [ ] figure out how to get drink, pour, and chuck all to do the right thing at the same time (sigh)
* [ ] add tests once behavior is correct
* [ ] move TODOs to some personal list, or just implement; maybe move over condition at the same time?

note also this is a potentially interesting case for https://github.com/cursorless-dev/cursorless/issues/1631

pokey:
Looking at your scope visualizer screenshots from slack, it appears that the problem is that tree-sitter-go includes the trailing newline in the node's range. I'd recommend using our shrink regex predicate operator to exclude it. Hopefully that will fix your various issues.

Keep in mind that "chuck branch" will leave an empty line, but pretty much every other scope type has the same problem, and we're planning to fix all tree-sitter whitespace removal using the #855 hammer

josharian:
Range:

Screenshot 2023-09-06 at 8 07 56 PM

Removal:

Screenshot 2023-09-06 at 8 08 07 PM

Iteration:

Screenshot 2023-09-06 at 8 08 17 PM

pokey (commented):
From your screenshots, it appears the newline is still being included. 

Fwiw I do wonder if we want to create a "trim" query predicate operator so we don't have to keep reinventing the wheel with whitespace / newline trimming

josharian:
PTAL at these visualizer screenshots. If it looks OK, I'll add more tests and then you can review the code.

I remain unhappy with the pour/drink behavior, but I'm realizing that fixing that is a much, much larger project that touches other scopes, and I'm pretty ready to get the basic functionality in.

Screenshot 2023-09-14 at 5 42 55 PM
Screenshot 2023-09-14 at 5 44 02 PM
Screenshot 2023-09-14 at 5 43 48 PM

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
```diff
@@ -150,6 +150,25 @@ class ShrinkToMatch extends QueryPredicateOperator<ShrinkToMatch> {
   }
 }
 
+/**
+ * A predicate operator that modifies the range of the match by trimming trailing whitespace,
+ * similar to the javascript trimEnd function.
+ */
+class TrimEnd extends QueryPredicateOperator<TrimEnd> {
```
pokey, line 157: I'd be tempted to just have a bidirectional trim. Then you could leverage our `trimRange` utility function. Curious why yours is so much shorter than that one tho 🤔

queries/go.scm
```diff
@@ -222,3 +222,63 @@
...
+  ) @_if
+  (#not-parent-type? @_if if_statement)
+)
+ @branch.start.startOf
+
+;; internal if in an if-else chain
+(if_statement
+  (
+    "else"
+  )
```
pokey, line 257: extraneous parens

queries/go.scm
pokey, line 1: please run autoformatter

pokey (commented):
had a quick look and tests / visualizer looks good! left some minor comments; didn't review the `.scm` thoroughly as it's a bit tough to read without autoformatter

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
josharian: > I'd be tempted to just have a bidirectional trim.

I guess the question is whether we are going to want `trim`, `trimStart`, and `trimEnd`. I opted for `trimEnd` on the grounds that it was precisely what I needed, and that we could add `trim` and `trimStart` if we need that behavior. (In this case `trim` and `trimEnd` are practically identical, but I preferred the precise semantics for documentation and clarity.) I chose the name to match the built-in javascript function it calls.

> Curious why yours is so much shorter than that one tho 

Why does `trimRange` use regexes instead of calling the javascript built-in `trim`/`trimStart`/`trimEnd`?

queries/go.scm
josharian: i did. if there's something in here that looks wrong, it's an autoformatter bug.

queries/go.scm
josharian: fixed

josharian:
added some tests. unfortunately, the behavior in

[packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/go/cloneBranch2.yml](file:///Applications/GitX.app/Contents/Resources/html/views/history/index.html#file_index_9)

and

[packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/go/cloneBranch3.yml](file:///Applications/GitX.app/Contents/Resources/html/views/history/index.html#file_index_10)

is bad: it generates invalid code.

I am not sure how to fix it. Looks like a line number off-by-one?

Because they're hard to read (maybe I really should finish up https://github.com/cursorless-dev/cursorless/pull/1805? there was so little enthusiasm for it), here's an example.

Start:

```go
if x {
	x++
} else if y {
	y++
} else {
	x--
}
```

Then `"clone branch"` with cursor at the beginning of the `y++` line. Result:

```go
if x {
	x++
else if y {
	y++
}
} else if y {
	y++
} else {
	x--
}
```

Best would be something like:

```go
if x {
	x++
} else if y {
	y++
} else if y {
	y++
} else {
	x--
}
```

But at a minimum, I'd expect valid syntax, so maybe:

```go
if x {
	x++
}
else if y {
	y++
} else if y {
	y++
} else {
	x--
}
```

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
josharian: WDYT about having a util something like this? (untested!)

```typescript
export function adjustPosition(doc: TextDocument, pos: Position, by: number): Position {
  const offset = document.offsetAt(pos);
  return document.positionAt(offset + by);
}
```

Then the final part of `trimRange` shrinks to:

```typescript
return new Range(
    adjustPosition(editor.document, range.start, leadingWhitespace.length),
    adjustPosition(editor.document, range.end, -trailingWhitespace.length),
  );
```

which is (IMHO) clear and pretty.

And the end of this new function is:

```typescript
    nodeInfo.range = new Range(
      range.start,
      adjustPosition(document, range.end, trimmed.length - text.length),
    );
    return true;
```

which is simple enough that it hardly seems worth attempting to DRY out.

[Go proverb: A little copying is better than a little dependency.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=568s)

josharian:
Sigh. The precommit failure is due to typescript code I added to the playground to experiment with `branch` scope behavior. Can we extent the playground exemption to include typescript?

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
josharian: `trimRange` also has a latent bug when the range is entirely whitespace. I'll go ahead and send a PR cleaning up a bunch of this.

josharian: lol. it's also dead code.

josharian: OK, sent a PR to cull the dead code, and added a commit making this function a bit clearer.

josharian:
Fixed clone. It's not quite perfect, but it's close enough. The weird setting of the insertion delimiter on `@branch.start.startOf` instead of just `@branch` is because we evaluate the insertion delimiter query before we rewrite `.start.startOf`, and it complains because plain old `@branch` doesn't exist at that point. We could reorder, or be more forgiving...the relevant code says:

>           // FIXME: We could allow some predicates to be forgiving,
>          // because it's possible to have a capture on an optional nodeInfo.
>          // In that case we'd prob just return `true` if any capture was
>          // `null`, but we should check that the given capture name
>          // appears statically in the given pattern.  But we don't yet
>          // have a use case so let's leave it for now.

But let's just get this in for now?

queries/go.scm
pokey: ah ok sorry i'll drop a pointer to the stuff that looks weird

queries/go.scm
```diff
@@ -222,3 +222,59 @@
...
+
+;; first if in an if-else chain
+(
+  (
+    (if_statement
+      consequence: (block) @branch.end.endOf
+    ) @_if
+    (#not-parent-type? @_if if_statement)
+  )
+ @branch.start.startOf
```
pokey, line 253: this is off. I'm not sure if it's an autoformatter bug or not, but I would just change your code this capture right after `@_if` and remove these surrounding parens. not sure what they're accomplishing?

queries/go.scm
```diff
@@ -222,3 +222,59 @@
...
+  ) @branch.end.endOf
+  (#insertion-delimiter! @branch.start.startOf " ")
+)
+
+;; iteration scope is always the outermost if statement
+(
+  (if_statement) @_if
+  (#not-parent-type? @_if if_statement)
+)
+ @branch.iteration
```
pokey, line 280: as above, autoformatter is doing strange stuff when you put a capture on a group, but I'm not sure what that does for you here either

queries/go.scm
josharian: done

queries/go.scm
josharian: done

josharian:
PTAL

josharian:
I believe this is ready for review.

josharian:
Ping.

queries/go.scm
```diff
@@ -222,3 +222,55 @@
...
+
+;; if-else-based branch
+
+;; first if in an if-else chain
+(
+  (if_statement
+    consequence: (block) @branch.end.endOf
+  ) @_if @branch.start.startOf
+  (#not-parent-type? @_if if_statement)
+  (#insertion-delimiter! @branch.start.startOf " ")
```
pokey, line 251: Is this necessary? I think a space is the default insertion delimiter.
```suggestion
```

queries/go.scm
```diff
@@ -222,3 +222,55 @@
...
+)
+
+;; internal if in an if-else chain
+(if_statement
+  "else" @branch.start.startOf
+  alternative: (if_statement
+    consequence: (block) @branch.end.endOf
+  )
+  (#insertion-delimiter! @branch.start.startOf " ")
```
pokey, line 260: As above
```suggestion
```

queries/go.scm
```diff
@@ -222,3 +222,55 @@
...
+  ) @_if @branch.start.startOf
+  (#not-parent-type? @_if if_statement)
+  (#insertion-delimiter! @branch.start.startOf " ")
+)
+
+;; internal if in an if-else chain
+(if_statement
+  "else" @branch.start.startOf
+  alternative: (if_statement
+    consequence: (block) @branch.end.endOf
```
pokey, line 258: I'm pretty sure you don't need the `.startOf` / `.endOf` if they're not overlapping and you can rely on the order
```suggestion
  "else" @branch.start
  alternative: (if_statement
    consequence: (block) @branch.end
```

queries/go.scm
```diff
@@ -222,3 +222,55 @@
...
+  (#insertion-delimiter! @branch.start.startOf " ")
+)
+
+;; final else branch in an if-else chain
+(
+  (if_statement
+    "else" @branch.start.startOf
+    alternative: (block)
+  ) @branch.end.endOf
+  (#insertion-delimiter! @branch.start.startOf " ")
```
pokey, line 269: as above
```suggestion
```

pokey (commented):
Looks good with some minor comments!

josharian:
> Made the final small tweaks to get this one merged in

Thanks! Sorry I'm such a mess at the moment. Really appreciated.

pokey:
Not at all!

PR #1856, opened 2023-09-06
fix typos

josharian:
My OCD finds them mildly distracting.

## Checklist

### DIFF
--- a/docs/user/experimental/keyboard/modal.md
+++ b/docs/user/experimental/keyboard/modal.md
@@ -99,7 +99,7 @@ To bind keys that do not have modifiers (eg just pressing `a`), add entries like
   },
 ```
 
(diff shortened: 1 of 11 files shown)


pokey (approved):
How did you find all these? Did you use a spellchecker?

auscompgeek:
Maybe we should run codespell/cspell in pre-commit? 🤔

josharian:
> How did you find all these? Did you use a spellchecker?

About half or so I noticed while reading the code; I saved them up for a while to avoid lots of one-char PRs. (Note that s/the/that/ is not in the remit of a spellchecker!) And then before sending I spent a few minutes skimming the many false positives emitted by vale. It found several thousand "mistakes" of which maybe three or four were legit.

Spellcheckers in pre-commit are usually annoying due to the high rate of terms of art leading to lots of false positives, even with an (ever-growing) allowlist.

PR #1859, opened 2023-09-06
add Windows specifics to test subset file

saidelike:
## Checklist

### DIFF
--- a/packages/test-harness/src/testSubset.ts
+++ b/packages/test-harness/src/testSubset.ts
@@ -8,12 +8,17 @@ import { getCursorlessRepoRoot } from "@cursorless/common";
  */
 export function testSubsetGrepString(): string {
   const inFile = testSubsetFilePath();
-  return fs
+  let grepStr = fs
     .readFileSync(inFile, "utf-8")
     .split(/\r?\n/)
     .map((line) => line.trim())
     .filter((line) => line.length > 0 && !line.startsWith("#"))
     .join("|");
+  if (process.platform === "win32") {
+    // Path separators are OS specific ("/" for Linux/MacOS and "\\" for Windows).
+    grepStr = grepStr.replaceAll("/", "\\\\");
+  }
+  return grepStr;
 }
 
 /**


r-tae:
Nitpick but *technically* its "MacOS" now

saidelike:
> Sorry for the nitpick but _technically_ its "MacOS" now

Thanks for feedback. Fixed

josharian:
I'd be tempted to normalize when reading this file instead, here: https://github.com/cursorless-dev/cursorless/blob/1df6f0255e33401fcac8a82b03e082bd3910a93f/packages/test-harness/src/testSubset.ts#L9-L17

pokey:
I'd actually argue to normalise when we construct the test name. See https://github.com/cursorless-dev/cursorless/pull/1690#issuecomment-1677327986. Looks like I forgot to file a follow-up issue for that 😅

AndreasArvidsson:
I'm very much in favor of normalizing instead of having to do backslashes

pokey:
Looks like maybe you missed my comment above. I would argue for normalising here https://github.com/cursorless-dev/cursorless/blob/63ea8e9076828b72ce198b58fcb060bd49f06ddc/packages/common/src/testUtil/getFixturePaths.ts

saidelike:
> Looks like maybe you missed [my comment above](https://github.com/cursorless-dev/cursorless/pull/1859#issuecomment-1708421754). I would argue for normalising here
> 
> https://github.com/cursorless-dev/cursorless/blob/63ea8e9076828b72ce198b58fcb060bd49f06ddc/packages/common/src/testUtil/getFixturePaths.ts#L31

Oh right I read your comment but missed it was a different place. I am not sure that place would be better as this is typically what we have on Windows:

```
name: recorded\actions\alternateHighlightHarp
path: C:\path\to\cursorless\packages\cursorless-vscode-e2e\src\suite\fixtures\recorded\actions\alternateHighlightHarp.yml
```

So you would want this changed to this right?

```
name: recorded/actions/alternateHighlightHarp
path: C:/path/to/cursorless/packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/alternateHighlightHarp.yml
```

I am not sure if there is an efficient way in typescript to apply the "\\" to "/" conversion for every single path/name. It seemed better (simpler) to me to do the conversion at the filter name and keep the Windows path names intact like I did in the commit?

AndreasArvidsson:
Agreed. If you ever print the path to the fixture and copy that you won't be able to use it on windows with forward slashes. It's the filter that should be separator agnostic not the file paths themselves that should be updated.

pokey:
No I was proposing to normalize just the name, not the path. The problem with normalizing when reading the subset file is that it's just filtering test names, so the `/` there doesn't necessarily refer to a path. For example, if we had a test suite named `check before / after`, and you wanted to filter to just that test suite, the `/` would get transformed to `\`. 

The fact that our test names look like paths is just a convenient way to name them

That being said, I don't feel strongly, and I'm guessing that all test names that have `/` in them today do in fact refer to paths, and even if they don't, it's unlikely we'd use the `/` when filtering. So if the Windows folks wanna push for this solution I'll cave

AndreasArvidsson:
At long it's just the name I'm totally fine with normalizing it.

AndreasArvidsson:
Pokeys suggested fix is now merged: https://github.com/cursorless-dev/cursorless/pull/2049
Closing this.

PR #1858, opened 2023-09-06
migrate Python scope 'comment'

saidelike:
## Checklist

- [/] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)
- [/] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [/] I have not broken the cheatsheet

- [x] I have confirmed `vizualize comment` work (it didn't work before)
- [x] I have confirmed `take comment` still work. And I ran the test subset on "languages\\python\\" and it passed.

PR #1862, opened 2023-09-07
Migrate Python to use next-gen scope handlers for text fragment extractors

saidelike:
- Depends on #1863 

## Checklist

- [x] merge https://github.com/cursorless-dev/cursorless/pull/1858 first as seems required
- [x] @pokey to fix a bug in the "processSurroundingPairCore" function atm that makes it not working atm (see commit message). It will be handled in a different commit before we can take this PR into account
- [x] record a test for the fix by @pokey by saying "cursorless record" and using "change string"

This PR also includes a few additional unit tests for Python strings.

### DIFF
--- a/data/playground/python/statements.py
+++ b/data/playground/python/statements.py
@@ -15,7 +15,6 @@ def statements():
     c = range(10)
     c.append(100)
     x = 1
-    two = 2
     x /= 2
     for i in range(10):
         if i == 0:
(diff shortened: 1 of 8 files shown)


saidelike:
I have confirmed that with this and #1863 that I can `take string air` and it works as expected (eg on triple quote strings)

pokey (approved):
Looks good! I removed some unnecessary parens in d30341921dfb462c41766e797426b44926989762 fwiw

Issue #1812, opened 2023-08-18
"take round" also takes leading $


jmegner:
"take round" on `$($env:APPDATA)` selects `$($env:APPDATA)`, but it should select `($env:APPDATA)`.

pokey:
That is intentional

AndreasArvidsson:
Defined here
https://github.com/cursorless-dev/cursorless/blob/3242176925e483e54a6e41c91ac651999fb924a2/packages/cursorless-engine/src/processTargets/modifiers/surroundingPair/delimiterMaps.ts#L23

To be honest I've never been one hundred percent sure about that decision.

pokey:
The idea is to include the full delimiter token, similar to the way we handle f-strings and triple quotes in Python. But if @AndreasArvidsson @josharian and @jmegner all want to push for this one I'll cave

jmegner:
....I now better appreciate the benefits of including the dollar sign, but I fear it's just going to cause more surprise and mistakes than benefit.  Let me play with it more now that I know it is deliberate.

AndreasArvidsson:
After discussion with @pokey we came to the following conclusion
* Matching pairs should be dumb and don't try to expand to anything else than they given pair
* the `string` scope type should be implemented per language(eg `f"foo"` would be implemented by python) with a fall back to matching pairs. This is similar to what we do for `idem`
* The `argument` scope type could be used for interpolated variables like `${foo}`

pokey:
> the string scope type should be implemented per language(eg f"foo" would be implemented by python) with a fall back to matching pairs. This is similar to what we do for item

Note that one drawback of this approach is that if you have something like

```
aaa = "bbb 'ccc' ddd"
```

Saying `"string cap"` will give you the whole string `"bbb 'ccc' ddd"` rather than `'ccc'`.

However, we have the same problem with `"item"`, where we'll leap to the syntactic item even if you're inside a smaller textual item. We should be able to fix both of these issues more easily once item and pair are migrated to next-gen scopes, so I don't think this problem should keep us from implementing the proposal in @AndreasArvidsson's comment in the meantime

AndreasArvidsson:
We could still have the string fragment concept. That would force a text based search inside the string.

AndreasArvidsson:
Yes if you just want to match on the `'` delimiters you should use a matching pair I'm not the string scope type. Now we have just overloaded the same word to mean both of them, but in the future we would have this distinction.

pokey:
[...]

PR #1863, opened 2023-09-07
Fix text fragment extractor fallbacks

pokey:
The fallback logic for determining whether to use text-based surrounding pairs or parse-tree-based was broken for text fragment extractors based on next-gen scope handlers (ie using the new `@textFragment` tag in a query file).

This breakage meant that in a string like `"""hello"""` in Python, the surrounding pair finder fell back to text-based, which resulted in it thinking `"hello"` was a string, rather than `"""hello"""`, as would happen if it fell back to parse-tree-based

This PR fixes the issue by unifying the fallback logic between legacy and next-gen text fragment extractors, so that they don't fall out of sync again

See also https://github.com/cursorless-dev/cursorless/issues/1812#issuecomment-1691493746; arguably that will make this PR irrelevant, but until then, it's better to have string not be broken when we migrate a language to use `@textFragment`

Note that the tests in this PR don't actually test the code yet, because Python is still using legacy text fragment extractors. The tests will start biting when this PR is merged into main and then merged into https://github.com/cursorless-dev/cursorless/pull/1862

- This PR is required by https://github.com/cursorless-dev/cursorless/pull/1862

## Checklist


packages/cursorless-engine/src/processTargets/modifiers/surroundingPair/index.ts
```diff
@@ -140,14 +102,41 @@ function processSurroundingPairCore(
...
+
+    if (textFragmentExtractor == null) {
+      // If the text fragment extractor doesn't exist, or if it explicitly is
+      // set to `null`, then we just use text-based algorithm on entire document
+      return document.range;
+    }
+
+    const selectionWithEditor = {
+      editor,
+      selection: new Selection(range.start, range.end),
```
josharian, line 131: i think we have a range.toSelection. although looking at it, it has a boolean arg, ick. (maybe we could make that optional/named and default to false?)

josharian (approved):
lgtm fwiw

packages/cursorless-engine/src/processTargets/modifiers/surroundingPair/index.ts
AndreasArvidsson: I think it's useful to force the user to think about the direction. If you don't care about a direction a selection is not that different from a range.

josharian: sure. but at the call site `.toSelection(false)` is inscrutable.

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/surroundingPair/parseTree/python/changeString.yml
```diff
@@ -0,0 +1,23 @@
...
+  action:
+    name: clearAndSetSelection
+    target:
+      type: primitive
+      modifiers:
+        - type: containingScope
+          scopeType: {type: surroundingPair, delimiter: string}
+  usePrePhraseSnapshot: true
+initialState:
+  documentContents: "\"\"\"hello\"\"\""
```
AndreasArvidsson, line 14: Wow the escaping here is horrid

AndreasArvidsson (approved):

packages/cursorless-engine/src/processTargets/modifiers/surroundingPair/index.ts
```diff
@@ -73,40 +71,6 @@ function processSurroundingPairCore(
...
-        containingScope[0].contentRange,
-        delimiters,
-        scopeType,
-      );
-      if (surroundingRange != null) {
-        // Found the pair within this text fragment or comment, e.g. "(abc)"
-        return surroundingRange;
-      }
-      // Search in the rest of the file, to find e.g. ("abc")
-      return findSurroundingPairTextBased(
```
pokey, line 101: Fwiw the specific bug that this PR is trying to fix was on this line: we were directly falling back to the text-based algorithm when the input range was not in a text fragment, when instead it should have fallen back to parse-tree-based

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/surroundingPair/parseTree/python/changeString.yml
pokey: haha yeah yaml ftw 😅

AndreasArvidsson: It's fine either way from me

PR #1868, opened 2023-09-08
Updated hat shapes

AndreasArvidsson:
* Updated hat shape svg for increased area
* svg all now have the same size and proportions which means that we don't have to do as much tweaking in typescript
* hats have an increased default height

main
image

pr1868
image

matt 2.1
image

matt 2.3
image

pr1868 vs 2.1 vs 2.3
image
image
image

## Checklist

- [x] Do a pass through looking at the hat sorting to see if we want to reorder them
- [x] Check that the order of hats in constants is same as order in package.json
- [ ] Send message to users indicating that if they have used the individual hat adjustments setting, they may want to reset it and start over because things have changed


images/hats/wing.svg
```diff
@@ -1 +1,10 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 371.81 234.76" width="1em" height="1em"><defs><style>.cls-1{fill:#666666;}</style></defs><g id="Layer_2" data-name="Layer 2"><g id="Layer_1-2" data-name="Layer 1"><path class="cls-1" d="M371.81,234.42h0s-153.27-57.29-185.9-54.08h0C154.36,177.24,0,234.76,0,234.76H0L158.55,101.69c.6-.49-.39,1.87-.23,1.12L186.24,0h0l26.84,101.37"/></g></g></svg>
\ No newline at end of file
+<svg width="12" height="9" viewBox="0 0 12 9" fill="none" xmlns="http://www.w3.org/2000/svg">
+<g clip-path="url(#clip0_25_59)">
+<path d="M6 0L12 9L6 7.5L0 9L6 0Z" fill="#FF0000"/>
+</g>
+<defs>
+<clipPath id="clip0_25_59">
+<rect width="12" height="9" fill="white"/>
```
pokey, line 7: This doesn't look right 🤔

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
```diff
@@ -203,8 +203,8 @@ export default class VscodeHatRenderer {
     }
 
     const svg = originalSvg
-      .replace(/fill="[^"]+"/, `fill="${color}"`)
-      .replace(/fill:[^;]+;/, `fill:${color};`);
+      .replace(/fill="#.+"/g, `fill="${color}"`)
```
pokey, line 206: What was wrong with the old regex? In theory it was more robust, eg if color name were something like `white`

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
```diff
@@ -203,8 +203,8 @@ export default class VscodeHatRenderer {
     }
 
     const svg = originalSvg
-      .replace(/fill="[^"]+"/, `fill="${color}"`)
-      .replace(/fill:[^;]+;/, `fill:${color};`);
```
pokey, line 207: How come we dropped this one? Still could be useful

Remember that once your other PR merges, we might get svgs with this way of doing fill

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
```diff
@@ -203,8 +203,8 @@ export default class VscodeHatRenderer {
...
+      .replace(/fill="#.+"/g, `fill="${color}"`)
+      .replace(/\r?\n/g, "");
```
pokey, line 207: Do we want to replace with space just in case?

packages/cursorless-vscode/src/ide/vscode/hats/shapeAdjustments.ts
```diff
@@ -7,49 +7,52 @@ export interface HatAdjustments {
...
 
-export const DEFAULT_HAT_HEIGHT_EM = 0.29;
+export const DEFAULT_HAT_HEIGHT_EM = 0.38;
 export const DEFAULT_VERTICAL_OFFSET_EM = 0.032;
 
 export const defaultShapeAdjustments: IndividualHatAdjustmentMap = {
   default: {
-    sizeAdjustment: -14,
-    verticalOffset: -0.35,
+    sizeAdjustment: -15,
```
pokey, line 15: Huh I thought the idea was to zero these all out. I guess easier to leave them full size and adjust here?

pokey (commented):
Nice! How long were you guys at it yesterday?

images/hats/wing.svg
AndreasArvidsson: That is the frame. It appears to not be rendering. I'm not quite sure why.

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
AndreasArvidsson: 1. It set a color for the outer svg that had `fill="none"` 
2. It wasn't global. Some often new shapes generated by Figma had multiple fill fields and all of them need to have the color of course.

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
AndreasArvidsson: See answer above. I don't know what format that would be?

packages/cursorless-vscode/src/ide/vscode/hats/shapeAdjustments.ts
AndreasArvidsson: Definitely. I could set zero for almost all of them since we're now actually taking the full viewport, but I still feel that actual "artistic" adjustments should be made in code and not in the svg themself. This is just so much simpler.

jmegner: > I still feel that actual "artistic" adjustments should be made in code and not in the svg themself.

Do/will people with custom hats have the ability to override the IndividualHatAdjustmentMap stuff?

AndreasArvidsson: Yes the same adjustments are available in your vscode settings.

jmegner: Ah, I see it, thanks!

```
Cursorless:
Individual Hat Adjustments
Separate adjustments for each hat shape
Edit in settings.json
```

And once you click the edit link, you get a filled out dict with all zeros, exactly what I wanted!

image
image

images/hats/wing.svg
pokey: Even on dark VSCode?

pokey: Oh right it's used as a clip path. So presumably just clips the shape if it extends outside the view box

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
pokey: > 1. It set a color for the outer svg that had `fill="none"` 

I'd be inclined to specifically exclude `none` then

> 2. It wasn't global. Some often new shapes generated by Figma had multiple fill fields and all of them need to have the color of course.

Yeah that makes sense

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
pokey: Seems more robust

packages/cursorless-vscode/src/ide/vscode/hats/shapeAdjustments.ts
pokey: Seems fine to me. Prob easier to tweak this way while developing

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
pokey:
```svg
<rect x="10" height="180" y="10" width="180" style="stroke: black; fill: red;"/>
```

pokey:
Also, I believe the same changes need to be made to the pre processor, and you'll want to run the preprocessor on these hats. Notice how all the hats in the [docs deploy preview](https://deploy-preview-1868--cursorless.netlify.app/docs/#shapes) are now red

Ideally we'd run the preprocessor in pre-commit or meta-updater

pokey:
In terms of shapes, at quick glance they all look nice except I don't love how wing and fox are nearly direct mirror images. As discussed, that has been an issue for users (myself included 😅)

AndreasArvidsson:
I feel the same, but it's hard to make wing have more area and still be distinct in shape.

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
```diff
@@ -203,8 +203,10 @@ export default class VscodeHatRenderer {
     }
 
     const svg = originalSvg
-      .replace(/fill="[^"]+"/, `fill="${color}"`)
-      .replace(/fill:[^;]+;/, `fill:${color};`);
+      .replace(/fill="none"/g, "")
```
pokey, line 206: Does this not cause fill to be black? 

> If the fill attribute property is not specified, the default is black.

From [here](https://docs.aspose.com/svg/net/drawing-basics/fills-and-strokes/#:~:text=if%20the%20fill%20attribute%20property%20is%20not%20specified%2C%20the%20default%20is%20black.)

pokey (commented):
Getting there. See inline comment, and icymi https://github.com/cursorless-dev/cursorless/pull/1868#issuecomment-1711898715

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
AndreasArvidsson: It looks good in vscode, but I updated the regex to make it more secure

AndreasArvidsson:
Preprocessor and other comments are now dealt with. I only have another crack at redesigning wing left.

pokey (commented):
Looks good! Maybe we should drive it for a couple days before merging?

Btw how'd you get the SVGs so clean? They were much messier before

AndreasArvidsson:
I removed the clip rectangle from being exported in Figma

pokey:
Just reposting my comments from Figma here for visibility:

- `default` feels a bit chunky relative to other hats (maybe easiest to solve this one in code)
- `hole` and `frame` look like they could be tough to distinguish. `hole` and `eye` may also be an issue. I'll see how this one plays out after a couple days' use. 
- `crosshairs` kinda just looks like a blob when it's that small. I'm not sure that's necessarily a showstopper, but we'll see if that messes with usability
- `frame` feels a little bit chunky relative to eg `fox`. Maybe the answer is slightly beef up `fox`? 🤷‍♂️
- worried about `curve` vs `wing`; maybe the pointiness on the top of wing will help things
- `curve` sits a bit high, esp compared to eg `wing` (maybe easiest to solve this one in code)

pokey:
If anyone wants to try these hats out locally and is comfortable piping curl to bash, here's a magic incantation to install them locally:

```bash
curl -sSf https://raw.githubusercontent.com/cursorless-dev/cursorless/main/packages/cursorless-vscode/scripts/install-from-pr.sh | bash -s -- 1868
```

Alternately, if you prefer not to curl into bash, check out this PR locally and run `pnpm -F cursorless-vscode install-local`

AndreasArvidsson:
> Just reposting my comments from Figma here for visibility:
> 
>     * `default` feels a bit chunky relative to other hats (maybe easiest to solve this one in code)
> 
>     * `hole` and `frame` look like they could be tough to distinguish. `hole` and `eye` may also be an issue. I'll see how this one plays out after a couple days' use.
> 
>     * `crosshairs` kinda just looks like a blob when it's that small. I'm not sure that's necessarily a showstopper, but we'll see if that messes with usability
> 
>     * `frame` feels a little bit chunky relative to eg `fox`. Maybe the answer is slightly beef up `fox`? 🤷‍♂️
> 
>     * worried about `curve` vs `wing`; maybe the pointiness on the top of wing will help things
> 
>     * `curve` sits a bit high, esp compared to eg `wing` (maybe easiest to solve this one in code)

It definitely is hard to do comparisons like this in Figma for a couple of reasons. 1 vscode doesn't actually render them in small sizes the same way that Figma does and 2 I have already applied size and vertical offset in code for a few hats. `default` and `frame` have been reduced in size and `curve` have been moved down. Looking at the two images in this pr and of course trying the out for yourself is more indicative of actual experience.

We definitely need to discuss this on the meetup

changelog/2023-09-updatedHatShapes.md
```diff
@@ -0,0 +1,6 @@
+---
+tags: [enhancement]
+pullRequest: 1868
+---
+
+- Updated the default Cursorless hat shapes. All the hat shapes have been reworked with the goal to improve visibility. Many of the shapes have been made ticker to enhance color recognition. See the [shape docs](https://www.cursorless.org/docs/#shapes) for more info.
```
phillco, line 6:
```suggestion
- Updated the default Cursorless hat shapes. All the hat shapes have been reworked with the goal to improve visibility. Many of the shapes have been made thicker to enhance color recognition. See the [shape docs](https://www.cursorless.org/docs/#shapes) for more info.
```

AndreasArvidsson:
main
main

pr1868
pr1868

matt latest
image

pokey:
Update from meet-up:

- [ ] use hats from https://github.com/pokey/dotfiles/tree/develop/cursorless-hat-shapes
- [ ] use settings from https://github.com/pokey/dotfiles/blob/6a7f521c7d328238ffd39e4134d4e35d92eebd92/vscode/settings.json#L165-L210, rounding to 0.5
- [ ] send these back to matt (@pokey)
- [ ] send message to slack gathering working group
- [ ] drive them for a couple days
- [ ] ship it

AndreasArvidsson:
That new default light color looks way to bright for me
image

pokey:
There should be no default light color. It's only for dark

pokey:
update from meet-up:

use this hat order:

1. default
2. bolt
3. curve
4. fox
5. frame
6. play
7. wing
8. hole
9. ex
10. crosshairs
11. eye

Then:

- [ ] increase fox and ex by 5%
- [ ] decrease crosshairs by 10%
- [ ] send to matt
- [ ] send to users for feedback

AndreasArvidsson:
@pokey ex, fox, frame have been increased in size.

AndreasArvidsson:
@pokey last commit reverted

pokey:
ok @AndreasArvidsson this should now show a message if the user has shapes enabled and has any active adjustments. It checks once on first update, and then never checks again, so that subsequent shape tweaks won't trigger a message. Lmk if it works for you locally

pokey (commented):
ok this looks good to me! @AndreasArvidsson lmk once you've had a chance to look at my hat adjustment message code and try it locally, then I'll ship it so I can coordinate with a message and be there in case a fire breaks out or something 😄

AndreasArvidsson:
@pokey Looks fine

AndreasArvidsson:
Correct

pokey (approved):

PR #1869, opened 2023-09-08
Update adding-a-new-language.md

saidelike:
Migrating from old scope handler to new one (tree-sitter based)

## Checklist


pokey (commented):
I would argue this should be its own page, no?

pokey:
At this point we will probably do the rest of the migrations ourselves, so I'm not sure it makes sense to have this blurb in the docs

PR #1873, opened 2023-09-08
Made interface names targetable as keys

### LINKED ISSUE #1790: Support for key in typescript interfaces
```ts
interface Hello {
    value: number
}
```
`value` here can be referred to as `name`, but not `key`. I would like to support both.
@pokey what do you think?

AndreasArvidsson:
Fixes #1790

## Checklist


auscompgeek (commented):
Are type aliases handled already?

AndreasArvidsson:
It is now!

auscompgeek (commented):
What about object types in function signatures? 😅

AndreasArvidsson:
It does now!!!

pokey (approved):

Issue #1795, opened 2023-08-15
Start writing changelog entries


pokey:
- [ ] Add a page to contributor docs about how to do it
- [ ] add checkbox to PR template
- [ ] Add an example on one of our PRs

Could eventually be used to automate part of https://github.com/cursorless-dev/cursorless/issues/491

auscompgeek:
Depending on how much we want multiple tags, there's tools like [towncrier] for this sort of thing

[towncrier]: https://towncrier.readthedocs.io/en/stable/index.html

PR #1875, opened 2023-09-08
Added text insertion action to Cursorless public api

### LINKED ISSUE #1754: Expose insert action in public api
# The problem

The question of how you insert text with Cursorless comes up quite often. People often want to insert text at a position away from their current cursor position.

# The solution
There is no way we can support all different ways people want to insert text. Different languages, math formulas and so on. The list is basically infinite. If we instead just exposed a insert action the user can insert text however they want. 

```python
def cursorless_insert(destination: CursorlessDestination, text: Union[str, List[str]]):
```

## Examples

For example, this could be used as follows:

```talon
word <user.word> <user.cursorless_destination>:
    user.cursorless_insert(cursorless_destination, word)
```

Then the user would be able to say eg `"word hello after air"`

AndreasArvidsson:
Fixes #1754

## Checklist

### DIFF
--- a/cursorless-talon-dev/src/cursorless_test.talon
+++ b/cursorless-talon-dev/src/cursorless_test.talon
@@ -7,6 +7,10 @@ test api command <user.cursorless_target>:
     user.cursorless_command("setSelection", cursorless_target)
 test api command bring <user.cursorless_target>:
     user.cursorless_command("replaceWithTarget", cursorless_target)
+test api insert <user.word> <user.cursorless_destination>:
+    user.cursorless_insert(cursorless_destination, word)
+test api insert <user.word> and <user.word> <user.cursorless_destination>:
+    user.cursorless_insert(cursorless_destination, word_list)
 test api insert snippet:
     user.cursorless_insert_snippet("Hello, $foo!  My name is $bar!")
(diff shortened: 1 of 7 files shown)


pokey (approved):
Looks good!

josharian:
This might be a good PR to use for the inaugural changelog note (https://github.com/cursorless-dev/cursorless/issues/1795).

AndreasArvidsson:
Agreed. Pokey maybe we should try to discuss tomorrow were and exactly how we want these files to look and this might be the first one?

pokey:
Sure, happy to discuss. Is there something specific from the description of #1795 that is unclear that you wanted to discuss?

AndreasArvidsson:
For once it doesn't even say where to place the files. Should I link the pull request, the creation date and so on?

pokey:
It provides an example of `changelog/2023-08-expanded-talon-api.md`. I would also put PR number and release date in frontmatter

AndreasArvidsson:
I made a try here
https://github.com/cursorless-dev/cursorless/pull/1886/commits/bdbfa75e070b73d05302f543c621be45b90d8152

PR #1877, opened 2023-09-08
Migrate python scope "value" to new scope handler

saidelike:
## Checklist

- [ ] I have executed the existing tests but they seem to fail for whatever reason. Not sure why (e.g. `takeValue.yml` from a quick glance
- [x] I have tested the `visualize value` command and all the `take value` from this file and they all work fine

image

If anyone has ideas why the unit tests don't work. 

NOTES:
- I haven't included the assignment symbols like ".", "=", "/=" as I don't think they are needed when using the tree-sitter but let me know if you think otherwise.
- the "return" in the "return_statement" is something that was ignored too as not relevant anymore afaict


saidelike:
And this is temporarily disabling the assignment values to see the domain of dictionary values:

image

pokey (commented):
Coming along! Missing

- [ ] Domains on assignments
- [ ] Leading delimiters on everything except return values (search in scm files for `leading`)

saidelike:
> Coming along! Missing
> 
>     * [ ]  Domains on assignments
> 
>     * [ ]  Leading delimiters on everything except return values (search in scm files for `leading`)

Ok domains on assignments done. Do you have examples for use cases where delimeters matter for the `value` scope? I understand the `,` in the `arg` scope case, but I don't see it for `value`...

pokey:
It's if you say "chuck value". For example if you wanted to convert a dict to a set by deleting all the values, you could say "chuck every value" and it would clean up the `:`. Not totally clear how useful it is for assignments, but we support it today so I'd be inclined to preserve that behaviour 

Also, speaking of "every" you'll want to add iteration scopes as well

saidelike:
Added value iteration scopes:

`visualize value`:

image

`visualize value iteration`:
image

queries/python.scm
pokey, line 1: It looks like you've lost a bunch of value captures by making the types more specific here. Search in [tree-sitter-python](https://github.com/tree-sitter/tree-sitter-python/blob/master/grammar.js) for `'value'`. Eg we lost [values within match statement patterns](https://www.inspiredpython.com/course/pattern-matching/mastering-structural-pattern-matching), values in [named expressions](https://realpython.com/lessons/assignment-expressions/), etc. I'd be inclined to keep a generic `value: (_)` of some sort as that is often what we want

pokey (commented):
Ok I made some small tweaks in 094c70e684c5f30c4a6c4af340d5d55ecbf7206e. Note in particular how I added `!` and `!!` marks to your comments and tweaked some of the example markers underneath those lines. See https://github.com/cursorless-dev/cursorless/issues/1524 for more on that syntax

pokey:
ok sorry thought you were sleeping so I force-pushed over my previous commit 😅

pokey:
this PR is good to go now from my perspective; see 22bf6ea7494c75d82d653d49fcbd6f1315add155

pokey:
all told, I made the following changes:

- Added those `!` and tweaked scope range markers as mentioned above
- Removed extraneous parens from groupings where there was no predicate
- Changed the way we do leading range to start from the end of previous named node instead of looking for `=`, `+=`, etc, as that simplified code and also fixed the failing tests that expected better whitespace cleanup
- Added some tests to catch a few other value examples

pokey:
The last issue remaining is that the new playground file has some invalid python syntax; see the complaining pre-commit step in CI

pokey:
We're also not supporting iteration scope for multi-value return values, eg `return a, b`. That is definitely extra credit, though, so I'd be tempted to do that in follow-up PR

pokey:
image

data/playground/python/values.py
```diff
@@ -37,9 +37,7 @@ def my_funk(value: str = "hello", other: bool=False) -> str:
 
 # we can say "change every value" to allow modifying all the values in one go
 def foo():
-    a = 0
-    b = 1
-    c = 2
+    pass
```
pokey, line 40: wow looks like we have some pretty aggressive "auto-formatting" going on here 😅. cc/ @auscompgeek any idea which formatter is doing this?

queries/python.scm
```diff
@@ -27,6 +27,58 @@
...
 
+;;!! a = 25
+;;!      ^^
+;;!   xxxxx
+;;!  ------
+(assignment
+  (_) @_.leading.start.endOf
+  .
+  right: (_) @value @_.leading.end.startOf
+) @_.domain
```
pokey, line 38: fwiw I do think this pattern and the following one (`augmented_assignment`) are potentially a compelling use case for `#has-type?` predicate, as it would avoid duplication. @AndreasArvidsson wdyt?

queries/python.scm
```diff
@@ -27,6 +27,58 @@
...
+  (#not-type? @_.domain subscript)
+)
+
+;;!! return 1
+;;!         ^
+;;!        xx
+;;!  --------
+;;
+;; NOTE: in tree-sitter, both "return" and the "1" are children of `return_statement`
+;; but "return" is anonymous whereas "1" is named node, so no need to exclude explicitly
```
pokey, line 77: This comment might be overkill. We leverage this kind of assumption all over the place. For example, I'm leveraging this kind of assumption in my leading matchers above, where I rely on the fact that `=` is anonymous whereas the name / type before it is not

saidelike:
> notice how the removal range extends all the way past the space before the `=` etc
> 
> image
> 

Looks great. What command can you say to see the removal range?

pokey:
"visualize value removal"

see [docs for visualizer](https://www.cursorless.org/docs/user/scope-visualizer/), although at this point I think you now know everything in that doc 😄

queries/python.scm
saidelike: To me it is the same as dictionary vs dictionary_comprehension as discussed in https://github.com/cursorless-dev/cursorless/commit/36c2f4c7e5227086cf9ed098a5c8d787fea063d5

pokey: Agreed it is quite similar. In that case we managed to avoid duplication another way, but I don't see a way out of it here other than `has-type`

pokey:
> The last issue remaining is that the new playground file has some invalid python syntax; see the complaining pre-commit step in CI

Once this is fixed I think we're good to merge

data/playground/python/values.py
auscompgeek: It's probably ruff's pyflakes (F) autofixes yeeting the unused variables, then the unused literal expressions?

josharian: I guess I would prefer a blanket exception for everything in the playground. I have several things in the playground that are not gofmt compliant (but are valid go code ) specifically so that I can experiment to see whether cursorless does the right thing with them.

pokey: I am very much on the fence, but lean slightly towards formatting playground but being liberal with comments to disable auto-formatter / linter where we want. @josharian is it not easy to disable gofmt on a per-line basis?

I don't feel strongly, tho. I could see the benefit to having full freedom in the playground

josharian: > is it not easy to disable gofmt on a per-line basis

no, by design gofmt is impossible to selectively disable and has no knobs whatsoever

(Happy to expound upon the thinking behind those design decisions over voice, if desired)

josharian: However, there are also no pre-commit hooks for Go at all, so far this is not a problem in the playground or anywhere. :)

pokey: I could see a case for blanket disabling formatter in playground. That being said, the formatter has detected real syntax errors in these new python playground files, so it is providing some value here

josharian: We could have the pre-commit step check for the actual property we want, namely valid syntax

josharian: It depends in part on how active the formatter is. There is some danger of writing queries that only match when using the same formatter as us, particularly if the formatter is not whitespace only. 🤷🏽

pokey: Yeah that's a good point. I think maximal flexibility is not necessarily a bad thing in the playground. Interesting idea to just check for valid syntax. Altho tbh we even might want to violate that in some cases because error tolerance is one of the reasons we like tree-sitter

pokey: I'm convinced. See 43939f76e080c3000e9d23df27b1004fe667f359. Any objections @auscompgeek @josharian @AndreasArvidsson ?

AndreasArvidsson: No sounds resonable

pokey (approved):
Looks good to me!

PR #1880, opened 2023-09-09
Target Python 3.11 (Talon 0.4)

auscompgeek:
Our previous Python 3.9 target matched up with Talon 0.3.

Since https://github.com/cursorless-dev/cursorless/pull/1757#discussion_r1314089443 (merged into #1738), cursorless-talon depends on Python 3.10 features. Attempting to use cursorless-talon with Talon 0.3.1 results in the following error:

```
2023-09-09 23:17:40 ERROR user.cursorless-talon.src.spoken_forms import failed
   13:                lib/python3.9/threading.py:937* # loader thread
   12:                lib/python3.9/threading.py:980*
   11:                lib/python3.9/threading.py:917*
   10:                   app/resources/loader.py:894|
    9:                   app/resources/loader.py:830|
    8:                   app/resources/loader.py:648|
    7:                   app/resources/loader.py:684|
    6:                   app/resources/loader.py:738|
    5:                   app/resources/loader.py:432| # [stack splice]
    4:       lib/python3.9/importlib/__init__.py:127|
    3:                   app/resources/loader.py:253|
    2:                   app/resources/loader.py:248|
    1: user/cursorless-talon/src/spoken_forms.py:3  | from typing import Callable, Concatenate, ParamSpec, TypeVar
ImportError: cannot import name 'Concatenate' from 'typing' (/Applications/Talon.app/Contents/Resources/python/lib/python39.zip/typing.pyc)
```

Talon 0.4 has Python 3.11. This bumps our target to match latest public Talon.

## Checklist

- [ ] I have run the Talon tests


cursorless-talon-dev/src/spoken_form_test.py
```diff
@@ -82,9 +82,7 @@ def private_cursorless_spoken_form_test_mode(enable: bool):
                 "Cursorless spoken form tests are done. Talon microphone is re-enabled."
             )
 
-    def private_cursorless_spoken_form_test(
-        phrase: str, mockedGetValue_: Optional[str]
-    ):
+    def private_cursorless_spoken_form_test(phrase: str, mockedGetValue_: str | None):
```
AndreasArvidsson, line 85: Finally a useful syntax!

AndreasArvidsson:
As long as the Talon tests pass I think this looks great

pokey (commented):
yep looks good as long as talon tests pass 👍

pokey:
Talon tests failing with

```
2023-09-15 15:01:12.464 ERROR TalonScript TypeError in: /Users/pokey/.talon/user/cursorless-talon-dev/src/cursorless_test.talon:15
2023-09-15 15:01:12.464 ERROR Line 15: user.cursorless_wrap_with_snippet(body: str, target: Union[user.cursorless-talon.src.targets.target_types.ListTarget, user.cursorless-talon.src.targets.target_types.RangeTarget, user.cursorless-talon.src.targets.target_types.PrimitiveTarget, user.cursorless-talon.src.targets.target_types.ImplicitTarget], variable_name: str | None = None, scope: str | None = None)
2023-09-15 15:01:12.464 ERROR Line 15:   parameter 3 (variable_name) expects type str | None, got <class 'str'>
2023-09-15 15:01:12.464 ERROR Line 15: user.cursorless_wrap_with_snippet(body: str, target: Union[user.cursorless-talon.src.targets.target_types.ListTarget, user.cursorless-talon.src.targets.target_types.RangeTarget, user.cursorless-talon.src.targets.target_types.PrimitiveTarget, user.cursorless-talon.src.targets.target_types.ImplicitTarget], variable_name: str | None = None, scope: str | None = None)
2023-09-15 15:01:12.464 ERROR Line 15:   parameter 4 (scope) expects type str | None, got <class 'str'>
```

auscompgeek:
😥 I was afraid of that. I'm slightly surprised it doesn't work though since `isinstance` would work correctly.

pokey:
Is that a talon bug then?

auscompgeek:
I'd definitely consider it a bug that Talon doesn't accept standard Python types. This makes me suspect Talon is internally checking `type(foo) is accepted_type` rather than `isinstance(foo, accepted_type)`.

For reference this is what I was referring to:

```pycon
>>> isinstance('', str | None)
True
>>> isinstance(None, str | None)
True
```

pokey:
So probably worth filing a bug against Talon then? Is there a reasonable path forward here in the meantime, until / when Talon fixes that bug?

auscompgeek:
We can disable the rule that upgrades the type union syntax in the meantime.

pokey:
Ok let's disable that rule for now. Filed https://github.com/talonvoice/talon/issues/634 to track the possible Talon bug

pokey:
Disabled `UP007` for now. Filed https://github.com/cursorless-dev/cursorless/issues/2151 to track re-enabling

pokey (approved):

lunixbochs:
> I'd definitely consider it a bug that Talon doesn't accept [standard Python types](https://peps.python.org/pep-0604/). This makes me suspect Talon is internally checking type(foo) is accepted_type rather than isinstance(foo, accepted_type).

Talon isn't checking `type(foo) is accepted_type` - Talon has a static type checker, which checks the code statically without actually running it. So at type check time, we don't have an object to `isinstance()`.

This issue is just a case of the new `A | B` union type being `types.UnionType` instead of the old `typing.Union`, so I had to add support for it.

PR #1891, opened 2023-09-12
Reduce extension dependencies

### LINKED ISSUE #1814: Register file extensions for languages we support
For some languages (eg talon script, scala, .scm), we require the user to have a separate VSCode extension to register the language id, as it doesn't ship with VSCode. We might want to just register these language ids ourselves

Then we could remove them from https://github.com/cursorless-dev/cursorless/blob/main/packages/common/src/extensionDependencies.ts

AndreasArvidsson:
Fixes #1814

There is no problem with registering `.scm`. If the user is using scheme there is still a good probability that their Scheme extension would win over Cursorless and if not vscode has a file association setting they can use.

## Checklist

### DIFF
--- a/packages/common/src/extensionDependencies.ts
+++ b/packages/common/src/extensionDependencies.ts
@@ -1,8 +1,6 @@
 export const extensionDependencies = [
   "pokey.parse-tree",
+
+  // Necessary for the `drink cell` and `pour cell` tests
   "ms-toolsai.jupyter",
-  "scalameta.metals",
-  "ms-python.python",
-  "mrob95.vscode-talonscript",
-  "jrieken.vscode-tree-sitter-query",
 ];
--- a/packages/cursorless-vscode/package.json
+++ b/packages/cursorless-vscode/package.json
@@ -1016,6 +1016,33 @@
         "extensions": [
           ".cursorless-snippets"
         ]
+      },
+      {
+        "id": "talon",
+        "aliases": [
+          "Talon"
+        ],
+        "extensions": [
+          ".talon"
+        ]
+      },
+      {
+        "id": "scala",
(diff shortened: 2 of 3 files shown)


pokey:
> there is still a good probability that their Scheme extension would win over Cursorless

Have you installed a scheme extension alongside this version of Cursorless to see what happens?

AndreasArvidsson:
> > there is still a good probability that their Scheme extension would win over Cursorless
> 
> Have you installed a scheme extension alongside this version of Cursorless to see what happens?

I have not

pokey:
> I have not. I have had other scenarios where I had the same extension registered to multiple ids and what determines which one wins out I have no idea.

In that case, I would lean towards not registering the extension. The others are fine. If you feel strongly, let's discuss

auscompgeek (approved):
LGTM. The PR description should also be updated before merging.

pokey:
For posterity's sake: there are drawbacks to registering extension and not registering it

Drawback to registering the extension is that we might silently break their syntax highlighting, lsp, etc. For example, if they have scheme extension installed, we might override that mapping and cause VSCode to treat it as query file

On the  other hand, if we don't register, then Cursorless just won't work in certain languages if they don't have extension installed that registers given file extension

We decided that it is better not to register the extension for the following reasons:

- They will likely have an extension installed for their given language
- We can get a decent middle ground with #1895
- Fundamentally it is not our jurisdiction to decide which file extensions to map to which languages; let's leave that to the editor / extensions
- When they install Cursorless, they only use a small subset of our supported languages; it's not great if we're altering file extension behaviour for multiple different languages they may not even know we support

pokey (approved):

PR #1894, opened 2023-09-13
Added playground action

### LINKED ISSUE #1893: Add playground command
In yesterday's meeting we were talking about implementing a tree sitter playground for cursorless. @josharian came with the brilliant idea of just using Cursorless to target arbitrary code and then display the results in an untitled document

1. Add a new action `playground`
2. Generate a Tree sitter playground output for each target content range
3. Open a new untitled the document with the result

None of the above should be particular hard. The only considerations are the following:
* Today the actions don't have access to the tree sitter implementation. Shouldn't be any limitations to actually passing it down to a particular action though.
* Today there is no canonical way of opening a new untitled document with a specific content. We would need to extend the ide api with something analogous to `function openTextDocument(options?: { language?: string; content?: string; })`

@josharian Did I capture your intent here?
@pokey Any thoughts?

AndreasArvidsson:
Fixes #1893

`"parse tree class"`

image

## Checklist


josharian:
Andreas gave me a live demo. This is super slick; I'm excited to use it.

I wonder whether the spoken form should be `syntax tree` or `treesitter` or even `inspect/dump/show syntax tree` rather than `playground`.

AndreasArvidsson:
@josharian Updated the image with the changes we talked about

AndreasArvidsson:
@pokey @josharian scm query added

cursorless-talon-dev/src/cursorless_test.talon
```diff
@@ -19,3 +19,5 @@ test api wrap with snippet <user.cursorless_target>:
     user.cursorless_wrap_with_snippet("Hello, $foo!  My name is $bar!", cursorless_target, "foo", "statement")
 test api wrap with snippet by name <user.cursorless_target>:
     user.cursorless_wrap_with_snippet_by_name("functionDeclaration", "body", cursorless_target)
+test snippet make <user.cursorless_target>:
+    user.private_cursorless_make_snippet_test(cursorless_target)
```
pokey, line 23: this is for recording a snippet make test; the commands in this file are only to be activated during Talon spoken form tests

cursorless-talon-dev/src/cursorless_dev.talon
```diff
@@ -26,5 +28,5 @@ tag: user.cursorless
 {user.cursorless_homophone} update cheatsheet:
     user.private_cursorless_cheat_sheet_update_json()
 
-test snippet make <user.cursorless_target>:
-    user.private_cursorless_make_snippet_test(cursorless_target)
+playground <user.cursorless_target>:
```
pokey, line 31: I think I agree with @josharian that we should possibly name this something else, esp since the "playground" we make will prob be more sophisticated. Could use eg "show parse", "parse tree", "tree sitter", "tree sit", etc

cursorless-talon-dev/src/cursorless_dev.talon
```diff
@@ -26,5 +28,5 @@ tag: user.cursorless
...
+    user.cursorless_command("private.playground", cursorless_target)
```
pokey, line 32: I'd also be tempted to propagate this rename all the way through for consistency

packages/cursorless-engine/src/test/fixtures/talonApi.fixture.ts
```diff
@@ -99,6 +103,7 @@ export const talonApiFixture = [
     "test api wrap with snippet by name this",
     wrapWithSnippetByNameAction,
   ),
+  spokenFormTest("playground air", playgroundAction),
```
pokey, line 106: 👍

packages/cursorless-engine/src/actions/Playground.ts
pokey, line 1: If we decide we like this thing, we should add a test at some point, but I'm fine with leaving it for now

packages/cursorless-vscode/src/ide/vscode/VscodeIDE.ts
```diff
@@ -126,6 +126,14 @@ export class VscodeIDE implements IDE {
     return this.fromVscodeEditor(await window.showTextDocument(textDocument));
   }
 
+  public async openUntitledTextDocument(options?: {
+    language?: string;
+    content?: string;
+  }): Promise<TextEditor> {
```
pokey, line 132: I'd import and use type `OpenUntitledTextDocumentOptions` from our `ide.types.ts`

pokey (commented):
left a few minor comments but looking good!

packages/cursorless-vscode/src/ide/vscode/VscodeIDE.ts
AndreasArvidsson: Yes missed that one. You like the name `openUntitledTextDocument`? Vscode just has a overloaded function that can take either a uri or these options.

cursorless-talon-dev/src/cursorless_dev.talon
AndreasArvidsson: `Parse tree` for both the spoken form and the action name sounds reasonable.

packages/cursorless-vscode/src/ide/vscode/VscodeIDE.ts
pokey: Maybe `newUntitledTextEditor`? I wouldn't say I strongly prefer your api or VSCode's, but I agree it's a bit cumbersome with their api that the client needs to call two functions

AndreasArvidsson: Yet that part is annoying, but what I was getting at is if we want to have two separate functions. `openTextDocument` and `openUntitledTextDocument` or do what vscode does and just have the single `openTextDocument` that does both depending on passed arguments?

cursorless-talon-dev/src/cursorless_test.talon
AndreasArvidsson: Okay I will move it back

packages/cursorless-engine/src/actions/Playground.ts
AndreasArvidsson: For now I recorded a test fixture. It of course doesn't show the result in then new editor, but at least we're capturing that we're not completely breaking even running the action.

AndreasArvidsson:
@pokey updated

packages/cursorless-vscode/src/ide/vscode/VscodeIDE.ts
josharian: I would vote for two separate functions; overloading is hard to read at the call site. (They can both dispatch to a shared, overloaded function under the hood, if desired.)

AndreasArvidsson: Ok then I guess we leave it as is :)

packages/cursorless-engine/src/actions/ParseTree.ts
```diff
@@ -0,0 +1,121 @@
+import { FlashStyle, Range, TextDocument } from "@cursorless/common";
+import * as path from "node:path";
+import type { Tree, TreeCursor } from "web-tree-sitter";
+import type { TreeSitter } from "..";
+import { ide } from "../singletons/ide.singleton";
+import type { Target } from "../typings/target.types";
+import { flashTargets } from "../util/targetUtils";
+import type { ActionReturnValue } from "./actions.types";
+
+export default class ParseTree {
```
pokey, line 10: It's a bit awkward to have an action name which is a noun. I think it's ok for a spoken form, but for the name of the class and for our action ID, I'd prefer a verb, eg "show parse tree". Also makes it easier to grep for

packages/cursorless-vscode/src/ide/vscode/VscodeIDE.ts
pokey: In terms of function name, I think I'd prefer `newUntitledTextEditor` or `openNewUntitledTextEditor`, to indicate it's returning an editor, not a document

pokey (commented):
ok looking good; left a couple final comments

packages/cursorless-vscode/src/ide/vscode/VscodeIDE.ts
AndreasArvidsson: The same is true for `openTextDocument` I think we should be consistent in the naming of these two.

packages/cursorless-engine/src/actions/ParseTree.ts
AndreasArvidsson: Sounds reasonable

AndreasArvidsson: Done

pokey (commented):
ok one last comment; I would have done it myself but wanted to make sure you're ok with the change

packages/cursorless-engine/src/actions/ShowParseTree.ts
```diff
@@ -0,0 +1,121 @@
...
+    document.getText(contentRange),
+    "```",
+    "",
+    "```js",
+    ...resultPlayground,
+    "```",
+    "",
+    "```scm",
+    ...resultQuery,
+    "```",
```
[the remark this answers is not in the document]
AndreasArvidsson: done

pokey (approved):

PR #1896, opened 2023-09-14
improve Range function signatures and docs

josharian:
If the first arg isn't reliably used as start
in a constructor, then don't call it that.
(This bit me hard.)

Call out swapping behavior in with.
(So did this.)

Make other docs more precise.

## Checklist

### DIFF
--- a/packages/common/src/types/Range.ts
+++ b/packages/common/src/types/Range.ts
@@ -12,29 +12,24 @@ export class Range {
   readonly end: Position;
 
   /**
-   * Create a new range from two positions. If `start` is not
-   * before or equal to `end`, the values will be swapped.
+   * Create a new range from two positions.
+   * The earlier of `p1` and `p2` will be used as the start position.
    *
-   * @param start A position.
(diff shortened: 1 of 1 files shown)


packages/common/src/types/Range.ts
```diff
@@ -12,29 +12,24 @@ export class Range {
...
-  constructor(
-    startLine: number,
-    startCharacter: number,
-    endLine: number,
-    endCharacter: number,
-  );
+  constructor(line1: number, char1: number, line2: number, char2: number);
 
   constructor(...args: any[]) {
     const [start, end]: [Position, Position] = (() => {
```
pokey, line 35: I think we should prob mirror the naming from the function signature?

pokey (commented):
Much clearer! Left one minor comment

packages/common/src/types/Range.ts
josharian: done

pokey (approved):

PR #1899, opened 2023-09-15
vendor enough of vscode's TextDocument implementation

### LINKED ISSUE #1882: add in-memory TextDocument implementation
Moving the conversation from https://github.com/cursorless-dev/cursorless/pull/1815, where it is buried in a comment thread: https://github.com/cursorless-dev/cursorless/pull/1815#discussion_r1299382008

For moving to an LSP, and for faster/easier tests not requiring the full VSCode API, it would be helpful to have an in-memory implementation of TextDocument.

#1815 adds a simple one, coded from scratch. But @pokey points out that https://github.com/microsoft/vscode-languageserver-node already has one.

Unfortunately, the TextEditor interface as used by cursorless internally (call this CTE, or Cursorless TextEditor) is larger than the TextEditor interface implemented by https://github.com/microsoft/vscode-languageserver-node (call this VSCTE, for VSCode TextEditor).

Attempts to use VSCTE hit a wall, because missing properties like `lineAt` and `range` are extensively used in CTE and are difficult to implement efficiently using VSCTE. Rewriting all of cursorless to use VSCTE directly would (I think) result in less readable, less efficient code.

Attempts to wrap VSCTE in a new object implementing CTE are also not promising: CTE's `lineAt` returns a TextLine, which VSCTE doesn't have. By the time you set up all the glue, there aren't many savings over just using a mock CTE directly. (VSCTE has some nice performance properties, like lazy line resolution, but IIUC those get destroyed the instant you need to wrap it to make a CTE.)

The question is: As a practical matter, how do I proceed from here? Input requested, although not urgently. (I can keep hacking atop #1815 with the existing implementation for the time being.)

cc @pokey 

Related issues:

* https://github.com/cursorless-dev/cursorless/issues/946
* https://github.com/cursorless-dev/cursorless/issues/1160

josharian:
- fixes #1882 

## Checklist

### DIFF
--- a/packages/common/src/types/text_document.ts
+++ b/packages/common/src/types/text_document.ts
@@ -1,0 +1,411 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Position, Range, TextDocument, TextLine } from "..";
+
+class MirrorTextModel {
+  protected _uri: URI;
+  protected _lines: string[];
+  protected _eol: string;
+  protected _versionId: number;
+  protected _lineStarts: PrefixSumComputer | null;
+  private _cachedTextValue: string | null;
+
+  constructor(uri: URI, lines: string[], eol: string, versionId: number) {
+    this._uri = uri;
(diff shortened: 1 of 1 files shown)


pokey:
A descendant of this code will land in #2256; we should probably use that as our new starting point, as it is being used in Cursorless tests, so a couple corner cases have been fixed

PR #1900, opened 2023-09-16
Added argument to call action

AndreasArvidsson:
`call air on bat`

## Checklist

### DIFF
--- a/cursorless-talon/src/actions/call.py
+++ b/cursorless-talon/src/actions/call.py
@@ -1,13 +1,22 @@
-from talon import actions
+from typing import Optional
+from talon import Module, actions
 
 from ..targets.target_types import CursorlessTarget, ImplicitTarget
 
+mod = Module()
+mod.list("cursorless_call_action", desc="Cursorless call action")
 
-def cursorless_call_action(target: CursorlessTarget):
-    actions.user.private_cursorless_command_and_wait(
-        {
-            "name": "callAsFunction",
-            "callee": target,
-            "argument": ImplicitTarget(),
-        }
-    )
+@mod.action_class
+class Actions:
+    def private_cursorless_call(
+        callee: CursorlessTarget,
+        argument: Optional[CursorlessTarget] = ImplicitTarget(),
+    ):
+        """Execute Cursorless call action"""
+        actions.user.private_cursorless_command_and_wait(
+            {
+                "name": "callAsFunction",
+                "callee": callee,
+                "argument": argument,
+            }
+        )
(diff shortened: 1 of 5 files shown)


AndreasArvidsson:
I've solved this without adding a new data class by both having it in `<user.cursorless_action_or_ide_command>` for the single target case and then adding a new command to `cursorless.talon` for the two target case. I think we should think of migrate bring, swap, move and so on to this format and getting rid of the data classes. This also have the benefit of actually utilizing `user.cursorless_command()` from our public api in the single target case which makes it so that we're always testing that path as well.

pokey:
How's DFA?

AndreasArvidsson:
> How's DFA?

Just restarted Talon which to vscode and issued a Cursorless command
`IO [audio]=2100.000ms [compile]=86.483ms [emit]=31.591ms [decode]=1.151ms [total]=119.226ms`

pokey:
nice 🙌. presumably that is not much slower than `main`?

AndreasArvidsson:
No looks to be within error marginal. I don't think I could even get a reliable figure if I don't rerun them more times than I have patience for.

cursorless-talon/src/actions/call.py
```diff
@@ -1,13 +1,23 @@
...
-            "name": "callAsFunction",
-            "callee": target,
-            "argument": ImplicitTarget(),
-        }
-    )
+@mod.action_class
+class Actions:
+    def private_cursorless_call(
+        callee: CursorlessTarget,
+        argument: Optional[CursorlessTarget] = ImplicitTarget(),
```
auscompgeek, line 14: Does this actually accept `None`?
```suggestion
        argument: CursorlessTarget = ImplicitTarget(),
```

AndreasArvidsson: We either call this function with one or two arguments. There is no reason to pass none.

auscompgeek: If there's no reason to pass `None` to this, then this shouldn't be type hinted to accept `None` :)

AndreasArvidsson: I don't follow?

auscompgeek: Whether something is optional (i.e. can be `None`) is orthogonal to whether an argument has a default value. https://docs.python.org/3/library/typing.html#typing.Optional

AndreasArvidsson: ` typing.Optional` `Note that this is not the same concept as an optional argument`. Of course it isn't! :D
Thanks!

pokey (approved):
Made small fixes to cheatsheet, otherwise looks good! Merge if you're happy with my fixes

PR #1910, opened 2023-09-23
Fix Black

### LINKED ISSUE #1909: Black is not formatting anything
Since #1877, Black is no longer formatting cursorless-talon.

Manually running `black` on _any_ Python file, or even the entire directory, outputs:

```console
> black cursorless-talon
No Python files are present to be formatted. Nothing to do 😴
```

Commenting out this line makes Black work as expected again:

https://github.com/cursorless-dev/cursorless/blob/7369ceaa0dadb877549e328b518ad29f996e4513/pyproject.toml#L3

```console
> black cursorless-talon
All done! ✨ 🍰 ✨
59 files left unchanged.
```

auscompgeek:
Fixes #1909

### DIFF
--- a/data/playground/python/values.py
+++ b/data/playground/python/values.py
@@ -26,7 +26,6 @@ val = 1.5
 val /= 25
 
 val1, val2 = 0, 5
-val1 = 0, val2 = 5
 val1 = val2
 
 def my_funk(value: str) -> str:
--- a/data/playground/python/values.py
+++ b/data/playground/python/values.py
@@ -5,6 +5,7 @@
 # "chuck value"
 # "chuck every value"
 
+
 # the argument value "True" that is part of "b=True"
 # is "value" scope
 def func(b=True, c=True):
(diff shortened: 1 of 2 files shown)


auscompgeek:
😕 do we intentionally have invalid Python syntax in that playground file?

https://github.com/cursorless-dev/cursorless/blob/7369ceaa0dadb877549e328b518ad29f996e4513/data/playground/python/values.py#L29

pokey:
I don't think it's necessarily a bad thing to have bad syntax in the playground

pokey (commented):
Is there not a way to make the exclusion work?

auscompgeek:
I don't know, it was 1am my time and I just wanted to get black working again

pokey (approved):
Ok looks like `force-exclude` was supposed to be a regex. It's strange that there was no error thrown, just silently failing to format any files 🤔

auscompgeek:
That's… quite amazing. Possibly worth filing an issue, but we should update black first.

auscompgeek:
Wait, I was running black on the command line wasn't I 😅

pokey:
> Wait, I was running black on the command line wasn't I 😅

yes, appears so, why?

PR #1911, opened 2023-09-24
initial nix language support

fidgetingbits:
EDIT: I removed the original PR text, as no longer relevant and will have made it harder to review.

This PR adds nix language support.

## Checklist

- [-] Supported` "type"` both for type annotations (eg `foo: string`) and declarations (eg `interface Foo {}`) (and added tests for this behaviour 😊)

| Supported | Tested | Term | Capture | Definition | Comment |
| - | - | - |  - | - | - |
| &check; | &check; | `list` | `@list` | List type equivalent | - |
| &check; | &check; | `inside list` | `@list.interior` | Inside of a list | - |
| &check; | &check; | `map` | `@map` | Dictionary type equivalent | - |
| &check; | &check; | `inside map` | `@map.interior` | Inside of a dictionary | - |
| &check; | &check; | `key` | `@collectionKey` | Dictionary key equivalent | - |
| &check; | &check; | `funk` | `@namedFunction` | A named function declaration | - |
| &check; | &check; | `inside funk` | `@namedFunction.interior` |  The inside of a lambda declaration | - |
| &check; | &check; | `funk name` | `@functionName` | Name of declared function | - |
| &check; | &check; | `lambda` | `@anonymousFunction` | A lambda declaration | - |
| &check; | &check; | `inside lambda` | `@anonymousFunction.interior` |  The inside of a lambda declaration | - |
| &check; | &check; | `name` | `@name` | Variable name | - |
| &check; | &check; | `value` | `@value` | Right-hand-side value of an assignment | - |
| _ | _ | `value` | `@value` | Value returned from a function | - |
| &check; | &check; | `value` | `@value` | Value of a key-value pair | - |
| &check; | &check; | `state` | `@statement` | Any single coded statement | - |
| &check; | &check; | `if state` | `@ifStatement` | An if conditional block | - |
| &check; | &check; | `condition` | `@condition` | Condition of an if block | - |
| _ | _ | `condition` | `@condition` | Condition of a while loop | - |
| _ | _ | `condition` | `@condition` | Condition of a do while style loop | - |
| _ | _ | `condition` | `@condition` | Condition of a for loop | - |
| _ | _ | `condition` | `@condition` | Condition of a ternary expression | - |
| &check; | &check; | `branch` | `@branch` | The resulting code associated with a conditional expression | - |
| &check; | &check; | `inside branch` | `@branch.interior` | The statements associated with the body of a conditional expression | - |
| &check; | &check; | `comment` | `@comment` | Single line code comment | - |
| &check; | &check; | `comment` | `@comment` | Multi-line code comment | - |
| &check; | &check; | `string` | `@string` | Single line strings | - |
| &check; | &check; | `string` | `@string` | Multi-line strings | - |
| &check; | &check; | - | `@textFragment` | Used to capture string-type nodes (strings and comments) | - |
| &check; | &check; | `call` | `@functionCall` | A function call (not a function definition) | - |
| &check; | &check; | `callee` | `@functionCallee` | Name of the function being called | - |
| &check; | &check; | `arg` | `@argumentOrParameter` | Arguments to function definition | - |
| &check; | &check; | `arg` | `@argumentOrParameter` | Arguments to function calls | - |
| &check; | &check; | `arg` | `@argumentOrParameter` | Interpolated variables in strings | - |
| _ | _ | `class` | `@class` | Class or structure declaration | - |
| _ | _ | `inside class` | `@class.interior` |  The inside of a class declaration | - |
| _ | _ | `class name` | `@className` | Name of class or structure declaration | - |
| _ | _ | `type` | `@type` | Type declarations | - |

## Checklist


fidgetingbits:
Not sure what changed, but seems `take map` no longer includes `;`.

Comments:
screenshot-2023-09-25T10-02-04

Lists:
screenshot-2023-09-25T10-02-35

Maps:
screenshot-2023-09-25T10-02-58

I played around with `@statement` a bit and realized adding all the `xxx_expression` nodes doesn't make sense. Seems `binding` is one of the main ones we want to keep though. I didn't bother removing the expressions that I don't know what they are yet, as they may still be relevant so will keep poking at it. Not showing visualize for statement, as I want to create a more exhaustive playground file for it first.

For strings I tried adding some more `@textFragment` assignments to various bits and trying a variety of string incantations, but nothing works yet. Worth noting that `indented_string_expression` node represents the `''` style strings (can think of it like python `"""` pretty much). Can't show visualization as it just doesn't work.

fidgetingbits:
91bc77d8a4c8a9eb335fe1626cc53a2d166527c5 is more testing with some new captures.

Adds if conditions which seem to work fine.

Adds function capture, but it doesn't currently over capture curried functions, so it's a little bit non-intuitive still probably.

Fixes statements to be more sane by removing expressions I added, and special case asserts as they have weird nesting. Probably some other stuff that will need to be special cased too.

Will update with some screenshots of the new captures when I get time.

fidgetingbits:
I'll break this out into multiple PRs with proper screenshots and tests soon, as I'm just kind of randomly adding/testing support for stuff as I read through other scm files and poke around, so it's already kind of all over the place and there are some questions/quirks for various pieces that make sense to have in specific PRs.

pokey (commented):
Looks like you're making some good progress! I would advise recording tests while you're developing rather than at the end. If something works like you want it to, record a test. That way you won't break it later. It's ok if a scope doesn't work as you'd like it to in some ways: I would still advise recording tests to capture the ways in which it works like you do want it to.
Recording tests is absolutely trivial once you're set up; takes about as long as manually trying a command. In addition, it's really hard for me to review this code without tests, so I won't be able to give you early feedback, because I'm not intimately familiar with tree-sitter-nix 😅

pokey:
Ok thanks for the tests, though we prefer "change" instead of "take" because it makes the tests much easier to read. See the [PR template](https://github.com/cursorless-dev/cursorless/blob/main/.github/PULL_REQUEST_TEMPLATE/new_programming_language.md?plain=1). Also, "chuck" tests are only necessary in cases where "chuck" is interesting, eg removing commas, etc

If you don't want to re-record the "take" tests, you can just find and replace `setSelection` to `clearAndSetSelection` in your recorded tests, and then run the "update test fixtures" launch config, which should handle the rest for you

auscompgeek:
I'd be happy to try this out on some real files once the corresponding parse-tree PR is merged.

fidgetingbits:
One thing I can't entirely decide is if `take lambda` from the right most lambda should capture all curried lambdas as one highlight or should individually select each (the way it works currently). Taking all curried lambdas would make it feel more like other languages when dealing with functions, which seems nice. 

I have no functional programming background, so someone with experience might think differently?

To be clear, in nix technically `x = a: b: a + b` is two lambda expressions. So if the cursor is in `a + b` I say `take lambda` I get:

image

But if the cursor is in the `a:` part I say `take lambda` I get:

image

As I alluded to above, this also gets a bit cumbersome if your cursor is in the `a + b` and say `take every arg` and expect it to highlight both `a` and `b`. Currently it would only select `b`. I think I'd rather it select all.

I think one possible exception to this should be if one of the lambdas is explicitly wrapped in `()` which to me implies some sort of explicit separation vs a regular multi-argument lambda?

Maybe this was already discussed/decided for other languages, but I didn't check yet. 

Worth noting that atm I also made it so above works as a single "named function" where `x` is the function name, so if you say `take funk` you get:

image

and `take funk name`:

image

Not sure what to think of that yet, or how it currently works in other languages. But my thinking is because nix is only lambda functions, meaning named functions won't be used elsewhere, this might be nice to use.

pokey:
huh interesting. so is currying the only way to define a lambda with multiple parameters? If so, then I think I'd lean towards treating `a: b: a+b` as lambda, and having "every arg" iterate over `a` and `b`. Lmk if you need help getting that to work

@auscompgeek @josharian @AndreasArvidsson curious to hear your thoughts tho

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeArgBat.yml
```diff
@@ -0,0 +1,39 @@
...
+    target:
+      type: primitive
+      mark: {type: decoratedSymbol, symbolColor: default, character: b}
+      modifiers:
+        - type: containingScope
+          scopeType: {type: argumentOrParameter}
+  usePrePhraseSnapshot: true
+initialState:
+  documentContents: |-
+    { lib, buildPackages, fetchurl, stdenv }:
```
pokey, line 16: this looks like an import. how come you're calling it "arg"? I'd lean towards just letting "item" handle this one, which should work automatically

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeArgBlueDot.yml
```diff
@@ -0,0 +1,39 @@
...
+    target:
+      type: primitive
+      mark: {type: decoratedSymbol, symbolColor: blue, character: .}
+      modifiers:
+        - type: containingScope
+          scopeType: {type: argumentOrParameter}
+  usePrePhraseSnapshot: true
+initialState:
+  documentContents: |-
+    { lib, buildPackages, fetchurl, stdenv, ... }:
```
pokey, line 16: what does the `...` do?

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeBranch.yml
pokey, line 1: whoah this one's pretty big 😅. These tests are a lot easier to read if they're minimal. If there's aspects of the general case that the minimal test misses, then it's worth adding another test as well

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeEveryIfState.yml
```diff
@@ -0,0 +1,37 @@
...
+    target:
+      type: primitive
+      modifiers:
+        - type: everyScope
+          scopeType: {type: ifStatement}
+  usePrePhraseSnapshot: true
+initialState:
+  documentContents: |
+    {
+      key = if a then b else c;
```
pokey, line 16: are all `if`s just inline expressions, or are there standalone `if` statements? This code looks like a ternary, which we don't match as an `if` statement in typescript, but it's worth looking at what we do in Rust, where the ternaries look more like this syntactically

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeEveryMap.yml
pokey, line 1: hmm we don't usually have an iteration scope for `map`, but I guess this is ok? cc/ @AndreasArvidsson

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeNextBranch.yml
pokey, line 1: as above; all these branch tests are really big. would be great to minimize them if possible

pokey: also, you don't need to test "next"

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeNextState.yml
pokey, line 1: as above, not necessary to test "state", as that mechanism is independent of your definitions, unless I'm missing something

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeThirdArg.yml
pokey, line 1: as with "next", I don't think you need to test ordinal scopes

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/chuckName.yml
pokey, line 1: if you're going to test "chuck name", it should do something interesting. Did you want to remove the `=`? That would make sense if the RHS could stand on its own without a LHS

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/chuckName2.yml
pokey, line 1: as above

queries/nix.scm
```diff
@@ -0,0 +1,146 @@
...
+;; This matches assert statements as stand-alone statements. This
+;; is because it's common to have a series of assert statements in a file, which
+;; would otherwise all match as a single statement. See the playground/nix/asserts.nix
+;; file for an example.
+(assert_expression
+    "assert" @statement.start
+    condition: (_)
+    ";" @statement.end
+    body: (_)
+)
```
pokey, line 15: are there tests for asserts?

queries/nix.scm
```diff
@@ -0,0 +1,146 @@
...
+[
+    (binding)
+] @statement
+;; This matches assert statements as stand-alone statements. This
+;; is because it's common to have a series of assert statements in a file, which
+;; would otherwise all match as a single statement. See the playground/nix/asserts.nix
+;; file for an example.
+(assert_expression
+    "assert" @statement.start
+    condition: (_)
```
pokey, line 12: not sure you need this line

queries/nix.scm
```diff
@@ -0,0 +1,146 @@
...
+    ";" @statement.end
+    body: (_)
```
pokey, line 14: or this line. unless a `;` can appear later in this pattern?

queries/nix.scm
```diff
@@ -0,0 +1,146 @@
...
+)
+(binding_set) @map.iteration @list.iteration @ifStatement.iteration
+(source_code) @statement.iteration @map.iteration @comment.iteration
+
+;;!! let
+;;!!   a = 1;
+;;!    <*****
```
pokey, line 21: is `a` included? If so the `<` should be one to the left

queries/nix.scm
```diff
@@ -0,0 +1,146 @@
...
+;;!!   b = 2;
+;;!    *****>
+;;!! in a + b
+(let_expression
+    "let"
+    (_) @statement.iteration @value.iteration @name.iteration
+)
+
+;; Conditionals
```
pokey, line 30: might be worth supporting `condition` as well

queries/nix.scm
```diff
@@ -0,0 +1,146 @@
...
+    condition: (_) @branch.end
+)
+;;!! key = if a then b else c;
+;;!             ^^^^^^
+(if_expression
+    "then" @branch.start
+    consequence: (_) @branch.end
+)
+;;!! key = if a then b else c;
+;;!                    ^^^^^^
```
pokey, line 48: in TS, we don't include the `else` in ternaries (ie the `:`), but we do in full `if` statements 🤔 cc/ @AndreasArvidsson

queries/nix.scm
```diff
@@ -0,0 +1,146 @@
...
+)
+
+;; Lists and maps
+
+;;!! foo = [ a b c ];
+;;!        ^^^^^^^^^
+(list_expression) @list
+
+;;!! foo = { x = 1; y = 2; };
+;;!        ^^^^^^^^^^^^^^^^^^
```
pokey, line 61: the `;` is excluded right? If not, I'd argue it should be for `map`

queries/nix.scm
```diff
@@ -0,0 +1,146 @@
...
+[
+    (attrset_expression)
+    (rec_attrset_expression)
+] @map @statement.iteration @value.iteration @name.iteration
+
+;;!! foo = { x = 1; y = 2; };
+;;!          ^
```
pokey, line 68: test please

queries/nix.scm
```diff
@@ -0,0 +1,146 @@
...
+(attrpath) @collectionKey
+
+;; Strings
+
+;;!! # foo
+;;!  ^^^^^
+(comment) @comment @textFragment
```
pokey, line 75: if you're defining text fragments, it's worth having a `"change round"` test inside a string and comment

queries/nix.scm
```diff
@@ -0,0 +1,146 @@
...
+    "''" @textFragment.start.endOf
+    (string_fragment)
+    "''" @textFragment.end.startOf
+) @string
+
+;; Hack to test these two are otherwise sane selectors
+;; [
+;;     (string_expression)
+;;     (indented_string_expression)
+;; ] @comment
```
pokey, line 93: what's this?

queries/nix.scm
```diff
@@ -0,0 +1,146 @@
...
+    universal: (identifier) @argumentOrParameter
+    body: (_)
+) @anonymousFunction @argumentOrParameter.iteration
+
+;; We define the named function as the full assignment of the lambda
+;; This means funk name becomes the variable holding the lambda
+(binding
+    (_) @functionName
+    "="
+    expression: (function_expression) @namedFunction.interior
```
pokey, line 111: if you're supporting `interior`, it's worth having a `"change inside funk"` test

queries/nix.scm
```diff
@@ -0,0 +1,146 @@
...
+;;!      ^^^^^^^^^^^ Func2 due to currying
+(function_expression
+    universal: (identifier) @argumentOrParameter
+    body: (_)
+) @anonymousFunction @argumentOrParameter.iteration
+
+;; We define the named function as the full assignment of the lambda
+;; This means funk name becomes the variable holding the lambda
+(binding
+    (_) @functionName
```
pokey, line 109: want `"funk name"` test for this

pokey (commented):
Looking really good! Excellent work. Left a bunch of small comments

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeArgBat.yml
fidgetingbits: It is because it is actually function args. A function is designated in nix by the ':'. Most nix files are a just function taking inputs and outputting an attrset.

There is an explicit import expression in nix, bit its part of an assignment.

fidgetingbits: I guess worth noting in this case the real argument to the function is actually the single set {}, but in these cases of a set being passed this way it gets destructured into directly referencable variables within the function, so effectively the sets entries passed in this way become individual args (or at least that's how I'm modeling it here). Open to discussion if this is not a good way to do it.

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeArgBlueDot.yml
fidgetingbits: It basically allows you to ignore/allow all non-explicitly defined attributes in the set. If you don't include it and someone passes a set to the function including any attribute not explicitly included, it will fail. So ... allows you to accept a set with unexpected attributes.

It actually gets a bit more complicated though, which I haven't figured it if/how I'll deal with yet. The set passed as an argument has special syntax like foo@{ a, b, ... } where after destructuring you can still refer to the set by the name foo, in which case you can then reference one of the variables swallowed by ... explicitly lol. I'm not sure if there is any scope in cursorless suited to that foo@ part or if I just have to make sure arg references still work when that is included.

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeBranch.yml
fidgetingbits: oops, ya that was a mistake lol. apparently forgot to cut out the parts I was playing with but not testing. no problem trimming all those.

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeEveryIfState.yml
fidgetingbits: Ya this is the only form of if in nix, always part of an assignment. The reason I was doing it this way is because, especially when people break up the assignment expressions into multi-line (which seems quite common), it "feels" like a normal if statement more than a ternary? But I'll also check how rust does it ya.

Not sure if @auscompgeek you have thoughts since you seem familiar with the nix language?

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/chuckName.yml
fidgetingbits: I tested in other languages iirc and chucking the name always leaves the =. Normally let can't stand alone, although there is a case where you are defining variables to be used inside of a set thats being returned as the output of a function where there is no LHS assignment, but in that case I _think_  you'd never first have the LHS name assignment in the first place, so probably okay to leave the = begore the let. Admittedly I don't know the language well enough to say it's definitely the right decision though hehe

queries/nix.scm
fidgetingbits: Not yet, I wanted to ask other peoples opinion about defining it as a standalone statement before I added a test, as the test implies it works the way we expect, but not sure what other people will expect.

queries/nix.scm
fidgetingbits: Basically me trying to test to make string scope work, which it doesn't currently (why there are no string tests atm). I think it might be a cursorless bug with the text fallback as I get an error about not finding a pair. The " " string expression works, but I still think even in that case it's not using the query, as if in the other multi-line string (indented string expression, which is like python """ but with double single ticks: '' '') type I include a regular " " string inside it will pick that " " string up only, which is wrong. The comment hack was to be sure if I don't use @string the two string expression queries work correctly as I'd expect, which they do (also why I think it may be a cursorless bug in string matching).

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeArgBat.yml
pokey: huh that's not now we do it in JS, but JS supports multiple args. For example, in JS:

```ts
function aaa({bbb, ccc}) {}
```

In this case `bbb` and `ccc` are referenceable in the body of `aaa`, but the whole `{bbb, ccc}` is considered an arg. But if destructuring and currying are the only way to have multiple args in nix, and your eye sees these as args, maybe it's fine? What about:

```nix
{aaa, bbb}: ccc:
```

ie currying plus destructuring

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeArgBlueDot.yml
pokey: interested. tempted to have `arg` refer to the entire `foo@{ a, b, ... }`. Then you could still use `item` for `a`, `b`, etc

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeEveryMap.yml
pokey: Ok yeah it doesn't seem like a problem to me

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/chuckName.yml
pokey: yeah I'd just remove the "chuck" test then

queries/nix.scm
pokey: cc/ @auscompgeek

queries/nix.scm
pokey: yeah the spoken "string" scope doesn't actually use the string scope type; it uses a surrounding pair scope. Is "change string" not working?

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeEveryIfState.yml
auscompgeek: Yeah, everything's an expression in Nix. I'd lean towards what we do in Haskell or Rust.

queries/nix.scm
fidgetingbits: Tested in rust and it seems the else in `let foo = if ... else...` is included for branches.

But rust doesn't have a separate ternary operators (like ? and :), nor does nix.

queries/nix.scm
fidgetingbits: Done. Also I tweaked this to add a `@_.domain` because accessing the key wouldn't work in the RHS which I think is wrong, so added a test for that as well.

queries/nix.scm
fidgetingbits: Done for comment, so will mark as resolved. Strings have to work first, before I add that test though :D

queries/nix.scm
fidgetingbits: No `change string` doesn't work. Basically strings will look like this:

```nix
{
  a = "double quoted string";
  b = ''two single quote string'';
  c = ''
    multi-line
    quote string'';
  d = "Interpolated ${value} string";
  e = "Escaped \${string} ";
  f = ''
    "Nested"
    string
    '';
}
```

However the only thing cursorless will pick up with the `change` command is the `"`-based ones, so `a`, `d`, and `e` assignments, and `"Nested"` (technically wrong). Anything `''` wrapped fails with `Couldn't find containing pair` error. My typescript fu is weak and so is my debugging in vscode, so I didn't get far in debugging it so far but it does seem to be related to falling back to text parsing.

It doesn't seem like there's no `visualize string` command, so it's also not clear how to visualize what it's finding.

queries/nix.scm
fidgetingbits: Added it, albeit it not sure how useful `@namedFunction.interior` actually is in the nix case, but I think it doesn't hurt... basically the same as `change value`. Not sure how you folks feel about that type of redundancy. I guess it may depend on whether or not ultimately others agree my calling the assignment of a lambda to a variable a named function is sane. If not I guess I may drop the named function stuff altogether.

queries/nix.scm
fidgetingbits: I was holding off until confirmation from someone that my decision to make the variable assigned the lambda the function name was sane, but added.

queries/nix.scm
fidgetingbits: right, was a typo

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeEveryIfState.yml
fidgetingbits: Playing with Rust it seems like what I've done for nix is pretty much the same. Although the rust nested if behavior seems a bit buggy/weird to me, so there are a few inconsistencies. Can discuss the rust questions here or in a separate thread if you prefer.

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeArgBat.yml
fidgetingbits: Ya good point. I took a look at some more complicated nix code than what ya usually working on, and I think probably it will make it too confusing. `item` seems like a better fit. I'll change that so the set contents aren't picked up as args.

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeArgBlueDot.yml
fidgetingbits: Ya I agree. That reminds me `item` is broken for nix lists, which are ` `-delimited not `,`-delimited. `item` does work well for `,`-delimited sets though. Perhaps that's better discussed in a separate comment thread though. Basically though `change item` in a nix list will just change the entire contents of the list, not a single item, I guess because there is no `,` found.

queries/nix.scm
pokey: oh right yeah that's a tough one. Our surrounding pair scope is likely viewing each `''` as an empty string. Unfortunately, there's not a lot you can do about that for now. See https://github.com/cursorless-dev/cursorless/issues/1812#issuecomment-1691493746

queries/nix.scm
```diff
@@ -0,0 +1,146 @@
...
+    _ @textFragment.start.endOf
+    (string_fragment)
+    _ @textFragment.end.startOf
+) @string
+
+(indented_string_expression
+    "''" @textFragment.start.endOf
+    (string_fragment)
+    "''" @textFragment.end.startOf
+) @string
```
pokey, line 90: I think you can replace this section with:
```suggestion
[
    (string_expression)
    (indented_string_expression)
] @string

(string_fragment) @textFragment
```

The `@textFragment` should point at the string without the quotes, which it appears the `string_fragment` node type is designed to do

queries/nix.scm
pokey: Have you tried "change round" inside a string? It's independent of the "string" scope used when you say "take string"

queries/nix.scm
fidgetingbits: Was a typo

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeArgBlueDot.yml
fidgetingbits: fwiw I added a file `complex.nix` to the `data/playground/nix/` to show the real-world complicated list usage that one will encounter when coding nix, and to just demonstrate that `item` doesn't currently work. See the `modules = [` on line 16 for example.

This file also has some nice adversarial lambda usage that highlights some weaknesses in the queries I have atm.

queries/nix.scm
pokey: Reopening this one to see whether "change round" works inside a string. The machinery to make that work is independent of the "string" scope used when you say "take string", so this discussion is orthogonal to https://github.com/cursorless-dev/cursorless/pull/1911#discussion_r1343335701, which is presumably the thread you marked this one as resolved in favour of

fidgetingbits: Just to clarify, do you mean try change round in a string with the cursor outside of the () or inside? I did test it today, and it works inside a string only if the cursor is inside (), and doesn't work if outside the (). I believe this is consistent with how it works with comment, but I'll have to confirm tomorrow.

fidgetingbits: Just tested again and for comments and strings of both types (`"` and `''`) you can `change round` with cursor inside parens and it will work (delete the `(...`) and place cursor there). But you can't do it with the cursor outside of the parens. I tested other languages and this seems to be consistent. For some reason I thought cursorless would jump to the closes `()` pair on the line if the cursor was outside, but I guess I'm thinking vim...

queries/nix.scm
fidgetingbits: You are right, I didn't for the assert statement match, however as I was about to remove it I realized I could add `@condition` there which is really nice, so will keep it.

queries/nix.scm
fidgetingbits: Right ya removing body is ok.

I'm still kind of back and forth what I want to do about `assert`, and I'm starting to lean towards not special casing these as statements and just leveraging `line` scope instead. But I guess I will explain it here a bit to see if you or @auscompgeek has any thoughts.

Basically the form of an assert is: `assert <condition>; <body>` where `body` is often another `assert` (see data/playground/asserts.nix). So if I don't explicitly add `;` as the `@statement.end`, if I say `take state` on the first `assert`, it will take all of the subsequent `asserts`. Maybe this is actually okay, but it felt weird at first.

Here is an example of what I originally thought I wanted, where I `take state` on the second `assert` in the chain of asserts and it just applies to the single line and basically ignores the `body` of the expression:

image

On the contrary, this is what happens if I don't force the `assert` statement to end early by marking the `;` and having it keep the `body`, and say `take state` on the second `assert:

image

Also the caveat of special casing assert as a statement the way I do atm, is it will hijack a `take state` on a simple variable assignment, which is probably not wanted, for instance:

image

Maybe that's okay, but also I suspect a bit unintuitive. I could special case to not have variable assignments with `assert` apply. But then say the variable assignment has chain of asserts like the one earlier in the function, I may suddenly want to access each assert as an individual statement :D 

I guess this is the type of thing that could always be tweaked in a future PR after getting some use out of it |or do you prefer not changing this type of behavior once they're in?

queries/nix.scm
fidgetingbits: See previous thread about removing `body: (_)` for more details on what I'm debating about

queries/nix.scm
fidgetingbits: I added a `change inside funk` and a `change inside lambda` test

queries/nix.scm
fidgetingbits: Not sure if there is anything for me to do here, so leaving as open, but if you think this is ok I will mark as resolved.

queries/nix.scm
fidgetingbits: fwiw for now I changed it so that assert doesn't match as a statement if part of a binding, so the `foo = assert` will not longer trigger.

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeArgBlueDot.yml
pokey: If you define the `collectionItem` scope in places where it's space delimiter, it should use that there, and fallback to comma delimiter elsewhere

queries/nix.scm
pokey: Yeah you're thinking of vim 😄. What you've described is functioning as I'd expect

queries/nix.scm
pokey: I'm on the fence here. I do often swap branches in ternaries in typescript, which would break here. Otoh when it looks like a big if-else statement this behaviour is more consistent with what we do In typescript. Idk cc/ @AndreasArvidsson

queries/nix.scm
fidgetingbits: Cool, thanks for the context. I'm willing to try to fix it, so happy to discuss. On the one hand it would be nice to get any nix support in mostly works but on the other hand proper string handling seems like it would be super useful to have right away, so seems worth trying even if it takes a bit of time.

pokey: Actually probably shouldn't be too hard; I'd have a look at where that map is referenced, and you should have access to the editor at that point, so you can do `editor.document.languageId` to get the language id and use a custom map if it's nix

fidgetingbits: Okay that's fixed and pushed. The only thing is naming/organization:

- I called returning `delimiterToText` as `getSimpleDelimiterMap` to kind of mirror `complexDelimiterMap`. 

- I'm not sure is if you still want a `delimiterMap.ts` standalone file, and then a `getSimpleDelimiterMap.ts` only for that function. Now `delimiterToText` isn't referenced anywhere else, so seemed maybe okay to keep them together. 

- I also thought about adding `getComplexDelimiterMap()`, but atm because `complexDelimiterMap` only ever uses the keys from the `delimiterToText` it won't change even if the language has different values, so seemed unnecessary. These are all things I could guess at your preferences, but may as well ask instead.

- I also noticed `leftToRightMap` in that file isn't actually used anywhere so can be deleted I think, but also not sure about doing that as part of a totally unrelated code change, to keep commits clean.

fidgetingbits:
> I'd be happy to try this out on some real files once the corresponding parse-tree PR is merged.

@auscompgeek fyi nix support merged into parse-tree a little while ago in case you didn't see.

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/nix/changeEveryIfState.yml
fidgetingbits: I already forget what the weird rust behavior I had wanted to mention was, so marking this as resolved since as is nix works the same as rust.

queries/nix.scm
```diff
@@ -0,0 +1,312 @@
...
+    expression: (function_expression) @namedFunction.interior
+) @namedFunction @functionName.domain
+
+;; Calls to functions are a bit finicky, because of everything being curried lambdas
+(
+    (apply_expression) @dummy @functionCall @argumentOrParameter.iteration
+    (#not-parent-type? @functionCall apply_expression)
+)
+
+;; This is gross, but not sure how to do fuzzy matching against an unknown number of
```
fidgetingbits, line 184: I'm curious if someone is aware of a better way to do this? I couldn't see anything in other .scm files, the tree sitter query documentation, or chatgpt. Assuming there isn't something I missed, it almost seems like something a special predicate could handle more elegantly?

queries/nix.scm
pokey: Ok a couple surprising things when exploring the `assert` flie:

## `"call"`

image

To be honest, I wasn't expecting asserts to match "call" at all. It feels more like a statement to me than calling an "assert" function. But I could see that going either way cc/ @auscompgeek

The more surprising thing is that while the range seems reasonable (assuming we want asserts to be calls), the domain was quite surprising. It seems to contain everything after the assert statement. Hence the increasingly light color of each subsequent assert in the above screenshot.

I'm guessing that has something to do with this whole thing being pure functional, so it's kinda like how `let x in ...` technically includes the `...` in langs like clojure. But I'm not sure it's desirable behaviour for the "call" scope

This "call" in particular confused me:

image

Should it not include `"a"`? If not, should it really include the `;`? Is that assert statement even valid? I made it up 😅

## `"state"`

image

Agreed that `assert` should be a statement, and I'm happy with your ranges here. However, as with `"call"` above, it seems to be exhibiting the same surprising nested domain behaviour

And agreed that it shouldn't be a statement if it's the rhs of an assignment. Tbh I have no idea what that's supposed to do, but it doesn't look like a statement to me 😅

## `"branch"`

image

This one was also surprising to me. I guess it makes more sense in a one-line assert? cc/ @auscompgeek 

## `"callee"`

image

As above, the fact this is a "callee", as well as the domain, were surprising to me

## `"condition"`

image

This one makes the most sense of any to me, even the domain. I guess if people see assert statements this way, then maybe the other domains make sense as well? Still not totally sure bout "call" tho

## Tests

Looks like there are no tests for assert

- [ ] Assert as statement
- [ ] Assert condition
- [ ] Assert as call, if we decide to keep that
- [ ] Assert callee, if we decide to keep that
- [ ] Assert branch, if we decide to keep that
- [ ] Negative test for assert not being statement when on rhs of assignment (say "cursorless record error" and then do "change state")

pokey:
## `"item"`

image

I was really expecting a lot more items here 😅. In particular, for anything that you're considering a "map", the things inside should be items

queries/nix.scm
```diff
@@ -0,0 +1,312 @@
...
+;;!          ^
+;;!          xx
+(list_expression
+    element: (_) @collectionItem
+)
+
+;;!! foo = [ a b c ];
+;;!        ^^^^^^^^^
+(list_expression
+    element: (_)+ @list.interior
```
pokey, line 100: Hmm I get an error message

```
A capture with the same name may only appear once in a single pattern: list.interior 1:9-1:12,list.interior 1:13-1:16
```

If I say `"clear inside list"` with the following code:

```nix
{
    a = ["0" "1"];
}
```

pokey:
Hmm nix really does blur the line between statement and item, doesn't it?

image

My mind kind of expects the single-line ones not to be statements 😅. But that doesn't feel like a good distinction to draw. Idk cc/ @auscompgeek . Looks like there's no difference between a "statement block" and a "map"?

pokey:
## `"call"`

image

I did not expect this to be a call, especially since there is no `"callee"`. I think a good rule of thumb is that if there's a `"call"`, there should be a `"callee"`

pokey:
## `"callee"`

image

This callee where the same callee range has two nested domains was surprising. I also didn't expect an "import" to be a "call" / "callee" at all, but if people see it that way I won't object too strongly 😊

pokey (commented):
Ok left some comments. Haven't done a thorough review, but given it's marked draft, I assume you're still just looking for early feedback?

pokey:
Also, would be good to make sure you're formatting with [@AndreasArvidsson 's formatter](https://marketplace.visualstudio.com/items?itemName=AndreasArvidsson.andreas-talon); I'm getting a big autoformat when I save one of your files locally

queries/nix.scm
fidgetingbits: This is now tracked in PR #2012

packages/cursorless-engine/src/processTargets/modifiers/surroundingPair/findSurroundingPairTextBased.ts
```diff
@@ -74,7 +74,10 @@ export function findSurroundingPairTextBased(
   const document: TextDocument = editor.document;
   const fullRange = allowableRange ?? document.range;
 
-  const individualDelimiters = getIndividualDelimiters(delimiters);
+  const individualDelimiters = getIndividualDelimiters(
+    editor.document.languageId,
```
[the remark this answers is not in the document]
fidgetingbits: Fixed in #2012 version

packages/cursorless-engine/src/processTargets/modifiers/surroundingPair/getDelimiterMaps.ts
```diff
@@ -25,6 +25,26 @@ export const delimiterToText: Record<
...
+  backtickQuotes: ["`", "`"],
+  curlyBrackets: [["{", "${"], "}"],
+  doubleQuotes: ['"', '"'],
+  escapedDoubleQuotes: ['\\"', '\\"'],
+  escapedParentheses: ["\\(", "\\)"],
+  escapedSquareBrackets: ["\\[", "\\]"],
+  escapedSingleQuotes: ["\\'", "\\'"],
+  parentheses: [["(", "$("], ")"],
+  singleQuotes: ["''", "''"],
+  squareBrackets: ["[", "]"],
```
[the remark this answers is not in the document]
fidgetingbits: Fixed in #2012 version

fidgetingbits:
> ## `"item"`
> image
> 
> I was really expecting a lot more items here 😅. In particular, for anything that you're considering a "map", any contained key-value pairs should be items

This is fixed now for attrsets.

fidgetingbits:
> Hmm nix really does blur the line between statement and item, doesn't it?
> image
> 
> My mind kind of expects the single-line ones not to be statements 😅. But that doesn't feel like a good distinction to draw. Idk cc/ @auscompgeek . Looks like there's no difference between a "statement block" and a "map"?

Ya. Really there isn't much similarity to what you'd normally think of as a statement in other (non functional) languages at least. Originally I was leaning towards statements because there are so many complex expressions and stuff when setting key/value pairs, but I think using items feels more natural after playing with it for awhile. I've disabled most statement matching for now to test more, and also added better item support.

queries/nix.scm
fidgetingbits: This is fixed. I had to switch to `interior.starting` and `interior.ending` to get rid of it, but didn't look into why.

fidgetingbits:
> Also, would be good to make sure you're formatting with [@AndreasArvidsson 's formatter](https://marketplace.visualstudio.com/items?itemName=AndreasArvidsson.andreas-talon); I'm getting a big autoformat when I save one of your files locally

Weird, I do have that installed and I do get auto-formatting in .scm files. I originally thought it might be prettier conflicting or something, but if I disable prettier and then format .scm document, I don't see any other changes. What kind of changes do you see?

pokey:
Also, I see this is still draft. Did you want me to give it another review or shall I hold off while you continue to play with it?

fidgetingbits:
> > > Also, would be good to make sure you're formatting with [@AndreasArvidsson 's formatter](https://marketplace.visualstudio.com/items?itemName=AndreasArvidsson.andreas-talon); I'm getting a big autoformat when I save one of your files locally
> > 
> > Weird, I do have that installed and I do get auto-formatting in .scm files. I originally thought it might be prettier conflicting or something, but if I disable prettier and then format .scm document, I don't see any other changes. What kind of changes do you see?
> 
> I think it may just be indentation level. You might need to install [the editorconfig plugin](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) so that your VSCode will respect our `.editorconfig`. I wonder if we should add that to recommended extensions cc/ @AndreasArvidsson @auscompgeek 

Thanks, ya that was it it seems. Installed now. I am using your pre-commit hooks as well, so kinda surprised that doesn't automatically fix it either.

fidgetingbits:
> Also, I see this is still draft. Did you want me to give it another review or shall I hold off while you continue to play with it?

can hold off for now. I'm gonna tweak some stuff and there are still a few bugs that I didn't fix yet. Am looking forward to this getting in eventually though. Has been a fun experience.

pokey:
> Thanks, ya that was it it seems. Installed now. I am using your pre-commit hooks as well, so kinda surprised that doesn't automatically fix it either.

Yeah scm autoformat doesn't run as a hook yet as it's just implemented in that VSCode extension. We need to turn it into a cli

fidgetingbits:
> This one was also surprising to me. I guess it makes more sense in a one-line assert? 

Things seem a lot nicer with most of these assert issues addressed (not pushed yet). I like the idea of having branch apply only if it's a one-liner, but from searching the repo I don't see how its possible to actually have that as a query predicate? Is there some existing example of where you've already done this or would I have to add a new predicate?

pokey:
> I like the idea of having branch apply only if it's a one-liner, but from searching the repo I don't see how its possible to actually have that as a query predicate? Is there some existing example of where you've already done this or would I have to add a new predicate?

Oh I wasn't suggesting actually changing definition of "branch" depending on whether it's one or multiple lines. I think Python has taught us that making whitespace significant is a deeply problematic notion 😅. I think it would be quite surprising if a scope behaved differently if it was on one or multiple lines. Unless nix devs see them as significantly different constructs? cc/ @auscompgeek 

Lmk if you want me to take another look, or to drop into a meet-up to discuss some of this stuff

fidgetingbits:
Fair enough. As much as I hate python whitespace, I actually think there are _maybe_ a couple cases in nix where it could sort of make sense to be whitespace dependent to feel "natural", but I don't have time to explain right atm.

I'm new to coding nix so I'd rather hold off implementing weird things I assume maybe make sense and instead just push something that has the basics covered and then wait until more people (hopefully more experienced devs) get to play with it.

You can feel free to take another look now ya. 

There are two issues that I think might require new query predicates but aren't show stoppers (maybe... will see how much you groan at how I handled call arguments in apply/select expressions lol), so happy to solve them as follow up. Will explain in more detail in a comment tomorrow. 

I do hope to actually make it to a meetup at some point :D

fidgetingbits:
So to elaborate on a few changes and issues:

The statement stuff has changed to only be lines within the first `...` part of `let ... in ...` statements, which feels a lot better. 

There are I think 3 things could be done that I'm not sure how without possibly building new query predicates, so I'll just list them here.

1) The call argument handling nesting is gross, but works. I suspect there could be a predicate introduced that repeats some node match nested up to N deep or something. I'd probably rather do that as a follow up. I suppose worth noting I also had tried to use `#any-of` to match both an apply and select expression in the same big query, but ran into some odd errors which I couldn't figure out, which may be a bug I guess. `#any-of` would only fix the query duplication, not the huge gross query.

2) Probably we just won't use `assert` branch scope at all. But just touching on the idea of certain scopes only working depending on white space Like `inherit n;` on a single line could almost be thought of as a statement (maybe I'm too general with the way I think of statement). So for example you often see short assignments that might look like this:

```
  pkgs = import pkgsLibPath { inherit lib; pkgs = null; };
```

In this case intuitively I don't think the `inherit` would be make sense to be its own statement. But then you often also see longer assignments that are laid out like this:
```
  pkgs = import pkgsLibPath {
    inherit lib;
    pkgs = null;
  };
```

In the latter I find it feels natural to want to target the `inherit` line if it were a statement. Of course can just use `line` as the target, but just clarifying why in my brain it feels sort of natural to think of it differently depending on the white space.

There is a similar case with the `with` were you can have like `maintainers = with maintainers; [ onny ] ++ teams.php.members;` and:

```
{ config, lib, pkgs, ... }:

with lib;

let

  cfg = config.security.sudo;
```

Anyway, I think you are right that avoiding whitespace-based differences, but I wanted to explain anyway.

3) I took a stab at adding `inside comment` support and ran into some issues. There doesn't seem to currently be a way to use a predicate to query the contents of the matched node. If I do something like:

```
  (
    (comment) @comment.interior
    (#shrink-to-match! @comment.interior "# ?(?<keep>.*)$")
  )
```

It works-ish, but the problem is that there are two types of comment both which match as `(comment)` nodes. So by having the above requiring `#strink-to-match` then I lose the ability to match on the multi-line comments. I tried something naive like:

```
[
  (
    (comment) @comment.interior
    (#shrink-to-match! @comment.interior "# ?(?<keep>.*)$")
  )
  (
    (comment) @comment.interior
    ;; This would need to be a multi-line match, which I'm not sure shrink-to-match can do
    (#shrink-to-match! @comment.interior "/\*(?<keep>.*)\*/$")
  )
] @comment
```

But this doesn't work since `#shrink-to-match` isn't actually a match predicate per se, it's just a modifier on what was already matched, so if you try to match on the multiline variant, since `(comment)` still matches in the first case, `#shrink-to-match` will error because it doesn't find the `#`.

So this made me think we probably want something like a `#starts-with` or even just a `#regex` predicate where we could just have `regex("^#")` for single line and `#regex("^/\*")` for double line, and then have the `#strink-to-match` after.

I think that's it. Definitely understand why you like lots of small PRs, as it gets hard to track all these different issues (even for me actively working on it, let alone you lol).

pokey:
> So to elaborate on a few changes and issues:
> 
> The statement stuff has changed to only be lines within the first `...` part of `let ... in ...` statements, which feels a lot better.
> 
> There are I think 3 things could be done that I'm not sure how without possibly building new query predicates, so I'll just list them here.
> 
>     1. The call argument handling nesting is gross, but works. I suspect there could be a predicate introduced that repeats some node match nested up to N deep or something. I'd probably rather do that as a follow up. I suppose worth noting I also had tried to use `#any-of` to match both an apply and select expression in the same big query, but ran into some odd errors which I couldn't figure out, which may be a bug I guess. `#any-of` would only fix the query duplication, not the huge gross query.

Here is a proposal for a new query predicate operator: /Users/pokey/src/cursorless/data/playground/nix/complex.nix. I think it's actually fairly general purpose

>     2. Probably we just won't use `assert` branch scope at all. But just touching on the idea of certain scopes only working depending on white space Like `inherit n;` on a single line could almost be thought of as a statement (maybe I'm too general with the way I think of statement). So for example you often see short assignments that might look like this:
> 
> 
> ```
>   pkgs = import pkgsLibPath { inherit lib; pkgs = null; };
> ```
> 
> In this case intuitively I don't think the `inherit` would be make sense to be its own statement. But then you often also see longer assignments that are laid out like this:
> 
> ```
>   pkgs = import pkgsLibPath {
>     inherit lib;
>     pkgs = null;
>   };
> ```
> 
> In the latter I find it feels natural to want to target the `inherit` line if it were a statement. Of course can just use `line` as the target, but just clarifying why in my brain it feels sort of natural to think of it differently depending on the white space.
> 
> There is a similar case with the `with` were you can have like `maintainers = with maintainers; [ onny ] ++ teams.php.members;` and:
> 
> ```
> { config, lib, pkgs, ... }:
> 
> with lib;
> 
> let
> 
>   cfg = config.security.sudo;
> ```
> 
> Anyway, I think you are right that avoiding whitespace-based differences, but I wanted to explain anyway.

I would be tempted to just err on the side of more scopes. So if it makes sense multi-line, let's include it always. Can always "grand state" your way out of it or target part of the scope not trapped

>     3. I took a stab at adding `inside comment` support and ran into some issues. There doesn't seem to currently be a way to use a predicate to query the contents of the matched node. If I do something like:
> 
> 
> ```
>   (
>     (comment) @comment.interior
>     (#shrink-to-match! @comment.interior "# ?(?<keep>.*)$")
>   )
> ```
> 
> It works-ish, but the problem is that there are two types of comment both which match as `(comment)` nodes. So by having the above requiring `#strink-to-match` then I lose the ability to match on the multi-line comments. I tried something naive like:
> 
> ```
> [
>   (
>     (comment) @comment.interior
>     (#shrink-to-match! @comment.interior "# ?(?<keep>.*)$")
>   )
>   (
>     (comment) @comment.interior
>     ;; This would need to be a multi-line match, which I'm not sure shrink-to-match can do
>     (#shrink-to-match! @comment.interior "/\*(?<keep>.*)\*/$")
>   )
> ] @comment
> ```
> 
> But this doesn't work since `#shrink-to-match` isn't actually a match predicate per se, it's just a modifier on what was already matched, so if you try to match on the multiline variant, since `(comment)` still matches in the first case, `#shrink-to-match` will error because it doesn't find the `#`.
> 
> So this made me think we probably want something like a `#starts-with` or even just a `#regex` predicate where we could just have `regex("^#")` for single line and `#regex("^/\*")` for double line, and then have the `#strink-to-match` after.

I would just include the `# `. Does that make your life easier?

> I think that's it. Definitely understand why you like lots of small PRs, as it gets hard to track all these different issues (even for me actively working on it, let alone you lol).

Ha yeah 😅

Also, looks like this PR has some of the same leading / trailing / removal issues I fixed in #1962, so would be awesome if you could take a crack at fixing them. Basically:

- Leading and trailing is designed for delimiters. We only remove one of them; not both. Think of a comma-separated list, where deleting two commas would smash the siblings together. If you need to remove something before and after, just make a removal range
- You no longer need to specify `@leading.end`; it's implied to be the scope itself
- Instead of `@leading.start.endOf`, you can now just say `@leading.endOf`; start is now implied
- See my tweaks to #1962 if you're not sure what I mean; hopefully that will make things clear

pokey:
> Here is a proposal for a new query predicate operator: [fidgetingbits#1](https://github.com/fidgetingbits/cursorless/pull/1). I think it's actually fairly general purpose

We discussed this one in a meet-up and decided it was simpler to just introduce a query predicate operator called `#recursively-expand-to-parent-of-type!` where you give it a node and a type and it recursively expands the range to direct parents of the given type. So in this case it woudl be `(#recursively-expand-to-parent-of-type! @functionCallee.domain "apply_expression")`. Make sense @fidgetingbits . Lmk if you want help implementing that

fidgetingbits:
Sounds good. I do really need to get more familar with typescript, so will try to give it a whirl when I find the time.

pokey:
As proposed above, I don't think we'll end up going the route of https://github.com/fidgetingbits/cursorless/pull/1, but I captured it in a tag just in case we want it in the future, and the PR disappears from @fidgetingbits's repo for whatever reason (not sure how things work with forks). Here's the commit https://github.com/cursorless-dev/cursorless/commit/52ce5b123b0c5bd66f2860218b602244adfb05bb

AndreasArvidsson:
Okay I will convert it to draft in the meantime. 

You can decide if you want to continue to work on this pull request or close it and make multiple smaller ones. I unfortunately don't have the time to review languages I don't know in the kind of detail that Pokey did. I will more or less entrust the behavior to you. ie when you say that it's ready and you are happy with it we will merged unless I find something I definitely don't like. Sounds good?

PR #1912, opened 2023-09-26
First attempt at making exit keys configurable

Kn0rk:
So unfortunately I have no idea about typescript or vscode extensions. 

For the keyboard mode the actions that exit the keyboard mode are currently fixed. I wanted to make that configurable. So I added an array to the package.json. And then simply read the setting in at exactly the place where it is needed rather than keeping it in file scope as it was before.

My questions are:
- Should the setting array be constrained with an enum. And if so do i basically copy the keybinds.actions enum list?
- How and where would you read in the settings?

So I am hoping for some input before I continue with this. 

I will try to attend one of the next cursorless meetings. Might be easier to discuss this in person.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -24,5 +24,8 @@
   ],
   "files.eol": "\n",
   "typescript.enablePromptUseWorkspaceTsdk": true,
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "[typescript]": {
+    "editor.defaultFormatter": "vscode.typescript-language-features"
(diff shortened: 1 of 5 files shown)


pokey:
Which actions did you want to have exit Cursorless mode? We generally introduce a setting if different users have different needs. Maybe your action list is just better, and we should make it the default 🤷

pokey:
This branch now appears to be a working branch for some experiments you're running. I'm marking as draft for now; you might want to close the PR tho if it's going to pivot heavily, which it seems it might?

Kn0rk:
Yes it will probably pivot quite a bit.

PR #1919, opened 2023-09-29
Support "name" and "value" in guard of Python for-each statements

pokey:
- Depends on https://github.com/cursorless-dev/cursorless/pull/1922, which migrates `"name"` in Python
- Depends on #1924, which migrates `value` in Typescript

## Checklist

### DIFF
--- a/queries/python.scm
+++ b/queries/python.scm
@@ -84,7 +84,7 @@
 ;;!             ^^^
 ;;!  ---------------
 (for_statement
-  right: (identifier) @value
+  right: (_) @value
   ":" @value.domain.end
 ) @value.domain.start.startOf
 
--- a/queries/python.scm
+++ b/queries/python.scm
@@ -80,6 +80,14 @@
   (_) @value
 ) @_.domain
 
+;;!! for aaa in bbb:
+;;!             ^^^
+;;!  ---------------
+(for_statement
+  right: (identifier) @value
+  ":" @value.domain.end
+) @value.domain.start.startOf
+
 (comment) @comment @textFragment
 
 (string
--- a/queries/python.scm
+++ b/queries/python.scm
@@ -84,9 +84,10 @@
 ;;!             ^^^
 ;;!  ---------------
 (for_statement
+  left: (_) @name
   right: (_) @value
-  ":" @value.domain.end
-) @value.domain.start.startOf
+  ":" @_.domain.end
+) @_.domain.start.startOf
 
 (comment) @comment @textFragment
 
--- a/queries/python.scm
+++ b/queries/python.scm
@@ -80,8 +80,13 @@
   (_) @value
 ) @_.domain
 
+;; value:
 ;;!! for aaa in bbb:
 ;;!             ^^^
+;;!  ---------------
+;; name:
+;;!! for aaa in bbb:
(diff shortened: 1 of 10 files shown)


pokey:
Make sense? I should prob do this for TS as well?

queries/python.scm
```diff
@@ -80,6 +80,14 @@
   (_) @value
 ) @_.domain
 
+;;!! for aaa in bbb:
+;;!             ^^^
+;;!  ---------------
+(for_statement
+  right: (identifier) @value
```
josharian, line 87: why only identifier?

josharian (commented):
On the question of whether we should do this, my 2c is that it would be useful to try to explain, in some vague human-oriented way, what "name" and "value" mean.

I still don't have a sense for it; I can't reliably predict which I should expect to show up where.

For example, in

```python
for aaa in bbb:
  pass
```

does `name fine` refer to `aaa`? (`value fine` refers to `bbb`, I believe.) If not, why not? If so, why?

queries/python.scm
pokey: oh nice catch. I guess that's the danger of the new "parse tree" action: we have a new vector for copy-paste errors 😅

pokey:
> On the question of whether we should do this, my 2c is that it would be useful to try to explain, in some vague human-oriented way, what "name" and "value" mean.
> 
> I still don't have a sense for it; I can't reliably predict which I should expect to show up where.
> 
> For example, in
> 
> ```python
> for aaa in bbb:
>   pass
> ```
> 
> does `name fine` refer to `aaa`? (`value fine` refers to `bbb`, I believe.) If not, why not? If so, why?

yeah I think "name" should def be `aaa` here if `bbb` is value. I can't say I have a super precise definition for what value should be, it just felt intuitively right. I'm not sure we need to be able to precisely state it, because a) they can see it with the visualizer, and b) it's gonna be controlled by system I in the end anyway. But I'm probably just making excuses 😄. Maybe we should discuss?

pokey:
Fwiw the field names they're using (`left` and `right`) are the same we expect for name / value pairs in assignments, so I think that's a good sign. And the "name" here supports destructuring, which we support with names in assignment statements.

Maybe a "value" is a source from which data flows? Eg return value flows out of function, assignment value flows into the variable, etc

AndreasArvidsson:
I like this change and would like to see it for more languages if you have the time. Personally I feel that typescript and java that has basically an identical syntax should support this as well.

pokey:
Cool. Do you agree the domain should be just the guard, or do we want the whole loop?

AndreasArvidsson:
That sounds reasonable, but is also a bit inconsistent with how we do name for functions for example.

pokey:
I dont have a strong feeling one way or the other. My thinking was that it doesn't feel strongly enough like a value / name that we'd want users to get trapped there from anywhere in the body. But agreed it's a bit inconsistent

pokey:
Ok I added Typescript, but not Java, as Java name and value scopes aren't migrated, so I'm holding off, as I don't want to mess with legacy matchers, and we prefer languages be migrated by people who use them

AndreasArvidsson (approved):

PR #1929, opened 2023-10-04
Fix value scope nesting in Typescript

pokey:
We only want nested value scopes if there is more than one declarator in a statement.

Before:

image

After:

image

Unfortunately we have no good way to test this change until we have https://github.com/cursorless-dev/cursorless/issues/1524

**Edit**: I found a way to test it, but it's a bit hacky.

## Checklist


packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/typescript/changeNextValue.yml
pokey, line 1: When you say "next value" from the start of the statement, if there's a nested value, you'll get it; if not, you'll end up with the value from the next line

AndreasArvidsson (approved):

PR #1939, opened 2023-10-12
Write spokenForms.json from Talon

pokey:
- This PR is the Talon side of #1940

## Checklist

- [-] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/) (will do in follow-up)
- [-] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [x] I have not broken the cheatsheet
- [x] have the json have its own version number

Issue #1023, opened 2022-10-10
Support Cursorless in vscode.dev


pokey:
In order to make it easier for people to use Cursorless, we'd like to support Cursorless when VSCode is running in a web browser.  The approach is roughly as follows:

- [ ] https://github.com/pokey/command-server/issues/13
- [ ] https://github.com/cursorless-dev/cursorless/issues/945
- [ ] https://github.com/cursorless-dev/vscode-parse-tree/issues/53 (though by this point we may be embedding parse tree extension in Cursorless via nx)
- [ ] Add support for Cursorless to bundle as a [web extension](https://code.visualstudio.com/api/extension-guides/web-extensions)

PR #1940, opened 2023-10-12
Get custom spoken forms from Talon

pokey:
- This PR is the VSCode side of https://github.com/cursorless-dev/cursorless/pull/1939

## Checklist

- [-] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/) (tests added in #1941)


packages/cursorless-engine/src/nodeCommon/README.md
pokey, line 1: Will move this dir into its own package as part of https://github.com/cursorless-dev/cursorless/issues/1023; for now just making a dir to start collecting these things

packages/cursorless-engine/src/nodeCommon/TalonSpokenFormsJsonReader.ts
pokey, line 1: I tried to keep this file totally focused on the transport layer (watching a json file). That way we can swap it out for Talon rpc, Rango rpc, etc depending on the context

pokey:
ok @AndreasArvidsson this one is also ready for review

packages/common/src/util/Disposer.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.push(...disposables);
+  }
+
+  public push(...disposables: Disposable[]) {
+    this.disposables.push(...disposables);
+  }
+
+  dispose(): void {
+    this.disposables.forEach(({ dispose }) => {
+      try {
```
AndreasArvidsson, line 22: I assume that `Disposable.from` doesn't have this behavior?

packages/cursorless-engine/src/generateSpokenForm/CustomSpokenFormGeneratorImpl.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  getCustomRegexScopeTypes() {
+    return this.customSpokenForms.getCustomRegexScopeTypes();
+  }
+
+  get needsInitialTalonUpdate() {
+    return this.customSpokenForms.needsInitialTalonUpdate;
+  }
+
+  dispose = this.disposer.dispose;
```
AndreasArvidsson, line 58: Not sure if I'm a fan of this syntax

packages/common/src/util/Disposer.ts
pokey: Not sure. That's a VSCode thing tho; we need this in the engine

packages/cursorless-engine/src/generateSpokenForm/CustomSpokenFormGeneratorImpl.ts
pokey: It does have its drawbacks. I think I ran into an issue at some point where depending on when the disposer got initialised the method could be undefined. I'll change to something more verbose

packages/common/src/util/Disposer.ts
```diff
@@ -0,0 +1,30 @@
...
+        dispose();
+      } catch (e) {
+        // do nothing; some of the VSCode disposables misbehave, and we don't
+        // want that to prevent us from disposing the rest of the disposables
```
pokey, line 26: log the error

packages/common/src/util/Disposer.ts
```diff
@@ -0,0 +1,30 @@
+import { Disposable } from "../ide/types/ide.types";
+
+/**
+ * A class that can be used to dispose of multiple disposables at once. This is
+ * useful for managing the lifetime of multiple disposables that are created
+ * together. It ensures that if one of the disposables throws an error during
+ * disposal, the rest of the disposables will still be disposed.
+ */
+export class Disposer implements Disposable {
```
pokey, line 9: see if we actually need to dynamically push anywhere. If not, just have a function

packages/cursorless-engine/src/nodeCommon/TalonSpokenFormsJsonReader.ts
```diff
@@ -0,0 +1,79 @@
...
+   * @returns A {@link Disposable} which will stop the callback from running
+   */
+  onDidChange = this.notifier.registerListener;
+
+  async getSpokenFormEntries(): Promise<SpokenFormEntry[]> {
+    let payload: TalonSpokenFormsPayload;
+    try {
+      payload = JSON.parse(await readFile(this.spokenFormsPath, "utf-8"));
+    } catch (err) {
+      if ((err as any)?.code === "ENOENT") {
```
auscompgeek, line 57: You _could_ type annotate the catch instead:
```suggestion
    } catch (err: any) {
      if (err?.code === "ENOENT") {
```
but the idiomatic thing would be to do an `instanceof` check

packages/cursorless-engine/src/spokenForms/CustomSpokenForms.ts
```diff
@@ -0,0 +1,182 @@
...
+       * like custom regexes, where the user can create arbitrary ids.
+       */
+      const ids = Array.from(
+        new Set([
+          ...Object.keys(defaultEntries),
+          ...Object.keys(customEntries),
+        ]),
+      );
+      // FIXME: How to avoid the type assertions here?
+      this.spokenFormMap_[entryType] = Object.fromEntries(
```
pokey, line 137: add check that spoken forms is not empty as sanity check

packages/cursorless-engine/src/spokenForms/CustomSpokenForms.ts
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  getCustomRegexScopeTypes(): CustomRegexScopeType[] {
+    return Object.keys(this.spokenFormMap_.customRegex).map((regex) => ({
+      type: "customRegex",
+      regex,
+    }));
+  }
+
+  dispose = this.disposer.dispose;
```
pokey, line 181: this one too

packages/cursorless-engine/src/spokenForms/CustomSpokenForms.ts
```diff
@@ -0,0 +1,182 @@
...
+       * because the Talon side is old. The default entry could be missing entries
+       * like custom regexes, where the user can create arbitrary ids.
+       */
+      const ids = Array.from(
+        new Set([
+          ...Object.keys(defaultEntries),
+          ...Object.keys(customEntries),
+        ]),
+      );
+      // FIXME: How to avoid the type assertions here?
```
pokey, line 136: look into removing casts. but not too hard. at least try to remove `as any`

packages/common/src/util/Disposer.ts
pokey: @AndreasArvidsson looks like we do use dynamic push [here](https://github.com/cursorless-dev/cursorless/pull/1940/files#diff-404dfa5929061d4e42ab1fa85c49e3b31291731fc6aeaac6602fa48f618f9802R38-R42). So I'd be tempted to leave it as a class. Shall I leave it as `Disposer`? Maybe `CompositeDisposable`? `RobustCompositeDisposable` to signal it proceeds if there's an error in one of them?

packages/cursorless-engine/src/nodeCommon/TalonSpokenFormsJsonReader.ts
pokey: c1bd390f54c05717b870e236b95b0277b6a520c4 contains an alternative approach, courtesy of chatgpt. wdyt?

packages/cursorless-engine/src/nodeCommon/TalonSpokenFormsJsonReader.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  constructor(private fileSystem: FileSystem) {
+    this.init();
+  }
+
+  private async init() {
+    const parentDir = path.dirname(
+      this.fileSystem.cursorlessTalonStateJsonPath,
+    );
+    await mkdir(parentDir, { recursive: true });
```
pokey, line 30: do this in extension init. might be able to remove some of this async stuff then too. and then maybe disposer could turn into function?

packages/common/src/util/Disposer.ts
pokey: ok managed to remove all dynamic usage, and turns out we don't actually use any composites in this PR; changed to composite function and moved it to the next PR

pokey:
ok @AndreasArvidsson all feedback addressed; this one is ready for review again. I managed to get down to 2 type assertions, and they're more reasonable ones. Factoring it out into a function and using a for loop did simplify things 👍

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
```diff
@@ -1,7 +1,14 @@
 import { Disposable, FileSystem, PathChangeListener } from "@cursorless/common";
 import { RelativePattern, workspace } from "vscode";
+import * as path from "path";
 
 export class VscodeFileSystem implements FileSystem {
+  constructor(public cursorlessDir: string) {}
+
+  public get cursorlessTalonStateJsonPath() {
```
AndreasArvidsson, line 8: I would just prefer an attribute set in the constructor instead of a getter.

AndreasArvidsson (approved):

Issue #1846, opened 2023-09-04
Bump vscode engine version


pokey:
- Depend on 1-year old engine version
- If user has on old VSCode version, they'll just stop getting Cursorless updates until they upgrade, which seems reasonable

auscompgeek:
Probably a good idea anyway since we no longer test the oldest version we declare support for (see investigation in https://github.com/cursorless-dev/cursorless/pull/1399#issue-1659467109)

Issue #1808, opened 2023-08-18
Switch file watcher to use VSCode impl when available


pokey:
I'm not in love with the polling implementation. I would prefer if we added support for file watchers if you have a recent enough version.

_Originally posted by @AndreasArvidsson in https://github.com/cursorless-dev/cursorless/pull/1802#issuecomment-1681879094_

PR #1945, opened 2023-10-12
Switch file watcher to stop use polling; bump vscode to 1.66.0

pokey:
- Fixes https://github.com/cursorless-dev/cursorless/issues/1808
- Fixes https://github.com/cursorless-dev/cursorless/issues/1846

## Checklist

### DIFF
--- a/packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
+++ b/packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
@@ -1,12 +1,12 @@
 import { Disposable, FileSystem, PathChangeListener } from "@cursorless/common";
 import * as vscode from "vscode";
-import { RelativePattern, Uri, workspace } from "vscode";
+import { RelativePattern, workspace } from "vscode";
 
 export class VscodeFileSystem implements FileSystem {
   watchDir(path: string, onDidChange: PathChangeListener): Disposable {
     // FIXME: Support globs?
     const watcher = workspace.createFileSystemWatcher(
-      new RelativePattern(Uri.file(path), "**"),
+      new RelativePattern(path, "**"),
     );
 
     return vscode.Disposable.from(
--- a/packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
+++ b/packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
@@ -1,5 +1,4 @@
 import { Disposable, FileSystem, PathChangeListener } from "@cursorless/common";
-import * as vscode from "vscode";
 import { RelativePattern, workspace } from "vscode";
 
 export class VscodeFileSystem implements FileSystem {
@@ -9,11 +8,10 @@ export class VscodeFileSystem implements FileSystem {
       new RelativePattern(path, "**"),
     );
 
-    return vscode.Disposable.from(
-      watcher,
-      watcher.onDidChange(onDidChange),
-      watcher.onDidCreate(onDidChange),
-      watcher.onDidDelete(onDidChange),
-    );
+    watcher.onDidChange(onDidChange);
+    watcher.onDidCreate(onDidChange);
+    watcher.onDidDelete(onDidChange);
+
+    return watcher;
   }
 }
--- a/packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
+++ b/packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
@@ -1,52 +1,15 @@
-import {
-  Disposable,
-  FileSystem,
-  PathChangeListener,
-  walkFiles,
-} from "@cursorless/common";
-import { stat } from "fs/promises";
-import { max } from "lodash";
+import { Disposable, FileSystem, PathChangeListener } from "@cursorless/common";
+import { RelativePattern, Uri, workspace } from "vscode";
 
 export class VscodeFileSystem implements FileSystem {
   watchDir(path: string, onDidChange: PathChangeListener): Disposable {
-    // Just poll for now; we can take advantage of VSCode's sophisticated
-    // watcher later. Note that we would need to do a version check, as VSCode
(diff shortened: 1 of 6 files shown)


packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
```diff
@@ -1,52 +1,15 @@
...
-import { stat } from "fs/promises";
-import { max } from "lodash";
+import { Disposable, FileSystem, PathChangeListener } from "@cursorless/common";
+import { RelativePattern, Uri, workspace } from "vscode";
 
 export class VscodeFileSystem implements FileSystem {
   watchDir(path: string, onDidChange: PathChangeListener): Disposable {
-    // Just poll for now; we can take advantage of VSCode's sophisticated
-    // watcher later. Note that we would need to do a version check, as VSCode
-    // file watcher is only available in more recent versions of VSCode.
```
auscompgeek, line 14: Do we remember what versions this was referring to?

I suppose the types must exist if the type checker is happy, so I'm sure we're not dropping support for any version 😄

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
```diff
@@ -1,52 +1,15 @@
...
-      this.maxMtimeMs = maxMtime;
-      this.onDidChange();
-    }
-  }
-
-  dispose() {
-    clearInterval(this.timer);
+    // FIXME: Support globs?
+    const watcher = workspace.createFileSystemWatcher(
+      new RelativePattern(Uri.file(path), "**"),
```
AndreasArvidsson, line 8: You don't need to convert the string path to a uri. The first argument can just be a string.

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
```diff
@@ -1,52 +1,15 @@
...
+    );
+    watcher.onDidChange(onDidChange);
```
AndreasArvidsson, line 10: We should probably handle these disposables as well

https://github.com/cursorless-dev/cursorless/blob/e2198dbd50faf9a38fe143916a285356ed14ef68/packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts#L419-L425

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
pokey: All of the examples [in the docs](https://code.visualstudio.com/api/references/vscode-api#RelativePattern:~:text=Out%20of%20workspace%20file%20watching) use `Uri.file`, so I'd be inclined to follow their lead

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
AndreasArvidsson: I couldn't find anything definitive so I would err on the side of caution.

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
pokey: Good call; I've bumped our min version to the one that supports watching arbitrary files. Note that the type system doesn't help us here, because it's the same api but it would fail at runtime if you tried to watch a file outside the user's current project

pokey:
I bumped min version in this PR to match the legacy version that we test (1.66.0), as that version is over 1 year old, and per https://github.com/cursorless-dev/cursorless/issues/1846, we're comfortable only supporting a year back, as users with older VSCode versions will just stop receiving Cursorless upgrades, which seems reasonable if they're on a very old VSCode version. Fwiw this PR requires version 1.64.0 to watch arbitrary paths

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
pokey: huh interesting. I guess that's working so sure I'll change it here

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
pokey: fwiw it could also backfire, because it's possible disposing one of these `onFoo` methods after disposing the main watcher could cause an error to be thrown. I wonder if we should at least dispose the watcher after the `onFoo` methods 🤔

pokey: k just went spelunking through VSCode code base, and the code uses an [`Emitter`](https://github.com/microsoft/vscode/blob/9e9c4e4bf713e143063414c59c8056b113f0263b/src/vs/base/common/event.ts#L946) for each event type (change / remove / add) and [disposes of its emitters](https://github.com/microsoft/vscode/blob/9e9c4e4bf713e143063414c59c8056b113f0263b/src/vs/workbench/api/common/extHostFileSystemEventService.ts#L104). From what I can tell there's no need to individually remove listeners when the emitter is disposed, and in the VSCode git extension, they only [dispose of the watcher](https://github.com/microsoft/vscode/blob/9e9c4e4bf713e143063414c59c8056b113f0263b/extensions/git/src/watch.ts#L19).

So long story short, we don't need to dispose of the watchers, and in fact I worry it could cause problems if we do dispose of them after the emitter is disposed

pokey: Assuming you have no issues with my conclusions here, I can fix your code while I'm here

AndreasArvidsson: Go ahead

pokey:
@AndreasArvidsson this one is ready to go from my perspective

AndreasArvidsson (approved):

PR #1954, opened 2023-10-16
Remove dependency on node-html-parser.

bjaspan:
This PR replaces the dependency on the node-html-parser module with a single regexp replace.

The real motivation is to make importing Cursorless into the Google internal web IDE much easier, because doing this removes dependencies on different major versions of multiple other modules (which is a problem in Google's monorepo (don't ask)).  However, there is also the general benefit that fewer dependencies means smaller code, fewer opportunities for bugs and supply chain security issues, etc.

I have not gotten my local build and test environment working yet (I'll ask for help for that soon), so this change is COMPLETELY untested beyond just compiling.

## Checklist


bjaspan:
I'm going to finish verifying this is right then re-open it.

pokey:
I don't think this code is going to be able to make it into the web ide anyway, because it's writing a local file. We'll need to abstract away this part of the code. Easiest for now is just not to support cheatsheet in web VSCode

Doesn't VSCode need to run in a web browser for it to work with the Google VSCode ide? We'll be using a different bundle for web browser, and so might have different dependencies

I would advise dropping into a meet-up to discuss this stuff. I would like to make sure we're both on the same page about the approach here

bjaspan:
Oh, right. I should have thought of that; right now I'm entirely focused on figuring out how to get the dependencies imported and working so it didn't occur to me that the cheatsheet approach wouldn't work at all.  In any case, this certainly means I can skip node-html-parser and its transitive dependencies.

bjaspan:
I do not need this PR for its original purpose anymore; I currently excise the entire cheatsheet functionality from the Google (and later, web) version.  However, I still wonder whether you'd rather replace a significant dependency with a simple regexp replace.  If not, that's fine, and I'll delete the PR.

bjaspan:
Not worth figuring out what is wrong.

Issue #1959, opened 2023-10-18
Cloning multiple items puts result into new line above


E5DR:
I am trying to add some parameters with `clone item` and want to get from
```
self:get_context_of_length(len_ctx_prev, "prev_context_table")
```
to
```
self:get_context_of_length(len_ctx_prev, "prev_context_table", len_ctx_prev, "prev_context_table")
```
where I will do some further editing to the new parameters.

Using `clone` here shows some unexpected behavior.
While `clone item air` adds an extra parameter just as expected:
```
self:get_context_of_length(len_ctx_prev, "prev_context_table", "prev_context_table", len_ctx_prev, "prev_context_table")
```
`clone two items air` brings the new items into a new line above instead:
```
len_ctx_next, "next_context_table"
local appended_content = self:get_context_of_length(len_ctx_next, "next_context_table")
```
The same happens for cloning three, four, or more items. `clone one item` works as expected though.
Possibly related to #1691?

This is in lua (no language support).

Issue #1691, opened 2023-07-27
cloning two args misses a comma


josharian:
`f(a, b)`

"clone arg air" generates `f(a, a, b)`

"clone two args air" generates `f(a, b a, b)`. Ideally it would generate `f(a, b, a, b)`.

pokey:
Strange. Which language?

josharian:
I tested in both go and typescript.

pokey:
Interesting. Nice catch!

pokey:
Does it only happen when they are the only two items?

PR #1961, opened 2023-10-19
Fix "clone" action with `ScopeTypeTarget`

pokey:
- Fixes https://github.com/cursorless-dev/cursorless/issues/1691
- Fixes https://github.com/cursorless-dev/cursorless/issues/1959
- Also strengthens typing for clone args even more cc/ @auscompgeek

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/targets/ScopeTypeTarget.ts
+++ b/packages/cursorless-engine/src/processTargets/targets/ScopeTypeTarget.ts
@@ -145,6 +145,9 @@ export default class ScopeTypeTarget extends BaseTarget<ScopeTypeTargetParameter
   protected getCloneParameters() {
     return {
       ...this.state,
+      delimiter: this.insertionDelimiter,
+      removalRange: undefined,
+      interiorRange: undefined,
       scopeTypeType: this.scopeTypeType_,
       contentRemovalRange: this.removalRange_,
       leadingDelimiterRange: this.leadingDelimiterRange_,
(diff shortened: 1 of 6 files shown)


packages/cursorless-engine/src/processTargets/targets/UntypedTarget.ts
```diff
@@ -60,6 +60,7 @@ export default class UntypedTarget extends BaseTarget<UntypedTargetParameters> {
   protected getCloneParameters() {
     return {
       ...this.state,
+      isToken: this.isToken,
```
pokey, line 63: caught this one along the way

pokey:
quick one @AndreasArvidsson

packages/cursorless-engine/src/processTargets/targets/ScopeTypeTarget.ts
```diff
@@ -145,6 +145,9 @@ export default class ScopeTypeTarget extends BaseTarget<ScopeTypeTargetParameter
   protected getCloneParameters() {
     return {
       ...this.state,
+      delimiter: this.insertionDelimiter,
+      removalRange: undefined,
+      interiorRange: undefined,
```
pokey, line 150: I wasn't entirely sure what to do with these tbh. For constructing ranges (which is what "two items" does), these parameters are ignored anyway. And this code is equivalent to what we were doing before, just more explicit about it

AndreasArvidsson (approved):

PR #1962, opened 2023-10-20
Initial lua support

fidgetingbits:
## What

Adds support for the `lua` programming language

## Checklist

- [-] Supported` "type"` both for type annotations (eg `foo: string`) and declarations (eg `interface Foo {}`) (and added tests for this behaviour 😊)

## Scope Support

| Supported | Tested | Term | Capture | Definition | Comment |
| - | - | - | - | - | - |
| &check; | &check; | `list` | `@list` | List type equivalent | - |
| &check; | &check; | `inside list` | `@list.interior` | Inside of a list | - |
| &check; | &check; | `map` | `@map` | Dictionary type equivalent | - |
| &check; | &check; | `inside map` | `@map.interior` | Inside of a dictionary | - |
| &check; | &check; | `key` | `@collectionKey` | Dictionary key equivalent | - |
| &check; | &check; | `funk` | `@namedFunction` | A named function declaration | - |
| &check; | &check; | `inside funk` | `@namedFunction.interior` |  The inside of a lambda declaration | - |
| &check; | &check; | `funk name` | `@functionName` | Name of declared function | - |
| &check; | &check; | `lambda` | `@anonymousFunction` | A lambda declaration | - |
| &check; | &check; | `inside lambda` | `@anonymousFunction.interior` |  The inside of a lambda declaration | - |
| &check; | &check; | `name` | `@name` | Variable name | - |
| &check; | &check; | `value` | `@value` | Right-hand-side value of an assignment | - |
| &check; | &check; | `value` | `@value` | Value returned from a function | - |
| &check; | &check; | `value` | `@value` | Value of a key-value pair | - |
| &check; | &check; | `state` | `@statement` | Any single coded statement | - |
| &check; | &check; | `if state` | `@ifStatement` | An if conditional block | - |
| &check; | &check; | `condition` | `@condition` | Condition of an if block | - |
| &check; | &check; | `condition` | `@condition` | Condition of a while loop | - |
| &check; | &check; | `condition` | `@condition` | Condition of a do while style loop | - |
| - | - | `condition` | `@condition` | Condition of a for loop | - |
| &check; | &check; | `condition` | `@condition` | Condition of a ternary expression | - |
| &check; | &check; | `branch` | `@branch` | The resulting code associated with a conditional expression | - |
| &check; | &check; | `comment` | `@comment` | Code comment | - |
| &check; | &check; | `string` | `@string` | Single line strings | - |
| &cross; | &cross; | `string` | `@string` | Multi-line strings | https://github.com/cursorless-dev/cursorless/pull/1962#issuecomment-1783674916 |
| &check; | &check; | - | `@textFragment` | Used to capture string-type nodes (strings and comments) | - |
| &check; | &check; | `call` | `@functionCall` | A function call (not a function definition) | - |
| &check; | &check; | `callee` | `@functionCallee` | Name of the function being called | - |
| &check; | &check; | `arg` | `@argumentOrParameter` | Arguments to functions and calls | - |
| _ | _ | `class` | `@class` | Class or structure declaration | - |
| _ | _ | `inside class` | `@class.interior` |  The inside of a class declaration | - |
| _ | _ | `class name` | `@className` | Name of class or structure declaration | - |
| _ | _ | `type` | `@type` | Type declarations | - |


fidgetingbits:
I have a query like this for matching names (and a similar one for values):

```
(assignment_statement
    (variable_list
        name: (_) @name
        ;; Handle multiple variable declarators in one statement, eg
        ;;!! local b, c = "Hello", "World"
        ;;!  ------^--^-------------------
        (#allow-multiple! @name)
    ) @_.leading.end.startOf
) @_.domain
(variable_declaration
    (assignment_statement
        (variable_list
            name: (_) @name
            ;; Handle multiple variable declarators in one statement, eg
            ;;!! local b, c = "Hello", "World"
            ;;!  ------^--^-------------------
            (#allow-multiple! @name)
        ) @_.leading.end.startOf
    )
) @_.domain
```

Originally I wanted to add an additional line `(variable_declaration) @name.domain`, but it seems that because `name` isn't actually part of the query for `variable_declaration` part itself it won't associate with names from a separate match inside of a `(variable_declaration)` node. My hope was that
because `assignment_statement` sometimes (not always) is an inner node of `variable_declaration`, then the `name` captures from that might get associated with the domain of `variable_declaration`. Is this possible somehow without repeating the duplication I've done above?

fidgetingbits:
I was starting to work on the `;;!!` style documentation for this and realized something interesting.

Atm for lua if I have `local a = 1` and chuck name, it will result in `local = 1`. 

This behavior is consistent with:
- rust `let a = 1` becomes `let = 1`
- cpp `string name = "foo";` becomes `string = "foo";`
 
However, javascript and typescript are different: `const a = 1;` becomes `1;`

Should we make sure all of these languages are actually consistent? It seems like a generic enough coding construct that it seems like it should be consistent across cursorless.

I think it's probably reasonable to just follow the way it works in js/ts, so I'm tempted to do that for lua as well, but maybe this discussion already took place when whoever implemented rust and cpp did it.

AndreasArvidsson:
Typescript is our reference language and ideally the other languages would mirror typescripts behavior went appropriate.

fidgetingbits:
Seems there isn't a way to target the 'let / const' part in ts or 'global' scope part in python. I couldnt find any issues talking about this from a quick scan either, just curious what the thoughts are on why that doesnt exist or if there are plans. Lua uses 'local' for variable scoping and it seems that type of syntax might be a useful target too? Like 'scope' or something.

fidgetingbits:
I think this is probably good for review now and has most of what's needed. Possibly some missing iterations and domains and weird edge cases I didn't think of.

Two small things:

First, I didn't bother trying to play with the `private.fieldAccess` stuff yet as I figure you folks can first settle how it will work first. Just noting it in case you want me to poke around with it or give feedback on how it might work in lua.

Second, even though lua doesn't support a ternary, there is an idiomatic equivalent which is of the form `local max = x > y and x or y`. This functions exactly like a ternary in practice, but I wasn't really sure if I should try to special case something so that it supports branches in the same way that a real ternary in another language would work. I just decided to leave it out for now. But if you think it should be added right away, let me know.

fidgetingbits:
I just realized I forgot to add multi-line string support, which won't work with the existing cursorless string parsing, similar to what I ran into for nix. lua uses `[[ abc ]]` for multi-lines.

I did actually add support for language-specific overrides in the nix PR #1911 . I could break that logic out into a separate PR so that this could benefit from it early, or can wait to implement lua multi-line string support after the nix PR lands.

pokey:
> First, I didn't bother trying to play with the `private.fieldAccess` stuff yet as I figure you folks can first settle how it will work first. Just noting it in case you want me to poke around with it or give feedback on how it might work in lua.

Up to you; if you think it would be useful, go for it

> Second, even though lua doesn't support a ternary, there is an idiomatic equivalent which is of the form `local max = x > y and x or y`. This functions exactly like a ternary in practice, but I wasn't really sure if I should try to special case something so that it supports branches in the same way that a real ternary in another language would work. I just decided to leave it out for now. But if you think it should be added right away, let me know.

Hmm. Agreed it feels a bit special. I could go either way

pokey:
> Seems there isn't a way to target the 'let / const' part in ts or 'global' scope part in python. I couldnt find any issues talking about this from a quick scan either, just curious what the thoughts are on why that doesnt exist or if there are plans. Lua uses 'local' for variable scoping and it seems that type of syntax might be a useful target too? Like 'scope' or something.

Maybe? You can always get it using "first token state" or just a hat

pokey:
I love your support tables; filed https://github.com/cursorless-dev/cursorless/issues/2010 to discuss how to formalize them

fidgetingbits:
> > I did actually add support for language-specific overrides in the nix PR #1911 . I could break that logic out into a separate PR so that this could benefit from it early, or can wait to implement lua multi-line string support after the nix PR lands.
> 
> I would be tempted to do that, because a) we prefer atomic PRs to the extent practical, and b) that PR has lots of interesting discussion points 😅

Created PR #2012, though may not be quite as easy as planned.

pokey:
## `"branch"`

image

Not sure bout this one 🤔

image

I can see the logic of this one, though, and it's consistent with the above. Idk @josharian @AndreasArvidsson wdyt?

AndreasArvidsson:
> ## `"branch"`
> image
> 
> Not sure bout this one ⬆️🤔
> image
> 
> I can see the logic of this one ⬆️, though, and it's consistent with the above. Idk @josharian @AndreasArvidsson wdyt?

I can definitely see the logic, but I do wonder if we want to include end

pokey:
## `"condition"`

image

This ⬆️ looks sound, tho I have mixed feelings about the domain of the top condition including the second two branches. I don't feel strongly, though; I guess there's a logic to it. 

image

These look wrong to me

pokey:
## `"map"`

image

Looks more like a list to me 🤔

pokey:
## `"name"`

image

In typescript and Python, I believe `"name"` would be `success, result`. But I know we have some open discussions on how to handle destructuring / multiple return values cc/ @AndreasArvidsson @josharian

pokey:
## `"state"`

image

We don't usually consider individual branches to be statements

queries/lua.scm
```diff
@@ -0,0 +1,260 @@
...
+    [
+        (function_declaration)
+        (function_call)
+    ] @statement
+    (#not-parent-type? @statement expression_list)
+)
+[
+    (block)
+    (chunk)
+] @statement.iteration @namedFunction.iteration @functionCall.iteration @ifStatement.iteration
```
pokey, line 30: you'll find that this isn't working for `ifStatement`. Annoyingly, tree-sitter labels silently get ignored after the third label on a single node 😳

pokey (commented):
Once again, fantastic first ante here. I left some comments using screenshots above. I haven't reviewed your `.scm` file in any detail as I figured we might want to discuss the behaviour mentioned in the screenshots first

fidgetingbits:
> > ## `"branch"`
> > image
> > Not sure bout this one ⬆️🤔
> > image
> > I can see the logic of this one ⬆️, though, and it's consistent with the above. Idk @josharian @AndreasArvidsson wdyt?
> 
> I can definitely see the logic, but I do wonder if we want to include end

Finally able to take a look at your responses.

My reasoning was that `if state` itself includes `end`, where as it seemed to me the branch should be distinct. I'm open to being consistent with other languages of course, but my questions are:

1) What's the issue with having the branch logic for `if` only? I will give an example why I think it's useful:

Say I have this:
image

I can say `bring branch blue red to air` and get something like this:

image

2) I don't understand why you would want to include `end`, given `if state` would already handle that? Certainly you wouldn't want it for `chuck` in the `else` branch case, but I'll give an example of why I think it's useful to not include `end` for non-`chuck` case (similar to above). Say I have this:

image

I can say `bring branch red sun after peck` and get:

image

If I include `end` this wouldn't allow that.

queries/lua.scm
fidgetingbits: Huh. So is the solution to have multiple instances of `[ ... ] @...` with up to 3 labels each or there is actually just no way to do it at all as it's a hard underlying limit?

Is that the case if you use `@_.iteration` where `@_` matches over 3 scopes as well?

Will have to do some experimenting...

pokey:
@josharian @AndreasArvidsson see above re "branch". I know you guys have stronger opinions on branch ranges than I do 😅

queries/lua.scm
pokey: > Huh. So is the solution to have multiple instances of `[ ... ] @...` with up to 3 labels each or there is actually just no way to do it at all as it's a hard underlying limit?

Yeah you can just duplicate the pattern with the remaining captures

> Is that the case if you use `@_.iteration` where `@_` matches over 3 scopes as well?

No that's fine

AndreasArvidsson:
This looks fine to me
image

fidgetingbits:
> ## `"condition"`
> image
> 
> This ⬆️ looks sound, tho I have mixed feelings about the domain of the top condition including the second two branches. I don't feel strongly, though; I guess there's a logic to it.

Ya I don't feel particularly strong about keeping it, so will make it only apply to the if part. The more I think about it the less it makes sense. I've switched it now to be a little bit more consistent and without overlaps.

> image
> 
> These look wrong to me

Thanks, that was a left over from something else I had been testing and forgot to remove.

> ## `"map"`
> image
> 
> Looks more like a list to me 🤔

Fixed. It was already a list, but map matching was too permissive (both are the same data type in lua, so they have the same node type).


> ## `"name"`
> image
> 
> In typescript and Python, I believe `"name"` would be `success, result`. But I know we have some open discussions on how to handle destructuring / multiple return values cc/ @AndreasArvidsson @josharian

I don't feel strongly about that, and I confirmed it works that way on python so I switched it to be consistent there.

> image
> 
> Looks like the second line might have the scope doubled up? I guess that's prob because two targets are getting created, as above. But I wonder if "chuck" will choke on that

I think this is fixed now, I don't see the doubling up.

All the other suggestions have been fixed now as well, in the latest force push.

I don't plan to add the fieldAccess and ternary at the moment, so may just wait for follow up PRs for those. I plan to try to finally tackle PR #2012 next.

As in aside, I have to say the scope visualizer is so so amazing. Well done!

packages/cursorless-engine/src/scopeProviders/getTargetRanges.ts
pokey, line 1: @AndreasArvidsson fyi; boundary / interior testing was broken

AndreasArvidsson: Good catch

pokey (commented):
Ok nearly there! I pushed [a bunch of changes](https://github.com/cursorless-dev/cursorless/pull/1962/files/99aa975c028d2c4bd7894b9b05f9b34b67c76add..4a530564a21fb77e339d3ab08153558593ca5090). Worth having a look; if they make sense, would be great if you could apply similar tweaks to your other PRs if they apply

I migrated a few of the tests to facet tests where the facet test was a dramatic improvement. We'd love for all tests to be facet tests, but I don't want to block PRs on that if the PR was created before facet tests existed

The last issue here is that the condition tests seem to be failing for some reason. Guessing that broke when you tweaked condition based on my earlier feedback?

packages/common/src/scopeSupportFacets/scopeSupportFacetInfos.ts
pokey, line 1: @AndreasArvidsson have a look at the facets I added here. I think it's worth distinguishing between an assignment (`foo = bar`) and a variable declarateion (`const foo = bar`) to make it clear we want to support both

pokey:
Ok I factored out a couple semi-independent changes:

- https://github.com/cursorless-dev/cursorless/pull/2218
- https://github.com/cursorless-dev/cursorless/pull/2219

I kept the code here too, but presuming those PRs don't change much before merge, we should be able to cleanly merge as the changes will be identical

fidgetingbits:
> Ok nearly there! I pushed [a bunch of changes](https://github.com/cursorless-dev/cursorless/pull/1962/files/99aa975c028d2c4bd7894b9b05f9b34b67c76add..4a530564a21fb77e339d3ab08153558593ca5090). Worth having a look; if they make sense, would be great if you could apply similar tweaks to your other PRs if they apply

Thanks for that, it's a lot cleaner.

I should finally have time to start revisiting those other PRs soon, so will try to apply this style to them asap.

> I migrated a few of the tests to facet tests where the facet test was a dramatic improvement. We'd love for all tests to be facet tests, but I don't want to block PRs on that if the PR was created before facet tests existed

I intend to keep porting the tests to the new facet tests, but as I know at least saidelike wants to start using the lua grammar asap, I feel like they can wait for a follow up PR. I plan to start porting the tests pretty much right away, so can setup a new PR.

> The last issue here is that the condition tests seem to be failing for some reason. Guessing that broke when you tweaked condition based on my earlier feedback?

I just fixed those, so hopefully it's good to go as is. I guess I learned by lesson not to force push lol, as I have no idea where that code went... I have a habit to git rebase and not merge, but my workflow doesn't work so well for PRs it seems. I guess you usually `git merge --squash upstream/main` for PR branches instead?

Suppose I should note there is still #2012, but imo that isn't really blocking for lua atm as I've been using it for a lua project for a bit and is working well.

pokey:
looks like there's a bunch of test failures. I think maybe you need to add an entry to https://github.com/cursorless-dev/cursorless/blob/main/packages/common/src/extensionDependencies.ts so that CI installs a Lua extension

fidgetingbits:
I've got something setup so I can run the github workflows locally using [act](https://nektosact.com/), so hopefully can try to jump into the docker instance and poke around a bit...

pokey:
> I noticed [here](https://code.visualstudio.com/docs/languages/identifiers) that lua is a default known identifier...

ah right in that case yeah I'd roll back your change. Nice find

pokey:
> One thing I realized though is that the nix lang support tests are passing

that's because it installs nix dependency https://github.com/cursorless-dev/cursorless/pull/1911/files#diff-22615f93d65e3ad864e07dbd196d6094287fc773d4daa41f6f275bf1ee029f7e

pokey:
> I've got something setup so I can run the github workflows locally using [act](https://nektosact.com/), so hopefully can try to jump into the docker instance and poke around a bit...

what happens when you run tests locally without act? Ie according to [instructions in contributor guide](https://www.cursorless.org/docs/contributing/#running--testing-extension-locally)?

fidgetingbits:
> what happens when you run tests locally without act? Ie according to [instructions in contributor guide](https://www.cursorless.org/docs/contributing/#running--testing-extension-locally)?

I get three test errors, but seemingly totally unrelated like:

```
TypeError: injectIde is not a function
```

and 

```
Error: Could not get pokey.cursorless extension
```

Not sure why yet.

fidgetingbits:
Ya, so after fixing the other errors (had to do with direnv and vscode issues again), all the lua tests pass locally. I also don't have any special extensions loaded into the dev profile for vscode, command-server, parse tree, and nix.. so seems to be a github workflow only issue

fidgetingbits:
I was going through the failed test log again and noticed a lot earlier than the actual flagged red errors there was lots of complaints from lua.scm about @dummy being invalid identifier, and then noticed you folks used @_dummy not @dummy in your .scm files, so I just switched it to @_dummy and it resolved the languageId problem errors later on. but now other specific lua tests are failing. 

maybe worth having a pre-commit checker or other check somewhere to prevent use of invalid identifiers like @dummy 

fwiw I think the new test failures are all for the new style scope facet tests, which my subset test run locally wasn't using, so will investigate that next. forgot you had to run those tests separately

fidgetingbits:
I'm out of ideas again. The facet coverage tests do all pass on my system (and were running):

```
    ✔ lua facet coverage
extensionHostProcess.js:147
    ✔ scopes/lua/branch.if (50ms)
extensionHostProcess.js:147
    ✔ scopes/lua/functionCallee (63ms)
extensionHostProcess.js:147
    ✔ scopes/lua/map (68ms)
extensionHostProcess.js:147
    ✔ scopes/lua/name.assignment (68ms)
extensionHostProcess.js:147
    ✔ scopes/lua/name.variable (58ms)
extensionHostProcess.js:147
    ✔ scopes/lua/namedFunction (71ms)
extensionHostProcess.js:147
    ✔ scopes/lua/value.assignment (64ms)
extensionHostProcess.js:147
    ✔ scopes/lua/value.variable (66ms)
extensionHostProcess.js:
```

The error output from the failing test log that show the actual and expected look identical to me too, so not sure what it's running into

pokey (approved):
let's ship it!

PR #1979, opened 2023-10-27
bump typescript => 5.2.2

pokey:
## Checklist

- [x] Check source maps are working for extension
- [x] Check source maps are working in test files
- [x] Don't depend on esbuild in all packages?
- [x] Try running vsix from CI locally
- [-] Use two .. for update recorded tests script


auscompgeek:
Heh, I anticipated the build failing the moment I saw the PR title 😂

auscompgeek:
We need to bump typescript-eslint and prettier before typescript anyway.

pokey:
> We need to bump typescript-eslint and prettier before typescript anyway.

would you recommend doing that in a separate PR or combine with this one?

tsconfig.base.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "module": "commonjs",
-    "moduleResolution": "nodenext",
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
```
auscompgeek, line 4: My 2c: it'll be less painful in future if we switch to ESM output (which we'd get with `module: nodenext` if we have `type: module` in each package.json).

Also `moduleResolution: bundler` would be more correct since we use esbuild, except for maybe whatever the test harness consumes, since we don't disable tsc emit?

auscompgeek:
Our typescript-eslint version disagrees with our current typescript version anyway, so we need to bump it regardless.

tsconfig.base.json
pokey: Yeah we should prob switch to bundler. Forgot bout that option

pokey: So if we switch to bundler, what should we use for `module`? Should we revert it, or use `NodeNext` anyway?

auscompgeek: Having the package.json dictate whether the emit is CJS or ESM sounds reasonable.

I'm not sure continuing to emit CommonJS is a great idea though given that locks us out of using ESM-only libraries.

pokey: Looks like we [should be using ESNext with bundler](https://www.typescriptlang.org/docs/handbook/modules/reference.html#implied-and-enforced-options-2)

pokey: Ok I eventually gave up on trying to get `type: "module"` to work. We are using `webpack.config.ts`, and importing from other packages in our monorepo. That caused all sorts of craziness becasue `webpack.config.ts` needs to be a commonjs module, so the fact that everything else had `package.json` saying it is es module caused it to lose its mind. Feel free to look through the CI failures as I flailed around in the commit history

If there's no obvious fix, I'd be inclined to ship what I have now

packages/cheatsheet/jest.config.js
```diff
@@ -2,4 +2,5 @@
 module.exports = {
   preset: "ts-jest",
   testEnvironment: "jsdom",
+  testPathIgnorePatterns: ["/node_modules/", "/out/"],
```
pokey, line 5: not sure why I had to add this one now, but seems to fix things and I confirmed that the tests are still running in this dir

tsconfig.base.json
pokey: ...and looks like module resolution bundler isn't working either; something to do how we launch our vscode tests. I would be inclined to revert to the last passing commit (63174732e525e090fa4a7f8540fc8e347ed84df6) and ship, as I've already spent the better part of a day on this 😅

packages/cheatsheet-local/tsconfig.json
```diff
@@ -8,13 +8,13 @@
...
-    "module": "es6",
     "target": "es5",
     "allowJs": true
   },
   "ts-node": {
     "compilerOptions": {
-      "module": "CommonJS"
+      "module": "NodeNext",
+      "moduleResolution": "NodeNext"
     }
```
auscompgeek, line 18: Is there a reason why there's separate ts-node compiler options here in the first place? Surely it should inherit?

packages/cheatsheet/jest.config.js
auscompgeek: I think this is a sign we should be turning on noEmit here.

tsconfig.base.json
pokey: @auscompgeek see above—unless you have some silver bullet to get `type: "module"` or `moduleResolution: "bundler"` to work, I'd be inclined to roll back to the last passing commit and tackle other stuff later. Would be nice to get TS 5.2 features and I don't think that last working commit is a step backward otherwise

packages/cheatsheet/package.json
```diff
@@ -26,7 +26,7 @@
     "@fortawesome/fontawesome-svg-core": "6.3.0",
     "@fortawesome/free-solid-svg-icons": "6.3.0",
     "@fortawesome/react-fontawesome": "0.2.0",
-    "jest": "29.5.0",
+    "jest": "29.7.0",
```
auscompgeek, line 29: The `jest` bump is already has a separate PR open:

- #1970

packages/test-harness/package.json
```diff
@@ -5,12 +5,17 @@
   "private": true,
   "main": "./out/index.js",
   "scripts": {
-    "test": "env CURSORLESS_TEST=true tsx --conditions=cursorless:bundler src/scripts/runTestsCI.ts",
-    "compile": "tsc --build",
+    "test": "env CURSORLESS_TEST=true my-ts-node src/scripts/runTestsCI.ts",
```
pokey, line 8: todo: Use `env` like this instead of `cross-env` in the places we're using `cross-env`

pokey:
@auscompgeek passing on all but windows now! What a battle. Now using type: module, module resolution: bundler, and emit declarations only 💪

Still need to fix windows and give it a final once over, but feeling pretty good about it

The trick was to bundle each package using esbuild but exclude their deps

pokey:
Oh yeah the one place I couldn't get type module working was in cursorless-org-docs (it's a docusaurus package). Prob not a big deal as it's an entry point, not a package for import, and it contains almost no code

.github/workflows/deploy.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
       - uses: pnpm/action-setup@v2
       - uses: actions/setup-node@v3
         with:
-          node-version: 16
+          node-version-file: package.json
```
pokey, line 23: This way we have a single source of truth for our node version.

packages/cursorless-vscode/package.json
```diff
@@ -71,7 +71,7 @@
     "onCommand:cursorless.showScopeVisualizer",
     "onCommand:cursorless.hideScopeVisualizer"
   ],
-  "main": "./extension.js",
+  "main": "./extension.cjs",
```
pokey, line 74: Necessary now that we are an ESM module.

packages/meta-updater/src/updatePackageJson.ts
```diff
@@ -111,13 +125,16 @@ async function getScripts(
     return scripts;
   }
 
-  const cleanDirs = ["./out", "tsconfig.tsbuildinfo"];
+  const cleanDirs = ["./out", "tsconfig.tsbuildinfo", "./dist", "./build"];
```
pokey, line 128: Didn't seem worth the trouble of keeping different clean dirs for every package; easier to just remove a standard list. Can revisit if we have a real need for it in the future.

packages/test-harness/scripts/build-tests.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+TESTS="$(find .. -name '*.test.ts')"
+
+pnpm run build:base --outdir=dist --out-extension:.js=.cjs $TESTS
```
pokey, line 6: The only thing I'm not entirely happy with about this PR is this line. This results in about 100 MBs of bundled output, because each test bundles up all of its dependencies. I couldn't come up with a workaround because our tests run in a fairly constrained environment. We can't customize the loader that node uses because we are running within VS Code, so node's loader has already been configured because it's configured at launch time.

pokey:
Ok I'm now at my wit's end on the Windows failure. For some reason [I'm getting](https://github.com/cursorless-dev/cursorless/actions/runs/6826176799/job/18565515623)

```
This version of D:\a\cursorless\cursorless\packages\test-harness\.vscode-test\vscode-win32-x64-archive-1.84.2\Code.exe is not compatible with the version of Windows you're running. Check your computer's system information and then contact the software publisher.
```

Any idea why that would that be happening? cc/ @auscompgeek @AndreasArvidsson

AndreasArvidsson:
I try to run this branch locally and got another error

```bash
> @cursorless/cursorless-vscode@0.28.0 populate-dist C:\Users\andre\repositories\cursorless\packages\cursorless-vscode
> my-ts-node ./src/scripts/populateDist/index.ts

bash: C:\Users\andre\repositories\cursorless\packages\cursorless-vscode\node_modules\.bin\..\@cursorless\common\scripts\my-ts-node.sh: No such file or directory
 ELIFECYCLE  Command failed.
```

This lines doesn't appear to be working on windows
https://github.com/cursorless-dev/cursorless/blob/b1edf2553bede43c40d9c0900b730a9c4d01ebb7/packages/cursorless-vscode/package.json#L1067

AndreasArvidsson:
I tried the alder package.json scripts and none of the `my-ts-node` lines are working

pokey:
Maybe try them again now? Still failing in CI tho 😕

pokey:
Make sure to do a pnpm install first tho

pokey:
...and we're green!! 🙌

But do give stuff a whirl locally on windows @AndreasArvidsson to make sure it works for you 😅

AndreasArvidsson:
No same error still

AndreasArvidsson:
No `my-ts-node` still isn't working for me
```
> @cursorless/cursorless-vscode@0.28.0 preprocess-svg-hats C:\Users\andre\repositories\cursorless\packages\cursorless-vscode
> my-ts-node src/scripts/preprocessSvgHats.ts

'C:\Program' is not recognized as an internal or external command,
```

AndreasArvidsson:
@pokey After I fixed the support for absolute paths in my latest commit I still can't run scripts like preprocess hats because CURSORLESS_REPO_ROOT is missing from the environmental variables

```
> @cursorless/cursorless-vscode@0.28.0 preprocess-svg-hats C:\Users\andre\repositories\cursorless\packages\cursorless-vscode
> my-ts-node src/scripts/preprocessSvgHats.ts

C:\Users\andre\repositories\cursorless\packages\common\src\testUtil\getCursorlessRepoRoot.ts:9
    throw new Error(
          ^


Error: CURSORLESS_REPO_ROOT environment variable must be set to run this script
```

AndreasArvidsson:
Ok that is fixed now. Since I must say I'm not a big fan of having to add these really complicated run scripts with env.

pokey:
Yeah it's the whole `__dirname` thing. I wish esbuild supported that properly. Basically if you're using a bundler, tho, you can't really rely on it, so it's always been a bit dicey for us. We could prob look into it in a follow up. Maybe there's some esbuild plugin we could use

pokey:
We could switch to using cross-spawn. That's more robust and I believe supports spaces. Or I could go back to my bash script for my-ts-node. Translating it to  a node script was a Hail Mary that I thought would help windows CI and turned out to be totally irrelevant but I left it because js is a bit easier to read

AndreasArvidsson:
I definitely prefer javascript over bash

pokey:
> Yeah it's the whole `__dirname` thing. I wish esbuild supported that properly. Basically if you're using a bundler, tho, you can't really rely on it, so it's always been a bit dicey for us. We could prob look into it in a follow up. Maybe there's some esbuild plugin we could use
> 
> Maybe something like https://github.com/evanw/esbuild/issues/859#issuecomment-829154955

I guess we could also use a heuristic like looking for a .git directory and a couple child files we know should be in root

But basically the old way was pretty dicey. Eg we were relying on the fact that transpiled code was at same level of nesting, etc

AndreasArvidsson:
> > Yeah it's the whole `__dirname` thing. I wish esbuild supported that properly. Basically if you're using a bundler, tho, you can't really rely on it, so it's always been a bit dicey for us. We could prob look into it in a follow up. Maybe there's some esbuild plugin we could use
> > Maybe something like [evanw/esbuild#859 (comment)](https://github.com/evanw/esbuild/issues/859#issuecomment-829154955)
> 
> I guess we could also use a heuristic like looking for a .git directory and a couple child files we know should be in root
> 
> But basically the old way was pretty dicey. Eg we were relying on the fact that transpiled code was at same level of nesting, etc

pnpm just looks for `pnpm-lock.yaml`

pokey:
Yeah that could work. Would make the function need to be async or I guess just read synchronously

Or we could bake computing the env var into the my-ts-node script now that we own it

pokey:
ok @AndreasArvidsson have a look at [my latest changes](https://github.com/cursorless-dev/cursorless/pull/1979/files/c49c2890af46d3e58dcfca8a4fc29daae46ec84f..d947dad0b88456c4fa49f9b60e2b3ab54fb18291) and lmk if things still work locally for you; I

- switched from vanilla node `spawn` to `cross-spawn` in an effort to make spawning more robust,
- set `CURSORLESS_REPO_ROOT` from `my-ts-node`
- added lint rule to ensure we keep up with all the esm stuff

AndreasArvidsson:
Works fine here

pokey:
great! have a look at the code when you get a minute; it's not nearly as long as the diff implies. the vast majority of the lines of code are in the pnpm lockfile. If you're happy, let's give @auscompgeek a chance to have a look and then let's ship it!

AndreasArvidsson (approved):

pokey:
ok @auscompgeek I'm going to go ahead and merge this one in, but feel free to leave comments when you get a minute and I can fix in follow-ups. Biggest remaining annoyance from my perspective is that every test file (eg `foo.test.ts`) gets bundled into its own output file including all of its dependencies, so locally you end up with a dir of generated files that's about 140MB

PR #1987, opened 2023-11-01
Switch from util.isDeepStrictEqual to lodash.isEqual.

bjaspan:
upgradeStrictHere.ts uses util.isDeepStrictEqual and using that version in the web version of Cursorless will be a hassle.  Use lodash.isEqual instead: ["Performs a deep comparison between two values to determine if they are equivalent."](https://lodash.com/docs#isEqual)


pokey (approved):
Looks good!

auscompgeek:
As a follow-up, since we have a long-term goal of getting cursorless-engine on the web, should we remove node from its types?

pokey:
> As a follow-up, since we have a long-term goal of getting cursorless-engine on the web, should we remove node from its types?

How would we go about doing that?

PR #1999, opened 2023-11-05
Support wrapping snippet placeholders

AndreasArvidsson:
Support wrapping placeholders/numbered holes in snippets

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/snippets/snippet.ts
+++ b/packages/cursorless-engine/src/snippets/snippet.ts
@@ -54,6 +54,12 @@ export function transformSnippetVariables(
         candidate.parent.replace(candidate, [placeholder]);
       }
     }
+
+    if (candidate instanceof Placeholder) {
+      if (candidate.index.toString() === placeholderName) {
+        candidate.parent.replace(candidate, [new Variable("TM_SELECTED_TEXT")]);
+      }
+    }
     return true;
   });
 }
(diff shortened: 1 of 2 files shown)


packages/cursorless-engine/src/snippets/snippet.ts
```diff
@@ -54,6 +54,12 @@ export function transformSnippetVariables(
         candidate.parent.replace(candidate, [placeholder]);
       }
     }
+
+    if (candidate instanceof Placeholder) {
```
pokey, line 58: Nitpicking here, but I believe this should be an else if

pokey (approved):
Looks good! I left a minor nitpick; feel free to merge once addressed

packages/cursorless-engine/src/snippets/snippet.ts
AndreasArvidsson: I mean there is no way of both branches running so logically it will be the same result, but if you prefer that syntax it's fine for me.

pokey: It's a tiny bit more efficient, and makes it clear they're mutually exclusive

AndreasArvidsson:
I just realize that there is no error message if we can't find the variable name or a substitution. They are just silently skipped. This was something you voiced in community so should we try to add some validation here?

pokey:
> I just realize that there is no error message if we can't find the variable name or a substitution. They are just silently skipped. This was something you voiced in community so should we try to add some validation here?

Yeah good point. We should prob do that both for the wrapper var and for all substitutions

pokey:
Looks like we also don't support substitutions for numbered variables; only named once. We should prob be consistent on that as well?

AndreasArvidsson:
Yes

AndreasArvidsson:
Issue created

PR #2001, opened 2023-11-06
Use dependency injection for TestCaseRecorder.

bjaspan:
Use dependency injection for getting TestCaseRecorder into CursorlessEngine. 

A future PR should perhaps move TestCaseRecorder into a new package, @cursorless/cursorless-dev or similar, to isolate developer-only dependencies that are needed in production code.


bjaspan:
Perhaps the new package should be cursorless-node, to go with another new one cursorless-web, for code specific to the host environment.

PR #2002, opened 2023-11-06
Use dependency injection for TestCaseRecorder

bjaspan:
Use dependency injection for getting TestCaseRecorder into CursorlessEngine.

A future PR should perhaps move TestCaseRecorder into a new package, @cursorless/cursorless-dev or similar, to isolate developer-only dependencies that are needed in production code.

This is an alternate approach to https://github.com/cursorless-dev/cursorless/pull/2001 requested in a comment.

It is also the most TypeScript I've written to date, so I probably did some things stylistically wrong. Class `TestCaseRecorderCommandRunnerDecorator` in particular feels very C++-like, not very TS-like, to me.

### DIFF
--- a/packages/cursorless-engine/src/api/CursorlessEngineApi.ts
+++ b/packages/cursorless-engine/src/api/CursorlessEngineApi.ts
@@ -1,19 +1,20 @@
 import { Command, HatTokenMap, IDE } from "@cursorless/common";
 import { Snippets } from "../core/Snippets";
 import { StoredTargetMap } from "../core/StoredTargets";
-import { TestCaseRecorder } from "../testCaseRecorder/TestCaseRecorder";
 import { ScopeProvider } from "@cursorless/common";
+import { CommandRunner } from "../CommandRunner";
+import { ReadOnlyHatMap } from "@cursorless/common";
 
 export interface CursorlessEngine {
   commandApi: CommandApi;
   scopeProvider: ScopeProvider;
   customSpokenFormGenerator: CustomSpokenFormGenerator;
-  testCaseRecorder: TestCaseRecorder;
   storedTargets: StoredTargetMap;
   hatTokenMap: HatTokenMap;
   snippets: Snippets;
   injectIde: (ide: IDE | undefined) => void;
   runIntegrationTests: () => Promise<void>;
+  addCommandRunnerDecorator: (commandRunnerDecorator: CommandRunnerDecorator) => void;
 }
 
 export interface CustomSpokenFormGenerator {
@@ -38,4 +39,14 @@ export interface CommandApi {
    * the command args are of the correct shape.
    */
   runCommandSafe(...args: unknown[]): Promise<unknown>;
+}
+
+export interface CommandRunnerDecorator {
+  /**
+   * @param commandRunner: A CommandRunner.
+   * @param readableHatMap: A ReadOnlyHatMap.
+   * @returns A new CommandRunner that invokes the provided CommandRunner in
+   *   addition to performing some other work.
+   */
+  wrapCommandRunner: (readableHatMap: ReadOnlyHatMap, commandRunner: CommandRunner) => CommandRunner;
 }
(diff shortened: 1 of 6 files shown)


packages/cursorless-engine/src/api/CursorlessEngineApi.ts
```diff
@@ -39,3 +40,13 @@ export interface CommandApi {
...
 }
+
+export interface CommandRunnerDecorator {
+  /**
+   * @param commandRunner: A CommandRunner.
+   * @param readableHatMap: A ReadOnlyHatMap.
+   * @returns A new CommandRunner that invokes the provided CommandRunner in
+   *   addition to performing some other work.
+   */
+  wrapCommandRunner: (readableHatMap: ReadOnlyHatMap, commandRunner: CommandRunner) => CommandRunner;
```
bjaspan: It feels wrong that all CommandRunnerDecorators should have to take a ReadOnlyHatMap. However, it is generated inside runCommand asynchronously from hatTokenMap and I did not want to have to generate it again inside the decorator.  Suggestions welcome.

packages/cursorless-vscode/src/extension.ts
```diff
@@ -48,6 +49,29 @@ import {
...
+  testCaseRecorder: TestCaseRecorder;
+
+  constructor(testCaseRecorder: TestCaseRecorder) {
+    this.testCaseRecorder = testCaseRecorder;
+  }
+  wrapCommandRunner(
+    readableHatMap: ReadOnlyHatMap,
+    commandRunner: CommandRunner,
+  ) {
+    if (this.testCaseRecorder.isActive()) {
```
pokey, line 65: I believe if you move this if statement into the wrap function in the test case recorder itself, you could do away with this whole glue class.

packages/cursorless-engine/src/cursorlessEngine.ts
```diff
@@ -65,18 +66,20 @@ export function createCursorlessEngine(
 
   ide.disposeOnExit(rangeUpdater, languageDefinitions, hatTokenMap, debug);
 
+  const commandRunnerDecorators = new Array<CommandRunnerDecorator>();
```
pokey, line 69: Nitpicking here, but we tend to prefer
```suggestion
  const commandRunnerDecorators: CommandRunnerDecorator[] = [];
```

packages/cursorless-engine/src/cursorlessEngine.ts
```diff
@@ -101,13 +104,15 @@ export function createCursorlessEngine(
...
-    testCaseRecorder,
     storedTargets,
     hatTokenMap,
     snippets,
     injectIde,
     runIntegrationTests: () =>
       runIntegrationTests(treeSitter, languageDefinitions),
+    addCommandRunnerDecorator: (decorator: CommandRunnerDecorator) => {
+      commandRunnerDecorators.push(decorator);
+    },
```
pokey, line 115: Feels odd not to return a disposable that can be used to remove the decorator. But I guess we don't have a use case for it so probably fine 🤷‍♂️

pokey (commented):
This looks great! Left a few minor comments

packages/cursorless-engine/src/cursorlessEngine.ts
bjaspan: Is there unit test infrastructure so I could add, e.g., a test to confirm that decorators added this way are actually invoked?  I ask because if I added removal I'd definitely want to test that, but of course that just points out the lack of tests for calling the decorators in the first place.

pokey: We have end-to-end tests for the test case recorder, so if your decorators weren't working, those tests would break. Does that work?

bjaspan:
Sure, that's fine. I just wanted to make sure I wasn't leaving anything
out. And I'm happier not writing the dispose code given lack of use case or
tests.

packages/cursorless-engine/src/cursorlessEngine.ts
bjaspan: Sure, that's fine. I just wanted to make sure I wasn't leaving anything
out. And I'm happier not writing the dispose code given lack of use case or
tests.

pokey (commented):
couple minor things and we're good!

pokey (approved):
Looks good!

PR #2009, opened 2023-11-08
cursorless-org: Move eslint/typescript to devDependencies

auscompgeek:
ESLint and TypeScript should definitely not end up in the bundled output for the cursorless.org website.


pokey (approved):
👍

PR #2011, opened 2023-11-08
Stop using default exports from packages/cursorless-engine/src/processTargets.

bjaspan:
This removes all default exports from packages/cursorless-engine/src/processTargets.

I'm not yet sure if this will actually help me achieve my larger goal but it is a small net improvement in the codebase in any case.


bjaspan:
I think the eslint rule we want prevent regression is

```
'no-restricted-syntax': [
  'error',
  'Literal[value=/export default/i]',
],
```

but it has to be in an .eslintrc file in packages/cursorless-engine/src/processTargets and I haven't figured out what other boilerplate it needs yet.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ParagraphScopeHandler.ts
```diff
@@ -7,11 +7,11 @@ import {
 } from "@cursorless/common";
 import { Direction, ScopeType } from "@cursorless/common";
 import { ParagraphTarget } from "../../targets";
-import BaseScopeHandler from "./BaseScopeHandler";
+import { BaseScopeHandler } from "./BaseScopeHandler";
 import { fitRangeToLineContent } from "./LineScopeHandler";
 import { TargetScope } from "./scope.types";
 
-export default class TokenScopeHandler extends BaseScopeHandler {
+export class ParagraphScopeHandler extends BaseScopeHandler {
```
pokey, line 14: Whoops nice catch 😅

pokey (approved):
🙌

bjaspan:
Here is the quick-hack shell script I wrote to make this PR. We can probably adapt it to remove default exports from everywhere in cursorless:

```
#!/bin/bash -x

cd /Users/bjaspan/cursorless/packages/cursorless-engine

git reset --hard

# "export default class <name>" is invalid syntax but gets found for
# $default_classes and cleaned up in a subsequent command.
find src/processTargets -name '*.ts' | xargs perl -MFile::Basename -pi -e 'my $f = basename($ARGV, ".ts"); s/export default class implements/export default class $f implements/'

default_classes=`find src/processTargets/ -name '*.ts' | xargs perl -lne '/export default(?: abstract)? (?:class|interface) (\w+)/ && print $1' | sort | uniq | perl -le 'print join("|", grep(chomp, <>))'`

find src/processTargets -name '*.ts' | xargs perl -pi -e 's/export default/export/'
find . -name '*.ts' | xargs perl -pi -e 's/import\s+('${default_classes}')\s+from/import { \1 } from/'

sed -i '' '/default as/d' src/processTargets/targets/index.ts
sed -i '' '/default as/d' src/processTargets/modifiers/scopeHandlers/index.ts
sed -i '' 's/TokenScopeHandler/ParagraphScopeHandler/' src/processTargets/modifiers/scopeHandlers/ParagraphScopeHandler.ts

cat >src/processTargets/modifiers/ItemStage/index.ts <<EOF
export * from "./ItemStage";
EOF
```

pokey:
> I think the eslint rule we want prevent regression is
> 
> 
> 
> ```
> 
> 'no-restricted-syntax': [
> 
>   'error',
> 
>   'Literal[value=/export default/]',
> 
> ],
> 
> ```
> 
> 
> 
> but it has to be in an .eslintrc file in packages/cursorless-engine/src/processTargets and I haven't figured out what other boilerplate it needs yet.

No we want https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-default-export.md

pokey:
Did you want to add rule in this PR or follow up? I don't feel strongly

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ParagraphScopeHandler.ts
bjaspan: Well, the compiler caught it for me. :-)

bjaspan:
Followup since I don't know how to create the appropriate .eslintrc.json file yet.

bjaspan:
Added lint rule.

pokey:
Nice! Have you double checked that it's actually biting?

bjaspan:
Yes, I confirmed that the lint rule works.

Issue #1992, opened 2023-11-03
Support markdown scope for targeting code blocks


fidgetingbits:
It would be useful to have a scope to target markdown blocks of code inside ``` pairs. I regularly find myself wanting to chuck them, change inside them, or move out of them, and it's often in a scenario where there's no visible ` characters to use instance with.

auscompgeek:
Wouldn't `change inside skis` work in this scenario?

fidgetingbits:
It doesn't work for me no.

josharian:
Bringing over some comments from Slack:

> I do a lot of work in a templating language with no cursorless support. There are lots of paired {{ and }}. I wonder whether more general text-based support for multiple paired delimiters would be helpful. That’d also handle Python’s “”” … “””. 

> ya I suppose I should try a regex scope in the meantime. haven't used one before so an excuse to dabble I guess.

some general text-based support seems like it coild be good. more examples: nix language uses double ticks ''foo'' for multi-line strings, lua uses [[foo]].

auscompgeek:
In light of #1980, it might make sense to call code blocks "notebook cells". That'd be consistent with various literate programming formats.

fidgetingbits:
Ya I think that makes sense.

fidgetingbits:
Has there been discussion of something like 'outside' or 'parent' modifier similar to 'inside' where you target the parent scope? 

I've never used jupyter so not sure, but say you were in a cell containing markdown containing a code block but the block was big enough you couldn't see any part of the parent jupyter cell but then wanted to target it, I believe atm you couldn't?

I wonder if something like that would make this type of problem you mention slightly more natural to deal with.

This type of modifier could help in other similar situations I suppose like if you are in a big python closure and have no hats from the containing function, but want to reference it? Similar how you can jump to a class from its method.

That also raises the question in relation to your counter example: if you are in a nested language that triggers a new treesitter parser, does that show up cleanly in the node tree in a way that we could query? And even if it did, I'm not sure if/how cursorless deals with nested languages atm.

auscompgeek:
> if you are in a nested language that triggers a new treesitter parser

I believe Cursorless currently doesn't support nested languages. Definitely a feature everyone wants, and _potentially_ doable after everything is moved to tree-sitter queries…

pokey:
> > if you are in a nested language that triggers a new treesitter parser
> 
[...]

PR #2012, opened 2023-11-09
Add language overrides for delimiter maps

fidgetingbits:
This is me breaking out some functionality that was part of PR #1911 since that PR has lots of work to be done and it may still benefit other languages that get in sooner. I've applied the suggested changes from pokey in that PR's code review.

I've also added a placeholder map for lua which uses `[[ ]]` for multiline support as noted in PR #1962, however I realized as I went to add it that it's not as simple as the way I did it for nix because I can't just reuse the existing `singleQuote` entry since lua actually uses single quotes. So this may actually be a bit harder, and reminds me of [this discussion](https://github.com/cursorless-dev/cursorless/issues/1992#issuecomment-1793242952). So before I go randomly hacking stuff I'm curious what you think the best approach here is.

These are the comments I had left in the Nix PR about these changes:

*    I called returning delimiterToText as getSimpleDelimiterMap to kind of mirror complexDelimiterMap.

*    I'm not sure is if you still want a delimiterMap.ts standalone file, and then a getSimpleDelimiterMap.ts only for that function. Now delimiterToText isn't referenced anywhere else, so seemed maybe okay to keep them together.

*    I also thought about adding getComplexDelimiterMap(), but atm because complexDelimiterMap only ever uses the keys from the delimiterToText it won't change even if the language has different values, so seemed unnecessary. These are all things I could guess at your preferences, but may as well ask instead.

*    I also noticed leftToRightMap in that file isn't actually used anywhere so can be deleted I think, but also not sure about doing that as part of a totally unrelated code change, to keep commits clean.

## Checklist


pokey:
I would be tempted to introduce a new delimiter type for `[[`, and then make complex delimiter list customisable so you can include it

WDYT @AndreasArvidsson

fidgetingbits:
See https://github.com/cursorless-dev/cursorless/pull/1911#discussion_r1386549043 for a comment about some code from this PR.

AndreasArvidsson:
> I would be tempted to introduce a new delimiter type for `[[`, and then make complex delimiter list customisable so you can include it
> 
> WDYT @AndreasArvidsson

I think I need a bit more info on your suggestion

pokey:
> > I would be tempted to introduce a new delimiter type for `[[`, and then make complex delimiter list customisable so you can include it
> > WDYT @AndreasArvidsson
> 
> I think I need a bit more info on your suggestion

the problem is that in lua, `[[` and `]]` can be used as delimiters for a quote. So I'm proposing to add those as a new delimiter type `doubleSquareBrackets`. Then we'd need `"string"` to include those. `"string"` is defined as a complex delimiter `string: ["singleQuotes", "doubleQuotes", "backtickQuotes"],`, so we'd add `doubleSquareBrackets` to that list, but only for Lua. 

One thing we'd need to think about is that that would cause `[[foo]]` to be viewed as a pair for our text-based matchers, and it wouldn't be considered a "square", which is probably not great. So we'd prob need a way for that pair not to be active outside of Lua

AndreasArvidsson:
> > > I would be tempted to introduce a new delimiter type for `[[`, and then make complex delimiter list customisable so you can include it
> > > WDYT @AndreasArvidsson
> > 
> > 
> > I think I need a bit more info on your suggestion
> 
> the problem is that in lua, `[[` and `]]` can be used as delimiters for a quote. So I'm proposing to add those as a new delimiter type `doubleSquareBrackets`. Then we'd need `"string"` to include those. `"string"` is defined as a complex delimiter `string: ["singleQuotes", "doubleQuotes", "backtickQuotes"],`, so we'd add `doubleSquareBrackets` to that list, but only for Lua.
> 
> One thing we'd need to think about is that that would cause `[[foo]]` to be viewed as a pair for our text-based matchers, and it wouldn't be considered a "square", which is probably not great. So we'd prob need a way for that pair not to be active outside of Lua

I agree, but we still need to make square actually work of course.

pokey:
Ok @fidgetingbits I think we're in agreement. Lmk if you need help figuring out how to put the above to practice

fidgetingbits:
> Ok @fidgetingbits I think we're in agreement. Lmk if you need help figuring out how to put the above to practice

Ya some guidance would be good, as much as I'd love to try to figure it out on my own. My typescript experience consists of what's in this PR. I suspect it would take me quote long time work out what to do.

pokey:
Ok I'll take this over when I get a minute

pokey (approved):
Ok @fidgetingbits I made some tweaks. Given some more recent discussions (https://github.com/cursorless-dev/cursorless/issues/1812#issuecomment-1691493746), I don't think this is the long-term solution, but:

- It's not too much code,
- It's not too bad an abstraction, and
- I need it for a tree-sitter upgrade I'm doing to handle strings in Python

Let me know if you're happy with my changes (see e82832f4922c328290f61aed6ca7554807c44dd0)

Also, @AndreasArvidsson would be good to get a second 👀 as I've made some changes here and we've discussed syntactic vs text pairs a lot

AndreasArvidsson (approved):

AndreasArvidsson:
Looks reasonable to me

fidgetingbits:
> Ok @fidgetingbits I made some tweaks. Given some more recent discussions ([#1812 (comment)](https://github.com/cursorless-dev/cursorless/issues/1812#issuecomment-1691493746)), I don't think this is the long-term solution, but:
> 
>     * It's not too much code,
> 
>     * It's not too bad an abstraction, and
> 
>     * I need it for a tree-sitter upgrade I'm doing to handle strings in Python
> 
> 
> Let me know if you're happy with my changes (see [e82832f](https://github.com/cursorless-dev/cursorless/commit/e82832f4922c328290f61aed6ca7554807c44dd0))

Added lua test looks good to me. Your changes look good too, reads cleaner.

PR #2016, opened 2023-11-10
Remove an import from cursorless-engine/src/util to src/processTargets.

bjaspan:
Generally, code in src/utils should not be importing from src/processTargets.  This cleans one instance of this up.  

It also changes some imports from using the index.ts file to the direct file exporting the thing.

I do not think this will actually lead to a solution to the circular-import problem I'm having because ultimately there are still clearly circular file dependencies in this code, but removing the utils -> processTargets import still seems like a small improvement.


packages/cursorless-engine/src/actions/Clear.ts
```diff
@@ -1,4 +1,4 @@
-import { PlainTarget } from "../processTargets/targets";
+import { PlainTarget } from "../processTargets/targets/PlainTarget";
```
pokey, line 1: As discussed, you should only switch from index file to direct import when both files are descendants of the folder containing the index file. This import is likely what is breaking the build, because it subverts the barrel hack that we have to serialize the imports in target dir

bjaspan:
I am revoking this PR, perhaps temporarily.

PR #2036, opened 2023-11-17
Remove most cyclic dependencies in cursorless-engine

bjaspan:
This PR removes most but not all cyclic dependencies in cursorless-engine.  The changes include:

* Import directly from implementation files, not from indexes, though not from outside the local/containing index file.
* Move utility functions outside the processTargets/targets barrel that are only used by the barrel inside the barrel.
* Move exported functions and their local dependencies that return new instances of classes to the file defining that class (since they are essentially static factory functions for that class).

This does not remove all cyclic dependencies, such as the one between BaseTarget and UntypedTarget, but with these changes the remaining cycles can be eliminated with some fairly simple textual transformations.  Perhaps a future PR can remove the remaining ones so no such transformations are needed before compiling in an environment that disallows cyclic dependencies.

There are probably other variations of these changes that would also work.


bjaspan:
Closing until I fix the broken build. It looks like I lost some import ordering.

pokey:
The build is broken because you've broken our workaround for circular deps in targets dir. We always import from the barrel within that dir so we can force a load order for those modules that ensures base target is loaded before the classes that inherit from it. You can't do those direct imports until you remove the base target / untyped target cycle somehow, possibly via async import

packages/cursorless-engine/src/processTargets/targets/UntypedTarget.ts
```diff
@@ -65,3 +63,23 @@ export class UntypedTarget extends BaseTarget<UntypedTargetParameters> {
     };
   }
 }
+export function createContinuousRangeUntypedTarget(
```
pokey, line 66: How come this function needs to be here? I know it produces an untyped target, but I'm not sure I see it as a factory

pokey (commented):
For the most part looks fine, with a couple comments, and seem my comment above for why the build is broken

PR #2037, opened 2023-11-17
Break the cyclic dependency between BaseTarget and UntypedTarget.

bjaspan:
This PR factors `createContinuousRangeTarget` out of BaseTarget into a new function, `createContinuousRangeOrUntypedTarget`, which is used by a variety of target types.  As a result, BaseTarget.ts no longer needs to import UntypedTarget (indirectly through createContinuousRange.ts).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -26,4 +26,4 @@
   "typescript.enablePromptUseWorkspaceTsdk": true,
   "typescript.tsdk": "node_modules/typescript/lib",
   "eslint.workingDirectories": [{ "pattern": "packages/*/" }]
- }
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -25,9 +25,5 @@
   "files.eol": "\n",
   "typescript.enablePromptUseWorkspaceTsdk": true,
   "typescript.tsdk": "node_modules/typescript/lib",
-  "eslint.workingDirectories": [
-    {
-      "pattern": "packages/*/"
-    }
-  ]
-}
+  "eslint.workingDirectories": [{ "pattern": "packages/*/" }]
+ }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -25,5 +25,9 @@
   "files.eol": "\n",
   "typescript.enablePromptUseWorkspaceTsdk": true,
   "typescript.tsdk": "node_modules/typescript/lib",
-  "eslint.workingDirectories": [{ "pattern": "packages/*/" }]
+  "eslint.workingDirectories": [
+    {
+      "pattern": "packages/*/"
+    }
+  ]
 }
--- a/packages/cursorless-engine/src/processTargets/targets/UntypedTarget.ts
+++ b/packages/cursorless-engine/src/processTargets/targets/UntypedTarget.ts
@@ -1,12 +1,13 @@
 import { Range } from "@cursorless/common";
 import { BaseTarget, CommonTargetParameters } from ".";
 import type { Target } from "../../typings/target.types";
-import { createContinuousRangeUntypedTarget } from "../targetUtil/createContinuousRange";
 import {
   getTokenLeadingDelimiterTarget,
   getTokenRemovalRange,
   getTokenTrailingDelimiterTarget,
 } from "../targetUtil/insertionRemovalBehaviors/TokenInsertionRemovalBehavior";
+import {createContinuousRange} from "../targetUtil/createContinuousRange";
+import {isSameType} from "../../util/typeUtils";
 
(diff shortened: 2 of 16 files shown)


auscompgeek:
Hmm, there's a fair amount of copy paste here.

Is there a reason we keep `createContinuousRange*`, `BaseTarget`, and `UntypedTarget` in separate files?

pokey:
Yeah I'm thinking maybe the right answer is to just put those in the same file? That seems less onerous than this approach. Still don't love it, but at least better than this.

@auscompgeek what's your opinion on circular deps?

bjaspan:
I tried merging BaseTarget and UntypedTarget as well.  It turns out that UntypedTarget depends on PlainTarget (via TokenInsertionRemovalBehavior) which also depends on BaseTarget.  So, to solve the problem that way, we'd have to merge BaseTarget.ts, UntypedTarget.ts, and PlainTarget.ts into a single file.  I'm fine with that solution; I went with this one only because I guessed that you would all prefer it to merging three files.

bjaspan:
Hold on. I don't think this PR even really fixes the problem. There is still a cyclic dependency between UntypedTarget.ts and createContinuousRange.ts.

I think we have to merge the three files.

bjaspan:
I removed the cyclic dependency between UntypedTarget.ts and createContinuousRange.ts.  It would be nice if we had a tool to create a dependency graph, but I believe what we have now is:

* BaseTarget.ts depends on nothing relevant.
* createContinuousRange.ts depends on nothing relevant.
* UntypedTarget.ts depends on BaseTarget.ts and createContinuousRange.ts.
* All target types depend on BaseTarget.ts, many depend on UntypedTarget.ts, and some depend directly on createContinuousRange.ts. 
*

bjaspan:
https://github.com/cursorless-dev/cursorless/pull/2047 is the merging-files approach.

bjaspan:
I've eliminated the copy-paste by introducing a new abstract base template class, CommonTarget, which provides the createContinuousRangeOrUntypedTarget functionality. By exporting it from UntypedTarget, the import graph described in my previous comment is preserved.

.vscode/settings.json
```diff
@@ -25,5 +25,9 @@
   "files.eol": "\n",
   "typescript.enablePromptUseWorkspaceTsdk": true,
   "typescript.tsdk": "node_modules/typescript/lib",
-  "eslint.workingDirectories": [{ "pattern": "packages/*/" }]
+  "eslint.workingDirectories": [
+    {
+      "pattern": "packages/*/"
+    }
+  ]
```
pokey, line 32: strange this formatting would have changed cc/ @auscompgeek

packages/cursorless-engine/src/processTargets/targets/UntypedTarget.ts
```diff
@@ -1,12 +1,38 @@
...
 } from "../targetUtil/insertionRemovalBehaviors/TokenInsertionRemovalBehavior";
+import { createContinuousRange } from "../targetUtil/createContinuousRange";
+import { isSameType } from "../../util/typeUtils";
+
+/**
+ * An abstract target base class who createContinueRangeTarget method returns either
+ * a createContinuousRange if this and endTarget have the same type or
+ * otherwise returns an UntypedTarget from this to endTarget.
+ */
+export abstract class CommonTarget<
```
pokey, line 17: I think I may call this `CommonBaseTarget`? Idk wdyt @auscompgeek

pokey (commented):
Yeah this looks find modulo a couple comments. Would be great to get high-level feedback from @auscompgeek if possible, but if too busy we can move forward anyway

.vscode/settings.json
auscompgeek: :shrug: looks like prettier is fine with either too? weird that this is the _only_ change to this file

packages/cursorless-engine/src/processTargets/targets/UntypedTarget.ts
auscompgeek: `CommonBase` sounds like a bit of a tautology to me 😄

.vscode/settings.json
bjaspan: I removed this change.

packages/cursorless-engine/src/processTargets/targets/UntypedTarget.ts
bjaspan: What about `CommonContinuousRangeTarget`?  Any name is fine with me. :-)

pokey:
closing in favour of https://github.com/cursorless-dev/cursorless/pull/2074

PR #2038, opened 2023-11-18
Migrate java scopes

### LINKED ISSUE #1926: Java: Support name and value in for each statements
First requires name and value to migrate from legacy scope

reference https://github.com/cursorless-dev/cursorless/pull/1919

AndreasArvidsson:
#Fixes #1926

## Checklist


queries/java.scm
```diff
@@ -0,0 +1,135 @@
...
+;;!  ------------------------------
+(for_statement
+    condition: (_) @condition
+) @_.domain
+
+;;!! switch (value) {}
+;;!          ^^^^^
+;;!  -----------------
+(switch_expression
+    condition: (_) @condition @switchStatementSubject
```
AndreasArvidsson, line 124: We seem to have a conflict here. I just preserve the existing behavior, but this should probably change.

queries/java.scm
```diff
@@ -0,0 +1,135 @@
...
+) @namedFunction @functionName.domain
+(constructor_declaration
+    (identifier) @functionName
+) @namedFunction @functionName.domain
+
+(lambda_expression) @anonymousFunction
+
+(
+    (if_statement) @ifStatement
+    (#not-parent-type? @ifStatement "if_statement")
```
pokey, line 54: This doesn't match what was there before. I'm assuming this is an improvement, but worth adding a test to capture the new behaviour if so

queries/java.scm
```diff
@@ -0,0 +1,135 @@
...
+;;!! if () {}
+;;!  ^^^^^^^^
+(
+    (if_statement
+        "if" @branch.start @condition.domain.start
+        condition: (_) @condition
+        consequence: (block) @branch.end @condition.domain.end
+    ) @dummy
+    (#not-parent-type? @dummy "if_statement")
+    (#child-range! @condition 0 -1 true true)
```
pokey, line 92: This includes any whitespace surrounding the condition inside the parens. Probably fine. Def makes sense for removal range. Prob good enough for content range

queries/java.scm
pokey: Does a test break if you drop `@condition` here? If not, I'd be tempted to drop it

pokey (commented):
Looks good! Left a few minor comments

queries/java.scm
AndreasArvidsson: done

queries/java.scm
AndreasArvidsson: Seem to work without it. But currently there is no support for condition in switch matches. This doesn't seem to work in typescript either. Is the point that condition should be the values in the match statement?

queries/java.scm
AndreasArvidsson: I think so

queries/java.scm
```diff
@@ -0,0 +1,161 @@
...
+    (#child-range! @switchStatementSubject 0 -1 true true)
+) @_.domain
+
+;;!! true ? 1 : 2
+(ternary_expression
+    condition: (_) @condition
+    consequence: (_) @branch
+) @condition.domain
+(ternary_expression
+    alternative: (_) @branch
```
AndreasArvidsson, line 160: Do we have a way to define a scope multiple places in the same pattern?

pokey: unfortunately not. The problem is that then things like `.leading`, `.start`, etc become ambiguous. I guess we could allow it if we don't see any of those things in the pattern? 🤔 cc/ @josharian

queries/java.scm
```diff
@@ -0,0 +1,311 @@
...
+;;!! new test()
+;;!  ^^^^^^^^
+;;!  ----------
+(object_creation_expression
+    (argument_list) @functionCallee.end.startOf
+) @functionCallee.start.startOf @_.domain
+
+;;!! new test().bar();
+;;!  ^^^^^^^^^^^^^^
+;;!  ----------------
```
pokey, line 264: does this actually exclude the `;`? If so, we should prob try to include that

queries/java.scm
```diff
@@ -0,0 +1,311 @@
...
+(method_invocation
+    (argument_list) @functionCallee.end.startOf
+) @functionCallee.start.startOf @_.domain
+
+;;!! super();
+;;!  ^^^^^
+;;!  -------
```
pokey, line 271: as above

pokey (commented):
Ok I didn't look super closely, as it's been awhile since my java days, we have lots of tests, and you're using it day-to-day so will notice oddnesses.  I left a few comments. In addition, the following looked weird to me for "value"

image

domain should include `;` above; might be same issue I point out in inline comment

image

these should be values

queries/java.scm
AndreasArvidsson: fixed

queries/java.scm
AndreasArvidsson: fixed

AndreasArvidsson:
@pokey Updated.

queries/java.scm
josharian: I have now forgotten all of the details but last time I tried playing games like that I ended up in a tangled mess of code. I will generally choose clear over clever/concise, personally.

AndreasArvidsson: Couldn't we number the scopes? 
```
(ternary_expression
    consequence: (_) @branch1
    alternative: (_) @branch2
)
```

If we can't use actual numbers we could use letters or roman numerals? :D

pokey: That's an interesting idea. Or `@branch.1`

AndreasArvidsson: Yeah probably even better

queries/java.scm
```diff
@@ -0,0 +1,330 @@
...
+;;!          ^
+;;!       xxxx
+;;!  ----------
+(_
+    (assignment_expression
+        left: (_) @value.removal.start.endOf
+        right: (_) @value @value.removal.end
+    ) @_.domain.start
+    ";"? @_.domain.end
+)
```
pokey, line 322: I'd argue for using leading / trailing instead of removal for all of these

pokey (commented):
Better; left one more minor comment and I think we're good to go

queries/java.scm
AndreasArvidsson: If you prefer I could do that. Personally I don't think I have every use leading and trailing for anything else than white spaces, but I'm not against it.

AndreasArvidsson: fixed

pokey: Ha yeah was kind of thinking the same thing in the back of my head 😅. Better to be consistent tho I guess 🤷‍♂️

pokey (approved):
👍👍

PR #2039, opened 2023-11-19
Remove jsx attribute from item scope

### LINKED ISSUE #2034: comma based item broken inside of html attributes
Can't target `"disabled": disabled` or `"invalid": invalid` since item now refers to an html attribute

```jsx
<div className={classNames({"disabled": disabled, "invalid": invalid})} />
```

AndreasArvidsson:
Fixes #2034

## Checklist


pokey (commented):
We should prob ask on slack whether anyone is using this. Or maybe you already did that?

AndreasArvidsson:
> We should prob ask on slack whether anyone is using this. Or maybe you already did that?
No I haven't

packages/cursorless-engine/src/languages/html.ts
pokey, line 1: 🎉

pokey (approved):
Looks good but please file an issue to track re-adding if there's not already one

AndreasArvidsson:
I'm not completely sure if we actually want that? it's one of those things that if we overload a single scope to much it gets imprecise. I can make an issue about thinking about it :D

PR #2043, opened 2023-11-20
Migrate markdown scopes

AndreasArvidsson:
## Checklist

### DIFF
--- a/queries/markdown.scm
+++ b/queries/markdown.scm
@@ -5,3 +5,15 @@
 ;;!! * hello * stuff
 ;;!  ^^^^^^^^^^^^^^^
 (list) @list
+
+;;!! # Title
+;;!    ^^^^^
+;;!! xxxxxxx
+;;!  -------
+(section
+    (atx_heading
+        (_)
+        heading_content: (_) @name
+    ) @_.removal
+    (#shrink-to-match! @name "^\\s*(?<keep>.*)$")
+) @_.domain
--- a/queries/markdown.scm
+++ b/queries/markdown.scm
@@ -1,0 +1,7 @@
+(document) @textFragment
+
+(html_block) @comment
+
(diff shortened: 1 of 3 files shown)


queries/markdown.scm
```diff
@@ -0,0 +1,19 @@
...
+;;!    ^^^^^
+;;!! xxxxxxx
+;;!  -------
+(section
+    (atx_heading
+        (_)
+        heading_content: (_) @name
+    ) @_.removal
+    (#shrink-to-match! @name "^\\s*(?<keep>.*)$")
+) @_.domain
```
pokey, line 19: Huh is it really this simple? I guess I only needed the leading sibling finder for when we were searching for a section heading of a specific level? If so I'd be tempted to just drop support for specific levels; I can't imagine anyone is using that 🤷‍♂️

pokey: Oh I guess this is technically wrong. I believe "grand name" would be incorrect, because the domain should include all subheadings

AndreasArvidsson: All the test passed. That was kind of my benchmark.

I would be happy to drop the specific section levels.

I think that is something we can worry about when we try to implement that logic?

AndreasArvidsson: Already looks like subsections are nested
```scm
(document
  (section
    (atx_heading
      (atx_h1_marker)
      heading_content: (inline)
    )
    (paragraph
      (inline)
    )
    (section
      (atx_heading
        (atx_h2_marker)
        heading_content: (inline)
      )
      (paragraph
        (inline)
      )
    )
  )
)
```

pokey: Oh amazing. I believe all that sibling logic is a holdover from the old markdown parser we used to use then. We should be able to drop most of that and do something much simpler now

PR #2044, opened 2023-11-20
Migrate html scopes

AndreasArvidsson:
## Checklist


queries/html.scm
```diff
@@ -0,0 +1,93 @@
...
+;;!       ^^
+(attribute
+    (attribute_name) @collectionKey
+) @_.domain
+
+;;!! <aaa value=2>
+;;!             ^
+(attribute
+    (attribute_name) @value.removal.start.endOf
+    (attribute_value) @value @value.removal.end.endOf
```
pokey, line 29: For consistency we should prob switch all the removals to leading / trailing where possible

AndreasArvidsson: Done

pokey (approved):
looks good!

pokey (approved):
looks good!

queries/html.scm
```diff
@@ -0,0 +1,93 @@
...
+  [
+    (start_tag)
+    (end_tag)
+  ] @xmlBothTags
+  (#allow-multiple! @xmlBothTags)
+) @_.domain
+
+(_
+  (start_tag) @xmlElement.iteration.start.endOf @xmlBothTags.iteration.start.endOf
+  (element)
```
pokey, line 91: why do we require a child element? we don't do that for jsx

queries/html.scm
```diff
@@ -0,0 +1,93 @@
...
+(_
+  [
+    (start_tag)
+    (end_tag)
+  ] @xmlBothTags
+  (#allow-multiple! @xmlBothTags)
+) @_.domain
+
+(_
+  (start_tag) @xmlElement.iteration.start.endOf @xmlBothTags.iteration.start.endOf
```
pokey, line 90: you're missing startTag and endTag

queries/html.scm
AndreasArvidsson: Just because if you are in the middle off a text only element I don't think wish used that as iteration scope. In this example if the cursor is at the pipe and you say take every element I want both of them to be selected. And I think we also have tests for that since before that broke originally.
```html
<a>hello|stuff</a>
<b>something</b>
```

queries/html.scm
AndreasArvidsson: added

PR #2048, opened 2023-11-20
Token scope prefers $ over other symbols

### LINKED ISSUE #1321: `$` should be treated as an identifier
$|;
`"clear token"` should prefer dollar since that's a valid identifier in many languages.

AndreasArvidsson:
Fixes #1321

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TokenScopeHandler.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TokenScopeHandler.ts
@@ -42,12 +42,28 @@ export class TokenScopeHandler extends NestedScopeHandler {
     } = scopeA;
     const { identifierMatcher } = getMatcher(document.languageId);
 
+    const textA = document.getText(scopeA.domain);
+
     // NB: Don't directly use `test` here because global regexes are stateful
     // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#finding_successive_matches
-    return testRegex(identifierMatcher, document.getText(scopeA.domain))
-      ? true
-      : testRegex(identifierMatcher, document.getText(scopeB.domain))
-      ? false
-      : undefined;
+    if (testRegex(identifierMatcher, textA)) {
+      return true;
+    }
+
+    const textB = document.getText(scopeB.domain);
+
+    if (testRegex(identifierMatcher, textB)) {
+      return false;
+    }
+
+    if (textA === "$") {
+      return true;
+    }
+
+    if (textB === "$") {
+      return false;
+    }
+
+    return undefined;
   }
 }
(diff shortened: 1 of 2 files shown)


packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TokenScopeHandler.ts
```diff
@@ -42,12 +42,28 @@ export class TokenScopeHandler extends NestedScopeHandler {
...
+    const textA = document.getText(scopeA.domain);
+
     // NB: Don't directly use `test` here because global regexes are stateful
     // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#finding_successive_matches
-    return testRegex(identifierMatcher, document.getText(scopeA.domain))
-      ? true
-      : testRegex(identifierMatcher, document.getText(scopeB.domain))
-      ? false
-      : undefined;
+    if (testRegex(identifierMatcher, textA)) {
```
pokey, line 49: Why not just add $ to identifierMatcher regex?

AndreasArvidsson: Because that is a larger change. We discussed this in the linked issue.

pokey: No I don't mean change the global identifier regex, just lump it in here. I'm not sure we should be preferring identifiers over $, as you're doing here, and it's simpler if we don't

AndreasArvidsson: I do think we should prefer alpha to dollar

pokey, line 49: I'm also a bit surprised by our logic here. Why do we immediately prefer `a` if it's an identifier; what if `b` is also an identifier? I would expect us to return `undefined` in that case 🤔. I guess you're just mirroring the same logic here but seems wrong to me

pokey: Ohh right yeah I guess we never hit that with two identifiers

False and undefined are different I believe. False means prefer b; undefined means no preference so let modifier decide

I would argue we should be correct here now that we actually can have them adjacent

AndreasArvidsson: I'm just saying that today we don't actually utilize that. We could simplify the function signature and still have the same behavior yesterday.

AndreasArvidsson: But I have now updated the code to reflect the actual logic

PR #2050, opened 2023-11-20
Added glyph scope

### LINKED ISSUE #54: Support jump to sub character by letter name
eg:
take glyph air in fine -> Select first instance of character "a" in token "fine"

_(@pokey edited to use "glyph" instead of "char" to reserve "char" as a modifier that converts a target to refer to its individual characters rather than by token)_

AndreasArvidsson:
`chuck glyph dollar red air`

Fixes #54

## Checklist


cursorless-talon/src/modifiers/glyph.py
```diff
@@ -0,0 +1,16 @@
...
+mod.list(
+    "cursorless_glyph_modifier",
+    desc="Cursorless glyph,  modifier",
+)
+
+
+@mod.capture(rule="{user.cursorless_glyph_modifier} <user.any_alphanumeric_key>")
+def cursorless_glyph_modifier(m) -> dict[str, str]:
+    return {
+        "type": "glyph",
```
pokey, line 14: I think `"glyph"` is a nice spoken form, but I wonder whether it is the clearest identifier. I don't feel strongly one way or another. Could use something like `instanceOfChar`, `specificChar`. Or stick with `glyph`. @josharian any opinions?

cursorless-talon/src/modifiers/glyph.py
```diff
@@ -0,0 +1,16 @@
...
+        "glyph": m.any_alphanumeric_key,
```
pokey, line 15: as above, maybe should be `character`, like we use in decorated mark?

packages/cursorless-engine/src/processTargets/modifiers/GlyphStage.ts
```diff
@@ -0,0 +1,53 @@
+import { GlyphModifier, Range } from "@cursorless/common";
+import { Target } from "../../typings/target.types";
+import { ModifierStageFactory } from "../ModifierStageFactory";
+import { ModifierStage } from "../PipelineStages.types";
+import { PlainTarget } from "../targets";
+
+export class GlyphStage implements ModifierStage {
```
pokey, line 7: wait...why is this a modifier? I can't imagine why it wouldn't be a scope 🤔. Then you could say "every dollar", "next dollar", "first dollar", etc, all of which are very useful

pokey (commented):
I reviewed until I realised this was a modifier, and then stopped, because that's a pretty fatal flaw imo

cursorless-talon/src/modifiers/glyph.py
josharian: `glyph` seems fine to me. Why have multiple things to learn when you could have just one?

pokey: One thing @AndreasArvidsson and I discussed is that this exact scope could be used for arbitrary text, not just single characters. Might be an argument for a different name. Might not. But could be something like `exactMatch`. 

Although the glyph scope should be case insensitive. But I guess there could be a flag on it eventually and then it could be for both

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/RegexScopeHandler.ts
```diff
@@ -61,3 +67,17 @@ export class CustomRegexScopeHandler extends RegexStageBase {
     super(scopeHandlerFactory, scopeType, languageId);
   }
 }
+
+export class GlyphScopeHandler extends RegexStageBase {
+  get regex() {
+    return new RegExp(escapeRegExp(this.scopeType.character), "gu");
```
pokey, line 73: Prob want case-insensitive, at least by default?

cursorless-talon/src/modifiers/glyph.py
josharian: literal? match? constantString? matchString?

AndreasArvidsson: After a long discussion I reached consensus with myself and decided to call it `literal` :D

pokey:
Not sure glyph and literal are the same. Arguably glyph should go through same grapheme normalisation as hat chars, whereas literal wouldn't necessarily

AndreasArvidsson:
Then do you argue that I should revert this last commit back to only matching a single character? Because the implementation python side of this is the same capture that we use for decorated marks.

pokey:
I'm just arguing maybe we keep the same naming convention you had before (glyph scope / character as field). I don't feel strongly, just a thought. Maybe let's discuss if you disagree / aren't sure?

AndreasArvidsson:
I'm happy with reverting

AndreasArvidsson:
Reverted to the old naming scheme

cursorless-talon/src/modifiers/glyph.py
AndreasArvidsson: Reverted to glyph

cursorless-talon/src/cheatsheet/sections/modifiers.py
pokey, line 1: Cheatsheet seems to still have it as a modifier

packages/cursorless-engine/src/spokenForms/CustomSpokenForms.ts
```diff
@@ -175,7 +168,7 @@ function updateEntriesForType<T extends SpokenFormType>(
             defaultSpokenForms,
             spokenForms: customSpokenForms,
             requiresTalonUpdate: false,
-            isCustom: isEqual(defaultSpokenForms, customSpokenForms),
+            isCustom: !isEqual(defaultSpokenForms, customSpokenForms),
```
pokey, line 171: whoops 😅. we don't have any use case for this yet, but should prob add a test at some point

pokey (commented):
Ok I added a test for custom spoken form generation. I left one more comment. Once that's addressed, and if you're happy with my changes, let's ship it!

packages/cursorless-engine/src/spokenForms/CustomSpokenForms.ts
AndreasArvidsson: Really proves that if something isn't tested it can easily be correct :)

pokey (approved):
Looks like `defaults.json` had rotted a bit. I fixed the updater and ran it. Also made some slight tweaks to the cheatsheet entry for glyph. If you're happy with my changes (bc474c956eb3bd00eb674597d03588cab7e0e534) let's ship it!

PR #2051, opened 2023-11-21
keyboard: Use parser for key sequences

pokey:
## Checklist

- [-] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet) (keeping this somewhat under wraps while we experiment
- [-] Refactor delete mapping to just issue token for delete action
- [x] Extract partial parameters
- [-] Add railroad to docs?


pokey:
ok @kn0rk here's the parser idea in action. Seems to have worked out well I think. The most interesting file to have a look at is [the grammar](https://github.com/cursorless-dev/cursorless/pull/2051/files#diff-fac638c2ca56fc65b2115d487678f5e4dcd0e5444ff0d19c469370c1711e2598)

packages/cursorless-vscode/src/keyboard/grammar/grammar.ne
pokey, line 1: This is kind of the punchline of this PR: the grammar that defines our modal keyboard interface

packages/cursorless-vscode/src/keyboard/grammar/generated/grammar.ts
pokey, line 1: this is the generated grammar. note how it's fairly small, and we keep it in a directory called `generated`

pokey, line 1: this is the generated grammar. note how it's fairly small, and we keep it in a directory called `generated`

packages/cursorless-vscode/src/keyboard/grammar/grammar.ne
pokey, line 1: This is kind of the punchline of this PR: the grammar that defines our modal keyboard interface

packages/cursorless-vscode/src/keyboard/grammar/generated/grammar.ts
```diff
@@ -0,0 +1,92 @@
...
+
+interface NearleyLexer {
+  reset: (chunk: any, info: any) => void;
+  next: () => NearleyToken | undefined;
+  save: () => any;
+  formatError: (token: any, message: string) => string;
+  has: (tokenType: any) => boolean;
+};
+
+interface NearleyRule {
```
josharian, line 33: So nearly PokeyRule.

patch it to make it so! a lovely easter egg...

(not sure if I'm joking)

packages/cursorless-vscode/src/keyboard/grammar/getAcceptableTokenTypes.ts
```diff
@@ -0,0 +1,83 @@
...
+ * so that we can display the command name in the list. We also include a partial
+ * argument for the command that wants the token type, which we could use to
+ * provide even more information to the user.
+ *
+ * @param parser The parser to get the acceptable token types of
+ * @returns A list of acceptable token types, along with information about which
+ * top-level rules want them
+ */
+export function getAcceptableTokenTypes(parser: nearley.Parser) {
+  return uniqWith(
```
josharian, line 19: sure you don't want to use our own uniqWithHash that is not quadratic? :)

packages/cursorless-vscode/src/keyboard/grammar/grammarHelpers.ts
```diff
@@ -0,0 +1,70 @@
...
+    };
+  }
+  ret.metadata = { type, argNames };
+  return ret;
+}
+
+/**
+ * Creates a postprocess function for a lower-level capture in our keyboard
+ * grammar. The output will be an object with the keys of {@link argNames}
+ * mapped to the values at the same positions in the parser rule's output.
```
josharian, line 59: an example input/output pair would help a lot here

packages/cursorless-vscode/src/keyboard/sectionKeyMapExtractors.ts
```diff
@@ -0,0 +1,95 @@
...
+    : vscode.workspace
+        .getConfiguration("cursorless.experimental.keyboard.modal.keybindings")
+        .get<KeyMap<T>>(sectionName) ?? {};
+  const defaultKeyMap: KeyMap<T> = defaultMap[sectionName] ?? {};
+  return merge({}, defaultKeyMap, userOverrides);
+}
+
+/**
+ * Returns a keymap with a subset of entries from a given config section. If
+ * `values` is not provided, all entries will be returned.
```
josharian, line 29: consider s/values/only/ or s/values/limit/

packages/cursorless-vscode/src/keyboard/sectionKeyMapExtractors.ts
```diff
@@ -0,0 +1,95 @@
...
+ *   getSingularSectionEntry(getSectionKeyMapRaw("misc"), "direction", ["forward", "backward"]),
+ *   {
+ *     "f": { type: "direction", value: "forward" },
+ *     "b": { type: "direction", value: "backward" },
+ *   },
+ * );
+ * ```
+ *
+ * @param keyMap The keymap for the given config section
+ * @param value The value to get the entries for
```
josharian, line 44: s/value/values/g?

packages/cursorless-vscode/src/keyboard/sectionKeyMapExtractors.ts
```diff
@@ -0,0 +1,95 @@
...
+export function getSectionKeyMap<
+  K extends keyof SectionTypes,
+  T extends TokenType,
+  V extends SectionTypes[K] & TokenTypeValueMap[T] = SectionTypes[K] &
+    TokenTypeValueMap[T],
+>(sectionName: K, type: T, values?: V[]): KeyMap<{ type: T; value: V }> {
+  return mapValues(
+    pickBy(
+      getSectionKeyMapRaw(sectionName),
+      (v): v is V => values?.includes(v as V) ?? true,
```
josharian, line 56: this is cute but it took me at least four reads to figure out what it does

packages/cursorless-vscode/src/keyboard/sectionKeyMapExtractors.ts
```diff
@@ -0,0 +1,95 @@
...
+ *     "r": { type: "makeRange" },
+ *   },
+ * );
+ * ```
+ *
+ * @param keyMap The keymap for the given config section
+ * @param value The value to get the entries for
+ * @returns A keymap with entries only for the given value
+ */
+export function getSingularSectionEntry<
```
josharian, line 85: suggest s/Singular/Single/

josharian (commented):
Done!

Feel free to simply ignore questions and suggestions that don't make sense. I'm sure that at least some of my thoughts are due to typescript confusion and/or naming philosophy differences with you. :)

I'd really love for this to include a sample VSCode config, to make it easy to play with. I can see that there are a bunch of things yet to be added to fill this out, so maybe it makes sense to hold off on the config until it is fully fleshed out? Any which way this seems like an excellent direction and solid technical foundation.

package.json
```diff
@@ -44,7 +44,9 @@
   },
   "pnpm": {
     "patchedDependencies": {
-      "@docusaurus/theme-search-algolia@3.0.0": "patches/@docusaurus__theme-search-algolia@2.3.1.patch"
+      "@docusaurus/theme-search-algolia@3.0.0": "patches/@docusaurus__theme-search-algolia@2.3.1.patch",
+      "@types/nearley@2.11.5": "patches/@types__nearley@2.11.5.patch",
```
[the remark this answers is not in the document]
pokey: Unfortunately, the project is dead, so probably not. No activity in 3 years 😕

packages/cursorless-vscode/package.json
```diff
@@ -1104,6 +1104,11 @@
     "preprocess-svg-hats": "my-ts-node src/scripts/preprocessSvgHats.ts",
     "hat-adjustment-add": "my-ts-node src/scripts/hatAdjustments/add.ts",
     "hat-adjustment-average": "my-ts-node src/scripts/hatAdjustments/average.ts",
+    "generate-grammar:base": "nearleyc src/keyboard/grammar/grammar.ne",
+    "ensure-grammar-up-to-date": "pnpm -s generate-grammar:base | diff -u src/keyboard/grammar/generated/grammar.ts -",
+    "generate-grammar": "pnpm generate-grammar:base -o src/keyboard/grammar/generated/grammar.ts",
+    "generate-railroad": "nearley-railroad src/keyboard/grammar/grammar.ne -o out/railroad.html",
```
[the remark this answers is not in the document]
pokey: Agreed, tho maybe we should do it in contributor docs for now? Trying to keep the more advanced keyboard stuff under wraps for now to give us more flexibility to experiment

packages/cursorless-vscode/src/keyboard/sectionKeyMapExtractors.ts
pokey: Note that there could be multiple entries, if they have multiple keys bound to the same thing. Assuming you'd still prefer your proposed name?

packages/cursorless-vscode/src/keyboard/KeyboardCommandsModal.ts
```diff
@@ -63,106 +54,41 @@ export default class KeyboardCommandsModal {
...
-      this.mergedKeymap[key] = entry;
-    }
+  private computeLayer() {
+    const acceptableTokenTypes = getAcceptableTokenTypes(this.parser);
+    // FIXME: Here's where we'd update sidebar
+    const sections = pick(
+      this.sections,
+      acceptableTokenTypes.map(({ type }) => type),
+    );
+    this.currentLayer = new KeyboardCommandsModalLayer(
```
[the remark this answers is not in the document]
pokey: Layers never get "popped". What do you mean by "clean up"? A modal layer has no resources to clean up. JS has a garbage collector, so just dropping the reference should free up the memory, if that's what you mean

packages/cursorless-vscode/src/keyboard/KeyboardCommandHandler.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  performSimpleActionOnTarget({
+    actionName,
+  }: {
+    actionName: SimpleKeyboardActionType;
+  }) {
+    this.targeted.performActionOnTarget(actionName);
+  }
+
+  targetScopeType(arg: { scopeType: ScopeType }) {
```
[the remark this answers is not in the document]
pokey: Hmm. Not sure bout that one. Maybe `modifyTargetContainingScope`?

package.json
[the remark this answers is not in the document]
josharian: 😿

packages/cursorless-vscode/package.json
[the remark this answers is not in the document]
josharian: Sure

packages/cursorless-vscode/src/keyboard/KeyboardCommandHandler.ts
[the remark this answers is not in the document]
josharian: works for me!

packages/cursorless-vscode/src/keyboard/KeyboardCommandsModal.ts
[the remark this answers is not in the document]
josharian: oh, i somehow believed there was a stack of layers, rather than the current layer getting systematically replaced. never mind. :)

packages/cursorless-vscode/src/keyboard/getTokenTypeKeyMaps.ts
```diff
@@ -0,0 +1,69 @@
...
+ *   },
+ *   ...
+ * }
+ * ```
+ * @returns A map from token type names to a keymap for that token type.
+ */
+export function getTokenTypeKeyMaps(): TokenTypeKeyMapMap {
+  return {
+    simpleScopeTypeType: getSectionKeyMap("scopes", "simpleScopeTypeType"),
+    color: getSectionKeyMap("colors", "color"),
```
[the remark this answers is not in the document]
josharian: i'm generally a fan of all singular, including database names. (plus it makes editing easier.) but i don't feel strongly enough to strongly advocate breaking backwards compat.

packages/cursorless-vscode/src/keyboard/sectionKeyMapExtractors.ts
josharian: yeah. but like all the naming nits, i don't feel strongly.

packages/cursorless-vscode/src/keyboard/KeyboardCommandTypeHelpers.ts
```diff
@@ -0,0 +1,40 @@
+import { KeyboardCommandHandler } from "./KeyboardCommandHandler";
+
+/**
+ * Maps from the name of a method in KeyboardCommandHandler to the type of its
+ * argument.
+ */
+export type KeyboardCommandArgTypes = {
```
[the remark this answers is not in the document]
pokey: Not particularly. This kind of TS type mapping stuff is never pretty, but the proof is usually in the pudding. Happy to walk you through it if you're curious tho

packages/cursorless-vscode/src/keyboard/TokenTypes.ts
```diff
@@ -0,0 +1,72 @@
...
+ * that section.
+ */
+export interface SectionTypes {
+  actions: KeyboardActionType;
+  colors: HatColor;
+  misc: MiscValue;
+  scopes: SimpleScopeTypeType;
+  shapes: HatShape;
+  vscodeCommands: ModalVscodeCommandDescriptor;
+  modifier: ModifierKeyMapValue;
```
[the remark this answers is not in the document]
pokey: It should actually be just `Modifier`, but that name is a bit too generic. Maybe `ModifierType`?

packages/cursorless-vscode/src/keyboard/grammar/lexer.ts
```diff
@@ -0,0 +1,57 @@
...
+ * to use when it is checking token types, and exposes a {@link transform}
+ * method that the parser will then use to transform the token into the actual
+ * value that will be used when constructing rule outputs.
+ */
+class MyLexer implements Lexer {
+  buffer: any[] = [];
+  bufferIndex = 0;
+  index = 0;
+
+  reset(data: Token[], state?: LexerState) {
```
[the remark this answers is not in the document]
pokey: Agreed. I think the idea is that each time you get a new chunk of input, this function is used to feed the lexer's input buffer

packages/cursorless-vscode/src/keyboard/sectionKeyMapExtractors.ts
pokey: Ok here's a draft of removing this function entirely: 41372116a0f74e1d236512249941030093d40e98. Slightly more verbose, but less special stuff to learn. WDYT?

packages/cursorless-vscode/src/keyboard/getTokenTypeKeyMaps.ts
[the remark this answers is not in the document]
pokey: Switched to singular, and added a deprecation warning

pokey:
ok @josharian I believe I've addressed / responded to all your comments. [Here's](https://github.com/cursorless-dev/cursorless/pull/2051/files/a44b2f90c907a3297b19e7d01b681373a9436c04..845f3d891a328e5439f0e6880402ccbea7811c6a) a diff from when you left your comments, in case helpful.

packages/cursorless-vscode/src/keyboard/TokenTypes.ts
```diff
@@ -0,0 +1,72 @@
...
+} from "./KeyboardActionType";
+
+/**
+ * Maps from modal keyboard config section name to the type of entry expected in
+ * that section.
+ */
+export interface SectionTypes {
+  actions: KeyboardActionType;
+  colors: HatColor;
+  misc: MiscValue;
```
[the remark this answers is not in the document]
josharian: the awkward state is where you have a lot of structure and also a misc. it is hard for people to predict what goes in misc, and it is hard to refactor misc because backwards compat. I'd be pretty comfortable with dropping it all in one giant blob, or in having lots of structure; the concern is mainly around the mix.

packages/cursorless-vscode/package.json
```diff
@@ -961,7 +961,7 @@
             ]
           }
         },
-        "cursorless.experimental.keyboard.modal.keybindings.shapes": {
+        "cursorless.experimental.keyboard.modal.keybindings.shape": {
```
josharian, line 964: WDYT about changing:

> cursorless.experimental.keyboard.modal.keybindings.*

to

> cursorless.keyboard.bind.*

I understand every single word and why it is there in the original, but it feels like we could get away with something much more concise.

packages/cursorless-vscode/src/keyboard/TokenTypes.ts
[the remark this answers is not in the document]
josharian: that seems like an improvement

packages/cursorless-vscode/src/keyboard/grammar/WorkQueue.ts
```diff
@@ -11,20 +12,15 @@ export class WorkQueue<T> {
...
-    // while there are still items that haven't been yielded
-    while ([...this.items.values()].some((value) => !value)) {
-      // find any item that hasn't been yielded yet
-      const currentItem = [...this.items.entries()].find(
-        ([_, visited]) => !visited,
-      )![0];
-      this.items.set(currentItem, true);
-      yield currentItem;
+    while (this.queue.length > 0) {
+      yield this.queue.shift()!;
```
josharian, line 23: do you want to remove item from items? (that is, do you want to be able to re-add an item for processing a second time after it has already been processed a first time?)

packages/cursorless-vscode/src/keyboard/grammar/grammar.ne
```diff
@@ -0,0 +1,54 @@
+@preprocessor typescript
+@{%
+import { capture, command } from "../grammarHelpers"
+import { lexer } from "../lexer";
+%}
+@lexer lexer
+
+# ===================== Top-level commands ===================
+# "air"
+main -> decoratedMark {% command("targetDecoratedMarkReplace", ["decoratedMark"]) %}
```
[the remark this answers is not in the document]
josharian: yes that seems a lot clearer

packages/cursorless-vscode/src/keyboard/grammar/grammar.ne
```diff
@@ -0,0 +1,54 @@
...
+
+# "past air"
+main -> %makeRange decoratedMark {%
```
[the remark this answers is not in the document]
josharian: yes let's discuss over voice. It's not obvious to me what to do here. but let's also not let that uncertainty block forward progress.

packages/cursorless-vscode/src/keyboard/grammar/grammar.ne
```diff
@@ -0,0 +1,54 @@
...
+%}
+
+# "chuck"
+main -> %simpleAction {% command("performSimpleActionOnTarget", ["actionName"]) %}
+
+# Custom vscode command
+main -> %vscodeCommand {% command("vscodeCommand", ["command"]) %}
+
+# ========================== Captures =========================
+scopeType -> %simpleScopeTypeType {% capture("type") %}
```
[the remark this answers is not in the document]
josharian: ahhh. makes sense. another alternative would be a synonym for type, like `{kind: "surroundingPair`, or `class`, or `category`, or the like. anyway, not a problem for today.

packages/cursorless-vscode/src/keyboard/grammar/grammar.ne
```diff
@@ -0,0 +1,54 @@
...
+%}
+decoratedMark -> %combineColorAndShape %shape %color {%
+  capture(null, "shape", "color")
+%}
+
+# Contains a direction and a number for use with relative and ordinal
+offset -> %direction:? number {% capture("direction", "number") %}
+offset -> number:? %direction {% capture("number", "direction") %}
+
+number -> %digit:+ {%
```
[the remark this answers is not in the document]
josharian: good call

packages/cursorless-vscode/src/keyboard/grammar/grammarHelpers.ts
josharian: this is a massive improvement. Thank you for doing at.

packages/cursorless-vscode/package.json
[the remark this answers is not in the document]
josharian: works for me

packages/cursorless-vscode/src/keyboard/grammar/grammar.ne
```diff
@@ -0,0 +1,54 @@
...
+%}
+@lexer lexer
+
+# ===================== Top-level commands ===================
+# "air"
+main -> decoratedMark {% command("targetDecoratedMarkReplace", ["decoratedMark"]) %}
+
+# "past air"
+main -> %makeRange decoratedMark {%
+  command("targetDecoratedMarkExtend", [null, "decoratedMark"])
```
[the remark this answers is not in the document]
josharian: Ah, i see. Yes, makes sense.

packages/cursorless-vscode/src/keyboard/KeyboardCommandTypeHelpers.ts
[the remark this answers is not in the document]
josharian: I am intellectually interested, but we can leave that for another day.

packages/cursorless-vscode/src/keyboard/KeyboardCommandsModalLayer.ts
```diff
@@ -0,0 +1,118 @@
...
+      if (!this.isPrefixOfKey(sequence)) {
+        const errorMessage = `Unknown key sequence "${sequence}"`;
+        vscode.window.showErrorMessage(errorMessage);
+        throw Error(errorMessage);
+      }
+
+      const nextKey = await this.keyboardHandler.awaitSingleKeypress({
+        cursorStyle: vscode.TextEditorCursorStyle.Underline,
+        whenClauseContext: "cursorless.keyboard.targeted.awaitingKeys",
+        statusBarText: "Finish sequence...",
```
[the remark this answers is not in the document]
josharian: Thanks!

packages/cursorless-vscode/src/keyboard/grammar/WorkQueue.ts
```diff
@@ -0,0 +1,30 @@
...
+    for (const item of items) {
+      if (!this.items.has(item)) {
+        this.items.set(item, false);
+      }
+    }
+  }
+
+  *[Symbol.iterator]() {
+    // while there are still items that haven't been yielded
+    while ([...this.items.values()].some((value) => !value)) {
```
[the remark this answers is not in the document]
josharian: much better

packages/cursorless-vscode/src/keyboard/sectionKeyMapExtractors.ts
josharian: I like it

packages/cursorless-vscode/src/keyboard/buildSuffixTrie.ts
```diff
@@ -0,0 +1,115 @@
...
+};
+
+/**
+ * Build a trie containing all the keymaps, as well as non-conflicting suffixes
+ * of each key. Also returns a list of all the conflicting keymaps. If any key is
+ * a prefix of any other key, we consider them to be conflicting.
+ * @param keyMaps The keymaps to build the trie from
+ * @returns The trie, and a list of conflicting keymaps
+ */
+export function buildSuffixTrie<T>(entries: [string, T][]): BuildTrieReturn<T> {
```
josharian, line 27: this is so much clearer. thanks for taking the time to do it and document it nicely. and test it so well. I have WAY more confidence now that this is correct and performant.

packages/cursorless-vscode/src/keyboard/grammar/WorkQueue.ts
josharian: To be fair, there are situations in which you want to re-use the work queue, and the structure of the problem guarantees that it is safe to remove from items. But fair enough. :P

packages/cursorless-vscode/src/keyboard/TokenTypes.ts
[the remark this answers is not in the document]
pokey: One drawback with giant blob for this specific case is that then they have no way to have two identifiers with the same key. If they can never both simultaneously be valid tokens, then it's ok for them to do that. Tho I guess they could accomplish the same thing by giving them each a different prefix and letting suffix magic shorten them. Idk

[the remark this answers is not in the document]
pokey: > can't decide whether "giving them each a different prefix and letting suffix magic shorten them" is elegant or horrible.

I feel exactly the same way 😅

> i don't feel strongly. maybe let's leave as-is and then revisit as needed?

Yeah this is why I'm leaving all the new fancy stuff in this PR out of the user docs, so we can iterate quickly without worrying about backwards compatibility

packages/cursorless-vscode/src/keyboard/grammar/WorkQueue.ts
pokey: No I mean the only reason I created this class is because I was infinite looping due to recursion in the grammar causing me to visit the same node infinitely many times

Maybe I should rename this class to make it clear that's what it's for. `UniqueWorkQueue`? `IdempotentWorkQueue`?

josharian: nah, it's fine. stet

pokey: Unless you object, I do think I will rename it. The uniqueness thing is critical, and would be annoying to debug if someone accidentally removes that property

josharian: go for it

packages/cursorless-vscode/src/keyboard/KeyboardCommandsModalLayer.ts
```diff
@@ -0,0 +1,87 @@
...
+        vscode.window.showErrorMessage(errorMessage);
+        throw Error(errorMessage);
+      }
+    }
+
+    let sequence = text;
+
+    // We handle multi-key sequences by repeatedly awaiting a single keypress
+    // until they've pressed something in the map.
+    while (values.length !== 1) {
```
pokey, line 62: Note that checking this way means that not only can the user press any unique suffix, they can press any unique prefix. So you can either structure your maps to use leading differentiator or trailing differentiator and get the same benefit, eg:

leading:

```json
{
  "sf": "namedFunction",
  "sl": "line",
  "st": "token",
}
```

or trailing:

```json
{
  "fs": "namedFunction",
  "ls": "line",
  "ts": "token",
}
```

In either case, if the only thing the current layer is expecting is a scope, then these will shorten to `f`, `l`, and `t` in either case

josharian: I think(?) that’s good.

The only potential downside I see is that it is hard to mentally model, so minor changes in your bindings can impact other distant bindings.

But let’s just try it out and see. If that proves to be a problem, we have options: Can change approach, or build tooling to generate all possible keymaps, which can be inspected and diffed.

pokey: Yeah, tho I think the suffix behaviour has the exact same problem, no?

pokey: Also, generating all possible keymaps isn't super easy, because it requires enumerating all possible paths through the parser, which I believe could be a challenge as Earley parsers support nondeterministic CFGs

packages/cursorless-vscode/package.json
pokey: I know we're leaning away from the "experimental" prefix, but it makes me feel a bit more comfortable with things like the gruff deprecation from plural to singular. If it weren't tagged experimental, I'd prob invest in auto-migration. Maybe we switch to that once we have a subset we're more confident in?

packages/cursorless-vscode/src/keyboard/KeyboardCommandsModalLayer.ts
josharian: /me waves hands and adopts Yiddish accent

enough worrying already, you’ll worry your mother to her grave. too many confused users? You should be so lucky.

packages/cursorless-vscode/package.json
josharian: Works for me.

packages/cursorless-vscode/src/keyboard/grammar/grammar.ne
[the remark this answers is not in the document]
pokey: Tracked in https://github.com/cursorless-dev/cursorless/issues/1002; tagged as https://github.com/cursorless-dev/cursorless/labels/to%20discuss

packages/cursorless-vscode/src/keyboard/grammar/grammar.ne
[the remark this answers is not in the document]
pokey: I'm going to do that in a fast follow where I create proper `mark` and `modifier` captures. Make sense?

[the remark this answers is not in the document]
josharian: Works for me.

AndreasArvidsson (approved):

Issue #2010, opened 2023-11-08
Language support tables


pokey:
# The problem

As a contributor, it is hard to know exactly what needs to be implemented for a new language, or what is missing from an existing language. As a user, it is hard to know what our support level is for an existing language.

I really like the support tables in https://github.com/cursorless-dev/cursorless/pull/1911 and https://github.com/cursorless-dev/cursorless/pull/1962. We should formalize these somehow with the following goals:

- make it easier for people to contribute languages
- make it easier for contributors to see what's missing for a given language
- make it harder to forget to write tests
- make it easier for users to know what a given language supports

# The solution

There are multiple levels of this solution, in increasing levels of investment

## 1. Add support tables to our [new language PR template](https://github.com/cursorless-dev/cursorless/blob/main/.github/PULL_REQUEST_TEMPLATE/new_programming_language.md)

This is the lowest overhead, and has the benefit that it's easy for the contributor and core maintainers to update as a PR progresses. 

The drawback is that it's not helpful for users, and doesn't help as much after the initial PR. It's also not formalized in any way

## 2. Manually add support tables to per-language pages (https://github.com/cursorless-dev/cursorless/issues/1642)

This solution is probably the highest overhead for new language contributors, but doesn't require much investment on our part, and is helpful both for users and for contributors seeing where a language is currently

It also isn't automatically enforced in any way, so could get stale / be wrong just like 1) above

We'd also want to add a checkbox to the PR template saying we've updated this page. Probably just the new-language template but could be the general template

## 3. Add structured representation of language scope support

We could add types that represent different aspects of scope support. Something like the following

```ts
const scopeSupportFacets = [
  "list",
  "list.interior",
  "map",
  "map.interior",
  "collectionKey",
  "namedFunction",
  "namedFunction.interior",
  "functionName",
  "anonymousFunction",
  "anonymousFunction.interior",
  "name",
  "value.assignment",
  "value.assignment.removal",
  "value.return",
  "value.return.removal",
  "value.collectionItem",
  "value.collectionItem.removal",
  "statement",
  "ifStatement",
  "condition.if",
  "condition.while",
  "condition.doWhile",
  "condition.for",
  "condition.ternary",
  "branch",
  "comment.line",
  "comment.block",
  "string.singleLine",
  "string.multiLine",
  "textFragment",
  "functionCall",
  "functionCallee",
  "argumentOrParameter.argument",
  "argumentOrParameter.argument.removal",
[...]

PR #2053, opened 2023-11-22
Scope tests

AndreasArvidsson:
Started implementing a new scope handler test representation

```
[Content]
const name = "Hello world";
              ^^^^^^^^^^^^^
```

Originally I was thinking of doing `^` for each character. That worked fine until @pokey asked the question about empty ranges which we actually has in a few places. So I switch to a multiline representation of `[---]` where the empty range would be `[]`. 

```
[Content]
 function myFunk() {
[-------------------
 
 }
 -]
```

I do wonder if we want to start multiline ranges above the source code and end it below. Makes it so the code is actually inside the range endings.

```
[Content]
[-------------------
 function myFunk() {
 
 }
 -]
```

Relevant issues
#1524
#2010

## Checklist


pokey:
> So I switch to a representation of `[---]` where the empty range would be `[]`.

As discussed in original issue, we'd like to remove the brackets when it's a single-line range. Why not just use the originally suggested format, modulo the notion that we do one range type per section?

> I do wonder if we want to start multiline ranges above the source code and end it below. Makes it so the code is actually inside the range endings.

I think that makes sense if there are only two lines, but what happens if there are more lines? Do the middle ones go above or below? I think I'd lean toward consistency

AndreasArvidsson:
> > So I switch to a representation of `[---]` where the empty range would be `[]`.
> 
> As discussed in original issue, we'd like to remove the brackets when it's a single-line range. Why not just use the originally suggested format, modulo the notion that we do one range type per section?

It's just a question about how we would represent an empty range?

> 
> > I do wonder if we want to start multiline ranges above the source code and end it below. Makes it so the code is actually inside the range endings.
> 
> I think that makes sense if there are only two lines, but what happens if there are more lines? Do the middle ones go above or below? I think I'd lean toward consistency

the lines outside of the range would not be affected by this.

pokey:
> It's just a question about how we would represent an empty range?

By "original format" I mean that specified in the Examples section of the description of https://github.com/cursorless-dev/cursorless/issues/1524, just modified so that 

- we have unmodified document at the top before `---`
- we reproduce document once per range type per target
- instead of varying the annotation style by range type, we always use the `{` / `^` style (but note that we drop the `{` for single-line targets, as per initial spec, in contrast to your suggestion here). I much prefer `^` to `-`, but I guess if we adopt your other suggestion then the start arrows are pointing the wrong direction 😅

> the lines outside of the range would not be affected by this.

no I don't mean lines outside range, I mean lines inside range. But I guess we can just leave out the annotations there because we know the range is contiguous? eg

```
[content]
[-----------
 if (true) {
     const foo = "bar";
 }
 -]
```

I think it's an interesting idea, tho I wonder if we drop a bunch of the `-`'s in this case because they're a bit misleading, because then there's no dashes above or below the other internal lines. Maybe we just drop the `-`s altogether and just use `[` / `]` (though I think I might prefer curlies)

AndreasArvidsson:
```
[content]
[-----------
 if (true) {
     const foo = "bar";
 }
 -]
```
That was what I was thinking. We can of course extend the top line all the way to the right if we think that is a better indicator?

AndreasArvidsson:
@pokey I have updated the fixture files. `[` vs `{` is something we can discuss further. Does the format look okay otherwise?

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/function.scope
```diff
@@ -0,0 +1,25 @@
...
+ 
+ }
+ -]
+
+[Domain]
+[-------------------
+ function myFunk() {
+ 
+ }
+ -]
```
pokey, line 25: as with removal, I wonder if we just leave out domain if it's equal to content

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
pokey, line 1: prob want this one in plaintext, no?

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
```diff
@@ -0,0 +1,117 @@
+
+aaa
+
+---
+
+[#1 Content]
+ 
+[]
```
pokey, line 8: huh this becomes slightly confusing because otherwise when we see a `[` it's above the first line it's annotating 🤔

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
```diff
@@ -0,0 +1,117 @@
...
+ aaa
+ 
+
+[#1 Removal]
+[
+ 
+ aaa
+ ]
```
pokey, line 16: this is nice, tho things def do get a bit confusing with empty lines. I do wonder if we want to move to using a prefix for content lines so it's crystal clear which lines are content. We could even use a prefix and a suffix, so then empty lines are very clear, because they'd be eg `><`

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
```diff
@@ -0,0 +1,117 @@
+
+aaa
+
+---
```
pokey, line 4: how do we handle a file with no trailing newline? I know we don't allow them in our source code, but we should be able to test our support for them in case users use them

pokey (commented):
Just had a quick look through a few of the fixture files, and left a few comments

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/assignment.name.scope
```diff
@@ -0,0 +1,30 @@
...
+ const name = "Hello world";
+ ^^^^^^^^^^^^^
+
+[Leading delimiter: Content]
+ const name = "Hello world";
+      ^
+
+[Leading delimiter: Removal]
+ const name = "Hello world";
+      ^
```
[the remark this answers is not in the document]
AndreasArvidsson: We could do that. The reason for keeping it is that we will have smaller diffs in the future if one of them change, but that might be an unnecessary optimization.

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/assignment.name.scope
```diff
@@ -0,0 +1,30 @@
...
+ const name = "Hello world";
+       ^^^^
+
+[Removal]
+ const name = "Hello world";
+ ^^^^^^^^^^^^^
+
+[Leading delimiter: Content]
+ const name = "Hello world";
+      ^
```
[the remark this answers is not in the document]
AndreasArvidsson: True

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
AndreasArvidsson: True

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
AndreasArvidsson: We could use pipes along the left and right hand side to create a box? Kinda like mark down tables?

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
AndreasArvidsson: I have disabled pre commit on these files.

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
AndreasArvidsson: Something like this? 
```
|[    |
|     |
| aaa |
| ]   |
```

Or without the padding. Note that the starting bracket is now part of the wall.

```
[   |
|   |
|aaa|
|]  |
```

pokey: I like the pipes, but

- they should hug the content on the right rather than lining up, otherwise they're not giving us any information
- there should be no pipe on the annotation lines, just content lines

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
pokey: yes but that only impacts the end of the file. I mean:

```
a
---
...
```

vs

```
a

---
...
```

I would think in the first case, the content is `a\n`, in the second the content is `a\n\n`, but then how do we represent just `a` (ie no trailing newline)?

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
AndreasArvidsson: Got it. Like this?

```
[
||
|aaa|
 ]
```

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
AndreasArvidsson: This is already `a`
```
a
---
```

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/assignment.name.scope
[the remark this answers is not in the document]
AndreasArvidsson: How would we even determine that in a simple way?

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
pokey: yep that's exactly what I'm proposing. We could tweak from there, but I do think something like that makes whitespace very clear

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
pokey: Huh. So a single line results in no trailing newline? Not sure how I feel about that. Files with no trailing newline are generally considered bad practice, so I'd lean towards the default being trailing newline cc/ @josharian

AndreasArvidsson:
@pokey Now scope support facets are implemented  :)

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
AndreasArvidsson: Yes the block above the dashes is the entire document.

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
AndreasArvidsson: I'm not sure if that would be easier to read.

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/function.scope
AndreasArvidsson: Done

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/assignment.name.scope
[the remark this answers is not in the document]
AndreasArvidsson: Grouped headers with equal ranges

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
pokey: discussed in meet-up:

- Only add the marker at the end of the line
- Use something like `◂` as the marker
- If it's too noisy, don't include the marker if the line ends with a nonwhitespace char (ie you'd only get it for empty lines or lines with trailing whitespace)

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/plaintext/line2.scope
```diff
@@ -0,0 +1,78 @@
+
```
josharian, line 1: I find the contents of this file hard to parse (as a human)

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/name.assignment.scope
```diff
@@ -0,0 +1,24 @@
+const name = "Hello world";
```
josharian, line 1: this is lovely

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/plaintext/line2.scope
josharian: oops, I see that there was an entire conversation about this above.

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
AndreasArvidsson: I made that change now when I think it turned out quite nice. You're are correct that using the same character which different alignment is confusing

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
AndreasArvidsson: This is now implemented

AndreasArvidsson: @josharian better?

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/name.assignment.scope
```diff
@@ -0,0 +1,24 @@
...
+
+[Content]
+ const name = "Hello world";
+       ^^^^
+
+[Removal]
+ const name = "Hello world";
+ ^^^^^^^^^^^^^
+
+[Leading delimiter: Content]
```
josharian, line 12: nit: I would rather organize like this:

```
[content]
[content leading]
[content trailing]
[removal]
[removal leading]
[removal trailing]
[domain]
```

etc

I think it is easier to reason about.

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/line.scope
josharian: I couldn't figure out how to understand these from just looking at the files, which would be the best case goal. I'm probably just missing something obvious, but why aren't we using `^` in these cases?

AndreasArvidsson: It's an empty line. The carets are on the character, but the squares are outside of the range and therefore works.

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/name.assignment.scope
AndreasArvidsson: Glad to hear it :)

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/name.assignment.scope
AndreasArvidsson: I don't think I agree. We now group the headers for identical ranges together. The content removal and domain is likely to be the same while the leading and trailing will never. Content and removal also belongs to a single target while leading and trailing is two separate targets.

pokey: I agrée with @AndreasArvidsson here. Leading content and leading removal are attributes of a single target: the leading target of the primary target here

Tbh I'd be tempted to drop leasing removal and trailing removal unless they differ from leading content / trailing content, because they're almost always the same and their very existence is confusing, and basically only useful for the block scope

AndreasArvidsson: I mean we do already group the headers if they are identical?

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/plaintext/line2.scope
pokey, line 1: Huh. Getting better. Still needs something. Glad you included this file as it's prob one of the more challenging cases 😅

Maybe let's discuss

packages/cursorless-vscode-e2e/src/suite/scopes.vscode.test.ts
pokey, line 1: Could prob factor out the serialisation stuff into another file

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/plaintext/line2.scope
AndreasArvidsson: Sounds good

packages/cursorless-vscode-e2e/src/suite/scopes.vscode.test.ts
AndreasArvidsson: done

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/name.assignment.scope
pokey: update: let's just call it `Leading delimiter` if content and removal range are equal

AndreasArvidsson: done

AndreasArvidsson:
@pokey @josharian Format updated according to yesterday's discussion
I do wonder if we want some symbol on the start of the range line. When we have a trailing empty line like this it's a bit confusing.

image

I also wonder if we want to update our a range.toString()? Poke is example with parentheses around the positions was quite nice.

josharian:
Give it a quick scan. These look fabulous to me.

> I also wonder if we want to update our a range.toString()? Pokeys example with parentheses around the positions was quite nice.

I feel quite strongly that we should not add parens, at least not in `range.toString`. We could certainly add them to the files.

> Pipes on all lines

This remains on my favorite, but I don't feel strongly.

AndreasArvidsson:
> > Pipes on all lines
> 
> This remains on my favorite, but I don't feel strongly.

There are quite a few ways to draw lines/boxes using unicode. May be some differentiation between code and range lines would be good?
https://en.wikipedia.org/wiki/Box-drawing_character

```
[#1 Content] = 0:0-0:0
0│
 ┊><
1│ aaa
 ┊
2│
 ┊
```

```
[#1 Content] = 0:0-0:0
0│
 ┆><
1│ aaa
 ┆
2│
 ┆
```

pokey:
Yeah that looks nice with the box drawing chars. The one thing I still don't totally love is that space before line start. Fine when there's content, but a bit tough to read for that empty line

What if we used a light [block element](https://en.wikipedia.org/wiki/Block_Elements?wprov=sfti1#) before and/or after line content? (░)

AndreasArvidsson:
> Yeah that looks nice with the box drawing chars. The one thing I still don't totally love is that space before line start. Fine when there's content, but a bit tough to read for that empty line
> 
> What if we used a light [block element](https://en.wikipedia.org/wiki/Block_Elements?wprov=sfti1#) before and/or after line content? (`░`)

Like this?

[#1 Content] = 0:0-0:0
0│░
 ┆><
1│░aaa
 ┆
2│░
 ┆

josharian:
We do need to make sure any Unicode we use renders well across a variety of fonts so people can always read their files locally in their editor and terminal.

An advantage to true whitespace is that tools like git usually have useful special handling of it.

we could try again ditching the leading space, and having the > occur underneath a | when needed.

josharian:
But I’m with Andreas, I’m fine with the leading whitespace.

Perhaps most importantly, this is a write-only data format, so we could just get it in and change our minds later.

AndreasArvidsson:
> We do need to make sure any Unicode we use renders well across a variety of fonts so people can always read their files locally in their editor and terminal.
> 
> An advantage to true whitespace is that tools like git usually have useful special handling of it.
> 
> we could try again ditching the leading space, and having the > occur underneath a | when needed.

True

```
[#1 Content] = 0:0-0:0
0|
‣><
1|aaa
‣
2|
‣
```

pokey:
Maybe let's just stick with current

AndreasArvidsson:
> Maybe let's just stick with current

I think I want some symbol on totally empty lines. It's quite hard to understand that there's a trailing empty line otherwise.

pokey:
> I think I want some symbol on totally empty lines. It's quite hard to understand that there's a trailing empty line otherwise.

Not sure I follow. Empty line has a line number and pipe on it

AndreasArvidsson:
> > I think I want some symbol on totally empty lines. It's quite hard to understand that there's a trailing empty line otherwise.
> 
> Not sure I follow. Empty line has a line number and pipe on it

Empty range line.

pokey:
But if the line has no markers on it, then I'm not sure why it's important?

josharian:
I think we’ve reached a point of rapidly diminishing returns. I propose that we get this in is as it is, and if it proves to be confusing overtime, we fix it.

AndreasArvidsson:
@pokey @josharian Just realize that we don't capture insertion delimiter here. That also feels like an important part of a scope?

AndreasArvidsson:
@pokey @josharian Insertion delimiter added

packages/cursorless-vscode-e2e/src/suite/serializeHeader.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Constructs a header string from a configuration object. For example:
+ *
+ * ```
+ * serializeHeader({
+ *  prefix: "Leading delimiter",
+ *  header: "Content",
+ *  scopeIndex: 1,
+ *  targetIndex: 2,
+ *  range: new Range(new Position(1, 2), new Position(3, 4)),
+ * }) === "[#1 Leading delimiter: Content 2] =1:2-3:4"
```
pokey, line 21: Did I get this right? If so, it feels a bit awkward to have the target index be all the way at the end; feels like it's a higher part of the hierarchy than the header and prefix. I'd argue something like

```
[#1.2 Leading delimiter: Content]
```

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/name.assignment.scope
```diff
@@ -0,0 +1,24 @@
+const name = "Hello world";
+---
+
+[Content] = 0:6-0:10
+0| const name = "Hello world";
```
pokey, line 5: The one thing I do find throwing me is the leading space. My brain keeps trying to see it as part of the document. Have we experimented with dropping it? Is it too busy?

pokey (commented):
I must say this represenation is actually quite beautiful. Very happy with where we landed. Did some refactoring and left some comments

packages/common/src/scopeSupportFacets/scopeSupportFacets.ts
```diff
@@ -0,0 +1,167 @@
...
+] as const;
+
+const textualScopeSupportFacets = [
+  "character",
+  "word",
+  "token",
+  "line",
+  "paragraph",
+  "document",
+] as const;
```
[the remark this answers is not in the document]
AndreasArvidsson: They are not. They will only be added to the language plaintext. Other languages are not supposed to implement this.

packages/cursorless-vscode-e2e/src/suite/scopes.vscode.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    assert.fail(`Missing test for scope support facets [${values}]`);
+  }
+}
+
+async function runTest(file: string, languageId: string, facetId: string) {
+  const { ide, scopeProvider } = (await getCursorlessApi()).testHelpers!;
+  const scopeType = getScopeType(facetId);
+  const fixture = (await fsp.readFile(file, "utf8"))
+    .toString()
+    .replaceAll("\r\n", "\n");
```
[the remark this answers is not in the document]
AndreasArvidsson: True

packages/cursorless-vscode-e2e/src/suite/serializeHeader.ts
pokey: Sure. Shall I make that change then?

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/name.assignment.scope
pokey: Yeah I guess they do look a bit cramped 😕. Would be cool to use those shading chars, but I guess prob not introducing strange unicode symbols for. Let's leave it for now; hopefully my brain will just adjust

packages/common/src/testUtil/getFixturePaths.ts
```diff
@@ -32,3 +36,17 @@ export function getRecordedTestPaths() {
...
+export function getScopeTestPaths() {
+  const directory = getScopeTestsDirPath();
+  const relativeDir = path.dirname(directory);
+
+  return walkFilesSync(directory)
+    .filter((p) => p.endsWith(".scope"))
+    .map((p) => ({
+      path: p,
+      name: path.relative(relativeDir, p.substring(0, p.lastIndexOf("."))),
+      languageId: path.dirname(path.relative(directory, p)),
```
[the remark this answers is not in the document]
AndreasArvidsson: fixed!

packages/cursorless-vscode-e2e/src/suite/serializeHeader.ts
AndreasArvidsson: fixed

pokey:
Hmm. Not sure. Maybe separate fixture? I wonder if iteration scope for each scope should be its own facet. Not a bad idea, to make people think about iteration scope for scopes for which it matters

AndreasArvidsson:
Yes I was thinking of separate facets. Probably a good idea?

pokey:
Yeah let's go with separate facets 👍

AndreasArvidsson:
@pokey Iterations scope facet and fixtures added

packages/common/src/scopeSupportFacets/html.ts
```diff
@@ -0,0 +1,19 @@
+import {
+  LanguageScopeSupportFacetMap,
+  ScopeSupportFacetLevel,
+} from "@cursorless/common";
+
+const { supported, notApplicable } = ScopeSupportFacetLevel;
+
+export const htmlSupport: LanguageScopeSupportFacetMap = {
+  ["key.attribute"]: supported,
+  ["tags.element"]: supported,
```
pokey, line 10: Why the `.element` here?

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/html/tags.element.scope
```diff
@@ -0,0 +1,20 @@
...
+[#1.1 Insertion delimiter] = " "
+
+[#1.2 Content] =
+[#1.2 Removal] = 0:10-0:16
+0| <div>hello</div>
+            >------<
+
+[#1.2 Insertion delimiter] = " "
+
+[Domain] = 0:0-0:16
```
pokey, line 18: Hmm it's a bit confusing that the above have a `#1` but this one doesn't. I wonder if we should do `.1` instead of `#1.1` if there's only one?

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/value.mapPair.iteration.scope
```diff
@@ -0,0 +1,13 @@
+{ value: 123 }
+---
+
+[#1 Range] =
+[#1 Domain] = 0:0-0:14
+0| { value: 123 }
+  >--------------<
```
pokey, line 7: Huh. Technically this iteration scope is not for this facet. Not sure there's much we can do about that tho 🤷‍♂️

packages/common/src/scopeSupportFacets/html.ts
AndreasArvidsson: We might have tags in another context in the future. I'm just trying to be thorough.

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/html/tags.element.scope
AndreasArvidsson: We can do that. `.1` or `#.1`?

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/value.mapPair.iteration.scope
AndreasArvidsson: I don't think so. This would be a reason to have more fine grain scopes as well.

pokey: Yeah fair point. I wonder how many of our facets should really be their own scopes 🤔

AndreasArvidsson: Probably quite a few.

packages/common/src/scopeSupportFacets/html.ts
josharian: idk either

how immutable do we need these facets to be? can we lower the cost of mistakes sufficiently that it doesn't matter much?

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/html/tags.element.scope
AndreasArvidsson: done

packages/common/src/scopeSupportFacets/html.ts
AndreasArvidsson: Were gonna have a lot of files with these facets as their name so it's not gonna be that easy to change them, but also definitely not impossible.

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/value.mapPair.iteration.scope
AndreasArvidsson: I think one clear indicator that a scope should split is that we have multiple iteration scopes as in the above example. For example all the different types of ways you can do functions in javascript is still the same function scope, but a value in a map pair is not really the same thing as a return statement value and they do have different iteration scopes. This is definitely going to be on a case by case basis, but I think that having multiple iteration scopes for the same scope is a bit of a warning sign.

packages/common/src/scopeSupportFacets/html.ts
pokey: Update from meet-up: let's go with just `tags`. It's ok to have the general thing and then sub-facets as well

AndreasArvidsson:
@pokey Updated after today's meeting discussion

packages/common/src/scopeSupportFacets/textual.ts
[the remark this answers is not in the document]
AndreasArvidsson: True. Removed

packages/common/src/scopeSupportFacets/javascript.ts
```diff
@@ -0,0 +1,19 @@
...
+  namedFunction: supported,
+  ["name.assignment"]: supported,
+  ["key.mapPair"]: supported,
+  ["key.mapPair.iteration"]: supported,
+  ["value.mapPair"]: supported,
+  ["value.mapPair.iteration"]: supported,
+  ["value.assignment"]: supported,
+
+  ["key.attribute"]: notApplicable,
+  ["tags"]: notApplicable,
```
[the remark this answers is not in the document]
AndreasArvidsson: I was thinking that we should test all jsx under javascript eact and not javascript. I would leave it undefined if anything. Saying that it's unsupported is not true either.

packages/common/src/scopeSupportFacets/textualScopeSupportFacetInfos.ts
```diff
@@ -0,0 +1,47 @@
...
+    label: "Character",
+    description: "A single character in the document",
+    scopeType: "character",
+    examples: ["a", "."],
+  },
+  word: {
+    label: "Word",
+    description: "A single word in a token",
+    scopeType: "word",
+    examples: ["foo_bar", "fooBar"],
```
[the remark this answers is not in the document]
AndreasArvidsson: I was thinking examples as a very general guide. If you want details you she'd look at the actual fixtures

packages/common/src/scopeSupportFacets/textualScopeSupportFacetInfos.ts
```diff
@@ -0,0 +1,47 @@
...
+    label: "Token",
+    description: "A single token in the document",
+    scopeType: "token",
+    examples: ["foo", "("],
+  },
+  line: {
+    label: "Line",
+    description: "A single line in the document",
+    scopeType: "line",
+    examples: ["foo"],
```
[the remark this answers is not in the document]
AndreasArvidsson: If you don't like them we can't ditch them

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
```diff
@@ -0,0 +1,82 @@
+import { SimpleScopeTypeType } from "../types/command/PartialTargetDescriptor.types";
+
+const scopeSupportFacets = [
```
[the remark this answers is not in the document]
AndreasArvidsson: I don't quite follow what you're suggesting? What exactly do you want to make more fine grained? I definitely prefer them fine grained

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
```diff
@@ -0,0 +1,87 @@
...
+  "token",
+  "line",
+  "paragraph",
+  "document",
+] as const;
+
+export interface ScopeSupportFacetInfo {
+  readonly label: string;
+  readonly description: string;
+  readonly scopeType: SimpleScopeTypeType;
```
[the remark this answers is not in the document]
AndreasArvidsson: I just copied the type from your issue. We can make it a scope type instead?

packages/common/src/scopeSupportFacets/scopeSupportFacetInfos.ts
```diff
@@ -0,0 +1,66 @@
...
+    label: "Assignment name",
+    description: "Name(LHS) of an assignment",
+    scopeType: "name",
+    examples: ["const foo = 1"],
+  },
+  ["key.attribute"]: {
+    label: "Attributes key",
+    description: "Key(LHS) of an attribute",
+    scopeType: "collectionKey",
+    examples: ['id="root"'],
```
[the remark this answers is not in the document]
AndreasArvidsson: Absolutely.

packages/common/src/scopeSupportFacets/scopeSupportFacetInfos.ts
```diff
@@ -0,0 +1,66 @@
...
+    description: "Key(LHS) of a map pair",
+    scopeType: "collectionKey",
+    examples: ["value: 0"],
+  },
+  ["key.mapPair.iteration"]: {
+    label: "Map pair key iteration",
+    description: "Iteration of map pair keys",
+    scopeType: "collectionKey",
+    isIteration: true,
+    examples: ["{ value: 0 }"],
```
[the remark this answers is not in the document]
AndreasArvidsson: The examples are and the content range. It's just example of the type of code that would have this facet

packages/common/src/scopeSupportFacets/scopeSupportFacetInfos.ts
```diff
@@ -0,0 +1,66 @@
...
+    examples: ["const foo = 1"],
+  },
+  ["key.attribute"]: {
+    label: "Attributes key",
+    description: "Key(LHS) of an attribute",
+    scopeType: "collectionKey",
+    examples: ['id="root"'],
+  },
+  ["key.mapPair"]: {
+    label: "Map key",
```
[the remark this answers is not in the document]
AndreasArvidsson: Sounds good to me

AndreasArvidsson:
I would be happy to remove label and examples, but I don't think we should make them optional.

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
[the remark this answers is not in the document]
AndreasArvidsson: Totally agree

packages/common/src/scopeSupportFacets/javascript.ts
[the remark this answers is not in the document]
pokey: Hmm. Probably fine for now. At some point we should prob figure out how to automatically run tests for all langs where they're applicable, eg run our js tests in ts, etc. We could maybe make subsets of languages and then define a way to import tests from one language to another. But we can do that in a follow-up

pokey (approved):

PR #2069, opened 2023-11-29
Public get text action

### LINKED ISSUE #452: Add `get_text` action to public cursorless api
Could use this api to implement https://github.com/pokey/cursorless-vscode/issues/101

AndreasArvidsson:
Fixes #452

## Checklist


AndreasArvidsson:
> we should prob have api test for this, no?

Extension side we actually have multiple tests for this already
eg
https://github.com/cursorless-dev/cursorless/blob/a723b253c61106dfc7e32638e2d20b53292d17a0/packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/findVest.yml

pokey:
Yes but I mean the talon-side api introduced by this PR. We have tests for our other talon-side APIs

cursorless-talon/src/actions/get_text.py
```diff
@@ -1,29 +1,32 @@
...
-        }
-    )
+
+@mod.action_class
+class Actions:
+    def cursorless_get_text(
+        target: CursorlessTarget,
+        show_decorations: Optional[bool] = None,
+        ensure_single_target: Optional[bool] = None,
+    ) -> list[str]:
```
pokey, line 16: this api will probably be a bit surprising for the most common use case of getting text from a single target. And I believe it will be awkward to use from a `.talon` file. I wonder if we want to

1. Change this action to return a string, and not expose `ensure_single_target`; we just set it to `true` before passing along to extension
1. Add a separate action `cursorless_get_text_list` or `cursorless_get_text_multiple` which returns a list, and doesn't set `ensure_single_target`

wdyt?

AndreasArvidsson: I like it. Updated!

cursorless-talon/src/actions/get_text.py
```diff
@@ -1,24 +1,46 @@
...
-
-    if show_decorations is not None:
-        options["showDecorations"] = show_decorations
-
-    if ensure_single_target is not None:
-        options["ensureSingleTarget"] = ensure_single_target
+    options: dict[str, bool] = {
+        "showDecorations": show_decorations,
+        "ensureSingleTarget": ensure_single_target,
+    }
```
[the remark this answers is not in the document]
AndreasArvidsson: Either is fine for me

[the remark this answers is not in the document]
AndreasArvidsson: I thought a bit about it and I now only pass it if it's needed.

cursorless-talon/src/actions/get_text.py
```diff
@@ -1,24 +1,48 @@
...
 from ..targets.target_types import CursorlessTarget
 
+mod = Module()
+
+
+@mod.action_class
+class Actions:
+    def cursorless_get_text(
+        target: CursorlessTarget,
+        hide_decorations: bool = False,
```
pokey, line 12: Not quite in love with this name. `no_decorations`? `silent`? `hide` feels a bit to active for what is really the absence of acting. Ie I might expect it to actually hide something

AndreasArvidsson: I had a similar thought. `no_decorations` works for me.

AndreasArvidsson: done

cursorless-talon/src/actions/get_text.py
```diff
@@ -1,24 +1,48 @@
...
 
-    if show_decorations is not None:
-        options["showDecorations"] = show_decorations
-
-    if ensure_single_target is not None:
-        options["ensureSingleTarget"] = ensure_single_target
+    if no_decorations:
+        options["showDecorations"] = False
+    if ensure_single_target:
+        options["ensureSingleTarget"] = True
```
pokey, line 45: seems a bit odd to use a boolean to determine whether to pass args. I would argue for either keeping them as they were originally, where we don't pass anything if nothing comes in, or always passing them. Silently dropping them based on current default behaviour feels like it's asking for trouble

AndreasArvidsson: Sure

pokey (approved):
made some changes:

- Stopped using the public api for our internal impl. I'm not in love with coupling those
- Switched `no_decorations` to `silent`. Double-negatives confuse me, so I prefer to avoid them if possible
- I made all args to our function explicit. If we have the decorations flag reversed between talon api and cursorless api, the only way to make heads or tails I think is to just be super explicit and pass them every time

Feel free to merge if you're happy

AndreasArvidsson:
I'm not in love with the argument `silent`. It's much less descriptive than `no_decorations`.

pokey:
> I'm not in love with the argument `silent`. It's much less descriptive than `no_decorations`.

I agree. But I also really don't like double negatives (ie `no_decorations = False`). Any better ideas? Chatgpt suggests

- `omit_decorations`
- `hide_decorations`
- `suppress_decorations`

cc/ @josharian

AndreasArvidsson:
> > I'm not in love with the argument `silent`. It's much less descriptive than `no_decorations`.
> 
> I agree. But I also really don't like double negatives (ie `no_decorations = False`). Any better ideas? Chatgpt suggests
> 
>     * `omit_decorations`
> 
>     * `hide_decorations`
> 
>     * `suppress_decorations`
> 
> 
> cc/ @josharian

`hide_decorations` sounds reasonable

pokey:
Ok made some final tweaks and confirmed talon tests work locally. I'll give you a chance for a quick look in case I did anything dumb; merge if happy

cursorless-talon-dev/src/cursorless_test.talon
```diff
@@ -9,8 +9,12 @@ test api command bring <user.cursorless_target>:
     user.cursorless_command("replaceWithTarget", cursorless_target)
 test api get text <user.cursorless_target>:
     user.cursorless_get_text(cursorless_target)
-test api get text list <user.cursorless_target>:
+test api get text list on <user.cursorless_target>:
```
AndreasArvidsson, line 12: Why on?

pokey: otherwise there's ambiguity with command on line above because "list" is a scope type

packages/cursorless-engine/src/test/fixtures/talonApi.fixture.ts
```diff
@@ -131,8 +130,26 @@ export const talonApiFixture = [
     wrapWithSnippetByNameAction,
   ),
   spokenFormTest("parse tree air", parseTreeAction),
-  spokenFormTest("test api get text air", getTextAction),
-  spokenFormTest("test api get text list air", getTextListAction),
+  multiActionSpokenFormTest(
```
AndreasArvidsson, line 133: How is this a multi action spoken form? All it does is getting the text in a single action?

cursorless-talon-dev/src/cursorless_test.talon
```diff
@@ -9,8 +9,12 @@ test api command bring <user.cursorless_target>:
...
     user.cursorless_get_text_list(cursorless_target)
+test api get text hide decorations <user.cursorless_target>:
+    user.cursorless_get_text(cursorless_target, 1)
```
AndreasArvidsson, line 15: I would prefer to use `true` instead of `1`

packages/cursorless-engine/src/test/fixtures/talonApi.fixture.ts
pokey: yeah that function is poorly named. That function has an arg that lets us mock the output of `get_text`, which is necessary for the test to work, as otherwise the `[0]` in python action breaks. Shall I rename the function / make a different function? I agree it's a bit confusing

cursorless-talon-dev/src/cursorless_test.talon
pokey: oh is that a thing now? i have a vague recollection of that. It's lowercase?

packages/cursorless-engine/src/test/fixtures/talonApi.fixture.ts
AndreasArvidsson: Yeah please rename it because I was thoroughly confused :D

cursorless-talon-dev/src/cursorless_test.talon
AndreasArvidsson: I guess, but the tests were fine the way it was before. Unless we actually want to test with the list scope type I don't believe there is a problem. Anyhow I'm fine with being more verbose.

cursorless-talon-dev/src/cursorless_test.talon
AndreasArvidsson: Yes

cursorless-talon-dev/src/cursorless_test.talon
pokey: Yeah I'm not sure why they started breaking for me tbh. But I think that it is better practice to make sure get text returns the right thing for the test anyway

packages/cursorless-engine/src/test/fixtures/talonApi.fixture.ts
pokey: better?

AndreasArvidsson: Yes

PR #2073, opened 2023-11-29
Migrated xml queries to new parser

AndreasArvidsson:
Parsed using:
https://github.com/ObserverOfTime/tree-sitter-xml

Still have a problem where `<?xml version="1.1" encoding="UTF-8" ?>` fails. Even though the parser library has these in their own ci tests. Going to have to do some more testing

https://github.com/pokey/tree-sitter-wasm-test

## Checklist


pokey (commented):
Looks good but see my comments on #2044; same apply here

pokey (approved):
👍

PR #2076, opened 2023-12-01
Remove more cyclic import dependencies

bjaspan:
* Changes many imports from index files to direct source.
* Removes cycle from notebook.ts and notebookLegacy.ts.
* Moves processTargets/targetUtil to processTargets/targets/util so that those files can import from direct source without breaking the barrel boundary.
* Moves tryConstructPlainTarget and tryConstructLineTarget to PlainTarget and LineTarget, respectively.

The Position/Range/Selection cycle is not touched.  There are still some other cycles remaining as well, but this gets a lot of them.


bjaspan:
It looks like my notebook.ts refactoring broke something but I do not yet understand the failure message:

```
  1) Cross-cell set selection
       Cross-cell set selection:
     Error: Extension 'pokey.cursorless' CANNOT use API proposal: notebookEditor.
Its package.json#enabledApiProposals-property declares: [] but NOT notebookEditor.
 The missing proposal MUST be added and you must start in extension development mode or use the following command line switch: --enable-proposed-api pokey.cursorless
```

pokey (commented):
Mostly looks fine; left a pointer for how to fix your notebook issue

One challenge / annoyance will be getting typescript auto-import not to use barrels when it finds them. Otherwise these will just keep popping back up, and even with CI check it will be annoying to have to keep fixing them. There are some settings we can use to tune auto-import; maybe one of them can help us here

Did some research and I see https://github.com/microsoft/TypeScript/issues/45953, which claims to have been solved by https://github.com/microsoft/TypeScript/pull/47516, but that's not what I'm observing 🤔

image

packages/cursorless-vscode/src/ide/vscode/notebook/notebookCurrent.ts
```diff
@@ -0,0 +1,23 @@
+import { NotebookCell, TextDocument, window } from "vscode";
+
+export function getNotebookFromCellDocumentCurrent(document: TextDocument) {
```
pokey, line 3:
```suggestion
/** Gets the notebook containing a text document using >1.68.0 VSCode notebook api **/ 
export function getNotebookFromCellDocumentCurrent(document: TextDocument) {
```

pokey (approved):

packages/cursorless-vscode/src/ide/vscode/notebook/notebookCurrent.ts
```diff
@@ -0,0 +1,24 @@
+import { NotebookCell, TextDocument, window } from "vscode";
+
+/** Gets the notebook containing a text document using >1.68.0 VSCode notebook api **/
```
pokey, line 3:
```suggestion
/** Gets the notebook containing a text document using >=1.68.0 VSCode notebook api **/
```

PR #2086, opened 2023-12-05
Return value for lambdas

AndreasArvidsson:
## Checklist


queries/javascript.core.scm
```diff
@@ -370,6 +370,14 @@
   (_) @value
 ) @_.domain
 
+;;!! str => str.length > 0
+;;!         ^^^^^^^^^^^^^^
+;;!  ---------------------
+(arrow_function
+  body: (_) @value
+  (#not-type? @value statement_block)
```
pokey, line 378: I agree this is prob the right place to start. We'll see if we end up wanting it for blocks as well in the future

pokey (approved):
Oh awesome yeah I've been wanting this for a while 🙌

queries/javascript.core.scm
AndreasArvidsson: That should probably be the `body` scope instead?

PR #2087, opened 2023-12-05
Support value for typescript destruct pattern

### LINKED ISSUE #2082: "value" broken for patterns in typescript
```ts
const { aaa: bbb } = ccc;
```

`"value air"` should be `bbb`.

AndreasArvidsson:
Fixes #2082

## Checklist

### DIFF
--- a/packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/typescript/changeValue3.yml
+++ b/packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/languages/typescript/changeValue3.yml
@@ -1,0 +1,25 @@
+languageId: typescript
+command:
+  version: 6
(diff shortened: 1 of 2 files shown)


pokey (commented):
looks good but def want leading delimiter here. I def chuck these. Also worth double-checking that we support name / chuck name here

pokey:
> If this looks like a value shouldn't it be key instead?

sorry right yeah that 😅

AndreasArvidsson:
> > If this looks like a value shouldn't it be key instead?
> 
> sorry right yeah that 😅

Honestly when its name and when its key is actually a bit confusing at times

AndreasArvidsson:
Fixed

pokey (approved):
👍

PR #2091, opened 2023-12-05
Don't focus search editor

### LINKED ISSUE #1722: cannot focus search results editor
To reproduce:

* use the editor-based VSCode search UI
* say `pre whale`, where whale is one of the hats in the search results

Result:

The search query is focused rather than the results. This makes it hard to use page up and page down to look through the search results.

AndreasArvidsson:
Partially addresses #1722

The problem is that when you focus the search editor focus is brought back to the import/query box. For now just don't focus the editor actually makes it so if the search editor is already in focus you can now use Cursorless in it. The simplest way to get to the search editor is probably use `search.action.focusNextSearchResult`. Unfortunately there is no command to focus the results generally. Focusing the next result we of course don't want to issue every time we do a Cursorless command.

## Checklist

### DIFF
--- a/packages/cursorless-vscode/src/ide/vscode/VscodeFocusEditor.ts
+++ b/packages/cursorless-vscode/src/ide/vscode/VscodeFocusEditor.ts
@@ -32,6 +32,11 @@ export default async function vscodeFocusEditor(
   ide: VscodeIDE,
   editor: VscodeTextEditorImpl,
 ) {
+  // Focusing the search editor brings focus back to the input field.
+  if (editor.document.uri.scheme === "search-editor") {
+    return;
+  }
+
   const viewColumn = getViewColumn(editor.vscodeEditor);
   if (viewColumn != null) {
     await commands.executeCommand(columnFocusCommands[viewColumn]);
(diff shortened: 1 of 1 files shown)


pokey:
note that if you say "partially fixes #nnn", github ignores the "partially" and will close the issue. I usually say "partially addresses" for that reason

packages/cursorless-vscode/src/ide/vscode/VscodeFocusEditor.ts
```diff
@@ -32,6 +32,11 @@ export default async function vscodeFocusEditor(
   ide: VscodeIDE,
   editor: VscodeTextEditorImpl,
 ) {
+  // Focusing the search editor brings focus back to the input field.
+  if (editor.document.uri.scheme === "search-editor") {
+    return;
+  }
+
```
pokey, line 39: is there a way to detect that the search editor is already focused? If it is not, maybe we still want to issue this command? Either way, I'd add a FIXME here because this code makes it look like it's doing the right thing, when really it's a hack. I'd like to the issue

AndreasArvidsson:
> note that if you say "partially fixes #nnn", github ignores the "partially" and will close the issue. I usually say "partially addresses" for that reason

Didn't know that. thanks!

packages/cursorless-vscode/src/ide/vscode/VscodeFocusEditor.ts
AndreasArvidsson: Comment added

pokey (approved):

PR #2094, opened 2023-12-05
Added visible modifier

### LINKED ISSUE #1607: Add "visible" scope

AndreasArvidsson:
Fixes #1607

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/VisibleStage.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/VisibleStage.ts
@@ -1,0 +1,19 @@
+import { VisibleModifier } from "@cursorless/common";
+import { Target } from "../../typings/target.types";
+import { ModifierStage } from "../PipelineStages.types";
+import { PlainTarget } from "../targets";
+
+export class VisibleStage implements ModifierStage {
+  constructor(private modifier: VisibleModifier) {}
+
+  run(target: Target): Target[] {
+    return target.editor.visibleRanges.map(
+      (range) =>
+        new PlainTarget({
+          editor: target.editor,
+          isReversed: target.isReversed,
+          contentRange: range,
+        }),
+    );
+  }
+}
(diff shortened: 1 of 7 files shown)


josharian (commented):
Yay, thanks! Excited to have this.

pokey (commented):
Looks good. Let's get a spoken form test; should be easy with our mechanism for custom spoken form tests eg add an entry to https://github.com/cursorless-dev/cursorless/blob/197bb16247a45dff3337a3f5a1488335d83a6682/packages/cursorless-engine/src/test/spokenForms.talon.test.ts#L42-L44

or just add a recorded test. Either way

AndreasArvidsson:
fixed

packages/cursorless-engine/src/processTargets/modifiers/VisibleStage.ts
```diff
@@ -0,0 +1,19 @@
+import { VisibleModifier } from "@cursorless/common";
+import { Target } from "../../typings/target.types";
+import { ModifierStage } from "../PipelineStages.types";
+import { PlainTarget } from "../targets";
+
+export class VisibleStage implements ModifierStage {
+  constructor(private modifier: VisibleModifier) {}
```
pokey, line 7: not worth changing for this PR, esp since we do this all over the place, but don't feel compelled to pass in modifier if it's not actually used. Gives a nice signal we don't use it if we don't even pass it in. I think one of the benefits of our DI setup is that we don't enforce uniformity like other DI mechanisms

pokey (approved):
👍

packages/cursorless-engine/src/processTargets/modifiers/VisibleStage.ts
pokey: Yeah I guess that's fair; hadn't thought bout that

Issue #473, opened 2022-01-13
Support "all" modifier


pokey:
Similar to "every", but targets range from first to last rather than emitting one selection per item

Will-Sommers:
Heyo — I'm looking at this ticket as one to take and work on. If you think it will be across too much of the project, let me know.

So, taking a look at this, here's where I would start.

### Processing the command on the talon side
It looks like we need to add something into cursorless-talon to pick up the `"all"` command just as we pick up the [`"every"` command](https://github.com/cursorless-dev/cursorless-vscode/blob/4c3117f1c393cc1a9088aa1dfa3cf570ab26ec0d/cursorless-talon/src/modifiers/containing_scope.py#L49-L58). I've only worked on/looked at code from the canonical command to `.ts` boundary. Since `includeSiblings` is a boolean, @pokey, I think that we'd either want:

- Add a new modifier type to be processed within [`processModifier.ts`](https://github.com/cursorless-dev/cursorless-vscode/blob/4c3117f1c393cc1a9088aa1dfa3cf570ab26ec0d/src%2FprocessTargets%2Fmodifiers%2FprocessModifier.ts#L34)
- Change the `includeSiblings` boolean to encode more data, such as "single selection" or "multiple selection".

### Processing the command on the cursorless side

Then within the cursorless side, support the changes above in the processModifier file, either adding a new scope type or working at the [`findNearestContainingAncestorNode`](https://github.com/cursorless-dev/cursorless-vscode/blob/4c3117f1c393cc1a9088aa1dfa3cf570ab26ec0d/src%2FprocessTargets%2Fmodifiers%2FprocessModifier.ts#L220-L243) level. 


### A few QQ's:
- When working on the cursorless-talon portion of the project, I'll need symlink my `~/.talon/user/cursorless-talon` to my dev folder, correct?
- Is there a policy on adding tests for languages. Should I target every language currently supported for this?

Will-Sommers:
Also, if this is blocking or not inline with the engine 2.0 work, lmk and I'll look at another ticket.

pokey:
Yep seems like a good one!  I'd proceed as follows:

- talon side: i'd either add a bit more info to the `includeSiblings` field or even just add another boolean.  we'll be reworking this representation so the details aren't super important.  We have a good translation layer for when we change this stuff
[...]

PR #2101, opened 2023-12-06
Contiguous scope

### LINKED ISSUE #1835: Add contiguous scope modifier
Could be used for example targeting multiple contiguous line comments.

`"comment slurp comments fine"`

```ts
// foo
// bar
```

## Potential solution
I was thinking that this could be implemented around the existing scope handler similar to how there every modifier works and just join scopes/targets based on leading and trailing delimiter ranges.

1. Get all scopes
```ts
const scopes = scopeHandler.generateScopes(...);
```

2. Iterate scopes and join contiguous targets
```ts
const leadingRange = leadingTarget.trailingDelimiterRange ?? leadingTarget.contentRange;
const trailingRange = trailingTarget.leadingDelimiterRange ?? trailingTarget.contentRange;
if (leadingRange.intersects(trailingRange) != null) {
    newContentRange = leadingTarget.contentRange.union(trailingTarget.contentRange);
}
```

## Thinks to discuss
* One problem is that most leading and trailing delimited ranges doesn't contain the eol. eg with comment the leading range would end with the previous line and the trailing range would start with the next line. We could of course say that that is valid for merging, but it would probably make `"take slurp tokens"` join all tokens in the file and therefore take the entire file.
* Do we want to support `"every slurp <scope>"`?

AndreasArvidsson:
Fixes #1835 

I decided to implement this as a scope type instead of a modifier. Now you can say things like `chuck fat comment` or `changed next fat alpha`. This behavior of having alternatives to existing scopes is already something we started with `BoundedNonWhitespaceSequenceStage` albeit in a less capacity.

## Checklist


packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
```diff
@@ -0,0 +1,191 @@
...
+      domain: leadingTarget.contentRange,
+      getTargets: () => [leadingTarget],
+    };
+  }
+
+  const range = leadingTarget.contentRange.union(trailingTarget.contentRange);
+  return {
+    editor: leadingTarget.editor,
+    domain: range,
+    getTargets: () => [leadingTarget.withContentRange(range)],
```
pokey, line 119: why not use our new `createContinuousRangeTarget` function?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
```diff
@@ -0,0 +1,191 @@
...
+  for (const scope of generator) {
+    for (const target of scope.getTargets(false)) {
+      if (first == null) {
+        first = target;
+      }
+
+      if (last != null) {
+        const [leadingTarget, trailingTarget] = isForward
+          ? [last, target]
+          : [target, last];
```
pokey, line 146: Tempted to switch to a `{proximal, distal}` obj to remove all these ternaries and confusing anonymous lists

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
```diff
@@ -0,0 +1,191 @@
...
+      last = target;
+    }
+  }
+
+  if (first != null && last != null) {
+    yield isForward ? [first, last] : [last, first];
+  }
+}
+
+function isAdjacent(leadingTarget: Target, trailingTarget: Target): boolean {
```
pokey, line 163: surprised you ignore domain here. what if their domains are adjacent?

packages/cursorless-engine/src/scopeProviders/ScopeInfoProvider.ts
```diff
@@ -176,6 +176,7 @@ function isLanguageSpecific(scopeType: ScopeType): boolean {
       return false;
 
     case "oneOf":
+    case "contiguous":
```
pokey, line 179: can't you just call this function on the constituent scope?

pokey (commented):
seems like a good implementation approach. left some comments.  also, please add some tests with "next" / "previous", and be sure to test boundaries on edges of scopes and between constituent scopes

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
AndreasArvidsson: Because I needed the continuous range for the domain before we even create the target

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
```diff
@@ -0,0 +1,191 @@
...
+  leadingTarget: Target,
+  trailingTarget: Target,
+): TargetScope {
+  if (leadingTarget.contentRange.isRangeEqual(trailingTarget.contentRange)) {
+    return {
+      editor: leadingTarget.editor,
+      domain: leadingTarget.contentRange,
+      getTargets: () => [leadingTarget],
+    };
+  }
```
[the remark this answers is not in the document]
AndreasArvidsson: Should we just make a union of the domains as well you mean?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
AndreasArvidsson: That is on purpose. I don't think that contiguous name should eat multiple functions just because the entire function is the domain for the name.

packages/cursorless-engine/src/scopeProviders/ScopeInfoProvider.ts
AndreasArvidsson: fixed

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
pokey: Not sure I follow. Why can't you use union for domain and `createContinuousRangeTarget` for the target?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
pokey: Ok, let's leave it for now; if we find we want that behaviour we can change that in the future

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
```diff
@@ -0,0 +1,209 @@
...
+}
+
+function* generateTargetRangesInDirection(
+  scopeHandler: ScopeHandler,
+  editor: TextEditor,
+  position: Position,
+  direction: Direction,
+): Iterable<[Target, Target]> {
+  const isForward = direction === "forward";
+  let first, last: Target | undefined;
```
pokey, line 148: are these in iteration order or document order? We really need to have solid conventions when we could be iterating backwards

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
AndreasArvidsson: I meant in this case where content range and domain is this same I don't want to have that calculation in two separate places. If we do a separate calculation for domain its no problem.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
AndreasArvidsson: Absolutely!

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
AndreasArvidsson: First and last is in document order. The pair is yielded in iteration order.

pokey: Ok I'd call them `proximal` and `distal` then. That's the convention I've used elsewhere. "first" and "last" may be expected to correspond to the modifiers of those names, which are document order

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
AndreasArvidsson: I don't mean where content range and domain is the same between two scopes. I mean the resulting content range and domain for the new scope is the same. We're lazily constructing the targets in a callback and I needed the content range to even create the scope in the first place

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
AndreasArvidsson: Switched to proximal and distal

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    const targetsBackward = getTargetsInDirection(
+      this.scopeHandler,
+      editor,
+      position,
+      "backward",
+    );
+
+    for (const targets of targetsBackward) {
+      if (targetsForward != null && isAdjacent(targets[1], targetsForward[0])) {
```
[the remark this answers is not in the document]
AndreasArvidsson: done

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
AndreasArvidsson: Anyhow I now use `createContinuousRangeTarget `

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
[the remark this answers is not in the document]
AndreasArvidsson: done

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
```diff
@@ -0,0 +1,169 @@
...
+  return {
+    editor: scope1.editor,
+    domain: scope1.domain.union(scope2.domain),
+    getTargets: (isReversed) => {
+      return constructScopeRangeTarget(isReversed, scope1, scope2);
+    },
+  };
+}
+
+function* generateTargetRangesInDirection(
```
pokey, line 94: I'm not convinced this function is correct. Eg corner cases like single scope not adjacent to others that's last in the file. To me it looks like if you've already yielded anything, then the final yield after the loop body won't fire.

Some proper unit tests for `ContiguousScopeHandler` would help. See https://github.com/cursorless-dev/cursorless/blob/main/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BaseScopeHandler.test.ts if you're not sure how to do unit tests for this kinda thing. There's a limit to how far you can get using existing scopes

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
```diff
@@ -0,0 +1,169 @@
...
+
+function isAdjacent(scope1: TargetScope, scope2: TargetScope): boolean {
+  if (scope1.domain.isRangeEqual(scope2.domain)) {
+    return true;
+  }
+
+  const [startTarget, endTarget] = getTargetsInDocumentOrder(
+    scope1.getTargets(false)[0],
+    scope2.getTargets(false)[0],
+  );
```
pokey, line 135: do we want to throw an error if there's more than one target? Should we have a utility function `function ensureSingleTarget(scope: Scope): Target` that checks there's only one target and returns it? That worked out well for `ensureSingleEditor` I think

packages/cursorless-engine/src/spokenForms/defaultSpokenFormMapCore.ts
```diff
@@ -130,6 +130,7 @@ export const defaultSpokenFormMapCore: DefaultSpokenFormMapDefinition = {
     next: "next",
     forward: "forward",
     backward: "backward",
+    contiguousScope: "fat",
```
pokey, line 133: this should prob go in `complexScopeTypes` with `glyph`, no?

pokey (commented):
Well this code simplified nicely 😄. Left a couple more comments

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
pokey: Actually might make sense to hold off on unit tests for the following reasons:

1. Your code actually looks at document text, unlike the test I linked, so might be harder to mock that properly
2. On second thought, I think your code is probably correct
3. I'm actually not convinced of the utility of this contiguous scope thing 😅. The only example I saw that was actually useful was "comment", and tbh I'd be tempted to just automatically merge adjacent single-line comments, as we could target a single one using "line". And I worry users will try "fat arg" to get something like https://github.com/cursorless-dev/cursorless/issues/473, and then be surprised when it breaks if there's a comment between args. So maybe we ship this with slightly fewer tests, maybe marked private? idk

pokey: Update from discussion today:

- We agreed "comment" is the only use case we can come up with, and we probably just want that to be default behaviour for line comments
- If @josharian wants to fight for supporting "fat" for arbitrary scopes, then we might keep it; otherwise it dies
- We potentially could reuse this code to make contiguous be default behaviour for comment

josharian: > If @josharian wants to fight for supporting "fat" for arbitrary scopes, then we might keep it; otherwise it diesau

looking at these test cases, i think it should die

pokey: Sounds good. Any opinion on making single-line comments behave this way by default? (Ie auto expand to all comments that aren't separated by an empty line)

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
AndreasArvidsson: Reused our existing and shore single target. They should really never trigger with our current implementation so am fine with using an error message that might not be perfect.

AndreasArvidsson:
@pokey @josharian The current implementation now only affects the command scope type and there is no spoken form for it.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
```diff
@@ -0,0 +1,160 @@
...
+    yield { proximal, distal };
+  }
+}
+
+function isAdjacent(scope1: TargetScope, scope2: TargetScope): boolean {
+  if (scope1.domain.isRangeEqual(scope2.domain)) {
+    return true;
+  }
+
+  const [startTarget, endTarget] = getTargetsInDocumentOrder(
```
AndreasArvidsson, line 123: This file is now general enough to be used with most scopes. If we wanted to we could remove a bunch of code here and rely on the fact that we only use it for comments. I could go either way.

AndreasArvidsson:
Yes unless we make this more comment specific and check for single line ranges. Ideally we would have single and multi-line as separate scopes and only use this for single line.

Or add a new predicate for contiguous scope. That is probably the best solution?

pokey:
hmm so the predicate would just set a flag on the capture. I think that sounds reasonable?

AndreasArvidsson:
Yes exactly.

AndreasArvidsson:
There is no problem really with this implementation except for the fact that in javascript there is no Tree sitter difference between block and line comments. I guess that contiguous predicate could take a boolean that requires single line ranges?

packages/cursorless-engine/src/languages/LanguageDefinition.ts
```diff
@@ -141,3 +151,15 @@ function readQueryFileAndImports(languageQueryPath: string) {
...
+ * Returns true if the given scope type should use a contiguous scope handler.
+ */
+function useContiguousScopeHandler(scopeType: ScopeType): boolean {
+  switch (scopeType.type) {
+    case "comment":
+      return true;
+    default:
+      return false;
+  }
+}
```
pokey, line 165: I thought you were using a query predicate for this now

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
```diff
@@ -185,6 +185,23 @@ class AllowMultiple extends QueryPredicateOperator<AllowMultiple> {
   }
 }
 
+/** Indicates that it's okay for this scope to extend contiguously it through its siblings. */
+class Contiguous extends QueryPredicateOperator<Contiguous> {
+  name = "contiguous!" as const;
+  schema = z.union([z.tuple([q.node]), z.tuple([q.node, q.string])]);
+
+  run(nodeInfo: MutableQueryCapture, pattern?: string) {
```
pokey, line 193: Yikes. This is a slippery slope. This operator merges to completely orthogonal concepts. I fear we're sleepwalking into the same problem we had with our old typescript-based matchers, where we had 5 custom functions per language.

Maybe we could add a predicate that rejects a pattern if it doesn't match a regex, and then define the pattern twice: once with positive match, one negative?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/scope.types.ts
```diff
@@ -32,6 +32,9 @@ export interface TargetScope {
    */
   readonly domain: Range;
 
+  /** Whether this scope could expand contiguously to its siblings. */
+  readonly contiguous?: boolean;
+
```
pokey, line 37: huh.  A bit strange to have this on the scope handler 🤔

pokey (commented):
left a few comments

packages/cursorless-engine/src/languages/LanguageDefinition.ts
AndreasArvidsson: We use both. I don't know how to check it at this stage. When we actually fetch the matched scope it's to late to inject this handler. Not without a lot of rewrites at least.

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
AndreasArvidsson: This is still not language specific. Probably more languages that dont separate block comments in tree sitter.

That could also work, but is that really easier to read and maintain?

packages/cursorless-engine/src/languages/LanguageDefinition.ts
pokey: Sorry I'm not following how you would use both. When do you use one vs the other?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/scope.types.ts
AndreasArvidsson: It's on the scope and not the handler itself.

packages/cursorless-engine/src/languages/LanguageDefinition.ts
AndreasArvidsson: This code decides to use the handler. The the scope.contiguous is used to determine when to merge two scopes. That way we can ignore block comments.
https://github.com/cursorless-dev/cursorless/blob/b6ac60f708834daa151b24cf4d3f86f7728c4c8c/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts#L119

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
AndreasArvidsson: Because the pattern matches a single line comment? look how I use it in js.

Lets.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/scope.types.ts
pokey: yeah I guess you're right 🤔

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
pokey: https://github.com/cursorless-dev/cursorless/labels/to%20discuss

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ContiguousScopeHandler.ts
pokey: Update from meet-up:

```scm

(
  (comment) @comment
  (#match? @comment "^//")
  (#contiguous! @comment)
)
(comment) @comment
```

we both don't love this option or the option in this PR. WE prefer something like 

```scm

(
  (comment) @comment
  ($if
    (#match? @comment "^//")
    (#contiguous! @comment)
  )
)
```

but that's a bit out of scope here 😅

packages/cursorless-engine/src/languages/TreeSitterQuery/queryPredicateOperators.ts
pokey: Update from meet-up:

- Remove `pattern` arg from this predicate
- use the built-in tree-sitter [`match` predicate](https://github.com/tree-sitter/tree-sitter/blob/master/lib/binding_web/binding.js#L906), and duplicate the pattern, once with `contiguous` and once without

AndreasArvidsson: done!

PR #2102, opened 2023-12-07
Added more scope support facets

AndreasArvidsson:
This pull request has these goals:
1. To flesh out the available facets and their definitions
2. To provide one example scope fixture of each facet
3. Update the domain of for each loops name/value/type
4. Update domain of iteration scope for formal arguments

Once the above is done I'm going to do the bulk fixture recorder we talked about to fill in the other languages.

## Checklist


queries/python.scm
```diff
@@ -344,14 +343,6 @@
   (#not-parent-type? @_.removal case_clause)
 ) @_.domain
 
-;;!! for name in value:
```
AndreasArvidsson, line 347: The scope fixtures showed me that we had two different definitions for this same pattern. Really shows why we need them :)

queries/javascript.core.scm
```diff
@@ -481,24 +480,33 @@
...
-[
-  (call_expression)
-  (new_expression)
-] @functionCall
+(
+  [
+    (call_expression)
+    (new_expression)
+  ] @functionCall @_.domain.start
+  ";"? @_.domain.end
```
AndreasArvidsson, line 488: Semi was missing from domain

queries/javascript.scm
```diff
@@ -36,7 +36,8 @@
   ;;!  ^^^-------
   (field_definition
     property: (_) @name
-  ) @name.domain.start
+    value: (_)? @value
```
AndreasArvidsson, line 39: Value scope was missing from fields in javascript

queries/javascript.jsx.scm
```diff
@@ -89,7 +89,8 @@
 ;;!! <aaa bbb="ccc" />
 ;;!       ^^^
 (jsx_attribute
-  (property_identifier) @collectionKey
+  (property_identifier) @collectionKey @_.trailing.start.endOf
```
AndreasArvidsson, line 92: Was missing trailing delimiter

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
```diff
@@ -66,6 +113,7 @@ export interface ScopeSupportFacetInfo {
 
 export enum ScopeSupportFacetLevel {
   supported,
+  supportedLegacy,
```
pokey, line 116: were you planning to do something with this field?

packages/common/src/scopeSupportFacets/typescript.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+
+import {
+  LanguageScopeSupportFacetMap,
+  ScopeSupportFacetLevel,
+} from "./scopeSupportFacets.types";
+
+const { supported } = ScopeSupportFacetLevel;
+
+export const typescriptScopeSupport: LanguageScopeSupportFacetMap = {
```
pokey, line 10: This feels like a lie to say we only support these 5 things in TS. We should create a way to "import" support from another language, including the tests, the way we do for our `.scm` files. Shall I file an issue for that?

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/functionCall.constructor.scope
```diff
@@ -0,0 +1,13 @@
...
+---
+
+[Content] =
+[Removal] = 0:0-0:14
+0| new Foo.bar(0);
+  >--------------<
+
+[Domain] = 0:0-0:15
+0| new Foo.bar(0);
+  >---------------<
```
pokey, line 11: This looks wrong

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/functionCall.scope
```diff
@@ -0,0 +1,13 @@
...
+---
+
+[Content] =
+[Removal] = 0:0-0:10
+0| foo.bar(0);
+  >----------<
+
+[Domain] = 0:0-0:11
+0| foo.bar(0);
+  >-----------<
```
pokey, line 11: and this too

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/functionCallee.constructor.scope
```diff
@@ -0,0 +1,13 @@
...
+---
+
+[Content] =
+[Removal] = 0:0-0:11
+0| new Foo.bar(0);
+  >-----------<
+
+[Domain] = 0:0-0:15
+0| new Foo.bar(0);
+  >---------------<
```
pokey, line 11: also wrong

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/functionCallee.scope
```diff
@@ -0,0 +1,13 @@
...
+---
+
+[Content] =
+[Removal] = 0:0-0:7
+0| foo.bar(0);
+  >-------<
+
+[Domain] = 0:0-0:11
+0| foo.bar(0);
+  >-----------<
```
pokey, line 11: wrong too

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/name.field.scope
```diff
@@ -0,0 +1,89 @@
...
+  value = 0;
+}
+---
+
+[#1 Content] = 0:6-0:13
+0| class MyClass {
+        >-------<
+1|   value = 0;
+
+2| }
```
pokey, line 11: This one is another good argument for fine-grained scopes

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/value.field.scope
```diff
@@ -0,0 +1,38 @@
...
+1|   value = 0;
+            >-<
+2| }
+
+
+[Removal] = 1:9-1:11
+0| class MyClass {
+
+1|   value = 0;
+           >--<
```
pokey, line 18: This looks wrong

queries/javascript.core.scm
pokey: Oh, right this was the bugs I pointed out in my earlier comments 😅. Why would semi be part of the domain? Semi is part of the containing statement; a call is an expression

queries/javascript.core.scm
```diff
@@ -481,24 +480,33 @@
...
-;;!  -----
-(call_expression
-  function: (_) @functionCallee
-) @_.domain
+;;!  ------
+(
+  (call_expression
+    function: (_) @functionCallee
+  ) @_.domain.start
+  ";"? @_.domain.end
```
pokey, line 498: as above

queries/javascript.core.scm
```diff
@@ -481,24 +480,33 @@
...
-;;!  ---------
-(new_expression
-  (arguments) @functionCallee.end.startOf
-) @functionCallee.start.startOf @_.domain
+;;!  ----------
+(
+  (new_expression
+    (arguments) @functionCallee.end.startOf
+  ) @functionCallee.start.startOf @_.domain.start
+  ";"? @_.domain.end
```
pokey, line 508: and here too

pokey (commented):
This is actually pretty comprehensive! I added a few more. In a follow-up, we should

- [ ] add more iteration scopes for name / value / type
- [ ] add `named.method`
- [ ] add some more `"statement"` scope facets

packages/common/src/scopeSupportFacets/scopeSupportFacetInfos.ts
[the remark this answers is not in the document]
AndreasArvidsson: Agreed

packages/common/src/scopeSupportFacets/scopeSupportFacetInfos.ts
```diff
@@ -7,42 +9,274 @@ export const scopeSupportFacetInfos: Record<
...
+  list: {
+    description: "A list/array",
+    scopeType: "list",
+  },
+  map: {
+    description: "A map/dictionary",
+    scopeType: "map",
+  },
+  statement: {
+    description: "A statement",
```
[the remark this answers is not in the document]
AndreasArvidsson: Agreed

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/name.field.scope
AndreasArvidsson: Definitely

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/value.field.scope
AndreasArvidsson: updated

packages/common/src/scopeSupportFacets/scopeSupportFacetInfos.ts
```diff
@@ -7,42 +9,275 @@ export const scopeSupportFacetInfos: Record<
...
+  },
+  anonymousFunction: {
+    description:
+      "An anonymous function, eg a lambda function, an arrow function, etc",
+    scopeType: "anonymousFunction",
+  },
+  "anonymousFunction.lambda": {
+    description: "A lambda function",
+    scopeType: "anonymousFunction",
+  },
```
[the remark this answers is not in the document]
AndreasArvidsson: I believe lambda is the correct technical description and exists in many languages. That javascript calls them arrow functions is besides the point. 
Personally I feel that lambdas is something quite different from just an anonymous function. I want facets for every single nuance if possible.

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
AndreasArvidsson: Yes it was used an should be used for every language that still has legacy support. Then we can generate documentation to that fact.

packages/common/src/scopeSupportFacets/typescript.ts
AndreasArvidsson: Please do

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
pokey: From what I can tell you're not using it anywhere. Prob worth using it at least once if you're going to introduce it? But maybe I just missed it

AndreasArvidsson: It was used for typescript before we migrated all the scopes. It's now used by python as an example.

queries/javascript.core.scm
pokey: I did some testing on this, and here's the behaviour this PR introduces

image

Note how it only extends the domain to include the semicolon for

- the outermost call, and
- only if the entire statement is a function call.

In particular, when it's the value of an assignment, or part of a bigger expression, it doesn't include the semicolon

There is a certain logic to that, but might be surprising? Also might be surprising if you say "previous call" and that doesn't work? I don't have a strong feeling one way or another. I also might consider going a step further and including the semicolon on the content range itself

If we do keep this behaviour, let's capture the semantics from that screenshot in a scope fixture. I tried putting all four of those lines in a scope fixture, and man we really aren't optimized for more than one scope per fixture 😅. I wonder if we should

1. Only keep the empty marker lines if they're inside a scope, or
2. Not bother to show lines outside of the scope; that's kinda what the line numbers are for

cc/ @josharian on both topics mentioned here 😊

AndreasArvidsson: For now I would be fine with to just reverting the domains so we can actually merge this. Changing the domain of call seems totally separate from this pull request.

Dropping lines outside of the scope is totally be fine for me

I still haven't heard of any way to actually determine token removal. And I I'm not sure if we not losing information if we do that.

packages/common/src/scopeSupportFacets/html.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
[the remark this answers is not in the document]
AndreasArvidsson: Personally I don't know if I even like naming convention for keys. I think you should be able to use whatever fits your needs.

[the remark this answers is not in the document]
pokey: Yeah that's fair. I wonder if we should grab one of the configs from https://typescript-eslint.io/rules/naming-convention/#ignore-properties-that-require-quotes. Eg the third one looks promising; make exceptions if key has certain chars in it, eg `[- .]`

queries/javascript.core.scm
AndreasArvidsson: Call domain reverted

queries/javascript.core.scm
AndreasArvidsson: fixed

queries/javascript.core.scm
AndreasArvidsson: fixed

pokey (commented):
Whoops looks like I forgot to submit these comments

queries/javascript.core.scm
josharian: > I would be supportive of that as a first step. @josharian any objections? If not, I'll file an issue to trackmr

Seems worth trying. If it creates other issues, we can always adjust from there.

packages/common/src/scopeSupportFacets/html.ts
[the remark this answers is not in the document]
AndreasArvidsson: Issue created
https://github.com/cursorless-dev/cursorless/issues/2132

packages/common/src/scopeSupportFacets/scopeSupportFacetInfos.ts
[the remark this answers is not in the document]
AndreasArvidsson: Lambda facet removed

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/functionCallee.scope
AndreasArvidsson: updated

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/functionCallee.constructor.scope
AndreasArvidsson: updated

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/functionCall.scope
AndreasArvidsson: updated

queries/javascript.core.scm
pokey: Filed

- https://github.com/cursorless-dev/cursorless/issues/2139
- https://github.com/cursorless-dev/cursorless/issues/2140

pokey (approved):
🎉

PR #2110, opened 2023-12-08
Relative exclusive scope should skip containing

AndreasArvidsson:
This frequently bites me. `"two states"` refers to the if statement and the second print statement, but `"next state"` refers to the first print statement. This change makes it so it skips containing and refers to the second one instead.

```ts
    if ( | true) {
        console.log(1)
    }

    console.log(2)
```
## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/RelativeExclusiveScopeStage.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/RelativeExclusiveScopeStage.ts
@@ -3,6 +3,7 @@ import type { Target } from "../../typings/target.types";
 import { ModifierStageFactory } from "../ModifierStageFactory";
 import type { ModifierStage } from "../PipelineStages.types";
 import { constructScopeRangeTarget } from "./constructScopeRangeTarget";
+import { getPreferredScopeTouchingPosition } from "./getPreferredScopeTouchingPosition";
 import { runLegacy } from "./relativeScopeLegacy";
 import { ScopeHandlerFactory } from "./scopeHandlers/ScopeHandlerFactory";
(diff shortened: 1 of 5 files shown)


pokey:
I think both today's behaviour and your proposed behaviour are useful. And I think both can be equally surprising. It would be quite surprising to pop out of a "curly" you can't see when you say "next curly"

Why don't we support both as separate modifiers? You can map it to "next", and I'll map it to "skip", eg "skip funk", "skip second funk", etc

If we find the other is more often useful, and less often surprising, we can change default spoken form

Make sense?

pokey:
> Should we make a brand new modifier or just add an additional parameter to the modifier? We could also make it a private setting and try it out?

yes I'd make it private. Not sure bout new modifier vs param. One thing to keep in mind: the impl is very very close to `RelativeInclusive`, so I would be inclined to fold it in there rather than trying to handle both in `RelativeExclusive`. Or make a separate class. But I don't think it belongs in `RelativeExclusive`

AndreasArvidsson:
> > Should we make a brand new modifier or just add an additional parameter to the modifier? We could also make it a private setting and try it out?
> 
> yes I'd make it private. Not sure bout new modifier vs param. One thing to keep in mind: the impl is very very close to `RelativeInclusive`, so I would be inclined to fold it in there rather than trying to handle both in `RelativeExclusive`. Or make a separate class. But I don't think it belongs in `RelativeExclusive`

Implementation wise it might be that, but its still most definitely relative exclusive since we have offset 1?

AndreasArvidsson:
We can always make utility function for that part

AndreasArvidsson:
Closed in favor of https://github.com/cursorless-dev/cursorless/pull/2133

PR #2113, opened 2023-12-09
Terser leading/trailing syntax in scm

### LINKED ISSUE #2108: Terser leading/trailing syntax in scm
It's unnecessary having to specify that the leading and trailing ranges should terminate at the content range. We could just specify their distal position and the range could be calculated for us.

from:
```scm
  (flow_node)? @collectionItem.leading.start.endOf
  .
  (flow_node) @collectionItem @collectionItem.leading.end.startOf @collectionItem.trailing.end.endOf
  .
  (flow_node)? @collectionItem.trailing.end.startOf
```

to:
```scm
  (flow_node)? @collectionItem.leading.endOf
  .
  (flow_node) @collectionItem
  .
  (flow_node)? @collectionItem.trailing.startOf
```

@pokey @josharian Any thoughts?

AndreasArvidsson:
Fixes #2108

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler/TreeSitterScopeHandler.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler/TreeSitterScopeHandler.ts
@@ -1,6 +1,4 @@
 import {
-  Position,
-  Range,
   SimpleScopeType,
   TextEditor,
 } from "@cursorless/common";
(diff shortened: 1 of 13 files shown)


pokey (approved):
Wow yeah way more readable

On an extremely minor note, I have very mixed feelings about the convention you're starting to introduce of functions that accept `undefined` on one of their params and return `undefined` in that case. Feels like it's conflating concerns. I tried rewriting it to use null-ish chaining and I think it worked well (f5dc4281f295a47961db47e6f867723ee939d6e7). I don't feel super strongly, so if you want to roll back that's fine, was just curious if it could be done with nullish chaining. If you do roll back, let's come up with a better name for that function. I think the convention could work if we come up with a solid naming convention to go along with it

PR #2115, opened 2023-12-10
Added command history

AndreasArvidsson:
setting:
```json
"cursorless.commandHistory": true
```

Monthly log file: `cursorlessCommandHistory_2023-12.jsonl`
```
{"date":"2023-12-10","cursorlessVersion":"0.28.0-40a6fee4","command":{"version":6,"action":{"name":"setSelection","target":{"type":"primitive","modifiers":[{"type":"containingScope","scopeType":{"type":"line"}}]}},"usePrePhraseSnapshot":true}}
```

Voice command: `"Cursorless analyze history"` opens a new untitled document:
```
[2023-12]

Total commands: 24

Actions (7):
  setSelection: 18
  clearAndSetSelection: 1
  remove: 1
  wrapWithPairedDelimiter: 1
  getText: 1
  replace: 1
  copyToClipboard: 1

Modifiers (2):
  containingScope: 20
  relativeScope: 2

Scope types (4):
  line: 18
  surroundingPair: 2
  comment: 1
  document: 1

```

## Checklist


packages/cursorless-engine/src/runCommand.ts
```diff
@@ -35,6 +40,7 @@ export async function runCommand(
   languageDefinitions: LanguageDefinitions,
   rangeUpdater: RangeUpdater,
   commandRunnerDecorators: CommandRunnerDecorator[],
+  commandHistory: CommandHistory,
```
pokey, line 43: Why can't this be a command decorator? (See the commandDecorators arg in the previous line (line 42))

pokey (commented):
Haven't looked in depth just quick comment

packages/cursorless-vscode/src/ide/vscode/VscodeCommandHistory.ts
```diff
@@ -0,0 +1,94 @@
...
+  private readonly cursorlessVersion: string;
+  private disposable: Disposable;
+  private active: boolean = false;
+
+  constructor(ide: VscodeIDE, fileSystem: FileSystem) {
+    this.evaluateSetting = this.evaluateSetting.bind(this);
+    this.cursorlessVersion = ide.cursorlessVersion;
+    this.dirPath = path.join(fileSystem.cursorlessDir, dirName);
+
+    mkdirSync(this.dirPath, { recursive: true });
```
bra1nDump, line 31: I would move the mkdir to the append method for these reasons:
- Ensures the directory is created right before writing to it. For example with the current mechanism if I delete the directory mid session, the append will throw I think
- No need to wait for directory to be created - currently blocking activate path

packages/cursorless-vscode/src/ide/vscode/VscodeCommandHistory.ts
```diff
@@ -0,0 +1,94 @@
...
+      if (event.affectsConfiguration(settingFullName)) {
+        this.evaluateSetting();
+      }
+    });
+  }
+
+  async append(command: CommandComplete): Promise<void> {
+    if (!this.active) {
+      return;
+    }
```
bra1nDump, line 45: We can use this to bail out instead of having DisabledCommandHistory. Seems cleaner to have a single path.

The current code will not handle enabling command history in the existing session - because we already initialized to the DisabledCommandHistory which has no workspace config watcher

AndreasArvidsson: Yes it does. Disabled command history is only used in testing. `vscode.workspace.onDidChangeConfiguration` watches for setting changes.

packages/cursorless-vscode/src/ide/vscode/VscodeCommandHistory.ts
AndreasArvidsson: The director is owned by Cursorless and the uses should not remove it.

packages/cursorless-engine/src/runCommand.ts
AndreasArvidsson: No reason I just wanted a stricter typing. Probably unnecessary and I have re implemented it as a command runner decorator instead.

AndreasArvidsson:
@pokey Updated and ready ;)

packages/cursorless-vscode/package.json
```diff
@@ -279,6 +279,11 @@
           "default": true,
           "description": "Whether to show decorations on vscode start."
         },
+        "cursorless.commandHistory": {
```
pokey, line 282: Shall we use the `order` field to push this one up a little bit? Or wait until we're more confident in it?

packages/cursorless-vscode/src/CommandHistory.ts
```diff
@@ -0,0 +1,147 @@
...
+  wrapCommandRunner(
+    readableHatMap: ReadOnlyHatMap,
+    runner: CommandRunner,
+  ): CommandRunner {
+    if (!this.isActive) {
+      return runner;
+    }
+
+    return {
+      run: async (commandComplete: CommandComplete) => {
```
pokey, line 54: Should we capture the errors? Might be useful to know how often people are running into inference issues, etc. I think just the error name would suffice. Worth quick look at how `TestCaseRecorder` wraps command in case helpful https://github.com/cursorless-dev/cursorless/blob/40a6fee4c4eec9b37cfb1ee7b4bf5ff78df59ea2/packages/cursorless-engine/src/testCaseRecorder/TestCaseRecorder.ts#L458-L483

packages/cursorless-vscode/src/CommandHistory.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    const historyItem: CommandHistoryItem = {
+      date: getDayDate(date),
+      cursorlessVersion: this.cursorlessVersion,
+      command: sanitizeCommand(command),
+    };
+    const data = JSON.stringify(historyItem) + "\n";
+
+    await fs.mkdir(this.dirPath, { recursive: true });
+    await fs.appendFile(file, data, "utf8");
```
pokey, line 75: Does this open file in exclusive mode? If not, would be worth doing so

packages/cursorless-vscode/src/CommandHistory.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  private async appendToLog(command: CommandComplete): Promise<void> {
+    const date = new Date();
+    const fileName = `${filePrefix}_${getMonthDate(date)}.jsonl`;
+    const file = path.join(this.dirPath, fileName);
+
+    const historyItem: CommandHistoryItem = {
+      date: getDayDate(date),
+      cursorlessVersion: this.cursorlessVersion,
+      command: sanitizeCommand(command),
```
pokey, line 70: Probably not worth changing, but `sanitizeCommand` and `sanitizeAction` would prob be cleaner using immer, like we do with canonicalization. You'd basically just change those functions to mutate the commands directly, calling them eg `sanitizeCommandInPlace`, then just use `produce` here like we do in https://github.com/cursorless-dev/cursorless/blob/40a6fee4c4eec9b37cfb1ee7b4bf5ff78df59ea2/packages/cursorless-engine/src/core/commandVersionUpgrades/canonicalizeAndValidateCommand.ts#L40-L46

packages/cursorless-vscode/src/CommandHistory.ts
```diff
@@ -0,0 +1,147 @@
...
+function sanitizeCommand(command: CommandComplete): CommandComplete {
+  const { spokenForm, action, ...rest } = command;
+  return {
+    ...rest,
+    action: sanitizeAction(action),
+  };
+}
+
+function sanitizeAction(action: ActionDescriptor): ActionDescriptor {
+  switch (action.name) {
```
pokey, line 99: I know it's a bit annoying, but I would be tempted to force us to enumerate all actions here. Otherwise we may inadvertently start logging sensitive information in the future

packages/cursorless-vscode/src/CommandHistory.ts
```diff
@@ -0,0 +1,147 @@
...
+// Remove spoken form and sanitize action
+function sanitizeCommand(command: CommandComplete): CommandComplete {
+  const { spokenForm, action, ...rest } = command;
+  return {
+    ...rest,
+    action: sanitizeAction(action),
+  };
+}
+
+function sanitizeAction(action: ActionDescriptor): ActionDescriptor {
```
pokey, line 98: Do we want to normalize the letter in decorated marks? Not sure how much you can glean from that, but feels like the right thing to do if people are to be comfortable sharing these things more widely. I believe we have some utilities for recursively descending commands so shouldn't be too hard I don't think

If we do, I'd argue for leaving punctuation chars, but changing all non-punctuation to just `x`

packages/cursorless-vscode/src/CommandHistory.ts
```diff
@@ -0,0 +1,147 @@
...
+import * as fs from "node:fs/promises";
+import * as path from "node:path";
+import * as vscode from "vscode";
+
+const filePrefix = "cursorlessCommandHistory";
+const settingSection = "cursorless";
+const settingName = "commandHistory";
+const settingFullName = `${settingSection}.${settingName}`;
+
+export class CommandHistory implements CommandRunnerDecorator {
```
pokey, line 22: jsdoc

pokey (commented):
Looks good! Left a few comments. Also

- [x] I think it would be nice to have at least one simple sunny-day test. You should be able to steal most of what you need from https://github.com/cursorless-dev/cursorless/blob/main/packages/cursorless-vscode-e2e/src/suite/testCaseRecorder.vscode.test.ts

packages/cursorless-vscode/package.json
AndreasArvidsson: Sounds reasonable. What would be a good order do you think?

packages/cursorless-vscode/src/CommandHistory.ts
AndreasArvidsson: I was thinking about that, but I also think that which letters you use is an interesting data point?

packages/cursorless-vscode/src/CommandHistory.ts
AndreasArvidsson: I don't know. Personally I find that pattern quite hard to read.

packages/cursorless-vscode/src/CommandHistory.ts
pokey: @josharian what did you mean when you said "open the file in exclusive mode"? From what I understand, exclusive just means fail if the file already exists, which I don't think is what we want here

packages/cursorless-vscode/src/CommandHistory.ts
AndreasArvidsson: updated

packages/cursorless-vscode/src/CommandHistory.ts
pokey: I find deeply nested spreads hard to read. What don't you like about the immer pattern?

Just to be clear, let's not block this PR on this discussion, but I'm interested to have it anyway

packages/cursorless-vscode/src/CommandHistory.ts
pokey: I could go either way. @josharian any thoughts?

packages/cursorless-vscode/src/CommandHistory.ts
AndreasArvidsson: Not quite sure what to write in this case but I added a small comment

packages/cursorless-vscode/src/CommandHistory.ts
pokey: Yeah I think generally we should try to avoid them if we can, but in this case I think it's justified

AndreasArvidsson:
> Looks good! Left a few comments. Also
> 
>     * [ ]  I think it would be nice to have at least one simple sunny-day test. You should be able to steal most of what you need from https://github.com/cursorless-dev/cursorless/blob/main/packages/cursorless-vscode-e2e/src/suite/testCaseRecorder.vscode.test.ts

I can't find the mechanism for mocking the file system paths.

pokey:
> I can't find the mechanism for mocking the file system paths.

The cursorless dir is already automatically set to a temp dir https://github.com/cursorless-dev/cursorless/blob/40a6fee4c4eec9b37cfb1ee7b4bf5ff78df59ea2/packages/cursorless-vscode/src/extension.ts#L175-L177 and then you can access it like this https://github.com/cursorless-dev/cursorless/blob/40a6fee4c4eec9b37cfb1ee7b4bf5ff78df59ea2/packages/cursorless-vscode-e2e/src/suite/scopeProvider/runCustomSpokenFormScopeInfoTest.ts#L12-L13

packages/cursorless-vscode/src/CommandHistory.ts
josharian: sorry about the confusion. I always mess this up. ideally we'd use something like a lockfile to coordinate access, e.g. https://www.npmjs.com/package/proper-lockfile (found via google search, no personal experience). it's either that or hope, which isn't necessarily a terrible approach here.

packages/cursorless-vscode/src/CommandHistory.ts
josharian: i don't feel strongly.

packages/cursorless-vscode/src/CommandHistory.ts
AndreasArvidsson: That is just so the code should be as compact as possible. You can definitely make it more verbose if you prefer that pattern. 

I think that immer adds a layer of abstraction that is quite esoteric. Instead of just returning or assigning a value all of the sudden you have a callback with an argument and no return value. Instead of constructing a new object without a field you should now do `delete draft.spokenForm`. In my mind immer solves a problem that we don't really have here and goes against the kiss principle.

packages/cursorless-vscode/package.json
AndreasArvidsson: That setting seems completely broken to me. I have shapes and hat stability at the top
image

packages/cursorless-vscode/src/CommandHistory.ts
AndreasArvidsson: But we are not logging the result? Should we log the command itself and then optionally log errors?

AndreasArvidsson:
> > I can't find the mechanism for mocking the file system paths.
> 
> The cursorless dir is already automatically set to a temp dir
> 
> https://github.com/cursorless-dev/cursorless/blob/40a6fee4c4eec9b37cfb1ee7b4bf5ff78df59ea2/packages/cursorless-vscode/src/extension.ts#L175-L177
> and then you can access it like this
> 
> https://github.com/cursorless-dev/cursorless/blob/40a6fee4c4eec9b37cfb1ee7b4bf5ff78df59ea2/packages/cursorless-vscode-e2e/src/suite/scopeProvider/runCustomSpokenFormScopeInfoTest.ts#L12-L13

Ok but I do wonder how I can make command history be active only for this single test. I don't want it logging every single test in our recorded suite.

packages/cursorless-vscode/package.json
pokey: Try saying "cursorless settings"

pokey:
> Ok but I do wonder how I can make command history be active only for this single test. I don't want it logging every single test in our recorded suite.

You can mock the setting. See eg https://github.com/cursorless-dev/cursorless/blob/8f88564b880c246ba226981218722d300dabb965/packages/cursorless-vscode-e2e/src/suite/keyboard/basic.vscode.test.ts#L139-L155

You may need to trigger a setting refresh. Worth testing also that nothing is written if the setting is off. That seems important from privacy perspective

packages/cursorless-vscode/src/CommandHistory.ts
pokey: Not sure I follow. You could just add an optional error field to the top-level payload

packages/cursorless-vscode/src/CommandHistory.ts
pokey: Yeah honestly just "hope" might be fine, since I'm not sure I see how you'd end up with multiple windows writing at once

josharian: yeah. talon and the command file are both serving as de facto external locking mechanisms here. if we end up with corrupt logs, we can reconsider.

pokey: And even by keyboard, you won't be typing into multiple windows

josharian: one certainly hopes. people do weird things. (and win weird prizes when they do.)

packages/cursorless-vscode/src/CommandHistory.ts
AndreasArvidsson: I just have my mind set on logging before executing the command. I guess we could add it afterwards.
Updated

packages/cursorless-vscode/package.json
AndreasArvidsson: Probably first time I have used at command. Now we got the desired order, but that also means that unless you get to the settings in a particular way the order doses matter. Weird.

packages/cursorless-vscode/src/CommandHistory.ts
AndreasArvidsson: Then I guess we're happy?

packages/cursorless-vscode/src/CommandHistory.ts
AndreasArvidsson: Then are we okay with leaving it as it is? I think it would be interesting to see which letters a user prefer. Or if there are letters/symbols that you don't use at all.

packages/cursorless-vscode/src/CommandHistory.ts
pokey: Yeah I think let's leave it and keep an eye out for corrupt json. Not seeing any better solution that's worth the effort

packages/cursorless-vscode/src/CommandHistory.ts
pokey: Still makes me nervous. Writing user's document text to file seems sketchy, even if it's a single character at a time. And I'm not sure what we'll glean from knowing which char it is. I would be inclined to let non-alphanumeric pass through untouched, and letters map to `x` and numbers map to `0`. I guess we could ask on slack to see how comfortable people are?

AndreasArvidsson: I don't count this as document text at all. The only thing we know is that the character exists. How many times and in which configurations we don't know. This is at maximum fingerprinting the user's language. I don't see a point to only keep non alphanumerics. We either remove everything or nothing is my vote.
Sounds reasonable to ask on slack.

packages/cursorless-vscode/src/CommandHistory.ts
pokey: Let's not block this PR on this discussion. I did a quick draft of what this would look like with immer in https://github.com/cursorless-dev/cursorless/pull/2124; let's discuss that one at our leisure. Not urgent at all

packages/cursorless-vscode/src/CommandHistory.ts
josharian: Given that it is opt-in and local-only, I think it’s probably OK to record the char.

pokey: Ok fine let's keep it

pokey:
I believe this is just waiting on tests and docs, right? Fwiw I reactivated the "tests" checkbox; curious why you disabled it in the first place?

AndreasArvidsson:
> I believe this is just waiting on tests and docs, right? Fwiw I reactivated the "tests" checkbox; curious why you disabled it in the first place?

Yes. Because I didn't think there was any way this could be tested.

AndreasArvidsson:
@pokey Tests added. Not sure where to put the documentation though?

AndreasArvidsson:
@pokey Added a small docs page

packages/cursorless-vscode/src/extension.ts
```diff
@@ -99,6 +100,8 @@ export async function activate(
     fileSystem,
   );
 
+  addCommandRunnerDecorator(new CommandHistory(context, vscodeApi, fileSystem));
```
pokey, line 103: How come this is VSCode-only? I know it's easier to put things in VSCode directory, but I think we should strive to put as much as we can in the engine

cursorless-talon/src/cursorless.py
```diff
@@ -15,3 +15,9 @@ def private_cursorless_show_settings_in_ide():
 
     def private_cursorless_show_sidebar():
         """Show Cursorless-specific settings in ide"""
+
+    def private_cursorless_analyze_command_history():
```
pokey, line 19: I think I'd call this `private_cursorless_show_statistics`. Also, not sure this needs an action, but I guess there's not real downside 🤷‍♂️

cursorless-talon/src/cursorless.talon
```diff
@@ -43,3 +43,6 @@ tag: user.cursorless
 
 bar {user.cursorless_homophone}:
     user.private_cursorless_show_sidebar()
+
+{user.cursorless_homophone} analyze history:
```
pokey, line 47: maybe
```suggestion
{user.cursorless_homophone} show stats:
```

packages/cursorless-engine/src/index.ts
```diff
@@ -6,3 +6,7 @@ export * from "./core/StoredTargets";
 export * from "./typings/TreeSitter";
 export * from "./cursorlessEngine";
 export * from "./api/CursorlessEngineApi";
+export * from "./CommandRunner";
+export * from "./util/getPartialTargetDescriptors";
+export * from "./util/getPrimitiveTargets";
+export * from "./util/getScopeType";
```
pokey, line 12: A bit of a smell to have to export all these; I really think as much of command history should be engine-side

packages/cursorless-vscode/src/CommandHistoryAnalyzer.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  getMonths() {
+    const months = Object.values(this.months);
+    months.sort((a, b) => a.month.localeCompare(b.month));
+    return months;
+  }
+}
+
+async function analyzeFilesAndReturnMonths(dir: string): Promise<Month[]> {
```
pokey, line 111: Why are all your abstractions in terms of months? Feels a bit arbitrary. The roll-up window should just be some knob we can tweak, rather than being deeply baked into the analysis code.

Shall we factor this analysis code into a separate PR so we can review that separately without blocking this one? I'd love to get the history part merged in

pokey (commented):
Made some tweaks and left a few comments; almost there

packages/cursorless-vscode/src/extension.ts
AndreasArvidsson: I had it in the engine originally, but since we changed it to a runner decorator there is no hard engine dependency anymore. It also needs a vscode config so the other option would be to migrate this to our own ide.configuration object?

cursorless-talon/src/cursorless.py
AndreasArvidsson: The setting is called command history and I wanted the command to reflect that they belong together? Without a spoken command people will not use this I'm afraid

packages/cursorless-vscode/src/CommandHistoryAnalyzer.ts
AndreasArvidsson: Just an arbitrary choice that I felt suited it. Easy to change if we want to.
There's no reason why we can't change the analysis code later?

packages/cursorless-engine/src/index.ts
AndreasArvidsson: I guess we could move the Cursorless version and the command history setting to the engine?

cursorless-talon/src/cursorless.talon
pokey: tho we might want to go object-verb? Eg "stats show" or "history stats"?

packages/cursorless-engine/src/index.ts
pokey: yeah I'd be tempted to just move it all to the engine. No reason to arbitrarily make it harder for other ides in the future

cursorless-talon/src/cursorless.py
AndreasArvidsson: Oh you mean we could just execute this rpc command directly? Yeah you're right we should probably just remove this action

packages/cursorless-vscode/src/CommandHistoryAnalyzer.ts
pokey: I guess it's very self-contained, but I just don't like landing code I haven't reviewed, and I'd love to get this thing in

cursorless-talon/src/cursorless.py
AndreasArvidsson: Unnecessary private action is now removed.

packages/cursorless-vscode/src/extension.ts
AndreasArvidsson: We can do that

packages/cursorless-vscode/src/CommandHistoryAnalyzer.ts
AndreasArvidsson: I've gonna have to take the time to move everything to the engine so you have time to look at this if you want to.

packages/cursorless-engine/src/index.ts
AndreasArvidsson: Agreed

pokey: Hmm. Let's put it there for now

cursorless-talon/src/cursorless.py
josharian: > Idk I find "show statistics" more intuitive. cc/ @josharian you have a leaning?af

My leaning is not to have a spoken command yet, while we mess around to find out what is useful, and then hope that once a handful of us have played a bit, it'll be clearer (a) what we want to show and thus (b) what a good spoken form for that is.

packages/cursorless-vscode/src/CommandHistoryAnalyzer.ts
pokey: Atomic PRs

pokey: Atomic PRs

pokey:
I think it would be good hygiene to include a command id and phrase id in each log entry.

Also, it would be good to test that our sanitisation actually works. I could see that breaking without us noticing 😅

AndreasArvidsson:
> I think it would be good hygiene to include a command id and phrase id in each log entry.
> 
> Also, it would be good to test that our sanitisation actually works. I could see that breaking without us noticing 😅

We don't really have the command id at this point in the pipeline, but I guess we could just use the constant with the knowledge that it will be correct?
What do you mean with a phrase id?

True sounds reasonable to test.

AndreasArvidsson:
> > We don't really have the command id at this point in the pipeline, but I guess we could just use the constant with the knowledge that it will be correct?
> 
> No I just mean generate one on the fly

Okay how would that id look and what would be the purpose? Why not just use the actual command id?

> > What do you mean with a phrase id?
> 
> Just generate an id, then use command server api to see if the phrase has changed, like we do for hat map, and if not reuse previous phrase id otherwise generate new one. Make sense?

Yes. If it's the same phrase use the same id otherwise generate new one. Using this we can group commands in the phrase.

pokey:
> Okay how would that id look and what would be the purpose? Why not just use the actual command id?

Just a uuid I guess? I thought you said we didn't have access to command id. The purpose is just hygiene. If we do transformations on the files, it's nice for each command to have an id so we can be sure it's the same. Just a data science best practice

> Yes. If it's the same phrase use the same id otherwise generate new one. Using this we can group commands in the phrase.

👍

AndreasArvidsson:
> > Okay how would that id look and what would be the purpose? Why not just use the actual command id?
> 
> Just a uuid I guess? I thought you said we didn't have access to command id. The purpose is just hygiene. If we do transformations on the files, it's nice for each command to have an id so we can be sure it's the same. Just a data science best practice

I don't really see how a uuid that gets generated for each command helps with anything? Or do you mean that we should just have a static one like a version number?
We don't have access to the command in the command runner itself, but we only have a single command to run Cursorless actions and that is it exported constant.
https://github.com/cursorless-dev/cursorless/blob/6718e84eb033e880980929098d4b21d2ff2613f2/packages/common/src/cursorlessCommandIds.ts#L1-L2

pokey:
> I don't really see how a uuid that gets generated for each command helps with anything?

It's just good practice. These IDs are very useful when doing postprocessing. But we can see if others feel strongly we should leave them out cc/ @phillco @josharian

packages/vscode-common/src/TestHelpers.ts
```diff
@@ -46,6 +46,7 @@ export interface TestHelpers {
   runIntegrationTests(): Promise<void>;
 
   cursorlessTalonStateJsonPath: string;
```
AndreasArvidsson, line 48: Should we return the file system instance instead of keep adding paths here?

AndreasArvidsson:
Made a few changes
* moved command history to the engine
* added Cursorless version string to the ide
* added phrase id to command history entry
* removed spoken command for analyzing command history

cursorless-talon/src/cursorless.py
AndreasArvidsson: Sounds good

pokey:
update from meet-up:

- add id for each payload

AndreasArvidsson:
id added

packages/vscode-common/src/TestHelpers.ts
pokey: Let's leave for now and revisit if it gets out of hand

pokey:
Removed analyzer; created https://github.com/cursorless-dev/cursorless/issues/2161 to track re-adding. Feel free to open PR with that patch once this one is in; lmk if you want help creating the patch @AndreasArvidsson

pokey (approved):
Ok I finished those last two steps, and tried installing locally to make sure it works as expected. Looking good! Feel free to merge if you're happy with my changes

Issue #1726, opened 2023-08-03
Improve snippet information in cheatsheet 


pokey:
- [ ] Add action entry for `snip <snippet> <destination>`
- [ ] Add action entry for `snip <snippet> <phrase>`
- [ ] Add action entry for `<wrapper_snippet> wrap <target>`
- [ ] Add box with list of wrapper snippets
- [ ] Add box with list of insertion snippets
- [ ] Add box with list of single-phrase insertion snippets (merge with insertion snippets box?)

AndreasArvidsson:
The Cursorless cements are now documented. Their rest is moved to community.

Issue #1510, opened 2023-05-30
add "common combinations" section to the cheatsheet


josharian:
It would be great to augment the cheatsheet with a list of commonly used combinations, such as `chuck bounds`.

This would help new users get off the ground quickly; examples and reference docs pair well.

Once the seed it planted, it can grow organically as recurring questions show up in Slack.

pokey:
Oh I was thinking docs. I guess cheatsheet could work? Not sure. Maybe have it live in docs and put a link to it from cheatsheet?

bra1nDump:
Oops, I thought this would only pin this for me :D

pokey:
shall we start a draft on the wiki for now? or just as comments here?

josharian:
Wiki is a great idea.

josharian:
Started https://github.com/cursorless-dev/cursorless/wiki/Combos. We can add from there.

PR #2116, opened 2023-12-10
Add simple usage stats to cheatsheet

bra1nDump:
Dependency: Command history logging #2115

As an avid cheatsheet user, it is the most intuitive place for me to show aggregate usage and would help me identify gaps.

# Implementation
- Read the raw usage file and create a rollup to actions, scopes, modifiers - simply usage count for each
- Inject a new variable `usageStatistics` similar to how we inject `cheatsheetInfo` in https://github.com/pokey/cursorless-vscode/blob/main/packages/cursorless-engine/src/core/Cheatsheet.ts'

# UI
image

# Discussions
- [ ] What UI do people expect from this?
- [ ] Gamifying completing features by coloring them different (after X number of uses) similar to keyboard shortcuts in Figma. Made harder by the table rows cycling different colors
- [ ] How to deal with sections with variants?

## Checklist

- [ ] Remove my personal notes from the code :D
- [ ] Remove changes contributed by my local autoformatter ... look into how is formatting setup on the project? I would expect it to do the right thing and play nicely with auto-save formatters (I have ESLint + prettier extensions)
- [ ] Add a default setting in VSCode (off by default), allow toggle on the cheatsheet side

[Outside the scope of this PR]
Would be most effective when combined with a re-work of the cheatsheet structure. Grouping by what commands do, popularity, common non-obvious commands, for instance 'change tree instances' etc. I imagine there have been a number of discussions on this already.


packages/cursorless-engine/src/core/Cheatsheet.ts
```diff
@@ -4,6 +4,7 @@ import produce from "immer";
 import { sortBy } from "lodash";
 import { ide } from "../singletons/ide.singleton";
 import path from "path";
+import { CheatsheetInfo } from "@cursorless/cheatsheet";
```
pokey, line 7: I'd add these to `packages/common` and import them from there. Having extension depend on a React component is probably ok, but we can avoid having that conversation by just moving the type to a shared dependency

packages/cursorless-engine/src/core/Cheatsheet.ts
```diff
@@ -41,6 +50,10 @@ export async function showCheatsheet({
 
   const root = parse(cheatsheetContent);
 
+  // Add usage stats here option #2
+  // Like ... document.cheatsheetUsageStats
+
+  // Inject data into the cheatsheet
```
pokey, line 56: Yeah this is prob where I'd do it. Just add them to the payload that gets injected below

pokey:
I'm curious what you expect to gain by this feature. Are you interested in finding actions / scopes you're not using?

bra1nDump:
@pokey Yes, the original idea is to discover features I'm not using. For new users this can serve as gamification for tracking their progress. 

While implementing this I think a more impactful contribution is to improve cheat sheet contents
- for example remove alphabetical sorting (example: tail and head should be next to each other)
- add snippets that are available #1726 (personally I feel this is where I leave most of the cursorless productivity on the table)
- add common combinations #1510

Call me lazy, but I just want the cheatsheet to be my one stop shop for pretty much anything after I learned the basics from documentation / watching YouTube videos

pokey:
I think we decided to hold off on this direction for now, right? Any reason not to close this one out for now?

PR #2122, opened 2023-12-11
Remove cyclic dependency between notebook.ts and notebookLegacy.ts.

bjaspan:
Once `isVscodeLegacyNotebookVersion` has determined that `getNotebookFromCellDocument` should call `getNotebookFromCellDocumentLegacy`, there is no need for `getNotebookFromCellDocumentLegacy` to determine that again, and doing so forces a cyclic import dependency.


pokey (approved):
👍

PR #2129, opened 2023-12-13
Improve vscode api get configuration

AndreasArvidsson:
I have on purpose not migrated all references to `vscode.workspace.getConfiguration` because we might need to have a discussion if this is a preferable implementation first.

## Checklist


pokey:
not sure I'm in love with having something on `vscodeApi` that doesn't exactly mirror vscode api, as that's kind of what that object is for. I don't feel super strongly tho, so maybe let's discuss. We could maybe put it somewhere else

packages/cursorless-vscode/src/vscodeApi.ts
```diff
@@ -9,12 +9,55 @@ import { env, window, workspace } from "vscode";
...
 export const vscodeApi: VscodeApi = {
-  workspace,
   window,
   env,
+
+  workspace: {
+    getConfiguration,
+    watchConfiguration,
+    onDidChangeConfiguration,
+  },
```
pokey, line 19: replace with top-level `usingConfiguration`, similar to https://github.com/cursorless-dev/cursorless/blob/pokey/keyboard-improvements/packages/cursorless-vscode/src/usingSetting.ts

AndreasArvidsson: Done. Unfortunately now will lost all improvements in mocking settings that I had originally introduced

PR #2130, opened 2023-12-13
Added grand scope

AndreasArvidsson:
`change grand statement`

## Checklist

### DIFF
--- a/cursorless-talon/src/modifiers/simple_scope_modifier.py
+++ b/cursorless-talon/src/modifiers/simple_scope_modifier.py
@@ -15,7 +15,23 @@ mod.list(
 )
 def cursorless_simple_scope_modifier(m) -> dict[str, Any]:
     """Containing scope, every scope, etc"""
+    if hasattr(m, "cursorless_simple_scope_modifier"):
+        modifier = m.cursorless_simple_scope_modifier
+
+        if modifier == "every":
+            return {
+                "type": "everyScope",
+                "scopeType": m.cursorless_scope_type,
+            }
+
+        if modifier == "grand":
+            return {
+                "type": "containingScope",
+                "scopeType": m.cursorless_scope_type,
+                "ancestorIndex": 1,
+            }
+
     return {
-        "type": "everyScope" if m[0] == "every" else "containingScope",
+        "type": "containingScope",
         "scopeType": m.cursorless_scope_type,
     }
--- a/cursorless-talon/src/spoken_forms.json
+++ b/cursorless-talon/src/spoken_forms.json
@@ -81,7 +81,7 @@
       "its": "inferPreviousMark",
       "visible": "visible"
     },
-    "simple_scope_modifier": { "every": "every" },
+    "simple_scope_modifier": { "every": "every", "grand": "grand" },
     "interior_modifier": {
       "inside": "interiorOnly"
     },
--- a/packages/cursorless-engine/src/generateSpokenForm/primitiveTargetToSpokenForm.ts
+++ b/packages/cursorless-engine/src/generateSpokenForm/primitiveTargetToSpokenForm.ts
@@ -49,6 +49,17 @@ export class PrimitiveTargetSpokenFormGenerator {
         throw new NoSpokenFormError(`Modifier '${modifier.type}'`);
 
       case "containingScope":
+        if (modifier.ancestorIndex != null) {
+          if (modifier.ancestorIndex === 1) {
+            return [
+              this.spokenFormMap.modifierExtra.grandScope,
+              this.handleScopeType(modifier.scopeType),
+            ];
+          }
+          throw new NoSpokenFormError(
+            `Modifier '${modifier.type}' with ancestor index != 1`,
+          );
+        }
         return [this.handleScopeType(modifier.scopeType)];
 
       case "everyScope":
--- a/packages/cursorless-engine/src/processTargets/modifiers/getContainingScopeTarget.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/getContainingScopeTarget.ts
@@ -36,8 +36,8 @@ export function getContainingScopeTarget(
       scope = expandFromPosition(
         scopeHandler,
         editor,
-        scope.domain.end,
-        "forward",
+        scope.domain.start,
+        "backward",
         ancestorIndex - 1,
       );
     }
--- a/packages/cursorless-engine/src/spokenForms/SpokenFormType.ts
+++ b/packages/cursorless-engine/src/spokenForms/SpokenFormType.ts
@@ -60,6 +60,7 @@ type ModifierExtra =
   | "previous"
   | "next"
   | "forward"
-  | "backward";
+  | "backward"
+  | "grandScope";
 
 export type SpokenFormType = keyof SpokenFormMapKeyTypes;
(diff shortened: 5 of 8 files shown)


packages/cursorless-engine/src/processTargets/modifiers/getContainingScopeTarget.ts
```diff
@@ -36,8 +36,8 @@ export function getContainingScopeTarget(
       scope = expandFromPosition(
         scopeHandler,
         editor,
-        scope.domain.end,
-        "forward",
+        scope.domain.start,
```
AndreasArvidsson, line 39: This change was necessary for python where the parent statement does not exist after each child due to the lack of a closing delimiter. See the python test for example that did not work without this change.

pokey: That fixes this specific case, but still leaves us with the same bug if we ever have scopes that start at the same position. I think you could iterate forward from the end using `allowAdjacentScopes: true` and `containment: "required"`, and then just count scopes that contain the current scope until you reach `ancestorIndex`. Does that make sense?

cursorless-talon/src/modifiers/simple_scope_modifier.py
```diff
@@ -15,7 +15,23 @@
...
+                "type": "everyScope",
+                "scopeType": m.cursorless_scope_type,
+            }
+
+        if modifier == "grand":
+            return {
+                "type": "containingScope",
+                "scopeType": m.cursorless_scope_type,
+                "ancestorIndex": 1,
+            }
```
pokey, line 32: Do we not want to support multiple levels? Doesn't seem much harder. Spoken form could be one of:

1. "great grand scope"
2. "grand grand scope"
3. "two grand scope"
4. "second grand scope"

I would lean towards 1) or 2) because 3) and 4) could conflict with modifiers I was thinking we could add in the future where they operate within the grand iteration scope. Eg in the following:

```ts
if (true) {
  const aaa = bbb;
  const ccc = ddd;
}

const eee = fff;
```

saying "second grand state air" would be `const eee = fff;`. I think "every grand state", "next grand state", etc could prob also work. I should file those as an issue. But long story short I'd lean towards "great grand" or "grand grand" to leave our grammatical options open, and because I think repeating the word is fine as you're very unlikely to go beyond 2 I think

The advantage of "grand grand" is that you could implement it super easily I believe: just put `{user.cursorless_modifier_grand}*` in the pattern and then use `ancestor_index = len(m.cursorless_modifier_grand_list)`

packages/cursorless-engine/src/generateSpokenForm/primitiveTargetToSpokenForm.ts
```diff
@@ -49,6 +49,17 @@ export class PrimitiveTargetSpokenFormGenerator {
         throw new NoSpokenFormError(`Modifier '${modifier.type}'`);
 
       case "containingScope":
+        if (modifier.ancestorIndex != null) {
+          if (modifier.ancestorIndex === 1) {
```
pokey, line 53: I think `modifier.ancestorIndex === 0` should behave the same as `modifier.ancestorIndex == null`

pokey (commented):
Nice! Glad to see it was *almost* as easy as I had hoped 😄. Left a couple comments

cursorless-talon/src/modifiers/simple_scope_modifier.py
AndreasArvidsson: Maybe we should just start with this one and see how we utilize it?

packages/cursorless-engine/src/generateSpokenForm/primitiveTargetToSpokenForm.ts
AndreasArvidsson: fixed

packages/cursorless-engine/src/processTargets/modifiers/getContainingScopeTarget.ts
AndreasArvidsson: Made a slight update now. Not sure if this is correct, but passes all the test.

pokey: I believe this approach will break down if cursor is between adjacent scopes. Let's say scopes look like the following (document text is `0123`):

```
   0123
a:   - 
b:    - 
c:  -- 
d:  --- 
```

And cursor is between `2` and `3` and scope preference breaks left, so containing scope is `a`. If you say `"grand grand scope"`, you will get `b`, when you should get `d`. I think you need to keep a `currentScope`, initialize it to the original containing scope, filter out scopes that are not a parent of the current scope, and keep updating it and increment ancestor count each time you find a matching scope. Not sure if that makes sense?

AndreasArvidsson: Do you actually have a a code example of this I can try?

pokey: No this is the kinda thing you'd want to use a mock scope for, like in that base target test I sent you

AndreasArvidsson: I can do that if you want to. At the same time if we can't actually come up with a real world code example that doesn't work isn't it good enough...at least for now?

cursorless-talon/src/modifiers/simple_scope_modifier.py
pokey: I guess it's consistent with all of our grammar proposals, and appears in the same csv as it would, so I guess it's fine to start here?

cursorless-talon/src/spoken_forms.json
```diff
@@ -81,7 +81,7 @@
       "its": "inferPreviousMark",
       "visible": "visible"
     },
-    "simple_scope_modifier": { "every": "every" },
+    "simple_scope_modifier": { "every": "every", "grand": "grand" },
```
pokey, line 84: Is this the internal identifier we want to use? I guess it's fine, but figured I'd raise it as these are really hard to change cc/ @josharian

packages/cursorless-engine/src/processTargets/modifiers/getContainingScopeTarget.ts
pokey: https://github.com/cursorless-dev/cursorless/issues/2138

cursorless-talon/src/spoken_forms.json
josharian: `grandparent`? `grandscope`? `ancestor`?

AndreasArvidsson: I think I prefer ancestor. If we in the future add ordinals or some other way to get even higher up in the hierarchy grand parent is the wrong term.

cursorless-talon/src/modifiers/simple_scope_modifier.py
AndreasArvidsson: I think so

cursorless-talon/src/spoken_forms.json
pokey: Ancestor it is

AndreasArvidsson: Updated

cursorless-talon/src/modifiers/simple_scope_modifier.py
pokey: https://github.com/cursorless-dev/cursorless/issues/2152

pokey (approved):
looks good!

PR #2133, opened 2023-12-14
Use iterations scope for relative modifier

### LINKED ISSUE #1597: Reevaluate iteration order for scopes that start or end at the same position
In the following Python code:

```py
def aaa():
    bbb = 0
    ccc = 0

ddd = 0
```

When you say `"two states backward"` from `ddd = 0`, you get `ccc = 0` past `ddd = 0`, but you probably expect to get the function `def aaa()` past `ddd = 0`. The reason for the current behavior is that both the function and `ccc = 0` end at the same position, and we yield the smaller scope first. For this case, we would prefer to yield the larger scope first.

However, in the desired implementation for #1057, we'd have the following:

```
    aaa(bbb, ccc)
#!1 ^^^^^^^^
#!2     ^^^^
#!3          ^^^
```

So if you said `"two small paints backward"` from `ccc`, today's behavior would give you `bbb, ccc`, which is probably what you want, but yielding larger scope first would give you `aaa(bbb, ccc`, which is probably not what you want.

There's something non-obvious here. It almost feels like you want to take into account iteration scope. In the latter example, you're crossing an iteration scope boundary, whereas in the former you're not

### LINKED ISSUE #2128: Change the way we iterate for relative scopes
# The problem

Today, the iteration semantics for relative scopes differs from the semantics for "every" and ordinal scopes. The latter first expand to iteration scope (if input target has no explicit range), and then get a list of all top-level scopes in that iteration scope. The former iterate directly from the input target, with no knowledge of iteration scopes. These semantics lead to the following problems:

- https://github.com/cursorless-dev/cursorless/issues/1597
- https://github.com/cursorless-dev/cursorless/pull/2110

# The solution

If the scope has no iteration scope, maybe we just leave today's behaviour?

Otherwise, we first want to find the proper iteration scope in which to iterate:

- For relative exclusive, we expand input target to smallest containing iteration scope
- For relative inclusive, we expand input target to smallest containing scope, then expand from there to smallest containing iteration scope. That handles the case where you're between scopes in the very smallest iteration scope, eg
    ```ts
    if (true) {
        console.log(1)
        |
        console.log(2)
    }

    console.log(2)
    ```

Then we do the same thing we do with "every" / ordinal, where we construct a list of all scopes in the iteration scope. These are the scopes we use instead of directly working with the scope handler. 

Note that if when iterating through the returned scopes, we run out, I think we move up to grand iteration scope and start over. Repeat that as many times as necessary

AndreasArvidsson:
Replaces https://github.com/cursorless-dev/cursorless/pull/2110

Fixes #2128
Fixes #1597

## Checklist

### DIFF
--- a/queries/javascript.core.scm
+++ b/queries/javascript.core.scm
@@ -681,6 +681,13 @@
   (#not-parent-type? @statement export_statement)
 )
 
+(program) @statement.iteration
+
+(statement_block
+  "{" @statement.iteration.start.endOf
+  "}" @statement.iteration.end.startOf
+)
+
 ;;!! foo(name: string) {}
 ;;!      ^^^^^^^^^^^^
 (
--- a/packages/cursorless-engine/src/processTargets/modifiers/RelativeExclusiveScopeStage.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/RelativeExclusiveScopeStage.ts
@@ -1,12 +1,20 @@
-import type { RelativeScopeModifier } from "@cursorless/common";
+import {
+  TextEditor,
+  type Range,
+  type RelativeScopeModifier,
+} from "@cursorless/common";
 import type { Target } from "../../typings/target.types";
 import { ModifierStageFactory } from "../ModifierStageFactory";
 import type { ModifierStage } from "../PipelineStages.types";
 import { constructScopeRangeTarget } from "./constructScopeRangeTarget";
+import { getContainingScopeTarget } from "./getContainingScopeTarget";
 import { runLegacy } from "./relativeScopeLegacy";
 import { ScopeHandlerFactory } from "./scopeHandlers/ScopeHandlerFactory";
 import { TargetScope } from "./scopeHandlers/scope.types";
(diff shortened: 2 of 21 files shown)


AndreasArvidsson:
This is definitely the direction we want to go, but on today's meet up me and @pokey realize that even on javascript the coverage of iteration scopes is quite poor. Once we have scope facets and good coverage across multiple languages this should reopen.

queries/javascript.core.scm
```diff
@@ -681,6 +681,13 @@
   (#not-parent-type? @statement export_statement)
 )
 
+(program) @statement.iteration
+
+(statement_block
+  "{" @statement.iteration.start.endOf
+  "}" @statement.iteration.end.startOf
+)
+
```
pokey, line 690: @AndreasArvidsson we should prob factor this out into its own PR with a couple tests, no?

AndreasArvidsson: done
https://github.com/cursorless-dev/cursorless/pull/2144

Issue #1775, opened 2023-08-13
create cursorless-dev extension


josharian:
let's start collecting here things we want in it

* scm formatter from andreas

pokey:
- should be able to roll `"debug generate subset"` and `"debug edit subset"` from https://github.com/cursorless-dev/cursorless/pull/1690 into a single lightning-fast command
- https://github.com/cursorless-dev/cursorless/issues/1777

PR #2136, opened 2023-12-14
add `cursorless.private.logQuickActions` internal command

pokey:
Useful for development as we add new Cursorless actions based on VSCode quick actions. We should prob put this in https://github.com/cursorless-dev/cursorless/issues/1775 once that exists, but until then, seems harmless here

## Checklist


AndreasArvidsson:
I don't quite follow the purpose of this. Why not just called the quick fix command and get the ui with available actions instead of printing them in the log?

pokey:
I print the ids and titles in the log. I can then use the id when automatically triggering a quick action. The things in the quick fix popup don't have ids. Unless I'm missing something?

AndreasArvidsson:
Is it the kind you are after? Because I got confused that you can actually pass the kind as an argument in.

AndreasArvidsson:
If you print the command instead we could just call that directly and don't have to bother with the quick fix menu?

pokey:
> Is it the kind you are after? Because I got confused that you can actually pass the kind as an argument in.

I don't actually pass that in. It's optional. You can pass that if you want to pass a general kind and see all the specific sub-kinds. But I could prob just drop that arg if it's confusing

pokey:
> If we print the command instead we could just call that directly and possibly don't have to bother with the quick fix menu?

Is there a command? Didn't realise that was returned. I could just print exactly what comes back; I believe the filtering was from when I was trying to show it in a notification in an earlier draft and then just switched to debug console

AndreasArvidsson:
> > Is it the kind you are after? Because I got confused that you can actually pass the kind as an argument in.
> 
> I don't actually pass that in. It's optional. You can pass that if you want to pass a general kind and see all the specific sub-kinds. But I could prob just drop that arg if it's confusing

Got it. Not that obvious that you can pass in a general/prefix. I can a thought that why would you pass the exact kind to then have just that printed back. Maybe the comment can be updated to better reflect this?

> > If we print the command instead we could just call that directly and possibly don't have to bother with the quick fix menu?
> 
> Is there a command? Didn't realise that was returned. I could just print exactly what comes back; I believe the filtering was from when I was trying to show it in a notification in an earlier draft and then just switched to debug console

Yes The code action class has an optional command field. Probably best to just print the entire return value. 


I do wonder if we should name this command something else; `showQuickAction` kinda sounds like it will show the quick pick ui. `logQuickActions` or something like that would make it more obvious what it's actually doing.

pokey:
> Got it. Not that obvious that you can pass in a general/prefix. I can a thought that why would you pass the exact kind to then have just that printed back. Maybe the comment can be updated to better reflect this?

Done

> Yes The code action class has an optional command field. Probably best to just print the entire return value.

I just tried that, and the command field is always either not present, or points to an internal command. For example:

```json
{
  "title": "Extract to inner function in function 'logQuickActions'",
  "kind": {
    "value": "refactor.extract.function"
  },
  "command": {
    "title": "Extract to inner function in function 'logQuickActions'",
    "command": "_typescript.didApplyRefactoring",
    "arguments": [
      {
        "action": "function_scope_0"
      }
    ]
  },
  "isPreferred": false
}
```

 I would be inclined to keep it the way it is, where I retain only the useful information (`kind.value`, `title` and `isPreferred`)

> I do wonder if we should name this command something else; `showQuickActions` kinda sounds like it will show the quick pick ui. `logQuickActions` or something like that would make it more obvious what it's actually doing.

Done

AndreasArvidsson (approved):

PR #2143, opened 2023-12-15
Added proper find in document implementation

### LINKED ISSUE #1878: scout it does not work with multiline selections
```
air
   bat
```

`"scout two tokens air"` only searches for "air". vscode can search for multiline queries.

Haven't investigated at all, heads down on other stuff.

AndreasArvidsson:
Learned of the existence of vscode command `editor.actions.findWithArgs`
https://github.com/microsoft/vscode/issues/200453

Exists in v1.66
https://github.com/microsoft/vscode/blob/release/1.66/src/vs/editor/contrib/find/browser/findModel.ts#L58

Fixes #1878

## Checklist


packages/common/src/ide/types/ide.types.ts
```diff
@@ -94,6 +94,12 @@ export interface IDE {
     listener: (event: TextDocumentChangeEvent) => void,
   ): Disposable;
 
+  /**
+   * Find occurrences of query string in the active document.
+   * @param query The string query to search for
+   */
+  findInDocument(query: string): Promise<void>;
+
```
pokey, line 102: Tempted to have the abstraction accept a text editor as input 🤔

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/scoutAir.yml
```diff
@@ -0,0 +1,34 @@
...
+    - anchor: {line: 0, character: 0}
+      active: {line: 0, character: 0}
+  thatMark:
+    - type: UntypedTarget
+      contentRange:
+        start: {line: 0, character: 0}
+        end: {line: 0, character: 3}
+      isReversed: false
+      hasExplicitRange: false
+returnValue: [aaa]
```
pokey, line 34: I don't see where the action returns anything 🤔

pokey (commented):
Nice!! Left a couple minor comments

packages/common/src/ide/types/ide.types.ts
AndreasArvidsson: Oh to open find widget in arbitrary splits?

pokey: Yeah, just feels like possibly a better abstraction. I don't feel strongly tho

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/scoutAir.yml
AndreasArvidsson: Hmm yeah this doesn't look right. Wrong action id even :p

packages/common/src/ide/types/ide.types.ts
AndreasArvidsson: I agree.

pokey: I guess prob will be optional anyway so could possibly leave it off for now until we need it

AndreasArvidsson: The question is if you target something in a split do you want to search that document or the active?

packages/cursorless-vscode-e2e/src/suite/fixtures/recorded/actions/scoutAir.yml
AndreasArvidsson: Tests updated

packages/common/src/ide/types/ide.types.ts
AndreasArvidsson: If we pass the editor as an argument we can change this behavior in the future without changing the api. If we don't pass the editor we have no control over this.

pokey: Yeah I'm happy for us to add it. Should be trivial, no?

Fwiw I'd lean towards searching in document containing cursor, but I could see either way

AndreasArvidsson: Fixed. For now we default to the active editor, but in the future we could change this behavior easily.

pokey (approved):
Looks good!

PR #2144, opened 2023-12-15
Added iteration scope for job script statements

AndreasArvidsson:
## Checklist

### DIFF
--- a/packages/common/src/scopeSupportFacets/javascript.ts
+++ b/packages/common/src/scopeSupportFacets/javascript.ts
@@ -10,11 +10,13 @@ const { supported, notApplicable } = ScopeSupportFacetLevel;
 export const javascriptScopeSupport: LanguageScopeSupportFacetMap = {
   list: supported,
   map: supported,
-  statement: supported,
   ifStatement: supported,
   regularExpression: supported,
   switchStatementSubject: supported,
   fieldAccess: supported,
+
+  statement: supported,
+  "statement.iteration": supported,
 
   class: supported,
   className: supported,
--- a/packages/common/src/testUtil/getFixturePaths.ts
+++ b/packages/common/src/testUtil/getFixturePaths.ts
@@ -32,10 +32,8 @@ export function getRecordedTestPaths() {
   return walkFilesSync(directory)
     .filter((p) => p.endsWith(".yml") || p.endsWith(".yaml"))
     .map((p) => ({
-      name: path
-        .relative(relativeDir, p.substring(0, p.lastIndexOf(".")))
-        .replaceAll("\\", "/"),
       path: p,
+      name: pathToName(relativeDir, p),
     }));
 }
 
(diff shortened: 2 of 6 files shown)


packages/common/src/testUtil/getFixturePaths.ts
```diff
@@ -47,8 +45,14 @@ export function getScopeTestPaths() {
     .filter((p) => p.endsWith(".scope"))
     .map((p) => ({
       path: p,
-      name: path.relative(relativeDir, p.substring(0, p.lastIndexOf("."))),
+      name: pathToName(relativeDir, p),
```
AndreasArvidsson, line 48: Had forgot to update the name scope test cases

packages/cursorless-vscode-e2e/src/suite/fixtures/scopes/javascript/statement.iteration.scope
```diff
@@ -0,0 +1,33 @@
+// Start
```
AndreasArvidsson, line 1: Included to make it very obvious that it's not the function, but the entire document

packages/common/src/scopeSupportFacets/javascript.ts
```diff
@@ -10,12 +10,14 @@ const { supported, notApplicable } = ScopeSupportFacetLevel;
...
   list: supported,
   map: supported,
-  statement: supported,
   ifStatement: supported,
   regularExpression: supported,
   switchStatementSubject: supported,
   fieldAccess: supported,
 
+  statement: supported,
+  "statement.iteration": supported,
```
pokey, line 19: This will absolutely get split up, and probably sooner rather than later. Should we do that now? Eg `statement.iteration.file` and `statement.iteration.block`? Tho `block` tends to mean something very different in our code base 🤔

pokey (commented):
looks good with minor comment

packages/common/src/scopeSupportFacets/javascript.ts
AndreasArvidsson: I was thinking about that, but there is no way to do separate scopes fixtures for that, both files will have 2 scopes I'm afraid.

Block is just the spoken form. We do call it paragraph as an id. I'm fine with block in this context.

pokey: > I was thinking about that, but there is no way to do separate scopes fixtures for that, both files will have 2 scopes I'm afraid.

Yeah I'm not sure of a way around that. But I still think making it super explicit what the iteration scopes should be is preferred

> Block is just the spoken form. We do call it paragraph as an id. I'm fine with block in this context.

Sounds good

AndreasArvidsson: Ok will split it!

AndreasArvidsson: Fixed!

pokey (approved):
👍

PR #2147, opened 2023-12-16
Validate query capture names

### LINKED ISSUE #1433: Validate capture names in queries
We should check that all capture names in queries (eg `@namedFunction.domain`) are valid. Today they just silently won't match anything

They should either start with a scope type or `_`, and have zero or more `.foo` suffixes, which can be one of a fixed set, eg `.domain`, `.iteration`, etc

AndreasArvidsson:
Show error message when an invalid capture name is used

Fixes #1433

## Checklist


pokey (approved):
Ok I made some tweaks based on our conversation; we no longer support `@dummy`, but instead support arbitrary `@_foo` captures. Feel free to merge if you're happy with my changes

PR #2148, opened 2023-12-17
Attempt at regex for validation

pokey:
Certainly a lot less code. Also allows things like `statement.startOf`, which should be allowed I think. We have zero-width captures, eg for name of start tag with react fragments

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/languages/TreeSitterQuery/validateQueryCaptures.test.ts
+++ b/packages/cursorless-engine/src/languages/TreeSitterQuery/validateQueryCaptures.test.ts
@@ -18,13 +18,13 @@ const testCases: { name: string; isOk: boolean; content: string }[] = [
     name: "Position captures",
     isOk: true,
     content:
-      "(if_statement) @statement.leading.startOf @statement.trailing.endOf",
+      "(if_statement) @statement.startOf @statement.leading.startOf @statement.trailing.endOf",
   },
   {
     name: "Range captures",
     isOk: true,
     content:
-      "(if_statement) @statement.removal.start @statement.interior.start.endOf",
+      "(if_statement) @statement.start @statement.start.endOf @statement.removal.start @statement.interior.start.endOf",
   },
   {
     name: "Dummy capture",
@@ -70,11 +70,6 @@ const testCases: { name: string; isOk: boolean; content: string }[] = [
     name: "Wildcard start",
     isOk: false,
     content: "(if_statement) @_.start",
-  },
-  {
-    name: "Leading start",
-    isOk: false,
-    content: "(if_statement) @statement.leading.start",
   },
   {
     name: "Text fragment removal",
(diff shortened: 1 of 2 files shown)


AndreasArvidsson:
I don't think you should able to do `leading.start` I think the notion that some captures are position and some are ranges is important.

pokey:
but if you do `scope.leading`, then it's actually a range corresponding to the node you've tagged

pokey:
added a couple more tests

AndreasArvidsson:
> but if you do `scope.leading`, then it's actually a range corresponding to the node you've tagged

We only use start. We don't care about the range.

pokey:
Ok fine; I added a few more tests on this branch. If you can pass all my tests then your code is fine 😊

The regex thing was just a quick 15-min experiment to see if it was clearer, because this task feels very much like what regexes were designed for, but after implementing it I'm not sure I'm convinced the regex approach is any clearer 😅

I think the clearest would be a parser, but that is most definitely overkill 😄

AndreasArvidsson:
I started with a regex, but I thought it became to hard to read and reason about that I went to the whitelist approach which is much simpler. You can even print the whitelist if you want :D

A parser would be nice because then we can complain about using wild card with no "named scope" and similar problems. For now I think it might be overkill.

pokey:
> I started with a regex, but I thought it became to hard to read and reason about that I went to the whitelist approach which is much simpler. You can even print the whitelist if you want :D

Makes sense 😄

> A parser would be nice because then we can complain about using wild card with no "named scope" and similar problems. For now I think it might be overkill.

Sorry what are the 'no "named scope" and similar problems'

> I don't think we should support `@_foo`. That kind of capture was only used in a single language `go` and wasn't even needed. For now I think we should just stick with `dummy`.

Oh ok I had thought we were moving in the direction of the `_` prefix after a discussion on a PR, but I guess maybe you weren't involved in that one. The reason to prefer `_` prefix is that it gives us more flexibility to be clearer / more verbose; "dummy" tells you nothing about what / why you're tagging something. It also feels a bit special / surprising that just that one special word works. Happy to discuss if you disagree tho

AndreasArvidsson:
> > A parser would be nice because then we can complain about using wild card with no "named scope" and similar problems. For now I think it might be overkill.
> 
> Sorry what are the 'no "named scope" and similar problems'?

A named scope capture like `@statement`. You can't just have a wildcard without a named scope in the same pattern.

> > I don't think we should support `@_foo`. That kind of capture was only used in a single language `go` and wasn't even needed. For now I think we should just stick with `dummy`.
> 
> Oh ok I had thought we were moving in the direction of the `_` prefix after a discussion on a PR, but I guess maybe you weren't involved in that one. The reason to prefer `_` prefix is that it gives us more flexibility to be clearer / more verbose; "dummy" tells you nothing about what / why you're tagging something. It also feels a bit special / surprising that just that one special word works. Happy to discuss if you disagree tho

I can't remember a discussion like that, but I can see the logic behind it. So we should allow any capture that starts with a underscore but they can't have any relationships or positions?

pokey:
Incorporated changes into #2147

PR #2153, opened 2023-12-19
Convert VscodeHatRendered to use vscode.workspace.fs.

bjaspan:
This is part of porting Cursorless to work in the web environment.


packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
```diff
@@ -132,11 +135,11 @@ export default class VscodeHatRenderer {
...
       await this.updateShapeOverrides(hatsDir);
 
-      if (fs.existsSync(hatsDir)) {
+      fs.access(hatsDir).then(() =>
         this.hatsDirWatcherDisposable = watchDir(hatsDir, () =>
           this.updateShapeOverrides(hatsDir),
-        );
-      }
+        ))
+        .catch(() => console.error("cannot watch hatsDir " + hatsDir));
```
bjaspan, line 142: Note to reviewers: I'm not entirely sure this change is correct.

pokey: also, are you sure it's ok to use `fs` directly? I'd think we would need to use the `vscode` fs stuff exclusively, like we do everywhere else in this file

pokey (commented):
mostly seems fine; left a few comments 😊

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
bjaspan: I thought I tried `await fs.access` and it didn't work/compile for some reason. I'll try again.

Yes, ideally we will remove all direct calls to `fs`. This PR does not completely port VscodeHatRenderer to the web runtime, it does just enough to make the hat images load. The remaining calls to fs are currently implemented by a shim library that just fails; so in the access/watchDir case, no watching happens, which is fine for now.

Also, of course, there is no watchDir equivalent vscode.workspace.fs (because HTTP). Ultimately this code, along with a variety of other cases, should be factored out into a native-runtime-specific injected dependency so that it can work locally and not throw errors in on web.

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
```diff
@@ -329,21 +344,23 @@ export default class VscodeHatRenderer {
...
-        this.extensionContext.extensionPath,
+      vscode.Uri.joinPath(
+        this.extensionContext.extensionUri,
         "images",
         "hats",
         `${shape}.svg`,
       );
-    const rawSvg = fs.readFileSync(iconPath, "utf8");
+    const rawSvg = this.decoder.decode(
+      await vscode.workspace.fs.readFile(iconPath),
```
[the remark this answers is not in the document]
bjaspan: Apparently yes. I am bundling the hat images using google3 build system tooling for Cider extensions; I haven't looked into how one bundles static resources for VS Code extensions built the normal way.

As a side note, I just saw this at https://code.visualstudio.com/api/working-with-extensions/publishing-extension#publishing-extensions: "Note: Due to security concerns, vsce will not publish extensions that contain user-provided SVG images."

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
bjaspan: The issue with `fs.access` is that it returns a `Promise<void>`, not a `Promise<boolean>`. The alternative to then/catch is try/catch.  See https://stackoverflow.com/questions/73674525/node-js-fs-access-returns-undefined-no-matter-what.

pokey: Yes I'd use try/catch

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
```diff
@@ -132,10 +135,13 @@ export default class VscodeHatRenderer {
...
       await this.updateShapeOverrides(hatsDir);
 
-      if (fs.existsSync(hatsDir)) {
+      try {
+        await fs.access(hatsDir);
         this.hatsDirWatcherDisposable = watchDir(hatsDir, () =>
           this.updateShapeOverrides(hatsDir),
         );
+      } catch (e) {
+        console.error("cannot watch hatsDir " + hatsDir);
```
auscompgeek, line 144: If we're going to log, we might as well log the exception too.
```suggestion
        console.error("cannot watch hatsDir", hatsDir, e);
```

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
```diff
@@ -75,6 +77,7 @@ export default class VscodeHatRenderer {
   ) {
     extensionContext.subscriptions.push(this);
 
+    this.decoder = new TextDecoder("utf-8");
```
auscompgeek, line 80: nit: do this on the same line as the declaration above?

auscompgeek (commented):
Looking good, just a couple of comments from me

packages/cursorless-vscode/src/ide/vscode/hats/VscodeHatRenderer.ts
bjaspan: Done.

bjaspan:
I seem to have blown git conflict resolution. I'll start over.

PR #2155, opened 2023-12-21
Add `implicit` special target

pokey:
- Depends on https://github.com/cursorless-dev/cursorless/pull/2170

## Checklist

### DIFF
--- a/cursorless-talon/src/spoken_forms.py
+++ b/cursorless-talon/src/spoken_forms.py
@@ -115,7 +115,11 @@ def update():
     )
 
     disposables = [
-        handle_csv("actions.csv"),
+        handle_csv(
+            "actions.csv",
+            extra_allowed_values=["private.setImplicitTarget"],
+            default_list_name="simple_action",
+        ),
         handle_csv("target_connectives.csv"),
         handle_csv("modifiers.csv"),
         handle_csv("positions.csv"),
--- a/packages/common/src/StoredTargetKey.ts
+++ b/packages/common/src/StoredTargetKey.ts
@@ -2,5 +2,6 @@ export const storedTargetKeys = [
   "that",
   "source",
   "instanceReference",
+  "implicit",
 ] as const;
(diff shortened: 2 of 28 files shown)


pokey:
There is a bit of a gotcha here:

A common use case might be `"from air and bat swap key with value"`, which is equivalent a cursorless version of `"take air and bat swap key with value"`. So the user will get in the habit of thinking of `"from air <action> key"` as equivalent to `"<action> key air"`

So the user might expect to be able to say `"from air take every instance key"`, and have it be equivalent to `"take every instance key air"`. But that's not the case, because when you use `"instance"`, the `"from"` action is setting the range in which to search for instances, so it would try to search in the token `"air"` for the `"key"` containing the cursor

cc/ @AndreasArvidsson https://github.com/cursorless-dev/cursorless/labels/to%20discuss

pokey:
update from meet-up: planning to try just using "take" to break up commands, so maybe this one isn't necessary. Only problem is we lose rich semantics. But let's try for a while

pokey:
update: I had a long one I wanted to break up: `"carve next sub three views first three subs gust"`. Slight annoyance that when breaking this up with "take", I need to do

1. "take first three subs gust"
2. "take three views _sub_" (because I lost the "sub" from the first command)
3. "carve next sub"

AndreasArvidsson:
That is an intriguing idea

josharian:
Interesting. I definitely see the appeal, but I am a little nervous about hidden state. It can make a system hard to reason about as a user. (For a very wordy example, see https://commaok.xyz/post/fork-point/.)

pokey:
Fwiw it wouldn't be hidden. We'd show the highlight

josharian:
Ahhh. In that case, concern withdrawn.

pokey:
closing for now as it's not clear we'll move in this direction. for now, let's just try to get in the habit of saying "take"

PR #2158, opened 2023-12-28
pre-commit: Update Python linters

auscompgeek:
- https://github.com/ikamensh/flynt/blob/master/CHANGELOG.md#v100
- https://github.com/astral-sh/ruff-pre-commit / https://github.com/astral-sh/ruff/blob/main/CHANGELOG.md#0111
- https://black.readthedocs.io/en/stable/integrations/source_version_control.html


pokey (approved):
Looks good!

PR #2169, opened 2024-01-10
Basic keyboard features

pokey:
- Depends on https://github.com/cursorless-dev/cursorless/pull/2168

## Checklist

- [ ] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)
- [ ] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [ ] I have not broken the cheatsheet

PR #2172, opened 2024-01-15
keyboard: fix partial arg code

pokey:
The original partial arg code constructed a partial arg where each field of the top-level command arg was either completely filled out or completely missing. Before #2169, that was fine, because the top-level commands were more specific, and thus their args had more detail and the command id itself told you more. For example, in `modifyTargetContainingScopeType`, you had a specific scope type that was missing, but you knew from command id that it was "containing". 

As of #2169, we do a lot more in the lower level captures. For example `modifyTargetContainingScopeType` and `modifyTargetEveryScopeType` are now just `modifyTarget`, which only has a `modifier` key at the top level, so you don't know until the very end what you're dealing with, because every modifier just looks like `{command: "modifyTarget", partialArg: {modifier: undefined}}`.

This PR deeply reconstructs the partial target, so that you can get much more information. See the new test cases for examples of these partial arguments

- Depends on https://github.com/cursorless-dev/cursorless/pull/2169

## Checklist

- [x] Switch to `CURRENT` for default arg
- [x] Remove unique queue file


pokey:
@AndreasArvidsson just need a stamp from you as this only touches keyboard; review will be from @josharian

packages/cursorless-vscode/src/keyboard/grammar/getAcceptableTokenTypes.ts
```diff
@@ -43,39 +66,80 @@ function getMetadata<T extends keyof KeyboardCommandHandler>(
     .metadata;
 }
 
+// Prevent infinite recursion
```
josharian, line 69: oof. why is 3 the right number?

(an alternative to the work queue and counts is also to use a fast/slow walk, like floyd warshall.)

packages/cursorless-vscode/src/keyboard/grammar/getAcceptableTokenTypes.ts
```diff
@@ -43,39 +66,80 @@ function getMetadata<T extends keyof KeyboardCommandHandler>(
...
+const MAX_VISITS = 3;
+
+/** Indicates that the given symbol hasn't been typed yet */
+export const MISSING = Symbol("missing");
+
+/** The next token to be consumed */
```
josharian, line 75: terminology/consistency: here you call it a token but three lines above you call it a symbol

packages/cursorless-vscode/src/keyboard/grammar/getAcceptableTokenTypes.ts
```diff
@@ -43,39 +66,80 @@ function getMetadata<T extends keyof KeyboardCommandHandler>(
...
-  const queue = new UniqueWorkQueue(state);
-  const roots: State[] = [];
+function computeRootStatePartialArgs(
+  state: nearley.State,
+  lastSymbol: any = NEXT,
+  visitCounts = new DefaultMap<State, number>(() => 0),
+  roots: { state: State; partialArg: any }[] = [],
+) {
+  const visitCount = visitCounts.get(state);
+  if (visitCount > MAX_VISITS) {
```
josharian, line 98: if perf matters, consider wrapping this around the actual call (with condition inverted), to avoid a bunch of leaf recursion calls

packages/cursorless-vscode/src/keyboard/grammar/getAcceptableTokenTypes.ts
```diff
@@ -43,39 +66,80 @@ function getMetadata<T extends keyof KeyboardCommandHandler>(
...
+
+  let partialArg: any;
+  try {
+    const argList = [...getCompletedSymbols(state), lastSymbol];
+    argList.push(
+      ...times(state.rule.symbols.length - argList.length, () => MISSING),
+    );
+    partialArg = state.rule.postprocess?.(argList) ?? argList;
+  } catch (err) {
+    partialArg = MISSING;
```
josharian, line 111: please add a comment about when this happens

packages/cursorless-vscode/src/keyboard/grammar/getAcceptableTokenTypes.ts
```diff
@@ -43,39 +66,80 @@ function getMetadata<T extends keyof KeyboardCommandHandler>(
...
-function computePartialArg<T extends keyof KeyboardCommandHandler>(
-  _state: nearley.State,
-): Partial<Record<T, any>> {
-  // FIXME: Fill this out
-  return {};
+function getCompletedSymbols(state: nearley.State) {
+  let currentState = state;
+  const completedSymbols: any[] = [];
+
+  while (currentState.dot > 0) {
```
josharian, line 139: naive question because my typescript is still weak: does mutating currentState here mutate state? (what is a value type and what is a reference type? sigh, sorry.)

josharian (approved):
insofar as I understand it, the code lgtm. it would be nice to have a top level comment somewhere (ideally pr description / commit message or issue) explaining what was wrong with the old approach.

packages/cursorless-vscode/src/keyboard/grammar/getAcceptableTokenTypes.ts
pokey: Sorry where do we mutate `currentState`?

josharian: oh, i see i misread something. never mind, ignore me.

AndreasArvidsson (approved):

packages/cursorless-vscode/src/keyboard/grammar/getAcceptableTokenTypes.ts
pokey: Fwiw primitives are value types, everything else is reference. (keeping in mind that strings are considered primitive in that statement). Tho I think strings are immutable so doesn't really matter. But I'd need to double check that

packages/cursorless-vscode/src/keyboard/grammar/getAcceptableTokenTypes.ts
pokey: It's not, just picked something that felt likely to be an upper bound on what we might reasonably get in practice; updated comment

packages/cursorless-vscode/src/keyboard/grammar/getAcceptableTokenTypes.ts
pokey: I think that would only help for things that hit the limit, and be pretty marginal gain, so I'd be inclined to hold off until we decide it matters. Tho I may be missing something here

packages/cursorless-vscode/src/keyboard/grammar/getAcceptableTokenTypes.ts
```diff
@@ -1,9 +1,28 @@
...
+  type: TokenType;
+
+  /**
+   * The command that wants the token type
+   */
+  command: keyof KeyboardCommandHandler;
+
+  /**
+   * A partial argument for the command that wants the token type, where
+   * {@link NEXT} indicates where the next token would go and {@link MISSING}
```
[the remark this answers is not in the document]
pokey: Yeah. It's special because we could potentially use it in an interesting way in the sidebar. When we construct all partial args in response to a hypothetical next token type, we could then enumerate the possible tokens for that type and slot them into `NEXT` to see what they'd look like. That way we could have a sidebar section for eg `color`, and then show what the command would be for each different color you could type

josharian:
lgtm, thanks

PR #2193, opened 2024-01-22
Add cursorless symbols to cursorless

jaresty:
Since there is symmetry between the cursorless commands and these symbols it makes sense to define them in the same repository.

Copied these from [pokey_talon](https://github.com/pokey/pokey_talon/blob/main/plugin/symbols/symbols.talon)

## Checklist


jaresty:
Not working quite right. Looking into it.

pokey:
yeah this impl is ~fine, tho I'd be tempted to insert using a snippet 🤔

but there are bigger questions around whether we want this stuff to be included in Cursorless at all or if it should be in community, as mentioned in https://github.com/cursorless-dev/cursorless-talon/pull/162#issuecomment-1904038995

jaresty:
Yep! Reiterating here for visibility: the main argument for me that putting these into cursorless makes sense is that this is where those delimiters are defined for cursorless itself. There is symmetry between the names used here and the symbols. If we defined it in community instead, it would be awkward to keep in sync as cursorless evolves and it would be questionable to use the user's custom abbreviations for these since they are in cursorless configuration which wouldn't exist for all community users.

jaresty:
Gotcha. One approach could be to adopt this temporarily and move it later if we decide to, since this is closer to current configuration, I think?

pokey:
Could work? Worth a discussion to try to figure out possible ramifications

pokey:
We discussed this PR in the meet-up on Sunday. I think we all agree that this approach is not the long-term solution we want, but that doesn't necessarily mean that this approach is a bad short-term solution. Fwiw, here is the long-term solution we landed on: https://github.com/talonhub/community/issues/546#issuecomment-1914536331.

When deciding whether to accept this PR, the questions are:

- How long will it be before we get to the long-term solution we desire?
- How much benefit is there to this PR in the interim?
- How much damage is there from this PR in the interim?
- How hard will it be to roll this PR back when we move to the long-term solution?

## How long will it be before we get to the long-term solution we desire?

This could be a while, as the migration path is going to require a bit of work

## How much benefit is there to this PR in the interim?

The primary benefit is that a user can watch one of my videos, see me saying `"round"`, and have that work out of the box. However:

- This command will only work in VSCode
- Dozens of other commands that I use won't be available. The real solution here is https://github.com/talonhub/community/issues/1221 and https://github.com/cursorless-dev/cursorless/discussions/547

So while it's clear that there is a short-term benefit here, it feels like it's addressing a tiny fraction of a much larger problem, in a way that doesn't move us towards the longer term goals

## How much damage is there from this PR in the interim?

The biggest risk from our perspective is that the fact that it only works in VSCode will encourage users to either copy this code, or fork and modify Cursorless, in order to get it to work elsewhere. We strongly discourage using internal Cursorless captures / actions, and we also discourage forking Cursorless, as we want the upgrade path to be simpler. So no matter how they proceed, we've induced them to make Cursorless harder to support. The only desirable way for them to get these things working outside of VSCode is for them to modify their community wrapper commands, which is what they'd have to do anyway today. 

## How hard will it be to roll this PR back when we move to the long-term solution?

Probably not too bad. We could fairly easily disable these commands when we detect the tag that indicates that they've upgraded to the version of community that includes the new pairs

# Conclusion

Based on the trade-off of benefit vs damage, from our perspective, the benefit doesn't outweigh the potential long-term damage of encouraging users to fork Cursorless or use our internal APIs

All that being said, opening a PR is probably the best way to get us to take a topic seriously, so from that perspective this PR has been a success 😊

I'll leave the PR open in case you want to make a case for it, but feel free to close if you agree

jaresty:
I'll respond in more depth later, but for now I'll start by sharing this PR in response: https://github.com/talonhub/community/pull/1375

pokey:
I would argue for closing this PR in favour of the approach proposed in https://github.com/talonhub/community/issues/546#issuecomment-1914536331

PR #2194, opened 2024-01-22
Convert VscodeHatRendered to use vscode.workspace.fs

bjaspan:
This is part of porting Cursorless to work in the web environment.

This is a re-roll of https://github.com/cursorless-dev/cursorless/pull/2153 which I messed up with bad git-fu.

## Checklist


pokey (commented):
looks good, but @AndreasArvidsson maybe you'd like to test locally to make sure it doesn't break your custom hats? I don't use custom hats

AndreasArvidsson:
I assume @bjaspan has tested it already?

bjaspan:
It works well enough for me in the web-based VSCode inside Google to load and display standard hats. I don't use custom hats, either, and they wouldn't work (yet) in the web environment since that part of the code still uses `fs/promises` instead of the vscode APIs.

pokey:
I'll test locally just to ensure that normal hats don't break, but I won't test custom hats because that's not an officially supported feature. @AndreasArvidsson I guess you can file an issue if custom hats break

PR #2199, opened 2024-01-23
Change LanguageDefinitions to use VscodeFilesystem.

bjaspan:
Remove some uses of `fs`, `fs/promises`, and `path` for the web port.


bjaspan:
This test is failing because it calls LanguageDefinitions.get directly, without an open-text-document callback which is what the class now uses to load languages: [https://github.com/cursorless-dev/cursorless/blame/08abb7f380952aa72c415c918ab5073[…]1df2d7934/packages/cursorless-engine/src/runIntegrationTests.ts](https://github.com/cursorless-dev/cursorless/blame/08abb7f380952aa72c415c918ab50731df2d7934/packages/cursorless-engine/src/runIntegrationTests.ts#L33).  

How do I inject an open-text-document to the FakeIDE?

bjaspan:
I fixed the failing test mentioned earlier.

packages/cursorless-engine/src/languages/LanguageDefinitions.ts
```diff
@@ -58,13 +64,34 @@ export class LanguageDefinitions {
...
+        languageId,
+      )) ?? LANGUAGE_UNDEFINED;
+
+    this.languageDefinitions.set(languageId, definition);
+  }
+
+  private async reloadLanguageDefinitions(): Promise<void> {
+    this.languageDefinitions.clear();
+    await Promise.all(
+      Array.from(this.languageDefinitions.keys()).map((languageId) =>
```
bjaspan, line 88: How can this work given that this.languageDefinitions was cleared on line 86?

pokey (changes requested):
Ok I pushed a couple changes to fix / simplify, but there's a problem with the way you're finding the queries. Here is the correct query dir: https://github.com/bjaspan/cursorless/blob/45261f41748e033c373366f8b40c8fddef9d5817/packages/cursorless-engine/src/languages/LanguageDefinitions.ts#L66-L71

That makes it so that in development mode we use the queries from the Cursorless source repo, rather than the copies we make when we bundle them into the extension `dist` folder. That enables us to do hot reloading of the queries during development

However, you instead always use the query dir bundled into `dist` when you read the queries https://github.com/bjaspan/cursorless/blob/45261f41748e033c373366f8b40c8fddef9d5817/packages/cursorless-engine/src/languages/LanguageDefinition.ts#L117

bjaspan:
I believe I fixed that but I'm not set up to test changing the development queries locally; please give it a try.

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
```diff
@@ -1,26 +1,90 @@
...
+    await vscode.workspace.fs.createDirectory(
+      vscode.Uri.file(this.cursorlessDir),
+    );
+  }
+
+  public join(path1: string, path2: string): string {
+    // FIXME: Look into whether we can use joinPath(...).path. The
+    // question is whether Uri will correctly remove the Uri.base's path
+    // from the returned path.
+    return `${path1}/${path2}`;
```
pokey, line 45: yeah we should try to do a proper join

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
```diff
@@ -1,26 +1,90 @@
...
+  }
+
+  public dirname(path: string): string {
+    return path.substring(0, path.lastIndexOf("/"));
```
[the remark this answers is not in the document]
bjaspan: I think it should. URIs are defined to use forward-slash, and vscode.Uri normalizes incoming filesystem paths to forward-slash as well: https://github.com/microsoft/vscode-uri/blob/6dec22d7dcc6c63c30343d3a8d56050d0078cb6a/src/uri.ts#L310.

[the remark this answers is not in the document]
pokey: Ok. I do wonder if we could solve all our problems with a Path abstraction. That way we could ask it for sub dirs, and we don't need to check for absolute path, etc. Lmk if you're not sure how that should look and I could take a swing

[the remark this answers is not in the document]
pokey: It could just have a `read` method. Like Python does

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
bjaspan: I confirmed that we can't use vscode.Uri.joinPath and then extract the path, as I said we should look into, because that will lose the extension's base path if there is one.  Since we also can't import `path` in the web version, and since vscore.Uri always uses forward-slash (see comment elsewhere), I think an array join on "/" is sufficient here.

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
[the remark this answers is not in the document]
bjaspan: I'm not sure what you have in mind and I'm over budget on time for this project this week, so go ahead. :-)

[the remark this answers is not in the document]
pokey: Sounds good I'll give it a whirl ☺️. I'll bail if it gets at all complicated as I do think this PR is in pretty good shape as is

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
```diff
@@ -1,26 +1,87 @@
-import { Disposable, FileSystem, PathChangeListener } from "@cursorless/common";
-import { RelativePattern, workspace } from "vscode";
-import * as path from "path";
+import {
+  Disposable,
+  FileSystem,
+  PathChangeListener,
+  RunMode,
+} from "@cursorless/common";
+import { isAbsolute } from "path";
```
bjaspan, line 7: I removed the import of `path` on purpose.  The goal is to remove dependencies on node modules so the extension will run in a web host with minimal effort. Do we really need isAbsolute vs startsWith("/")?

pokey: Will that work on windows?

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
```diff
@@ -1,26 +1,87 @@
...
+import { TextDecoder } from "util";
```
pokey, line 8: I'm pretty sure this is a node import, no?

pokey: update: looks like this import isn't necessary, as `TextDecoder` is globally available in both node and browser

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
pokey: I think I'd prefer to use a polyfill. See eg https://github.com/evanw/esbuild/issues/85#issuecomment-624346338. Is that going to cause a problem for your compiler?

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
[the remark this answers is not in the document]
bjaspan: I've gone back to path.dirname, so this is no longer relevant.

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
bjaspan: I've gone back to path.join, so this is no longer relevant.

packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
```diff
@@ -1,26 +1,90 @@
...
+  public async initialize(): Promise<void> {
+    await vscode.workspace.fs.createDirectory(
+      vscode.Uri.file(this.cursorlessDir),
+    );
+  }
+
+  public join(path1: string, path2: string): string {
+    // FIXME: Look into whether we can use joinPath(...).path. The
+    // question is whether Uri will correctly remove the Uri.base's path
+    // from the returned path.
```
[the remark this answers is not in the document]
bjaspan: No longer relevant.

packages/cursorless-engine/src/languages/LanguageDefinition.ts
```diff
@@ -33,19 +36,24 @@ export class LanguageDefinition {
...
+  ): Promise<LanguageDefinition | undefined> {
+    const languageQueryPath = fileSystem.join(queryDir, `${languageId}.scm`);
+
+    let rawLanguageQueryString;
+    try {
+      rawLanguageQueryString = await readQueryFileAndImports(
+        fileSystem,
+        languageQueryPath,
+      );
+    } catch (err) {
```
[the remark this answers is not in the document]
bjaspan: Done.

bjaspan:
I believe all comments so far are addressed.

pokey (approved):
Ok looks good to me. I made a couple more tweaks. The biggest problems were that:

1. Your code was failing to distinguish between a non-existent top-level language definition file vs a broken import in a language definition file. The former is a perfectly valid situation indicating that a language isn't supported, whereas the latter is a likely a typo made during cursorless development that we want to surface as quickly as possible
2. You were relying on the vscode-specific error code in the cursorless engine, which is supposed to be agnostic to ide

I'm not sure how I feel about returning undefined when file doesn't exist, but it does force the client to consider the possibility that the file might not exist, so I think is a win?

I'm happy if you are

We can look into the path abstraction thing I was talking about later. I think we'll need to revisit some of this stuff anyway when you remove the `os.tmpdir` stuff in extension, which I'm guessing you'll need to do

bjaspan:
I tested this on Cider (web environment) and found that `mkdir(cursorlessDir)` was failing (no surprise) and causing extension activation to fail.  There are ways we can improve this, such as by having cursorlessDir be a workspace.Uri and on web environments using extensionContext.storageUri. Of course that won't help for communicating with Talon (e.g. for spoken forms state.json), but I have some other ideas about that.  In the meantime, I just wrapped a try/catch around it so we log the error instead of failing activation. With that, this code at least executes in a web environment; I can't really test it until I have language definitions, and that has to wait for me to figure out how to get WASM to work.

Anyway, yes, I am happy with this PR and ready for it to be committed.

pokey:
> I tested this on Cider (web environment) and found that `mkdir(cursorlessDir)` was failing (no surprise) and causing extension activation to fail. There are ways we can improve this, such as by having cursorlessDir be a workspace.Uri and on web environments using extensionContext.storageUri. Of course that won't help for communicating with Talon (e.g. for spoken forms state.json), but I have some other ideas about that.

Yeah the cursorless dir is a bit overloaded. I removed it from the public interface

I agree we can prob use `extensionContext.storageUri` for command history, or we can just not support command history on web. That seems fine to me

Re communication with Talon, I think the right place to start would be to have a look at [`TalonSpokenFormsJsonReader`](https://github.com/bjaspan/cursorless/blob/main/packages/cursorless-engine/src/nodeCommon/TalonSpokenFormsJsonReader.ts). Note how it's in a directory called `nodeCommon`. I was imagining that we could just inject this from `extension.ts` when we're in regular VSCode, and inject something else conforming to the simple [`TalonSpokenForms`](https://github.com/bjaspan/cursorless/blob/f577cee9d4839a8beeb366a64dd4267b5a186f15/packages/cursorless-engine/src/scopeProviders/TalonSpokenForms.ts#L7-L14) interface when we're on web. At some point we'll prob want to migrate `nodeCommon` to its own package in the monorepo and add lint rules that ensure the cursorless engine itself doesn't refer to any node packages

> In the meantime, I just wrapped a try/catch around it so we log the error instead of failing activation. With that, this code at least executes in a web environment; I can't really test it until I have language definitions, and that has to wait for me to figure out how to get WASM to work.

Yeah I think that's fine. Tbh cursorless dir isn't critical to Cursorless behaviour, so probably better if the whole extension doesn't crash if for some reason we can't create that dir

PR #2202, opened 2024-01-25
Stop using `backspace` for `remove` action; tweak maps

pokey:
This broke when we reworked our keyboard actions setup. It's also not properly integrated with new keyboard parser. If we wanted to support special keys like `backspace` with our keyboard parser, we should define a VSCode command that sends a keyboard token to our parser, and map `backspace` to that

We also add `remove` to default config, and tweak some maps while we're there

## Checklist

### DIFF
--- a/docs/user/experimental/keyboard/modal.md
+++ b/docs/user/experimental/keyboard/modal.md
@@ -29,18 +29,12 @@ Paste the following into your [VSCode `keybindings.json`](https://code.visualstu
         "key": "escape",
         "command": "cursorless.keyboard.escape",
         "when": "cursorless.keyboard.listening && editorTextFocus && !suggestWidgetMultipleSuggestions && !suggestWidgetVisible"
-    },
-    {
-        "key": "backspace",
-        "command": "cursorless.keyboard.targeted.runActionOnTarget",
-        "args": "remove",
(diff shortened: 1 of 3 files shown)


docs/user/experimental/keyboard/modal.md
```diff
@@ -86,7 +80,8 @@ To bind keys that do not have modifiers (eg just pressing `a`), add entries like
     "ac": "copyToClipboard",
     "ax": "cutToClipboard",
     "ap": "pasteFromClipboard",
-    "ad": "followLink"
+    "ad": "followLink",
+    "aw": "remove",
```
josharian, line 84: aw is already wrap, i thought.

aa? as in "aaaaaah it's horrible delete it quick"?

edit: i see you stole "x" below. that seems good. eXcise.

josharian: unfortunate to have this info duplicated here. maybe for the moment just link to the fixture file?

josharian (approved):
i've also mirrored these to my local settings

docs/user/experimental/keyboard/modal.md
josharian: Oh right, duh, I forgot we were in dueling keyboard mode. Ship it.

PR #2203, opened 2024-01-25
Don't clobber keyboard target after executing action

pokey:
## Checklist

### DIFF
--- a/packages/cursorless-vscode-e2e/src/suite/keyboard/basic.vscode.test.ts
+++ b/packages/cursorless-vscode-e2e/src/suite/keyboard/basic.vscode.test.ts
@@ -70,7 +70,7 @@ const testCases: TestCase[] = [
     finalContent: "(a)",
   },
   {
-    name: "reset keyboard target",
+    name: "preserve keyboard target",
     initialContent: "a\n",
     // round wrap air; round wrap <keyboard target>
     keySequence: ["da", "aw", "wp", "aw", "wp"],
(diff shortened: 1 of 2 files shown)


packages/cursorless-vscode/src/keyboard/KeyboardCommandsTargeted.ts
```diff
@@ -261,6 +261,18 @@ export default class KeyboardCommandsTargeted {
       // For some Cursorless actions, it is more convenient if we automatically
       // exit modal mode
       await this.modal.modeOff();
+    } else {
+      // If we're not exiting cursorless mode, preserve the keyboard mark
+      // FIXME: Better to just not clobber the keyboard mark on each action?
```
josharian, line 266: i think it is funny that we outlawed TODO so now just use FIXME.

josharian (approved):
thanks! i just hit this myself, appreciate the fix

packages/cursorless-vscode/src/keyboard/KeyboardCommandsTargeted.ts
pokey: TODO is for things that must be fixed before merge; FIXME is for things we don't like but are ok to merge. I guess we should prob add a note about that to contributor docs

Fwiw we've always tried to use this convention, but now it's enforced so those TODOs can actually give peace of mind

packages/cursorless-vscode-e2e/src/suite/keyboard/basic.vscode.test.ts
```diff
@@ -69,6 +69,13 @@ const testCases: TestCase[] = [
     keySequence: ["da", "aw", "wp"],
     finalContent: "(a)",
   },
+  {
+    name: "reset keyboard target",
```
pokey, line 73:
```suggestion
    name: "preserve keyboard target",
```

PR #2219, opened 2024-02-01
Add name.variable and value.variable

pokey:
I think it's worth distinguishing between an assignment (`foo = bar`) and a variable declarateion (`const foo = bar`) to make it clear we want to support both. I think the name `name.variable` and `value.variable` are clear enough and consistent with `name.function`

## Checklist


AndreasArvidsson:
I don't see you're actually using both assignment and declaration here? You're renaming assignment to variable. Shouldn't we just keep assignment and add variable as an extra facet?
variable => `const foo = 2`
assignment => `foo = 2`

pokey:
For name and value I did keep assignment, and added declaration as additional. Maybe you missed it? For type I just renamed assignment because I don't think type makes sense for assignments

AndreasArvidsson:
Ah right.

AndreasArvidsson (approved):

PR #2221, opened 2024-02-01
Add "paste" how-to guide

pokey:
## Checklist

### DIFF
--- a/docs/user/how-to.md
+++ b/docs/user/how-to.md
@@ -1,0 +1,13 @@
+# How-to guides
+
+This section contains some how-to guides for common tasks.
+
+## How do I paste to a target?
+
+To replace a target with the contents of the clipboard, say `"paste to <target>"`. For example, `"paste to air"`.
+
+:::tip
+
+`"to <target>"` is an example of a Destination. See [Destinations](reference/destinations.md) for more information.
+
+:::
--- a/docs/user/how-to.md
+++ b/docs/user/how-to.md
@@ -4,7 +4,9 @@ This section contains some how-to guides for common tasks.
 
 ## How do I paste to a target?
 
-To replace a target with the contents of the clipboard, say `"paste to <target>"`. For example, `"paste to air"`.
+`"paste to air"`
+
+To replace a target with the contents of the clipboard, say `"paste to <target>"`.
 
 :::info
 
--- a/docs/user/how-to.md
+++ b/docs/user/how-to.md
@@ -6,7 +6,7 @@ This section contains some how-to guides for common tasks.
 
 To replace a target with the contents of the clipboard, say `"paste to <target>"`. For example, `"paste to air"`.
 
-:::tip
+:::info
 
 `"to <target>"` is an example of a Destination. See [Destinations](reference/destinations.md) for more information.
 
--- a/docs/user/reference/destinations.md
+++ b/docs/user/reference/destinations.md
@@ -1,0 +1,19 @@
+# Destinations
+
+## `"to <target>"`
+
+Replaces the target.
+
+For example, `"paste to air"` replaces the token with a hat over the `a` with the contents of the clipboard.
+
+## `"after <target>"`
+
+Inserts after the target, adding delimiters as appropriate.
+
+For example, `"paste after air"` inserts a space after the token with a hat over the `a`, followed by the contents of the clipboard.
+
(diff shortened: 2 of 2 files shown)


docs/user/how-to.md
```diff
@@ -0,0 +1,13 @@
+# How-to guides
+
+This section contains some how-to guides for common tasks.
+
+## How do I paste to a target?
+
+To replace a target with the contents of the clipboard, say `"paste to <target>"`. For example, `"paste to air"`.
```
josharian, line 7: Suggestion: Through, lead with the raw example. So:

> `"paste to air"`
>
> To replace a target with the contents of the clipboard, say `"paste to <target>"`.

(https://en.wikipedia.org/wiki/BLUF_(communication))

josharian: Also, suggestion: make each spoken form word link to reference.

`paste` -> reference for `paste`
`to` -> reference for destination
`air` -> reference for targets, which starts with "air is the generic target used in examples" and a link to talon alphabet and then more

(Or does `paste to` link to ref for `paste`?)

docs/user/reference/destinations.md
```diff
@@ -0,0 +1,19 @@
+# Destinations
+
```
josharian, line 2: Say a sentence or two about what a destination is? (Something along the lines of: "Destinations combine a preposition and a target, to define a place and behavior for a Cursorless action.")

docs/user/reference/destinations.md
```diff
@@ -0,0 +1,19 @@
+# Destinations
```
josharian, line 1: Will there be a page per reference? As a lover of monopages, this makes me sad. But I'll get over it.

We thought the existing docs were mostly reference. Will those...remain? Have their content migrated to these new pages?

Don't feel strongly, just curious about the plan.

josharian (approved):
Yay! The first PR is the hardest. Nice start.

docs/user/how-to.md
pokey: > Suggestion: Throughout, lead with the raw example. So:
> 
> > `"paste to air"`
> > To replace a target with the contents of the clipboard, say `"paste to <target>"`.
> 
> (https://en.wikipedia.org/wiki/BLUF_(communication))

done.

> Also, suggestion: make each spoken form word link to reference.
> 
> `paste` -> reference for `paste` `to` -> reference for destination `air` -> reference for targets, which starts with "air is the generic target used in examples" and a link to talon alphabet and then more
> 
> (Or does `paste to` link to ref for `paste`?)

This is a nice idea, but a) I'm not sure it will be clear that each separate word is a link, and b) I'm not sure exactly how to do that within an inline code snippet. I guess I can not use inline code snippet here. We could prob create a special element we use for all voice commands. And we might even use a different one for voice commands that are just part of a command vs a full command you can say

Another issue is that linking to "paste" is brittle right now, because it's just a header. If we use proper pages for each action / scope / whatever, then vscode will automatically update references if I move a file

docs/user/reference/destinations.md
pokey: As mentioned in other comment, there are advantages to each concept having its own page. It can be a canonical URI, and even if we do move it, VSCode will automatically update references. But I agree it hurts spontaneous discovery / casual scrolling

docs/user/how-to.md
josharian: > I'm not sure it will be clear that each separate word is a link

If it is underlined, I think it'll be clear enough?

All reasonable concerns. We can revisit later as desired; retrofitting this in would be only mildly annoying.

docs/user/reference/destinations.md
josharian: We can also always write a little script that accumulates all the little pages into a monopage for folks who like that sort of thing.

Let's move forward with the path of least resistance and edit from there.

AndreasArvidsson: I think I prefer a single page that has in index and maybe a very short summary. I prefer if each concept and its description has a single page with a canonical uri.

AndreasArvidsson (approved):

docs/user/how-to.md
pokey: https://github.com/cursorless-dev/cursorless/issues/2232

PR #2222, opened 2024-02-02
WIP: Support Elm Programming Language

anmolitor:
## What

Adds support for the Elm programming language.
I have tried to bring https://github.com/cursorless-dev/cursorless/pull/891 up to date.

Is there a good way to unit test the generated tree sitter queries?
I am currently struggling to extend the functionality - Elm is quite different in its syntax from most currently supported languages. For example, `record_expr[field][expression]` for `change value` inside of a map/struct { a = 1, b = 2 } currently always results in { a = , b = 2 }, even if the cursor/target is on "b".
I read https://www.cursorless.org/docs/contributing/parse-tree-patterns/ and set up tree-sitter-node locally with tree-sitter-elm to easily inspect the trees, but there are a lot of utility functions in use in the cursorless codebase regarding NodeMatchers that I do not fully understand. Is there some additional documentation/examples that I haven't found yet?

EDIT:
I have looked at the open Haskell MR which has a similar syntax. Scm Queries seem to be easier to debug to I will be trying that for now.

## Checklist

- [ ] Added `@textFragment` captures. Usually you want to put these on comment and string nodes. This enables `"take round"` to work within comments and strings.
- [ ] Supported` "item"` both for map pairs and list entries (with tests of course)

### DIFF
--- a/docs/user/experimental/keyboard/modal.md
+++ b/docs/user/experimental/keyboard/modal.md
@@ -29,18 +29,12 @@ Paste the following into your [VSCode `keybindings.json`](https://code.visualstu
         "key": "escape",
         "command": "cursorless.keyboard.escape",
         "when": "cursorless.keyboard.listening && editorTextFocus && !suggestWidgetMultipleSuggestions && !suggestWidgetVisible"
-    },
-    {
-        "key": "backspace",
-        "command": "cursorless.keyboard.targeted.runActionOnTarget",
-        "args": "remove",
-        "when": "cursorless.keyboard.modal.mode && editorTextFocus"
     }
 ```
 
 Any keybindings that use modifier keys should go in `keybindings.json` as well, with a `"when": "cursorless.keyboard.modal.mode` clause.
 
-The above allows you to press `ctrl-c` to switch to Cursorless mode, `escape` to exit Cursorless mode, and `backspace` to issue the delete action while in Cursorless mode.
+The above allows you to press `ctrl-c` to switch to Cursorless mode and `escape` to exit Cursorless mode.
 
 If you're already in Cursorless mode, pressing `ctrl-c` again will target the current selection, which is useful if you have moved the cursor using your mouse while in Cursorless mode, and want to target your new cursor position.
 
@@ -86,7 +80,8 @@ To bind keys that do not have modifiers (eg just pressing `a`), add entries like
     "ac": "copyToClipboard",
     "ax": "cutToClipboard",
     "ap": "pasteFromClipboard",
-    "ad": "followLink"
+    "ad": "followLink",
+    "aw": "remove",
   },
   "cursorless.experimental.keyboard.modal.keybindings.color": {
     "d": "default",
--- a/packages/common/src/extensionDependencies.ts
+++ b/packages/common/src/extensionDependencies.ts
@@ -6,6 +6,7 @@ export const extensionDependencies = [
   "scala-lang.scala", // scala
   "mrob95.vscode-talonscript", // talon
   "jrieken.vscode-tree-sitter-query", // scm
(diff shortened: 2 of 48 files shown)


pokey (commented):
Thanks for the PR! It would be great if you could migrate these patterns to the new scm query syntax, as we're in the process of removing the old pattern language completely. See [the docs](https://www.cursorless.org/docs/contributing/adding-a-new-language/), other [example PRs](https://github.com/cursorless-dev/cursorless/pulls?q=is%3Apr+is%3Aclosed+label%3Ascope-migration), eg [json](https://github.com/cursorless-dev/cursorless/pull/2042/files) and [java](https://github.com/cursorless-dev/cursorless/pull/2038/files), and feel free to drop into a meet-up!

anmolitor:
I added most of the scm queries that I understood. Currently I am struggling with scopes - many of the previously recorded tests are now failing because of "NoContainingScopeError"s.
I could fix some of them by adding @_.domain on the outer query, which I saw in other languages (and still don't fully understand), but others were not so easy to fix. 
In particular, I have no idea how to fix argumentAndParameter tagging.
If I match all params using (_)*, cursorless throws an error because there are multiple matches when it expected only one.
If I instead list every possible param by hand as seperate queries, I get "NoContainingScopeError: Couldn't find containing argumentOrParameter".
Using just (_) inside of the `functionDeclarationLeft` node gives the same scope related error.

For reference, here is what I thought should work:
```
;; namedFunction

;;!! f x y = x
;;!  ^^^^^^^^^
(value_declaration 
  functionDeclarationLeft: (function_declaration_left
    (lower_case_identifier) @functionName @name
    pattern: (_) @argumentOrParameter
  )
  body: (_) @namedFunction.interior
) @namedFunction @functionName.domain
```

Can you give me some hints? I've tried looking at other languages, but most have a mixed bag of scm and ts queries.

tankorsmash:
I think `@.domain` represents the domain of the scope, which is sort of like the area around the actual scope, as a sort of way of getting to the the scope that you're in, without actually literally being in the scope. there's a video of it [here](https://youtu.be/BbXEzUrf5lU?t=113)

pokey:
@anmolitor I think you're running into https://github.com/tree-sitter/tree-sitter/issues/2984. You can work around it as follows:

```scm
(function_declaration_left (_) @argumentOrParameter (#not-type? @argumentOrParameter lower_case_identifier))
```

pokey:
also, be sure to use [andreas's extension](https://marketplace.visualstudio.com/items?itemName=AndreasArvidsson.andreas-talon) to format your code

tankorsmash:
@anmolitor I forked this PR to see if I could help. I couldn't get far, but I did manage to the linked formatter going. It took a while to run on the entire repo but it seems like all it needed to format was the `elm.scm` file. https://github.com/anmolitor/cursorless/compare/main...tankorsmash:cursorless:main

steps:
 * install that extension
 * right click `elm.scm` in tree
 * very bottom of the dropdown is 'Format' which opens, formats and closes the file

anmolitor:
Wow thanks for the help everyone!
I'm currently dedicating my free time to prepare
 https://github.com/anmolitor/protoc-gen-elm 4.0.0,
but that should only take a few more days. I'll come back to work on this PR right afterwards :)

pokey:
@anmolitor if you don't think you'll have capacity to bring this one home, maybe @tankorsmash should open a new PR so that the formatting changes can be incorporated and push things forward? (assuming @tankorsmash has time)

tankorsmash:
I'll take another stab at this this weekend, if @anmolitor is busy!

anmolitor:
Sorry for the delay. I think I'm currently too busy to complete this PR so I'll close this for now. @tankorsmash feel free to use my code as a starting point :)

PR #2227, opened 2024-02-08
add simple head/tail modifiers

josharian:
## Checklist


josharian:
cc @pokey -- not for review yet, a 📌 for a conversation for a possibly terrible idea i had about handling non-simple head/tail

pokey:
I believe we discussed this and decided that it would be better to have one token to indicate the start of a head or tail modifier and one token to indicate the end. If nothing falls in between then it is a head or tail of line, otherwise it's a head or tail of whatever modifier sequence falls in between. Does that line up with your recollection?

pokey:
Decided to opt for making it easy to tail one modifier because I think "head inside" is at least as common as "head line" for me tbh. See https://github.com/cursorless-dev/cursorless/pull/2428

PR #2235, opened 2024-02-10
Fallback to Talon actions when focus is not on the text editor

AndreasArvidsson:
Edit operations supported by community will now work in vscode outside of the text editor. eg the search widget
`take line`
`chuck token`

Everything appears to be working when I have tested it. With that said I have not tested on community and we should probably have a discussion about some of the finer details of this.

## Checklist


cursorless-talon/src/fallback.py
```diff
@@ -0,0 +1,76 @@
...
+    "clearAndSetSelection": actions.edit.delete,
+    "remove": actions.edit.delete,
+    "applyFormatter": actions.user.reformat_selection,
+    "editNewLineBefore": actions.edit.line_insert_up,
+    "editNewLineAfter": actions.edit.line_insert_down,
+    "nextHomophone": actions.user.homophones_cycle_selected,
+    # "replaceWithTarget": replace_with_target,
+}
+
+scope_callbacks = {
```
pokey, line 18: `modifier_callbacks`

pokey:
notes from meet-up:

- [ ] Add tests by adding field called `fallback` to recorded test case fixture
- [ ] Return untouched modifier list from extension rather than creating a whole new modifier representation. Some tweaking might be necessary to handle things like implicit destination and auto-token expansion
- [ ] Add modifier called eg `expandToTokenIfEmpty`
- [ ] Return `actions` from `createCommandRunner` so that we can call `getText` action directly instead of running it as a command

cursorless-talon/src/fallback.py
```diff
@@ -0,0 +1,76 @@
...
+    if action == "insert":
+        return lambda: actions.insert(fallback["text"])
+
+    if action == "callAsFunction":
+        return lambda: callAsFunction(fallback["callee"])
+
+    if action == "wrapWithPairedDelimiter":
+        return lambda: actions.user.delimiters_pair_wrap_selection_with(
+            fallback["left"], fallback["right"]
+        )
```
pokey, line 59: figure out how this should work in community

cursorless-talon/src/fallback.py
```diff
@@ -0,0 +1,76 @@
...
+    "setSelection": actions.skip,
+    "copyToClipboard": actions.edit.copy,
+    "cutToClipboard": actions.edit.cut,
+    "pasteFromClipboard": actions.edit.paste,
+    "clearAndSetSelection": actions.edit.delete,
+    "remove": actions.edit.delete,
+    "applyFormatter": actions.user.reformat_selection,
+    "editNewLineBefore": actions.edit.line_insert_up,
+    "editNewLineAfter": actions.edit.line_insert_down,
+    "nextHomophone": actions.user.homophones_cycle_selected,
```
pokey, line 14: figure this out in community

packages/cursorless-engine/src/api/CursorlessEngineApi.ts
```diff
@@ -34,15 +34,28 @@ export interface CommandApi {
...
    * the command args are of the correct shape.
    */
-  runCommandSafe(...args: unknown[]): Promise<unknown>;
+  runCommandSafe(...args: unknown[]): Promise<CommandResponse | unknown>;
 }
 
+export type CommandResponse = { returnValue: unknown } | { fallback: Fallback };
+
+export type Fallback =
+  | { action: string; scope: string | null }
```
pokey, line 49: maybe make the `string` into a stronger type

AndreasArvidsson: done

cursorless-talon/src/fallback.py
AndreasArvidsson: Added our own implementation file. Maybe we should just add this wont to community?

packages/cursorless-engine/src/core/getCommandFallback.ts
```diff
@@ -0,0 +1,210 @@
...
+function getScopeFromDestination(
+  destination: DestinationDescriptor,
+): string | null {
+  if (destination.type === "primitive") {
+    return getScopeFromTarget(destination.target);
+  }
+  return null;
+}
+
+function getScopeFromTarget(target: PartialTargetDescriptor): string | null {
```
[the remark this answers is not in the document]
AndreasArvidsson: don

cursorless-talon/src/fallback.py
AndreasArvidsson: done

AndreasArvidsson:
@pokey Recording test was unfortunately not as easy as we assumed. We never actually run the command runner when we do a fallback. ergo the test case recorder is never called. 

https://github.com/cursorless-dev/cursorless/blob/407dc945a32ab14ee92c1f8b01a19bc4fc894ea1/packages/cursorless-engine/src/runCommand.ts#L83-L89

packages/cursorless-engine/src/cursorlessEngine.ts
```diff
@@ -85,18 +88,21 @@ export function createCursorlessEngine(
...
           hatTokenMap,
           snippets,
           storedTargets,
           languageDefinitions,
           rangeUpdater,
           commandRunnerDecorators,
-          ensureCommandShape(args),
+          command,
         );
+        return unwrapCommandResponse(command, response);
```
pokey, line 105: do this in runCommand

packages/cursorless-engine/src/cursorlessEngine.ts
```diff
@@ -117,6 +123,19 @@ export function createCursorlessEngine(
   };
 }
 
+async function unwrapCommandResponse(
+  command: Command,
+  response: CommandResponse,
+): Promise<CommandResponse | unknown> {
+  if (useFallback(command)) {
```
pokey, line 130: rename `useFallback` to `clientSupportsFallback`

packages/cursorless-engine/src/testCaseRecorder/TestCaseRecorder.ts
pokey, line 1: in this file, detect what is focused, and add a field to payload saying temrinal or other if not editor

then check that in tests rather than looking for fallback

cursorless-talon/src/fallback.py
```diff
@@ -0,0 +1,107 @@
+from typing import Callable
+
+from talon import actions
+
```
pokey, line 4: import client version from wherever it's defined here in talon, and add assertion here that it is equal to 7, and add a comment. That will ensure we remember to update this stuff if we change request payload shape

packages/cursorless-engine/src/cursorlessEngine.ts
AndreasArvidsson: done

packages/cursorless-engine/src/testCaseRecorder/TestCaseRecorder.ts
AndreasArvidsson: done

cursorless-talon/src/fallback.py
AndreasArvidsson: done

AndreasArvidsson:
@pokey ready for review

cursorless-talon/src/fallback.py
```diff
@@ -0,0 +1,112 @@
...
+action_callbacks = {
+    "getText": lambda: [actions.edit.selected_text()],
+    "setSelection": actions.skip,
+    "copyToClipboard": actions.edit.copy,
+    "cutToClipboard": actions.edit.cut,
+    "pasteFromClipboard": actions.edit.paste,
+    "clearAndSetSelection": actions.edit.delete,
+    "remove": actions.edit.delete,
+    "editNewLineBefore": actions.edit.line_insert_up,
+    "editNewLineAfter": actions.edit.line_insert_down,
```
pokey, line 19: What do we do about `"pour token"`?

cursorless-talon/src/fallback.py
```diff
@@ -0,0 +1,112 @@
+from typing import Callable
+
+from talon import actions
+
+from .versions import COMMAND_VERSION
+
+# This ensures that we remember to update fallback if the response payload changes
+assert COMMAND_VERSION == 7
+
+action_callbacks = {
```
pokey, line 10: Would be cool to support clone; fine leaving out of scope for this PR tho. We should prob file with some other improvements we'd like to make

packages/cursorless-engine/src/runCommand.ts
```diff
@@ -36,7 +44,7 @@ export async function runCommand(
   rangeUpdater: RangeUpdater,
   commandRunnerDecorators: CommandRunnerDecorator[],
   command: Command,
-): Promise<unknown> {
+): Promise<CommandResponse | unknown> {
```
pokey, line 47: Unfortunate that we will have to keep adding to this union as we change our return shape. Would be cool to use type mapping with discriminated union based on input command version number, but prob not worth the trouble. We can revisit if it becomes annoying

cursorless-talon/src/fallback.py
AndreasArvidsson: Not supported at the moment

pokey (commented):
Looking good! I made a few tweaks, and left a few minor comments. I tested locally and it works well, tho I find it's a bit slow. Is it pretty snappy for you?

pokey:
so we'll just delete `nextHomophone`

pokey:
@pokey to do some timestamps to see what's causing slowness on his computer

packages/common/src/types/command/CommandV7.types.ts
```diff
@@ -0,0 +1,8 @@
+import type { CommandV6 } from "./CommandV6.types";
+
+export interface CommandV7 extends Omit<CommandV6, "version"> {
```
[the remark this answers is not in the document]
AndreasArvidsson: I think so. Otherwise we just have to copy identical types

cursorless-talon/src/fallback.py
AndreasArvidsson: I think I will leave this for a follow up.

packages/cursorless-engine/src/runCommand.ts
AndreasArvidsson: Agreed

AndreasArvidsson:
> so we'll just delete `nextHomophone`

done

AndreasArvidsson:
> @pokey to do some timestamps to see what's causing slowness on his computer

I haven't noticed any problems

saidelike:
> > @pokey to do some timestamps to see what's causing slowness on his computer
> 
> I haven't noticed any problems

I haven't noticed any problem either fwiw

saidelike:
Not sure if you want to include it in this PR but I added support for
- pre/post actions
- this target
https://github.com/saidelike/cursorless-talon/blob/main/src/andreas/fallback.py#L33

AndreasArvidsson:
`this` is already supported
https://github.com/cursorless-dev/cursorless/blob/4a9386cc0dc7120b106ee4e2b917a8c39a4b190a/cursorless-talon/src/fallback.py#L43

AndreasArvidsson:
pre and post fallback actions added

pokey:
Ok I did some profiling, and it's actually my `select_word` that's slow. Prob has to do with my character dance

pokey (commented):
🚀

pokey (approved):
🚀

PR #2254, opened 2024-02-27
Added spread support for relative and ordinal scopes

### LINKED ISSUE #1514: Create `"spread"` version of relative / ordinal modifiers
Eg `"take spread three tokens"`, `"take spread first three tokens"`, etc, which instead of creating a range between scopes, yields each scope in the range.  It's just shorthand for `"every <scope> <modifier>"` but is common enough to be useful.

Might consider just using the word `"every"`, eg `"take every three tokens"`?  Or `"split"`?

Could be extra boolean field on these scopes in the Cursorless api

Could steal code from `"instance"`, and potentially would allow us to have `"instance"` reuse more of our existing code, as one of the primary differences in the `"instance"` pseudo-scope is that it defaults to a `"spread"` scope

AndreasArvidsson:
`"change every two tokens"`

Fixes #1514

## Checklist

- [x] I have updated the [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [x] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) 


pokey:
update from meet-up:

- switch spoken form to `"every two tokens"`. In the future, if we want every other token, that would be `"every second token"`, and every pair of tokens would be `"two tokens every second token"`
- switch internal identifier to `isEvery`

AndreasArvidsson:
Updated!

pokey (commented):
just posting old pending comments; ignore for now

pokey (commented):
Ok made some minor tweaks. Also, it looks like your relative exclusive code is broken; it only uses the first and the last targets. I added some failing test cases so you can see what I mean. Fwiw I ran talon tests locally after my changes and they passed.

pokey (approved):

pokey:
- [ ] update cheatsheet

pokey (approved):
ok should be good to go @AndreasArvidsson ! Was a good thing I updated cheatsheet—it was broken due to the fact you factored `simple_modifier` into two lists 😬

PR #2260, opened 2024-03-06
Fix FontMeasurementsImpl to work in a web environment

bjaspan:
FontMeasurementsImpl currently uses an inline script to compute font size ratios which is not compatible with a web environment due to Content Security Policy (CSP) rules.  This PR switches to a separate script file and adds a CSP rule for it according to the instructions in https://code.visualstudio.com/api/extension-guides/webview#content-security-policy.

To test, change the `Editor: Font Family` setting (you can just append `, x` to the end or make any other change).  Success is seeing a tab flash and disappear; failure is the tab showing "Loading Cursorless..." remaining open.


pokey (approved):
works fine on desktop; I'll take your word it works on web

PR #2263, opened 2024-03-11
initial changes to make document.getText() async

saidelike:
## Checklist

Base commit is f92f8890c3a21e6b74b470411de1c253236a2cc7

I created this draft PR in order to discuss the impact of trying to just change one function to be "async". This is not finished yet, but this gives an idea of the number of changes. 

It also makes the code less readable when dealing with "map", "flatMap", etc. 

There also edge cases where I am not sure how to deal with it yet. 


packages/cursorless-engine/src/actions/BringMoveSwap.ts
```diff
@@ -63,16 +63,16 @@ abstract class BringMoveSwap {
...
     destinations: Destination[],
-  ): ExtendedEdit[] {
+  ): Promise<ExtendedEdit[]> {
     const usedSources: Target[] = [];
     const results: ExtendedEdit[] = [];
     const shouldJoinSources =
       sources.length !== destinations.length && destinations.length === 1;
 
-    sources.forEach((source, i) => {
+    sources.forEach(async (source, i) => {
```
pokey, line 75:
```
for (let i = 0; i < source.length; i++) {
    const source = sources[i];
```

packages/cursorless-engine/src/processTargets/modifiers/InstanceStage.ts
```diff
@@ -45,6 +47,7 @@ export class InstanceStage implements ModifierStage {
     }
   }
 
+  // TODO: async I don't know how to deal with that
```
saidelike, line 50: @pokey do you know how to deal with that? it is a promise of iterator

image

packages/cursorless-engine/src/processTargets/targets/DestinationImpl.ts
```diff
@@ -31,12 +33,21 @@ export class DestinationImpl implements Destination {
       indentationString ?? this.isLineDelimiter
         ? getIndentationString(target.editor, target.contentRange)
         : "";
-    this.insertionPrefix =
-      target.prefixRange != null
-        ? target.editor.document.getText(target.prefixRange)
+  }
+
+  // TODO: we can't have "async/await" in a constructor so we initialize elsewhere
```
saidelike, line 38: @pokey do we like this way to call async functions when constructing an object?

packages/cursorless-engine/src/processTargets/targets/InteriorTarget.ts
```diff
@@ -12,20 +16,28 @@ export class InteriorTarget extends BaseTarget<InteriorTargetParameters> {
...
-  constructor(parameters: InteriorTargetParameters) {
-    super({
+  // TODO: don't use the constructor, instead we need to create the object with create()
+  // Are we happy about that?
+  constructor(parameters: InteriorTargetParameters & CommonTargetParameters) {
+    super(parameters);
+    this.fullInteriorRange = parameters.fullInteriorRange;
+  }
+
+  public static async create(
```
saidelike, line 26: @pokey or do we prefer this way to call async functions when constructing an object?

packages/cursorless-engine/src/processTargets/targets/DestinationImpl.ts
pokey: I wonder if we could get the insertion prefix lazily when we actually need it?

packages/cursorless-engine/src/processTargets/targets/InteriorTarget.ts
pokey: this is probably better, as it avoids client needing to call initializer. I'd also make the constructor private to force them to call `create`

packages/cursorless-engine/src/actions/BringMoveSwap.ts
```diff
@@ -82,16 +82,16 @@ abstract class BringMoveSwap {
...
-            .map((source, i) => {
-              const text = source.contentText;
+            .map(async (source, i) => {
+              const text = await source.contentText;
               const delimiter =
                 (destination.isRaw ? null : destination.insertionDelimiter) ??
                 (source.isRaw ? null : source.insertionDelimiter);
               return i > 0 && delimiter != null ? delimiter + text : text;
             })
             .join("");
```
conartist6, line 92: Isn't this join happening on an array of promises now?

pokey:
update from meet-up:

- Let's start by just grabbing the full buffer for active editor from neovim right before we execute the command. Not the most efficient thing, but seems like easiest way forward for now. We'll grab implementation from https://github.com/cursorless-dev/cursorless/pull/1899/files

saidelike:
closing as over complicated and won't be needed after all

PR #2273, opened 2024-03-26
Upgrade all dependencies

pokey:
This PR upgrades all of our dependencies to the latest versions, including a few major dependency version bumps. Figured it was good housekeeping to do once in a while

I pinned node and vscode, but otherwise everything is upgraded

## Checklist


packages/common/src/ide/fake/FakeConfiguration.ts
pokey, line 1: Something changed in lodash types for `get`, or maybe typescript 5.4, and I couldn't get these types to work. I don't think it's too bad of a cast, but @auscompgeek / @SimeonC if you wanna bail me out I wouldn't complain 😅

pokey:
Managed to get down to one cast; ready for a review @AndreasArvidsson

packages/common/src/ide/fake/FakeConfiguration.ts
SimeonC: I think it's probably a typescript change - `get` used to return `any` I think and I know ts 5.4 has done a lot with inference so it probably change the return type to `unknown` instead of `any`. I think this one is pretty normal for `get` type functions which are super hard to type anyway.

AndreasArvidsson (approved):

PR #2285, opened 2024-04-05
enforce new selection for "remove" action

saidelike:
## Checklist

From debugging the vscode cursorless extension, it seems the selection is automatically updated by vscode itself when dealing with the "remove" action while calling editBuilder.delete() (TextEditorEdit). 

However it is not the case for neovim so it was using the old selection.

I fixed it so cursorless always enforce selection for that "remove" action. Tested on neovim and vscode.

### DIFF
--- a/packages/cursorless-engine/src/actions/Remove.ts
+++ b/packages/cursorless-engine/src/actions/Remove.ts
@@ -1,17 +1,24 @@
-import { FlashStyle } from "@cursorless/common";
+import {
+  FlashStyle,
+  RangeExpansionBehavior,
+  Selection,
+  TextEditor,
+} from "@cursorless/common";
 import { flatten, zip } from "lodash";
 import { RangeUpdater } from "../core/updateSelections/RangeUpdater";
-import { performEditsAndUpdateRanges } from "../core/updateSelections/updateSelections";
+import { performEditsAndUpdateSelectionsWithBehavior } from "../core/updateSelections/updateSelections";
 import { RawSelectionTarget } from "../processTargets/targets";
 import { ide } from "../singletons/ide.singleton";
 import { Target } from "../typings/target.types";
 import { flashTargets, runOnTargetsForEachEditor } from "../util/targetUtils";
 import { unifyRemovalTargets } from "../util/unifyRanges";
 import { SimpleAction, ActionReturnValue } from "./actions.types";
+import { setSelectionsWithoutFocusingEditor } from "../util/setSelectionsAndFocusEditor";
 
 export default class Delete implements SimpleAction {
   constructor(private rangeUpdater: RangeUpdater) {
     this.run = this.run.bind(this);
+    this.runForEditor = this.runForEditor.bind(this);
   }
 
   async run(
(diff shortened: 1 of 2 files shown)


packages/cursorless-engine/src/actions/Remove.ts
```diff
@@ -28,28 +35,45 @@ export default class Delete implements SimpleAction {
...
+
+  private async runForEditor(editor: TextEditor, targets: Target[]) {
+    const edits = targets.map((target) => target.constructRemovalEdit());
+
+    const cursorSelections = { selections: editor.selections };
+    const editSelections = {
+      selections: edits.map(
+        ({ range }) => new Selection(range.start, range.end),
+      ),
+      rangeBehavior: RangeExpansionBehavior.closedClosed,
```
pokey, line 52: I don't think behaviour makes a difference for deletions; only insertions. Tho I would assume you switched to this more complex api because something wasn't working? Or maybe you copy-pasted this code from an action that performs insertions? In which case I'd try to switch back to the simpler api

pokey (commented):
Good stuff. @AndreasArvidsson and I had discussed at one point having a version of `performEditsAndUpdateSelections` that does this automatically, as I believe it would be useful in other contexts, ie if you copy-pasted this code from somewhere else they could both use that abstraction

packages/cursorless-engine/src/actions/Remove.ts
saidelike: The original code was using `performEditsAndUpdateRanges` https://github.com/cursorless-dev/cursorless/blob/main/packages/cursorless-engine/src/actions/Remove.ts#L35C39-L35C66 which calls `rangesToSelectionInfos` https://github.com/cursorless-dev/cursorless/blob/main/packages/cursorless-engine/src/core/updateSelections/updateSelections.ts#L309 which uses `RangeExpansionBehavior.closedClosed` https://github.com/cursorless-dev/cursorless/blob/main/packages/cursorless-engine/src/core/updateSelections/updateSelections.ts#L100 so didn't want to change that behaviour from original code.

OK I guess I could use `performEditsAndUpdateSelections` instead of `performEditsAndUpdateSelectionsWithBehavior`?

saidelike: And yes I had copied it from the "insert" action.

pokey: Yeah I'd try using `performEditsAndUpdateSelections` just because it's a simpler api

packages/cursorless-engine/src/actions/Remove.ts
```diff
@@ -28,28 +30,41 @@ export default class Delete implements SimpleAction {
...
 
     return { thatTargets };
   }
+
+  private async runForEditor(editor: TextEditor, targets: Target[]) {
+    const edits = targets.map((target) => target.constructRemovalEdit());
+
+    const cursorSelections = editor.selections;
+    const editSelections = edits.map(
+      ({ range }) => new Selection(range.start, range.end),
```
pokey, line 44: I believe there's just a `range.toSelection` function you can use for this

packages/cursorless-engine/src/actions/Remove.ts
```diff
@@ -28,28 +30,41 @@ export default class Delete implements SimpleAction {
...
+    );
+    const editableEditor = ide().getEditableTextEditor(editor);
+
+    const [updatedEditorSelections, updatedEditSelections]: Selection[][] =
```
pokey, line 48: I think I'd use `updatedCursorSelections` for consistency

pokey (commented):
Looks good with a couple more minor tweaks

Btw what would happen if you just moved this work into the `performEdits` function? Then we could remove it from the place you stole this code from. @AndreasArvidsson do you see any reason not to always update the cursor position the way we want? I guess for eg `"change" / "clear"` it's a waste of work. But my leaning would be an option to disable the behaviour when you know you're going to move the cursor anyway, but to do it by default if not

saidelike:
> Looks good with a couple more minor tweaks
> 
> Btw what would happen if you just moved this work into the `performEdits` function? Then we could remove it from the place you stole this code from. @AndreasArvidsson do you see any reason not to always update the cursor position the way we want? I guess for eg `"change" / "clear"` it's a waste of work. But my leaning would be an option to disable the behaviour when you know you're going to move the cursor anyway, but to do it by default if not

I propose to look into that in a future PR.

packages/cursorless-engine/src/actions/Remove.ts
saidelike: done

AndreasArvidsson:
> Looks good with a couple more minor tweaks
> 
> Btw what would happen if you just moved this work into the `performEdits` function? Then we could remove it from the place you stole this code from. @AndreasArvidsson do you see any reason not to always update the cursor position the way we want? I guess for eg `"change" / "clear"` it's a waste of work. But my leaning would be an option to disable the behaviour when you know you're going to move the cursor anyway, but to do it by default if not

The only reason I can think is if you make multiple edits and just want to update the selection at the end. I have no idea if we do that in any place and if we find that we need that behavior we could always add a parameter. Updating selection by default sounds reasonable.

pokey (approved):

PR #2289, opened 2024-04-07
github: Create ADR discussion category form

auscompgeek:
This codifies our ADR template (#1210) using [GitHub discussion category forms](https://docs.github.com/en/discussions/managing-discussions-for-your-community/creating-discussion-category-forms).

Note that this changes the heading levels (they'll all be `h3` I think), which means the **Options considered** heading will no longer be a subheading of the **Context**. This is probably fine – a lot of the other Markdown ADR templates have this section at the same level anyway.


pokey (approved):
neat! didn't know this was a thing

PR #2217, opened 2024-02-01
Bump CI node version to 20

pokey:
To get rid of deprecation warnings, eg https://github.com/cursorless-dev/cursorless/actions/runs/7742075291. See https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/ for more info

## Checklist

- [-] I have added [tests](https://www.cursorless.org/docs/contributing/test-case-recorder/)
- [-] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)
- [-] I have not broken the cheatsheet

PR #2297, opened 2024-04-17
Fix Windows CI by pinning node version

pokey:
Windows CI started failing due to a node.js regression. We should really be pinning node, like we pin pnpm and all our pnpm dependencies. The only things we don't want to pin are OS version and vscode version

This PR pins node and adds some debug output for the future

## Checklist

- [x] Pin node using technique removed from #2217
- [x] Double-check pnpm pinning
- [x] Keep some of the print statements here
- [x] Remove code that slices args
- [ ] Add checklist somewhere of versions to check when this kind of thing happens
  - [ ] node version (note this will be pinned)
  - [ ] OS version
  - [ ] vscode version


pokey:
@AndreasArvidsson on second thought, I'm not convinced adding a checklist with two things in it somewhere in our docs will be helpful, but if your gut still says this will be useful I'm happy to drop it somewhere

AndreasArvidsson:
> @AndreasArvidsson on second thought, I'm not convinced adding a checklist with two things in it somewhere in our docs will be helpful, but if your gut still says this will be useful I'm happy to drop it somewhere

Maybe not. I could go either way.

AndreasArvidsson (approved):

AndreasArvidsson (approved):

Issue #1984, opened 2023-10-31
Use or explicitly ignore all promises


bjaspan:
From Google's internal Typescript conformance checks (I'll look to see if these are open-source somewhere):

"When using async / await, it can be easy to forget to await a promise which can result in async tasks running in an unexpected order. Thus, we require that every promise in an async function is consumed in some way, so that there's a well-defined order in which async tasks are resolved. To fix this check, you can do one of the following:
* Remove async from the function, and instead use normal promise chaining.
await the promise.
* Assign the promise to a variable and await it later.
*  If you really want to use async and don't want to await a particular Promise, you can use the void operator to signal your intent.
   ```
   void returnsPromise();
   ```
   This explicitly makes your intention to not await the promise clear."

```
third_party/vscode_ext/cursorless_vscode/packages/cursorless-engine/src/actions/Find.ts:50:7 - error TS21225: [tsetse] All Promises in async functions must either be awaited or used in an expression.
	See http://go/ts-conformance#must-use-promises

 50       showWarning(
          ~~~~~~~~~~~~
 51         ide().messages,
    ~~~~~~~~~~~~~~~~~~~~~~~
... 
 53         "Search text is longer than 200 characters; truncating",
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 54       );
    ~~~~~~~
third_party/vscode_ext/cursorless_vscode/packages/cursorless-engine/src/actions/GenerateSnippet/GenerateSnippet.ts:90:5 - error TS21225: [tsetse] All Promises in async functions must either be awaited or used in an expression.
	See http://go/ts-conformance#must-use-promises

90     flashTargets(ide(), targets, FlashStyle.referenced);
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
third_party/vscode_ext/cursorless_vscode/packages/cursorless-engine/src/actions/ShowParseTree.ts:52:5 - error TS21225: [tsetse] All Promises in async functions must either be awaited or used in an expression.
	See http://go/ts-conformance#must-use-promises

52     ide().openUntitledTextDocument({
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
53       language: "markdown",
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~
54       content: results.join("\n\n"),
[...]

PR #2300, opened 2024-04-18
some missing await

saidelike:
## Checklist

### DIFF
--- a/packages/cursorless-vscode/src/extension.ts
+++ b/packages/cursorless-vscode/src/extension.ts
@@ -92,7 +92,7 @@ export async function activate(
     runIntegrationTests,
     addCommandRunnerDecorator,
     customSpokenFormGenerator,
-  } = createCursorlessEngine(
+  } = await createCursorlessEngine(
     treeSitter,
     normalizedIde,
     hats,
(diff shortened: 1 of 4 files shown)


AndreasArvidsson:
We should really put `void` before a promise we don't want to await

packages/cursorless-vscode/src/extension.ts
```diff
@@ -92,7 +92,7 @@ export async function activate(
     runIntegrationTests,
     addCommandRunnerDecorator,
     customSpokenFormGenerator,
-  } = createCursorlessEngine(
+  } = await createCursorlessEngine(
```
saidelike, line 95: Ok we don't want to wait on hat and snippets but maybe we want this one which is to wait that the cursorless engine is created? :)

auscompgeek:
> We should really put `void` before a promise we don't want to await

there's a lint rule for that :smile: #1984

packages/cursorless-vscode/src/extension.ts
auscompgeek: If we have no `await`s in `createCursorlessEngine` then it doesn't need to be `async`. If it's not `async` it won't return a Promise.

AndreasArvidsson:
> > We should really put `void` before a promise we don't want to await
> 
> there's a lint rule for that 😄 #1984

Of course there is :D

packages/cursorless-vscode/src/extension.ts
pokey: Yes, as @auscompgeek points out, the function is not async on `main`, only in this PR for those two `await` statements that you added here. So no need to `await` it

PR #2301, opened 2024-04-18
isTesting versus ide.runMode

saidelike:
## Checklist

I initially wanted to get rid of `isTesting()` entirely an only rely on `ide.runMode === "test"` and `ide.runMode === "production"` but it seems that some tests actually don't run as vscode extensions we need the `CURSORLESS_TEST` environment variable (see `launch.json`)

I thought about moving the `isTesting()` to `cursorless-engine` since it holds the ide and export that specific function to the rest of the world. However there is at least one instance of `isTesting()` in a package that doesn't import `cursorless-engine` (`packages\cursorless-vscode\src\ide\vscode\VscodeEnabledHatStyleManager.ts`) so we can't do that. 

That is why I came up with this following approach, even though I'm not entirely happy about it. I can't think about any better way, but at least I think it's better than the current way.  

it is all the more not ideal because for neovim I'm using `CURSORLESS_MODE` set to `development` or `test` to set the neovimIde.runMode since it's not given by vscode. consequently I'm never gonna use `isTesting()`. Consequently we could potentially move this one to vscode specific packages.

### DIFF
--- a/packages/common/src/testUtil/isTesting.ts
+++ b/packages/common/src/testUtil/isTesting.ts
@@ -1,1 +1,8 @@
-export const isTesting = () => process.env.CURSORLESS_TEST != null;
+import { IDE } from "../ide/types/ide.types";
+
+// we rely on the ide.runMode when running as a vscode extension
+// but we need an environment variable for other types of tests
+export const isTesting = (ide?: IDE) =>
+  ide?.runMode === "test" || process.env.CURSORLESS_TEST != null;
+
+export const isProduction = (ide: IDE) => ide.runMode === "production";
--- a/packages/cursorless-engine/src/actions/GenerateSnippet/GenerateSnippet.ts
+++ b/packages/cursorless-engine/src/actions/GenerateSnippet/GenerateSnippet.ts
@@ -1,4 +1,4 @@
-import { FlashStyle, isTesting, Range } from "@cursorless/common";
+import { FlashStyle, Range, isTesting } from "@cursorless/common";
 import { Offsets } from "../../processTargets/modifiers/surroundingPair/types";
 import { ide } from "../../singletons/ide.singleton";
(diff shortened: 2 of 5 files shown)


pokey:
hmm if we can't get rid of it entirely, i'd be inclined to just leave things as they are?

saidelike:
> hmm if we can't get rid of it entirely, i'd be inclined to just leave things as they are?

Yes we could tbh. It didn't go as planned lol but i thought i would explain it in this PR anyway. But i also dont really like the current approach but i can't think of a better way atm anyway.

Feel free to close if you want to stick to old method.

pokey:
might be worth switching to use `CURSORLESS_MODE` everywhere then instead of `CURSORLESS_TEST`?

saidelike:
Ya the advantage of CURSORLESS_MODE is that it is consistent if we vscode extension or not and for other apps than vscode.
The drawback is that we need to specify it for every single task or launch config (as development or test) as otherwise it will be treated as production (when the env variable is not set).

pokey:
but that's the same as today right?

saidelike:
Yes when outside of a vscode extension.

But in the case of a vscode extension, the vscodeIDE.runMode gives by default that it is production hence not test or development. But it is in other places where isTesting() is not used.

And that was that inconsistency that i wanted to fix initially.

pokey:
hmm not sure i'm totally following. maybe let's discuss at a meet-up?

pokey:
@AndreasArvidsson and I briefly discussed this, and both agree that having an environment variable called `CURSORLESS_MODE` is the way to go. If it's unset, it means production, because we don't have an easy way to set it in production and prob best not to have extra env vars in production just in case

pokey:
closing for now as it needs significant changes. See https://github.com/cursorless-dev/cursorless/issues/2408

PR #2325, opened 2024-05-11
Added function scope test to python

AndreasArvidsson:
## Checklist


pokey (approved):
Nice. Do we want to remove corresponding recorded tests? Seems redundant to have recorded and scope testing the same thing

AndreasArvidsson:
Maybe?

pokey:
Why wouldn't we?

AndreasArvidsson:
I was just thinking that our tests might test different syntax patterns, but I'm not against removing them.

pokey:
I think the idea should be for our scope tests to cover what our recorded tests used to cover. Maybe we should migrate one scope-language pair at a time? Ie for this PR we look at what recorded tests we have for Python lambda and function, add scope tests that are at least as comprehensive, and remove the recorded tests?

It's a bit more work, but would be nice to get our test suite a bit more principled

AndreasArvidsson:
Sounds good. Tests updated.

pokey (approved):
🚀

PR #2339, opened 2024-05-17
css scope migration

AndreasArvidsson:
## Checklist

### DIFF
--- a/packages/cursorless-engine/src/languages/scss.ts
+++ b/packages/cursorless-engine/src/languages/scss.ts
@@ -108,7 +108,6 @@ function unitMatcher(
 const nodeMatchers: Partial<
   Record<SimpleScopeTypeType, NodeMatcherAlternative>
 > = {
-  ifStatement: "if_statement",
   condition: conditionMatcher("condition"),
   statement: cascadingMatcher(
     patternMatcher(...STATEMENT_TYPES),
@@ -117,10 +116,6 @@ const nodeMatchers: Partial<
       childRangeSelector([], ["attribute_name", "string_value"]),
     ),
   ),
-  functionCall: "call_expression",
-  functionCallee: "call_expression.function_name!",
-  namedFunction: ["mixin_statement", "function_statement"],
-  functionName: ["mixin_statement.name!", "function_statement.name!"],
   argumentOrParameter: cascadingMatcher(
     matcher(
       patternFinder("arguments.*!", "parameters.*!"),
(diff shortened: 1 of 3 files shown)


data/fixtures/scopes/scss/index.json
pokey, line 1: The fact that this was missing inspired https://github.com/cursorless-dev/cursorless/pull/2370, which also includes this file

pokey (approved):
Looks like you forgot to add iteration scopes. I added them, along with some fixture tests. I also added a few new fixture types while I was here. Worth having a look at those fixture ID's. I think they're logical but they may warrant some discussion. See my changes in https://github.com/cursorless-dev/cursorless/pull/2339/files/13d759c3b090dd3872d2e170701521811b76436d..9496ae08a6f2dc5456e8a15ffca676affd52dc70

Merge if you're happy

packages/common/src/scopeSupportFacets/scopeSupportFacetInfos.ts
```diff
@@ -124,6 +124,21 @@ export const scopeSupportFacetInfos: Record<
     description: "A named method declaration in a class",
     scopeType: "namedFunction",
   },
+  "namedFunction.method.iteration.class": {
+    description: "Iteration scope for named functions: class bodies",
+    scopeType: "namedFunction",
+    isIteration: true,
+  },
+  "namedFunction.iteration": {
```
AndreasArvidsson, line 132: @pokey There's no indicator in the name or description what this iteration scope should be. Shouldn't this be `namedFunction.iteration.block`?

pokey: Yeah so the issue is that we don't necessarily want the iteration scope for functions to always include blocks. For example, that's not what we do in typescript. I added this as a catch-all so that I could include blocks for css. Might be worth discussing?

AndreasArvidsson: We can us gore had and merge this because we can always update later?

pokey: Slightly mixed feelings about merging in facets that we will want to change, as I think it could be a bit painful to rename them, especially as they get used more

AndreasArvidsson: Possibly. Should be punt on this then?

pokey: Hmm. Would be good to get this PR in, but I think it's going to be easier to get this one right, as we use this facet for this PR. I'd be inclined to discuss Sunday

AndreasArvidsson: Lets

pokey: On second thought, let's get this thing in. I filed https://github.com/cursorless-dev/cursorless/issues/2375; let's discuss Sunday

PR #2343, opened 2024-05-18
Ruby scope migration

AndreasArvidsson:
## Checklist

### DIFF
--- a/queries/ruby.scm
+++ b/queries/ruby.scm
@@ -38,7 +38,7 @@
 ] @textFragment
 
 (class
-  name: (_) @className @name
+  name: (_) @name
 ) @_.domain
 
 (method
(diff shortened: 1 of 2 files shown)


queries/ruby.scm
```diff
@@ -36,3 +36,21 @@
   (string_content)
   (heredoc_content)
 ] @textFragment
+
+(class
+  name: (_) @className @name
```
pokey, line 41: `className` is already defined above

pokey (commented):
Looks good with minor comment. Also when we continue migrating Ruby, it's probably worth seeing how much we want to steal from https://github.com/cursorless-dev/cursorless/compare/main...mkandalf:cursorless:ruby-scm

queries/ruby.scm
AndreasArvidsson: done

pokey (approved):

PR #2358, opened 2024-05-25
Added reformat action to public api

AndreasArvidsson:
## Checklist

### DIFF
--- a/changelog/2024-05-reformatPublicAPI.md
+++ b/changelog/2024-05-reformatPublicAPI.md
@@ -1,0 +1,7 @@
+---
+tags: [enhancement]
+pullRequest:
+---
+
+- Added public `user.cursorless_reformat` action to the Cursorless api
+  eg: `user.cursorless_reformat(cursorless_target, "ALL_CAPS")`
--- a/changelog/2024-05-reformatPublicAPI.md
+++ b/changelog/2024-05-reformatPublicAPI.md
@@ -1,6 +1,6 @@
 ---
 tags: [enhancement]
-pullRequest:
+pullRequest: 2358
 ---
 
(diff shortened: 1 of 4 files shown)


pokey (commented):
Nice! I don't think we need tests, because the talon action is already tested

- [x] Would be good to add an example to the docs of how to use it in a command which takes a formatter via spoken form

AndreasArvidsson:
> Nice! I don't think we need tests, because the talon action is already tested
> 
>     * [ ]  Would be good to add an example to the docs of how to use it in a command which takes a formatter via spoken form

I was planning to do that, but I can't find anywhere in our documentation where we have "recipes". Where should I place this?

AndreasArvidsson:
@pokey I don't like that precommit removes trailing whitespaces from markdown files. I prefer to have it as two separate lines.

pokey (approved):

auscompgeek:
> I don't like that precommit removes trailing whitespaces from markdown files. I prefer to have it as two separate lines.

You can use a backslash to hard line break in Markdown: https://spec.commonmark.org/0.31.2/#hard-line-breaks

pokey:
> @pokey I don't like that precommit removes trailing whitespaces from markdown files. I prefer to have it as two separate lines.

Sorry I'm not following. Why do you want the trailing whitespace?

pokey:
Ah right I wasn't familiar with that syntax. And yeah I'd def go for the backslash if you want a hard line break.

Although I'll generally always just go for two new lines to give a paragraph break. Eg that's what I would have done here

AndreasArvidsson:
I've always used the syntax with trailing whitespaces for new lines, but of course since it's markdown you can do everything multiple ways :p

PR #2361, opened 2024-05-27
typescript: Support "value" for type alias

pokey:
I wouldn't say I feel super strongly, but it looked like "value" to my brain so I tried to use "value" here

## Checklist

### DIFF
--- a/data/fixtures/scopes/typescript.core/value.typeAlias.scope
+++ b/data/fixtures/scopes/typescript.core/value.typeAlias.scope
@@ -1,0 +1,20 @@
+type Aaa = Bbb;
+---
+
+[Content] = 0:11-0:14
+             >---<
+0| type Aaa = Bbb;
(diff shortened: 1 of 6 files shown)


data/fixtures/scopes/typescript.core/value.typeAlias.scope
```diff
@@ -0,0 +1,20 @@
+type Aaa = Bbb;
+---
+
+[Content] = 0:11-0:14
+             >---<
+0| type Aaa = Bbb;
+
+[Removal] = 0:10-0:14
+            >----<
+0| type Aaa = Bbb;
```
pokey, line 10: I'm pretty sure `type Aaa;` would not be valid, so I don't think there's any point in supporting "chuck value"

AndreasArvidsson (approved):

PR #2369, opened 2024-05-30
Recorded test refactor

saidelike:
## Checklist

### DIFF
--- a/packages/common/src/ide/normalized/NormalizedIDE.ts
+++ b/packages/common/src/ide/normalized/NormalizedIDE.ts
@@ -1,7 +1,6 @@
 import { getFixturePath } from "../../index";
 import { GeneralizedRange } from "../../types/GeneralizedRange";
 import { TextEditor } from "../../types/TextEditor";
-import FakeClipboard from "../fake/FakeClipboard";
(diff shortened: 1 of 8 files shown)


saidelike:
Ready for review. Refactoring was also tested against neovim in https://github.com/cursorless-dev/cursorless/actions/runs/9310586370/job/25628241888

pokey (commented):
Ok I made some changes. In particular:

- I tried to remove as many parameters to `runRecordedTest` as possible
- As part of the above, I removed both the `suite` parameter and the `shouldRunTest`. I would just skip the test at the top level in neovim. I think this may require you to load the test fixture twice, but I didn't love having to pass the suite in
- I changed the list of arguments to an `opts` object. When there are more than about 3 arguments, I find it easier to use an object so that it's clearer what each argument does
- I strengthened the types. Please please please avoid `any` and type casts if at all possible. Losing types makes developing much harder, and results in confusing bugs like the ones you had in CI earlier

If you're happy with my changes, try to get it working with neovim

saidelike:
thanks for cleaning up what I wrote. I'm happy with the changes and I tested them with neovim so good to go

pokey (approved):

PR #2380, opened 2024-06-01
Added constructor facets

AndreasArvidsson:
## Checklist


packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
```diff
@@ -32,12 +32,16 @@ const scopeSupportFacets = [
   "namedFunction",
   "namedFunction.method",
   "namedFunction.method.iteration.class",
+  "namedFunction.constructor",
+  "namedFunction.constructor.iteration.class",
```
pokey, line 36: This seems redundant. What's the difference between this and the method iteration class one? They're both just a class

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
```diff
@@ -32,12 +32,16 @@ const scopeSupportFacets = [
...
   "namedFunction.iteration",
   "namedFunction.iteration.document",
   "anonymousFunction",
   "functionName",
   "functionName.method",
   "functionName.method.iteration.class",
+  "functionName.constructor",
+  "functionName.constructor.iteration.class",
```
pokey, line 44: As above

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
```diff
@@ -118,6 +134,10 @@ const scopeSupportFacets = [
   "type.class",
   "type.typeArgument",
   "type.typeArgument.iteration",
+  "type.typeArgument.method",
+  "type.typeArgument.method.iteration",
+  "type.typeArgument.constructor",
+  "type.typeArgument.constructor.iteration",
```
pokey, line 140: Not sure I understand what these are supposed to be

AndreasArvidsson: Eg `new MyClass()`

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
AndreasArvidsson: It's a different node in the parse tree so you need to define it separately and I kind of wanted us to force them to have that scope test?

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
AndreasArvidsson: Copied the wrong thing. Updated now.

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
AndreasArvidsson: done

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
AndreasArvidsson: done

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
```diff
@@ -104,10 +118,19 @@ const scopeSupportFacets = [
   "value.resource.iteration",
   "value.argument.formal",
   "value.argument.formal.iteration",
+  "value.argument.method.formal",
+  "value.argument.method.formal.iteration",
+  "value.argument.constructor.formal",
+  "value.argument.constructor.formal.iteration",
```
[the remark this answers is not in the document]
AndreasArvidsson: done

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
```diff
@@ -104,10 +118,19 @@ const scopeSupportFacets = [
...
   "value.typeAlias",
 
   "type.variable",
   "type.formalParameter",
+  "type.formalParameter.iteration",
+  "type.formalParameter.method",
+  "type.formalParameter.method.iteration",
+  "type.formalParameter.constructor",
+  "type.formalParameter.constructor.iteration",
```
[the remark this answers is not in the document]
AndreasArvidsson: Agreed. It's only referenced in a very few places I think I will just go ahead and rename the existing one in this pull request.

pokey (commented):
Looks good 👍

AndreasArvidsson:
New facets are now added to javascript/typescript. This pull request really shows the power of the fine grained facets since I actually found multiple bugs with js/ts.

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
```diff
@@ -26,20 +26,26 @@ const scopeSupportFacets = [
...
+  "anonymousFunction",
+
   "namedFunction",
-  "namedFunction.method",
-  "namedFunction.method.iteration.class",
   "namedFunction.iteration",
   "namedFunction.iteration.document",
-  "anonymousFunction",
+  "namedFunction.method",
+  "namedFunction.method.iteration.class",
```
AndreasArvidsson, line 40: I think I would prefer to be more terse with this facet. Since there only is one iteration scope for a class in method I don't think we need to have class in the identifier itself.

pokey: We discussed this at length but looks like we forgot to capture the outcome. Do you remember where we landed here?

AndreasArvidsson: I think we landed in that we should just leave it as it is for now?

pokey: Ok good well at least we have our conversation captured now 😂

pokey (approved):
Good stuff! I made some changes in https://github.com/cursorless-dev/cursorless/pull/2380/files/b3c8926812b3f146924363edf0fe5d3938f3da40..13990780a0c2c07bdd305579c5c4828c5ea18e76

Merge if you're happy with my changes

PR #2381, opened 2024-06-01
Text fragment scope

AndreasArvidsson:
Adds text fragment as a hidden scope. Allows us to use scope visualizer and facet scope fixtures.

Additional languages can be done in a follow up. This just implements the machinery and a single language as proof of concept.

Screenshot 2024-06-01T15-42-38 - Visual Studio Code

## Checklist


packages/common/src/scopeSupportFacets/scopeSupportFacetInfos.ts
```diff
@@ -225,6 +225,23 @@ export const scopeSupportFacetInfos: Record<
...
+    scopeType: "comment",
+  },
+  "textFragment.string.singleLine": {
+    description: "A single-line string",
+    scopeType: "string",
+  },
+  "textFragment.string.multiLine": {
+    description: "A multi-line string",
+    scopeType: "string",
+  },
```
pokey, line 243: These scope types and descriptions should be updated, no?

AndreasArvidsson: Well the description is still apt, but I guess i can eaborate a bit

packages/cursorless-engine/src/languages/TreeSitterQuery/validateQueryCaptures.test.ts
```diff
@@ -81,11 +81,6 @@ const testCases: { name: string; isOk: boolean; content: string }[] = [
     isOk: false,
     content: "(if_statement) @statement.leading.start",
   },
-  {
-    name: "Text fragment removal",
-    isOk: false,
-    content: "(comment) @textFragment.removal",
-  },
```
pokey, line 88: Why do we want to support text fragment removal? Guessing you were forced into this one, but seems like a bit of a smell if so

packages/common/src/scopeSupportFacets/scopeSupportFacetInfos.ts
pokey: Yeah would be good to indicate that they're text fragments. And the scope types are wrong, no?

AndreasArvidsson: You're definitely right. Done

packages/cursorless-engine/src/languages/TreeSitterQuery/validateQueryCaptures.test.ts
AndreasArvidsson: Yeah now a text fragment is just like any scope so potentially you could add a removal range for it even if it's not used. I don't know if I think that is a problem or not by itself?

pokey: I guess that's fine. You can prob remove the special casing code for text fragments in the validator now too I believe https://github.com/cursorless-dev/cursorless/blob/820732a3dbc449508b0a6ae00998cf8d4630f8f4/packages/cursorless-engine/src/languages/TreeSitterQuery/validateQueryCaptures.ts#L5

data/fixtures/scopes/c/textFragment.string.singleLine.scope
pokey, line 1: this is actually quite nice. def better than just relying on surrounding pair scope tests, which is what I originally had in mind

packages/common/src/types/command/PartialTargetDescriptor.types.ts
```diff
@@ -173,6 +173,8 @@ export const simpleScopeTypeTypes = [
   "notebookCell",
   // Talon
   "command",
+  // Secret scope types
```
pokey, line 176:
```suggestion
  // Private scope types
```

pokey (commented):
left a couple minor comments; otherwise looks good

packages/cursorless-engine/src/languages/TreeSitterQuery/validateQueryCaptures.test.ts
AndreasArvidsson: done!

pokey:
Looks good! One last thought: we should probably have instructions about text fragments somewhere in our contributor docs and mention how to activate this secret scope here. I'll file a follow-up for that

pokey:
https://github.com/cursorless-dev/cursorless/issues/2382

pokey (approved):
🚀

PR #2392, opened 2024-06-08
Split define action

jaresty:
This pull request adds a command to split the window and follow in one command.

## Checklist


packages/cursorless-engine/src/actions/SimpleIdeCommandActions.ts
```diff
@@ -103,6 +103,11 @@ export class RevealDefinition extends SimpleIdeCommandAction {
   ensureSingleTarget = true;
   restoreSelection = false;
 }
+export class RevealDefinitionAside extends SimpleIdeCommandAction {
+  command: CommandId = "revealDefinitionAside";
+  ensureSingleTarget = true;
+  restoreSelection = false;
```
pokey, line 109: I'd remove this so that it restores the selection
```suggestion
```

data/fixtures/recorded/actions/splitDefineVest.yml
```diff
@@ -0,0 +1,39 @@
+languageId: typescript
+command:
+  version: 1
+  spokenForm: split define vest
+  action: executeCommand
```
pokey, line 5: This doesn't look right. Are you sure this wasn't recorded using your personal version from custom actions csv?

packages/common/src/types/TextEditor.ts
```diff
@@ -196,6 +196,12 @@ export interface EditableTextEditor extends TextEditor {
    */
   revealDefinition(range?: Range): Promise<void>;
 
+  /**
+   * Reveal definition aside
+   * @param range A {@link Range range}
+   */
+  revealDefinitionAside(range?: Range): Promise<void>;
```
pokey, line 203: I would be tempted to just make this an option passed to `revealDefinition`. Slightly mixed feelings about it being on the editor, as it opens a new editor, but probably fine. Cc/ @AndreasArvidsson

packages/cursorless-vscode/src/ide/vscode/VscodeTextEditorImpl.ts
```diff
@@ -173,6 +173,10 @@ export class VscodeTextEditorImpl implements EditableTextEditor {
     await vscode.commands.executeCommand("editor.action.revealDefinition");
   }
 
+  public async revealDefinitionAside(_range?: Range): Promise<void> {
+    await vscode.commands.executeCommand("editor.action.revealDefinitionAside");
```
pokey, line 177: Thinking on it, I think it would be better to have `followLinkAside`, because I haven't used `revealDefinition` in probably a year. Follow link is much more robust. Here is where that is implemented: https://github.com/cursorless-dev/cursorless/blob/main/packages/cursorless-vscode/src/ide/vscode/VscodeOpenLink.ts

You should be able to just do a case-matched find/replace on your PR to change the name, then there will be adding the options object and opening aside in VscodeOpenLink.ts

We could pass an options object indicating to open aside, as suggested in another of my comments

cursorless-talon/src/spoken_forms.json
```diff
@@ -17,6 +17,7 @@
       "decrement": "decrement",
       "dedent": "outdentLine",
       "define": "revealDefinition",
+      "split define": "revealDefinitionAside",
```
pokey, line 20: I'd be tempted to swap this: `"define split"`

pokey (commented):
Nice job! This looks like it should be working. See my comments below, though

data/fixtures/recorded/actions/splitDefineVest.yml
AndreasArvidsson: `version: 1`? I assume you just copied an existing file instead of recording a new one?

packages/common/src/types/TextEditor.ts
AndreasArvidsson: I'm fine with either way. You are revealing a definition of something that exists in a particular editor so I think it's fine that it's on the editor type.

cursorless-talon/src/spoken_forms.json
AndreasArvidsson: Totally agree

packages/common/src/types/TextEditor.ts
pokey: Makes sense. WDYT about options object on existing function instead of new function. See also below re adding it to followLink instead of revealDefinition

AndreasArvidsson: As a said I'm fine with either adding an argument to existing function or adding a new function. With regards to the editor declaring capabilities it might be better with a separate function?

packages/cursorless-vscode/src/ide/vscode/VscodeTextEditorImpl.ts
AndreasArvidsson: Same here; I don't use reveal definition at all anymore. Almost like we should deprecate that one?

data/fixtures/recorded/actions/splitDefineVest.yml
jaresty: Yes, I copied the existing file because I wasn't able to figure out how to run the command from the debug editor-the test was passing so I was pretty sure the command was working, though.

pokey: This doesn't actually test any of your code, though

packages/cursorless-vscode/src/ide/vscode/VscodeTextEditorImpl.ts
pokey: Yeah I don't think there are any examples where it is better than follow link

data/fixtures/recorded/actions/splitDefineVest.yml
jaresty: Ok, then I misunderstood how it works. I don't really understand how to do the local recording for test cases yet.

cursorless-talon/src/spoken_forms.json
jaresty: Should it be "follow split"?

data/fixtures/recorded/actions/splitDefineVest.yml
pokey: Assuming you found the docs for the test recorder?

cursorless-talon/src/spoken_forms.json
pokey: Yeah

data/fixtures/recorded/actions/splitDefineVest.yml
jaresty: I did. Will take a look when I have some more time. Thanks!

packages/cursorless-engine/src/actions/SimpleIdeCommandActions.ts
jaresty: Changed.

jaresty:
I think I responded to all your comments. I wasn't sure if I should remove the define split commands. Everything seems to be working at the moment.

AndreasArvidsson:
We probably gonna deprecate define in favor of follow. No sense adding another define then.

jaresty:
I removed the define command. Only the follow command remains.

packages/cheatsheet/src/lib/sampleSpokenFormInfos/defaults.json
```diff
@@ -158,6 +158,16 @@
...
+        {
+          "id": "followLinkAside",
+          "type": "action",
+          "variations": [
+            {
+              "spokenForm": "follow aside <target>",
+              "description": "Follow link in a split"
+            }
+          ]
+        },
```
pokey, line 170: This is supposed to be generated automatically, but it's not easy to trigger today. For now I'd just manually change it to what the generated output would be
```suggestion
        {
          "id": "followLinkAside",
          "type": "action",
          "variations": [
            {
              "spokenForm": "follow split <target>",
              "description": "Follow link aside"
            }
          ]
        },
```

packages/cursorless-engine/src/actions/FollowLink.ts
```diff
@@ -24,11 +29,12 @@ export default class FollowLink implements SimpleAction {
       .openLink(target.contentRange);
```
pokey, line 29: You should pass an options object here to indicate link should be opened aside, and then handle that [here](https://github.com/cursorless-dev/cursorless/blob/main/packages/cursorless-vscode/src/ide/vscode/VscodeOpenLink.ts#L66) by passing an options object there to indicate to open aside

pokey (commented):
Looking pretty good; left a couple comments

packages/cursorless-engine/src/actions/FollowLink.ts
```diff
@@ -24,11 +29,12 @@ export default class FollowLink implements SimpleAction {
...
-        targets,
-        "editor.action.revealDefinition",
-        { restoreSelection: false },
-      );
+      let commandId = this.openInSplit
+        ? "editor.action.revealDefinitionAside"
+        : "editor.action.revealDefinition";
+      await this.actions.executeCommand.run(targets, commandId, {
+        restoreSelection: !this.openInSplit,
+      });
```
pokey, line 37: Actually I think we should just move this whole `if` statement to somewhere in https://github.com/cursorless-dev/cursorless/blob/main/packages/cursorless-vscode/src/ide/vscode/VscodeOpenLink.ts? Feels pretty vscode-specific

WDYT @AndreasArvidsson

AndreasArvidsson: Totally agree

pokey: @jaresty I think makes sense to do that in this PR as it's a small change and you'll be touching that code already

jaresty: Sounds good. I'll take a look soon.

jaresty: I'm looking at this but I'm having a hard time understanding what the change needs to be. It seems that the VSCode follow link method currently only handles URIs. I'm not quite sure how to make the change without doing a significant restructure here, which makes me think that I'm misunderstanding something about the intention or the request. Can you give me a little bit more specific details about what you're hoping to see change? I also don't mind if you want to make the change yourself (in case it's easier to do it than to explain).

jaresty: I think I figured it out. One thing that I'm still unsure about is that in the previous version I needed to pass the targets to the execute action, but now it doesn't seem to be necessary. The functionality does seem to work however.

jaresty:
I think I addressed all the comments. Anything I missed?

data/fixtures/recorded/actions/followSplitVest.yml
```diff
@@ -0,0 +1,40 @@
+languageId: typescript
+command:
+  version: 7
+  spokenForm: follow split urn
+  action:
+    name: executeCommand
+    commandId: editor.action.revealDefinitionAside
```
AndreasArvidsson, line 7: This doesn't test the new implementation

packages/cursorless-engine/src/actions/FollowLink.ts
```diff
@@ -10,8 +10,13 @@ import { Actions } from "./Actions";
 import { SimpleAction, ActionReturnValue } from "./actions.types";
 
 export default class FollowLink implements SimpleAction {
-  constructor(private actions: Actions) {
+  openInSplit = false;
+  constructor(
+    private actions: Actions,
+    { openInSplit }: { openInSplit: boolean },
```
AndreasArvidsson, line 16: I would probably just do `private options : { openInSplit: boolean }`.

data/fixtures/recorded/actions/followSplitVest.yml
jaresty: Shoot. Apparently it was picking up my custom action instead. I'm having some trouble getting it to recognize this new action now. Do you have any tips to debug if it's not seeming to recognize the action at all?

jaresty: I forgot that I needed to symlink the cursor list settings in place of my own settings. Updated with the new test case.

packages/cursorless-engine/src/actions/FollowLink.ts
```diff
@@ -10,8 +10,13 @@ import { Actions } from "./Actions";
...
 
 export default class FollowLink implements SimpleAction {
-  constructor(private actions: Actions) {
+  openInSplit = false;
+  constructor(
+    private actions: Actions,
+    private options: { openInSplit: boolean },
+  ) {
     this.run = this.run.bind(this);
+    this.options = options;
```
AndreasArvidsson, line 19: When you use private on a constructor argument you don't need to assign it yourself; that is done automatically
Just remove lines `openInSplit = false;` and `this.options = options;`

packages/cursorless-vscode/src/ide/vscode/VscodeOpenLink.ts
```diff
@@ -17,7 +19,10 @@ export default async function vscodeOpenLink(
   }
 
   if (filteredLinks.length === 0) {
-    return false;
+    let commandId = options?.openInSplit
```
AndreasArvidsson, line 22: const instead of let

packages/cursorless-engine/src/actions/FollowLink.ts
```diff
@@ -21,15 +26,7 @@ export default class FollowLink implements SimpleAction {
 
     const openedLink = await ide()
```
AndreasArvidsson, line 27: This return value is not used anymore. Should probably be removed from the openLink type.

packages/cursorless-vscode/src/ide/vscode/VscodeOpenLink.ts
```diff
@@ -17,7 +19,10 @@ export default async function vscodeOpenLink(
...
+      ? "editor.action.revealDefinitionAside"
+      : "editor.action.revealDefinition";
+    await vscode.commands.executeCommand(commandId);
```
AndreasArvidsson, line 25: Also you still need a return here. Otherwise you will run open link a few lines below on undefined.

AndreasArvidsson (changes requested):

packages/cursorless-vscode/src/ide/vscode/VscodeOpenLink.ts
AndreasArvidsson, line 25: I don't think this will actually work. You are executing reveal definition on the current selection always and not on the actual target. I would probably still use our own `actions.executeCommand`

AndreasArvidsson (changes requested):

packages/cursorless-vscode/src/ide/vscode/VscodeOpenLink.ts
jaresty: I think you're right. I didn't change this though-don't really understand what you mean by 'use our own actions.executeCommand'-can you help me understand?

AndreasArvidsson: The one you removed in this pull request

https://github.com/cursorless-dev/cursorless/blob/766864d9fa979124283b236bbb4738204a13fe45/packages/cursorless-
engine/src/actions/FollowLink.ts#L27-L31

But since we are already setting selection for the `editor.action.openLink` command it's probably best to merge these behaviors

jaresty: oh okay I see it now

jaresty: I wasn't sure exactly what you meant but I gave it a shot; let me know how it looks. Thanks!

AndreasArvidsson: You still need the if statement to check if we actually have found a link. The set selection code could be extracted to ear separate function that is used in both cases.
Also I would remove the return boolean type and just have void.

jaresty: 1. Extracted the set selection code
2. the return type is harder-that type is defined in the interface. Are you sure I should remove it?

pokey (approved):
I made some tweaks in f5d58fe2a9c0ea158551b93a704a3dc6d53d420b. I'm happy, but I believe will require approval from @AndreasArvidsson before Github will let it merge as he's requested changes

jaresty:
Thanks for cleaning it up! @pokey

AndreasArvidsson:
Nice exactly what I wanted! :)

pokey:
> Nice exactly what I wanted! :)

I didn't actually read any of your comments, so it's a good sign that we independently arrived at the same conclusion 😄

AndreasArvidsson:
> > Nice exactly what I wanted! :)
> 
> I didn't actually read any of your comments, so it's a good sign that we independently arrived at the same conclusion 😄

Heh don't know if I should be hurt or impressed by that :D

pokey:
well they had all been marked as resolved so I assumed they'd already been addressed 😅

jaresty:
Oops, sorry. I was using them kind of like a todo list-should I have left them open?  Appreciate all the guidance-thanks for helping me get this done 😀

pokey:
> Oops, sorry. I was using them kind of like a todo list-should I have left them open?  Appreciate all the guidance-thanks for helping me get this done 😀

That's ok. If you're confident you've resolved them, feel free to mark them resolved. If you have any doubt, though, best to leave them open and let us have a look to see if they've been resolved

PR #2402, opened 2024-06-11
Add `cursorless_create_destination` action to talon api

pokey:
This came up during a pairing session. I wanted to do something like:

```talon
transform <user.cursorless_target>:
    old = user.cursorless_get_text(cursorless_target, 1)
    new = some_transformation(old)
    destination = user.cursorless_create_destination(cursorless_target)
    user.cursorless_insert(destination, new)
```

Notice I can't use our existing `user.cursorless_insert` out of the box because it expects a destination, so we need a way to make a destination from a target

## Checklist

### DIFF
--- a/docs/user/customization.md
+++ b/docs/user/customization.md
@@ -162,7 +162,7 @@ Cursorless exposes a couple talon actions and captures that you can use to defin
   Insert text at destination.
   eg: `user.cursorless_insert(cursorless_destination, "hello")`
 - `user.cursorless_create_destination(target: CursorlessTarget, insertion_mode: Literal["to", "before", "after"] = "to")`:
-  Create a destination from a target. The insertion mode can be `to`, `before`, or `after`, and defaults to `to`, which will replace the target.
+  Create a destination from a target. The insertion mode can be `to`, `before`, or `after`, and defaults to `to`, which will replace the target. See [How do I run a custom Python transformation on a target?](./how-to.md#how-do-i-run-a-custom-python-transformation-on-a-target) for example usage.
 - `user.cursorless_reformat(target: CursorlessTarget, formatters: str)`
   Reformat target with specified formatters.
   eg: `user.cursorless_reformat(cursorless_target, "ALL_CAPS")`
--- a/docs/user/customization.md
+++ b/docs/user/customization.md
@@ -161,7 +161,7 @@ Cursorless exposes a couple talon actions and captures that you can use to defin
 - `user.cursorless_insert(destination: CursorlessDestination, text: Union[str, List[str]])`:
   Insert text at destination.
   eg: `user.cursorless_insert(cursorless_destination, "hello")`
-- `user.cursorless_create_destination(target: CursorlessTarget, insertion_mode: Literal["to", "before", "after"] = "to")`:
+- `user.cursorless_create_destination(target: CursorlessTarget, insertion_mode: Literal["to", "before", "after"] = "to") -> CursorlessDestination`:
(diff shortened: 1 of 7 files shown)


pokey:
This still needs tests and docs, but I wanted to get your thoughts first @AndreasArvidsson

AndreasArvidsson:
I wanted the same.

Couldn't insert just accept a destination or target instead? Is it really simpler with another action?

pokey:
Oh that's an interesting idea. Slightly mixed feelings as it might cause people to conflate target and destination, but could be simpler. I don't have a strong feeling. Maybe let's discuss tomorrow?

AndreasArvidsson:
Yup

pokey:
update from meet-up:

- Add `cursorless_create_destination(target: ListTarget | RangeTarget | PrimitiveTarget, insertion_mode: "before"|"after"|"to" = "to") -> CursorlessDestination`
- This will go into a file just for our public api; for now just include this but in the future move other functions there
- In the future, if it becomes annoying boilerplate, just automatically do conversion in `cursorless_insert`

pokey:
ok @AndreasArvidsson this one is good to go

pokey:
wait one sec i'll add release notes

pokey:
ok good to go

docs/user/customization.md
```diff
@@ -158,6 +161,8 @@ Cursorless exposes a couple talon actions and captures that you can use to defin
 - `user.cursorless_insert(destination: CursorlessDestination, text: Union[str, List[str]])`:
   Insert text at destination.
   eg: `user.cursorless_insert(cursorless_destination, "hello")`
+- `user.cursorless_create_destination(target: CursorlessTarget, insertion_mode: Literal["to", "before", "after"] = "to")`:
```
AndreasArvidsson, line 164: Do we want to add the return type to the function signature?

docs/user/how-to.md
```diff
@@ -46,3 +46,44 @@ You can now say eg `"echo air past bat"`.
...
+        def hello(text: str) -> str:
+            """Returns a greeting for the given text."""
+            return f"Hello, {text}!"
+   ```
+
+2. Add a spoken form to your `vscode.talon`:
+
+   ```talon
+   hello <user.cursorless_target>:
+       old = user.cursorless_get_text(cursorless_target, 1)
```
AndreasArvidsson, line 71: Doesn't `True` work on public Talon?

jaresty:
This rings a bell for me too. Great idea!

docs/user/how-to.md
pokey: didn't work for me 🤔

docs/user/customization.md
pokey: Idk the line is pretty long as is, but happy to if you prefer

docs/user/how-to.md
AndreasArvidsson: It's already used in community actually. Note that in Talon script its true with lower case t. ie not same case as python.
https://github.com/talonhub/community/blob/93c9261c240078c46290ae1a500f2c9cfc749833/settings.talon#L34

docs/user/customization.md
AndreasArvidsson: I mean since it's actually returning something important I would prefer that. We do have return type on the get text action for example.

docs/user/how-to.md
pokey: Yeah I tried with lowercase. Maybe I was doing something dumb. I'll try again

docs/user/customization.md
pokey: Sounds good will add

pokey:
ok @AndreasArvidsson feedback addressed

AndreasArvidsson (approved):

PR #2426, opened 2024-06-17
Make "inside" fall back to pair when the target doesn't have an interior

pokey:
This one bit me a lot. See the test case for what now works. Basically "inside value" when value is a string is the most common one

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/scopeProviders/getTargetRanges.ts
+++ b/packages/cursorless-engine/src/scopeProviders/getTargetRanges.ts
@@ -1,5 +1,4 @@
 import {
-  NoContainingScopeError,
   TargetRanges,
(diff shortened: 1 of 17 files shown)


packages/cursorless-engine/src/scopeProviders/getTargetRanges.ts
pokey, line 1: wow we were really going to great lengths to work around the fact that it threw an exception. good to see this new api simplifies this code a bunch

AndreasArvidsson (approved):

pokey:
Could use a quick re-review as I made some big changes

AndreasArvidsson (approved):

Issue #2390, opened 2024-06-05
Support migrating old recorded tests using bulk facet recorder


pokey:
Once https://github.com/cursorless-dev/cursorless/pull/2383 is merged, we should simplify the process of migrating old recorded tests to scope fixture tests using the new bulk fixture recorder.

We proceed as follows:

1. Find all recorded tests with the given language id
2. Canonicalize and inspect the command payload. We may want to filter to tests that use a specific set of actions and particular targets. We may also only want to migrate a specific subset of scopes, eg based on if they're marked as supported for the given language
3. For each chosen test, get initial document state
3. Skip if we've already processed a fixture with the same document state (or if there already exists a fixture with same document state?)
4. Generate an entry in the bulk recorder file with a candidate header for each possible facet that has the given scope. For example:

    ```
    [[python]]

    [branch.if]
    [branch.if.iteration]
    [branch.try]
    [branch.try.iteration]
    [branch.loop]
    if aaa:
        pass
    ---

    [branch.if]
    [branch.if.iteration]
    [branch.try]
    [branch.try.iteration]
    [branch.loop]
    try:
        pass
    except:
        pass
    ---
    ```

5. The user will then delete the unnecessary lines and clone the block as necessary eg to test both scope and iteration scope
6. When the user saves, we should probably delete the recorded tests that they actually ended up migrating

## To discuss: which tests to migrate?

We need to figure out how to decide which recorded tests to migrate. Some ideas:

- All fixtures for a given language, regardless of whether they're marked as supported
- Only ones marked as supported
- User picks specific scope type

We also need to figure out if we want to drop recorded tests if their command payload doesn't match a particular shape. Eg skip if multiple scope types are mentioned, or of the command isn't one of a few expected commands

pokey:
update from meet-up:

- Only facets marked as supported
- Only recorded tests in the folder `languages/{languageId}`
- Only recorded tests with action of "chuck", "take", "pour", "clone", "drink", "clone up", "clear", primitive target, containing scope modifier

Migrate the above, and then see what's left and maybe make filter more lenient or just do them manually

fidgetingbits:
Does this mean I should hold off an updating old tests for now to see how this goes? Porting the lua, bash, and nix tests are still on my list todo and was thinking recently about circling back to to it, but if it can mostly be automated that works too :D

pokey:
[...]

PR #2432, opened 2024-06-18
Add support for Kotlin.

wolfmanstout:
## What

Adds support for the `kotlin` programming language

## Checklist

- [ ] Supported` "type"` both for type annotations (eg `foo: string`) and declarations (eg `interface Foo {}`) (and added tests for this behaviour 😊)


wolfmanstout:
I think this is now ready for review. Some open questions:

- I left one item on the checklist unchecked, because the example given doesn't actually seem to align with conventions in either Java or Python: "take type" on a class (or interface) doesn't take the class name. There are many places in Kotlin where a type can appear, and I've covered the vast majority of them.
- Thus far I added support for @private.switchStatementSubject but I haven't tested it. Let me know if I should simply drop support or bind this and add a test.

queries/kotlin.scm
```diff
@@ -0,0 +1,634 @@
...
+(jump_expression
+  [
+    "return"
+    "throw"
+  ]
+  .
+  (_) @value
+) @_.domain
+
+;; Disabled due to Cursorless error ("invalid capture") caused by "return@"
```
wolfmanstout, line 447: I can file a bug for this.

queries/kotlin.scm
```diff
@@ -0,0 +1,627 @@
...
+    .
+    (function_body)
+  ) @_.domain
+)
+(
+  (function_declaration
+    ":"
+    .
+    (_) @type.start
+    (_) @type.end
```
wolfmanstout, line 219: If I could just do `(_)? @type.end` I wouldn't need this redundant query, but Cursorless complains when it's missing because I refer to it in a predicate below. I think Cursorless could safely ignore constraints when they don't appear in cases such as this where they are tied to an optional match. My reasoning is that if the predicate must be confirmed, then it shouldn't be an optional match. Should I file a bug for this? There are several places throughout this file where I could simplify my code if this were possible.

wolfmanstout:
Also, I noticed that "drink/pour arg" don't automatically work to add commas. How can I add support for these?

queries/kotlin.scm
```diff
@@ -0,0 +1,634 @@
...
+  (call_suffix
+    (type_arguments)? @functionCallee.end
+  )
+) @_.domain.start
+
+(call_suffix
+  (annotated_lambda) @argumentOrParameter
+)
+
+;; Note: trailing lambda mixed with regular arguments doesn't work due to bad tree sitter parse.
```
wolfmanstout, line 509: Filed bug for this: https://github.com/fwcd/tree-sitter-kotlin/issues/127

queries/kotlin.scm
wolfmanstout: Filed: https://github.com/cursorless-dev/cursorless/issues/2447

AndreasArvidsson:
> Also, I noticed that "drink/pour arg" don't automatically work to add commas. How can I add support for these?

Here is the implementation for javascript you can have a look at
https://github.com/cursorless-dev/cursorless/blob/089ba0da1461dd14ea644a6f000526d920e0110a/queries/javascript.core.scm#L692-L718

wolfmanstout:
Thanks @AndreasArvidsson, I added argument delimiter support.

pokey:
Ok before you get too much further, it's worth having a look at our updated [docs for adding a new language](https://www.cursorless.org/docs/contributing/adding-a-new-language/). In particular, we have a new way of writing tests, and we now advocate for many small PRs rather than one big PR.

Given that I added those new docs after you had already opened this PR, I won't make it a hard requirement for merging, but going forward, let's

a) not add any more scopes to this PR, and
b) use the new test format for subsequent PRs

We're planning to add a migration assistant for migrating tests to the new format (see #2390), so I don't think it's worth migrating these manually

I'll have a look at this PR in the next couple of days, but figured I'd give you a chance to look at those docs in case there's anything you want to tweak before I have a look

wolfmanstout:
Thanks! I don't think I will have time to make any changes to this pull request within the next couple days, so you should feel free to review. 

Unfortunately it's a little late to stop adding more scopes because all the scopes have been added 😂. I'm happy to add the metadata files before submitting that describe what scopes are covered.

pokey:
> Thanks! I don't think I will have time to make any changes to this pull request within the next couple days, so you should feel free to review.

Sounds good; will do 😊

> Unfortunately it's a little late to stop adding more scopes because all the scopes have been added 😂

Ha no worries

> I'm happy to add the metadata files before submitting that describe what scopes are covered.

You'd need to migrate your recorded tests to new scope tests as well, because we have checks to ensure that all scope facets marked as supported actually have tests. So if you're not planning to manually migrate those tests (which I'm not sure I'd recommend), I'd leave out the support tables for now

wolfmanstout:
Okay I won't plan to add the tables then. You can just let me know in the review if there are any other small things I should add as part of this change.

AndreasArvidsson:
There are so many test and I'm not a kotlin developer so I most likely have missed something, but I think this looks fine. If you're appy with it then we could merge this so you can start using it. Anything you find we could fix in a follow up.

AndreasArvidsson (approved):

jaresty:
I'm excited to try this one out! Thanks for putting it together.

wolfmanstout:
Fantastic I'll merge it! I've been using it plenty and the only issues I've seen are upstream in the tree sitter grammar.

wolfmanstout:
Ah I don't have merge permissions so you'll need to do that part.

AndreasArvidsson:
Okay here we go! :)

pokey:
> Fantastic I'll merge it! I've been using it plenty and the only issues I've seen are upstream in the tree sitter grammar.

@wolfmanstout def worth filing those upstream and also keeping a tracker issue here. We can often work around oddnesses / bugs with upstream tree-sitter grammars using tricks in our patterns

wolfmanstout:
> > Fantastic I'll merge it! I've been using it plenty and the only issues I've seen are upstream in the tree sitter grammar.
> 
> 
> 
> @wolfmanstout def worth filing those upstream and also keeping a tracker issue here. We can often work around oddnesses / bugs with upstream tree-sitter grammars using tricks in our patterns

Yep, filed upstream and even sent out PR fixes. I worked around a lot of shortcomings here but these would be extremely awkward to work around and probably introduce other issues. I can create a tracker issue too (a single one for these issues I suppose?).

pokey:
> > > Fantastic I'll merge it! I've been using it plenty and the only issues I've seen are upstream in the tree sitter grammar.
> > 
> > 
> > 
> > @wolfmanstout def worth filing those upstream and also keeping a tracker issue here. We can often work around oddnesses / bugs with upstream tree-sitter grammars using tricks in our patterns
> 
> Yep, filed upstream and even sent out PR fixes. I worked around a lot of shortcomings here but these would be extremely awkward to work around and probably introduce other issues. I can create a tracker issue too (a single one for these issues I suppose?). 

Yeah just a single tracker issue would be awesome 🙏

PR #2436, opened 2024-06-19
Extending custom grammar

AndreasArvidsson:
Extending custom grammar with actions

https://github.com/cursorless-dev/cursorless/issues/492

## Checklist


AndreasArvidsson:
@pokey I wonder where we want to put the place where we evaluate the input text and merge with eventual input target? `TargetMark` expects a full target descriptor not a partial one. That kind of means it needs to go into [TargetPipelineRunner.ts](https://github.com/cursorless-dev/cursorless/blob/ee3e44635844c588a514a1f17481e15f3544fbc0/packages/cursorless-engine/src/processTargets/TargetPipelineRunner.ts)

pokey:
Can't we just do it right after inference? Or during inference?

AndreasArvidsson:
That is probably the earliest possible.

pokey:
Ignore the rest of this message; I have an idea. I'll take a swing after lunch so leave this for now unless you are already working on it or have a good way forward

---

Ok I think what you want to do is add a new PartialTargetMark that is equivalent to TargetMark but with a partial target

Then if you encounter a parsed action, you will get an ActionDescriptor back from the parser. You look in that descriptor for placeholders, and fill them in with the partial targets you have from the parsed action, then recursively call run action on the ActionDescriptor

In inferFullTargets, you

AndreasArvidsson:
I don't have time to look at this today so please go ahead!

pokey (approved):
Ok I'm happy with this PR; have a look at [my changes](https://github.com/cursorless-dev/cursorless/pull/2436/files/79581b73cb5062158b95dbf7a93471d44f2172db..0f6110ff20173c8da097d4bd998dabbf80c3816f)

I filed some follow-up issues, but I think it's good to merge like this. The one thing to think about here is in terms of how public / private we want to make this right now. The compromise I've gone for is:

- Don't add docs or release notes, but
- Also don't add `private_` prefix to action name

My thinking is that it's pretty unlikely we'll remove anything we support in this version, but the grammar is still fairly limited so might be frustrating for users

But I'd be happy to consider either making it more private (ie adding `private_`) or more public (adding docs / release notes). Curious what your thoughts are

packages/cursorless-engine/src/generateSpokenForm/generateSpokenForm.test.ts
```diff
@@ -55,6 +56,31 @@ suite("Generate spoken forms", () => {
...
+            index: 0,
+          },
+        },
+      },
+      version: LATEST_VERSION,
+      usePrePhraseSnapshot: false,
+    });
+
+    assert(spokenForm.type === "success");
+    assert.equal(spokenForm.spokenForms, "take block <target>");
```
pokey, line 82: Note the similarity to the cheatsheet here. I had a fantasy of using this machinery to auto generate parts of cheatsheet and docs

pokey:
Thought about it some more and I wonder if the way we automatically infer placeholder index is too fancy. I wonder if it will make it harder to do things where we reuse a target. For example, what if we wanted to do "take <target> past next line <target>". We could do that by just passing the same target twice in the input target list, but that feels a bit strange, and makes it slightly more awkward to do other fancy things I was thinking like using target templates to create custom modifiers, eg "<target> past next line <target>" could be used as a modifier via a special csv, or in keyboard config

We could handle that in the same way, by just passing a bunch of copies of the input, but it feels weird

pokey:
I'd be tempted to just switch to talon-style placeholder numbering

packages/cursorless-engine/src/customCommandGrammar/CommandLexer.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (this.skipToken(rawToken)) {
+      return this.next();
+    }
+
+    return rawToken?.type === "placeholderMark"
+      ? { type: "placeholderMark", value: this.placeholderIndex++ }
+      : rawToken;
+  }
+
+  transform({ value }: MooToken) {
```
AndreasArvidsson, line 70: @pokey When is this actually called?

packages/cursorless-engine/src/generateSpokenForm/generateSpokenForm.test.ts
AndreasArvidsson: Neat!

packages/cursorless-engine/src/customCommandGrammar/CommandLexer.ts
pokey: it's called by the nearley parser

AndreasArvidsson: Shouldn't that then be on the `NearleyLexer` interface?

packages/cursorless-engine/src/customCommandGrammar/grammarUtil.ts
pokey, line 1: here's where this system breaks down: several of these types are lies, because they really should now be `WithPlaceholders<Foo>` instead of just `Foo`. So it gives us the illusion of strong typing without actually checking the types

pokey: ok I fixed the types in e755f83001bcd753f0d8c3219313fdde9e1b6393, but I have no idea whether I got it right because they're not really checked 🤷‍♂️. Ie typescript was happy before, even though they were wrong

packages/cursorless-engine/src/util/getPartialTargetDescriptors.ts
```diff
@@ -23,6 +23,9 @@ export function getPartialTargetDescriptors(
     case "replace":
     case "editNew":
       return getPartialTargetDescriptorsFromDestination(action.destination);
+    case "parsed":
+      // FIXME: The args might actually be targets; not sure how to handle
+      return [];
```
pokey, line 28: ok this is annoying. It's causing pre-commit failures because we are now ignoring the args, because we don't know if they're targets. So when it tries to extract targeted marks, it doesn't see them. We could have a hack to try to see if they're targets, but it feels pretty wrong

packages/cursorless-engine/src/util/getPartialTargetDescriptors.ts
```diff
@@ -23,11 +23,20 @@ export function getPartialTargetDescriptors(
...
       return [action.target];
   }
 }
 
+function isPartialTargetDescriptor(
+  arg: unknown,
+): arg is PartialTargetDescriptor {
+  return typeof arg === "object" && arg != null && "type" in arg;
+}
+
```
pokey, line 39: this is pretty brutal. I do kind of feel like targets are a bit special, and our system kinda breaks down when we have args and targets and destinations all mixed up

we could try the idea we discussed where we inject expected type when we parse placeholders, but that is also kind of rough

Idk. I'd be tempted to ship in either private or semi-private and sit on it

pokey (approved):
ok I'm done with the changes we discussed. It is definitely more powerful, in that you can reuse args and it should support non-target args (eg scope types and numbers), but something doesn't quite sit totally right with me

packages/cursorless-engine/src/util/getPartialTargetDescriptors.ts
AndreasArvidsson: If you're not happy with the solution then definitely go private

packages/cursorless-engine/src/customCommandGrammar/grammarUtil.ts
AndreasArvidsson: This system only gives us the shape of the return value. That the arguments in is unknown we already knew.

AndreasArvidsson:
Let's make it private and ship it so we can try it out

pokey:
idk something's not right here. I almost feel like the `$1` token should indicate what type of thing it is. For example, what if the user just uses `"take $1 $2 $3"` as the content to parse. How is our parser supposed to know what type each of those things are supposed to be? It could be `"take <number> <scope>s <target>"` or `"take <scope> <scope> <target>"` or probably plenty of other things. Idk something doesn't feel right

AndreasArvidsson:
Yeah if this makes the grammar become ambiguous we definite to have a problem

pokey (approved):
Ok I think this is a decent compromise for now. I'd be happy to ship it like this

I re-added the `target` prefix so our grammar is no longer ambiguous. We can figure out how to handle arguments of other types later

Issue #2401, opened 2024-06-11
Keyboard: Create a setting for the style of the cursor in keyboard mode


C-Loftus:
Currently there is no way to customize the type/style of cursor that Cursorless uses in keyboard mode. Having a separate cursor type would help to add extra contrast if needed. For accessibility reasons, the standard cursor style that the users prefers (i.e. line or outline block) may not provide enough contrast for some users.  

According to discussion with Pokey, there does not seem to be a way to change cursor color. And with regards to styling this is probably unnecessary and should be controlled by the user's theme.

PR #2443, opened 2024-06-20
Keyboard cursor customization

C-Loftus:
Addresses https://github.com/cursorless-dev/cursorless/issues/2401, the ability to have custom cursor styling exclusively when in cursorless keyboard mode.

Tried to follow the repo style as best as possible but please let me know if another way besides a record mapper is more idiomatic for TS.

## Checklist

    - _I could add docs in the visual a11y section but unclear if experimental keyboard support belongs there_

## Example 
An image showing the cursor start out with a custom type, then change into `block` when keyboard mode is activated, thus adding more contrast.

### DIFF
--- a/packages/cursorless-vscode/package.json
+++ b/packages/cursorless-vscode/package.json
@@ -861,6 +861,18 @@
           "description": "Directory containing snippets for use in Cursorless",
           "type": "string"
         },
+        "cursorless.experimental.keyboard.modal.cursorStyle": {
+          "description": "Color to use for the additional cursor when in keyboard mode",
+          "type": "string",
+          "enum": [
+            "line",
+            "block",
+            "underline",
+            "lineThin",
+            "blockOutline",
+            "underlineThin"
+          ]
+        },
         "cursorless.experimental.keyboard.modal.keybindings.action": {
           "description": "Define modal keybindings for actions",
           "type": "object",
@@ -877,7 +889,7 @@
                   "editNew",
                   "editNewLineAfter",
                   "editNewLineBefore",
-                  "executeCommand",
+                  "exiecuteCommand",
                   "extractVariable",
                   "findInDocument",
                   "findInWorkspace",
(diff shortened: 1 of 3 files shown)


pokey (commented):
This is a good start, but won't update the cursor until the next restart, I believe. We pretty much always reflect changes to settings in real time, so eg if you are in cursorless keyboard modal mode and change this setting, the cursor should update immediately. It's a little bit of work to get that to happen, so lmk if you need help and we can discuss

C-Loftus:
- Fixed this up a bit. Reads every time `modeOn` is called. Works for me locally.
    - I tested to be sure again, and you are right Pokey, this did not update previously and had the same cursor style for an entire session even if it was changed 
- Was not sure on watching the setting and updating in real time. Probably not needed since keyboard mode would be toggled on/off very frequently. 
    - As discussed in the meetup this works fine at least for me and it doesn't need to touch other parts of the codebase.  
 
- Changed the setting names so they match the same formatting as the names within the standard vscode cursor style setting.
    -  I had to disable an eslint setting to allow for kebab'd property names but I think this is better than the alternative of having a different name.

pokey:
> * Fixed this up a bit. Reads every time `modeOn` is called. Works for me locally.

👍

>       * I tested to be sure again, and you are right Pokey, this did not update previously and had the same cursor style for an entire session even if it was changed

👍

>     * Was not sure on watching the setting and updating in real time. Probably not needed since keyboard mode would be toggled on/off very frequently.

Yeah fine for now I think

>       * As discussed in the meetup this works fine at least for me and it doesn't need to touch other parts of the codebase.

👍

>     * Changed the setting names so they match the same formatting as the names within the standard vscode cursor style setting.

We always use camel case, but I think you're right to mirror VSCode ones

>       * I had to disable an eslint setting to allow for kebab'd property names but I think this is better than the alternative of having a different name.

See my workaround

PR #2444, opened 2024-06-21
Csharp scope migration

AndreasArvidsson:
## Checklist


pokey (approved):
Made some tweaks in 2aa4fd381f1b712bd6dbb0846c2319d4df62381d; merge if happy

AndreasArvidsson:
I must say I'm not really a fan of having new be part of calleee. But appears to be what we do in other language yes so I don't think we should change it here.

pokey:
> I must say I'm not really a fan of having new be part of calleee. But appears to be what we do in other language yes so I don't think we should change it here.

yeah i was somewhat surprised by that too, but better to be consistent

PR #2450, opened 2024-06-24
Migrate python collection item scope

AndreasArvidsson:
## Checklist


packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
```diff
@@ -85,6 +85,13 @@ const scopeSupportFacets = [
   "branch.ternary",
   "branch.loop",
 
+  "collectionItem.import",
+  "collectionItem.import.iteration",
+  "collectionItem.importFrom",
+  "collectionItem.importFrom.iteration",
+  "collectionItem.globalVariablesDeclaration",
+  "collectionItem.globalVariablesDeclaration.iteration",
```
pokey, line 93: These feel a bit specific to python. I think the key here is to handle collection items when there is no enclosing delimiter pair. Should we just have `collectionItem.unenclosed` ahd `collectionItem.unenclosed.iteration`?

pokey (commented):
Looks good with minor comment

packages/common/src/scopeSupportFacets/scopeSupportFacets.types.ts
AndreasArvidsson: I do think we should be a bit more specific. Specifically separating collection items with other type of items. Import for example is probably something that we would find in many languages?
I guess we could boil it down to just two different scope facets but the point of the facets is to be quite specific isn't it?

pokey: I think the key thing is making sure that unenclosed comma-separated lists are supported. Everything else we get for free. So eg if the language requires wrapping their puppets in brackets, or doesn't support global decls with lists, or supports multiple assignment without parens, etc, the key thing is unenclosed lists. I don't feel super strongly, so if you feel strongly maybe compromise would be import statement but not global? The global thing seems pretty specific and really is more about the multiple decls rather than anything specific about global

AndreasArvidsson: updated

pokey (approved):

PR #2451, opened 2024-06-24
Added per language scope support docs

AndreasArvidsson:
## Checklist


AndreasArvidsson:
@pokey Looks like meta updater have some problem on windows. If I run the `lint:meta` script on my own computer it works fine and works fine on Linux and mac, but windows on ci is failing. Probably have something to do with line endings?

AndreasArvidsson:
Also I would prefer to remove the `unsupported` scope facet level. Since there's isn't really any difference between explicitly saying unsupported and not specifying the level at all.

pokey:
I would argue for keeping unsupported. Otherwise there's no way to know if it is unspecified because it's unsupported or if we just haven't gotten around to indicating its support level yet

AndreasArvidsson:
Yes I'm just doubting that we are ever going to actually do that :p
But sure let's keep it for now.

pokey:
> Yes I'm just doubting that we are ever going to actually do that :p

Sorry not sure I'm following: doubting whether we'll do what?

AndreasArvidsson:
> > Yes I'm just doubting that we are ever going to actually do that :p
> 
> Sorry not sure I'm following: doubting whether we'll do what?

Getting around to actually going through all scope facets and specifying unsupported.

pokey:
> Getting around to actually going through all scope facets and specifying unsupported. 

Exactly. That's why it's important to have the ability to distinguish between unsupported and just not having been specified yet. If we just take unspecified to mean it's unsupported, we won't be able to tell which ones are actually unsupported or just haven't gotten around to it

pokey:
Eg we could render the missing ones with a question mark in the docs to encourage people to contribute, and so they don't get the false impression it's not supported when really it is

AndreasArvidsson:
@pokey Any ideas how to fix the lint:meta problem on windows?

pokey:
Not sure but I'm not sure it's a good idea to use meta-updater for generating this .md file anyway. My instinct would be to do that during web build step. Is there some reason to prefer checking this file into source control?

AndreasArvidsson:
I always prefer if documentation is in markdown. If I need to navigate to a webpage to read the docs I'm probably not gonna do it as often.

pokey:
Yeah but this file is huge. Are you really going to read this file anyway?

AndreasArvidsson:
I guess that works as well. Where do you want this placed instead?

pokey:
Maybe we just generate it into docs folder during website build step and add that to gitignore? That's what we used to do with our api docs

AndreasArvidsson:
Where do we have existing configuration for that I can look at?

pokey:
I don't believe we do anything like that today. On second thought, though, we should prob do it with an MDX react component that imports the scope support tables

AndreasArvidsson:
That sounds reasonable. Have you done anything like that before or do you have some documentation I could look at how to do that with docusaurs?

pokey:
Just Google docusaurus MDX. We have some MDX stuff in your next.js root site as well. Look for files ending in mdx

If you get stuck let's discuss at meetup

pokey:
You just put it where you'd put a markdown file but use mdx file ending. Then import react component and use it. The react component would import support table from common and use it to render support table

AndreasArvidsson:
Oh you're support to put it in the roots/docs folder? How is it supposed to get access to the Cursorless common package from there? I'm afraid I don't really understand the build config here.

pokey:
Have a look at our other mdx files. As mentioned, they're in the next js part of our site but should be similar

pokey:
And check out docusaurus mdx docs

pokey:
Have the docusaurus mdx docs not been of any help?

pokey:
Sounds like you're stuck. Let's discuss at meetup

packages/common/src/util/sleep.ts
```diff
@@ -8,6 +6,6 @@ import { promisify } from "util";
  * on failure, and this function will sleep the same amount of time every time
  * the test is retried.  Prefer {@link sleepWithBackoff} instead.
  */
-const sleep = promisify(setTimeout);
-
-export default sleep;
+export default function sleep(timeout: number): Promise<void> {
```
AndreasArvidsson, line 9: util import is not available in docusaurus. It was either changing this or installing a polyfill

packages/cursorless-org-docs/docusaurus.config.mts
```diff
@@ -42,7 +42,7 @@ function remarkPluginFixLinksToRepositoryArtifacts(): Transformer<Root> {
         "../..",
       );
       const artifact = resolve(file.dirname!, url);
-      const artifactRelative = relative(repoRoot, artifact);
+      const artifactRelative = relative(repoRoot, artifact).replace(/\\/g, "/");
```
AndreasArvidsson, line 45: On windows the web path would contain backslash and therefor wouldn't work.

packages/cursorless-org-docs/plugins/node/index.js
```diff
@@ -0,0 +1,18 @@
+// eslint-disable-next-line
+module.exports = function (context, options) {
```
AndreasArvidsson, line 2: This plugging is necessarily to instruct webpack that docusaurus uses to ignore the imports off path and fs that we have in commons.

AndreasArvidsson:
@pokey documentation is now moved into the packages folder and we are using a proper react component with a per language mdx file.
`docusaurus start` works fine, but `docusaurus build` fails with the following errors :
```
Error: Docusaurus server-side rendering could not render static page with path /user/languages/yaml/ because of error: Cannot read properties of undefined (reading 'jsx')
```

pokey:
> @pokey documentation is now moved into the packages folder and we are using a proper react component with a per language mdx file. `docusaurus start` works fine, but `docusaurus build` fails with the following errors :
> 
> ```
> Error: Docusaurus server-side rendering could not render static page with path /user/languages/yaml/ because of error: Cannot read properties of undefined (reading 'jsx')
> ```

After a merge, this working fine for me, both locally and in CI, so I guess it just went away 🤷‍♂️

packages/cursorless-org-docs/plugins/node/index.js
pokey: I removed this hack now that all node references have been removed from commons 🎉

packages/cursorless-org-docs/src/docs/user/experimental/keyboard/images/keyboardBring.gif
pokey, line 1: it's a bit odd for all these markdown files, images, etc to be in our `src` directory 🤔. Is that necessary so that our `.mdx` files will be under our tree-sitter `root`?

packages/cursorless-org-docs/src/docs/contributing/adding-a-new-language.md
pokey, line 1: man these are some pretty deep drills 😅. See below re whether we could at least go one level shallower. But I guess if they live in the docs package there's not that much we can do

We could move `cursorless-org-docs` out of `packages`, but that would be pretty dramatic, as it would be the first TS package that doesn't live in `packages`, so probably not worth it

packages/meta-updater/src/metaUpdater.ts
pokey, line 1: so cool! 😍

packages/meta-updater/src/updateLanguageScopeSupportConfig.ts
```diff
@@ -0,0 +1,25 @@
...
+  actual: string | null,
+  options: FormatPluginFnOptions,
+): string | null {
+  if (options.manifest.name !== "@cursorless/cursorless-org-docs") {
+    return null;
+  }
+
+  if (actual != null) {
+    return actual;
+  }
```
pokey, line 14: huh interesting. I guess this means we're just auto-generating at the start but not trying to clobber if they change. Seems ok. Will be annoying if we change the `<Language>` element, but useful to be able to be a bit more custom with languages I guess

packages/cursorless-org-docs/tsconfig.json
pokey, line 1: oh man I don't envy you the struggle to get this one right 😅. Good work

pokey:
This is very cool, but I wonder if it is premature. I worry users might see the "supported" list at the top and assume everything else is unsupported. Maybe we should wait until we've migrated our recorded test fixtures using https://github.com/cursorless-dev/cursorless/issues/2390? I don't want this to pile up merge conflicts tho, esp since you've gotten mdx working in our docs site

I wonder if there's a way to ship this in a way that these lang pages are harder to find / hidden

packages/cursorless-org-docs/src/docs/user/experimental/keyboard/images/keyboardBring.gif
AndreasArvidsson: Our mdx files import typescripts they need to be in the source directory at least. When it comes to building webpages I don't think it's at all odd that we have images and stuff like that in source?

packages/cursorless-org-docs/src/docs/contributing/adding-a-new-language.md
AndreasArvidsson: Yeah I think they should live in packages still.

packages/meta-updater/src/updateLanguageScopeSupportConfig.ts
AndreasArvidsson: Exactly that was my intent. If we don't want anything custom per language then we don't really need mdx at all surfing this is the right direction.

AndreasArvidsson:
We could just hide them from the side menu I assume? I wanted this to be in the developer docs for that exact reason.

pokey:
> We could just hide them from the side menu I assume? I wanted this to be in the developer docs for that exact reason.
> 
> I would prefer to get this in.

oh yeah is that easy to do? I think just hiding them from sidebar is good for now

packages/meta-updater/src/updateLanguageScopeSupportConfig.ts
pokey: Yeah seems reasonable

packages/cursorless-org-docs/src/docs/user/experimental/keyboard/images/keyboardBring.gif
pokey: 🤷‍♂️

AndreasArvidsson:
> > We could just hide them from the side menu I assume? I wanted this to be in the developer docs for that exact reason.
> > I would prefer to get this in.
> 
> oh yeah is that easy to do? I think just hiding them from sidebar is good for now

I don't think so as long was we used out to generate the sidebar. We problem need to generate the list of items ourself if we want to have more control. Which is definitely something we can do and automate, but the question is if it's worth it? Should we just for now move languages to developer docs? We can always move them to user later.

pokey:
@AndreasArvidsson why did you downgrade docusaurus?

AndreasArvidsson:
> @AndreasArvidsson why did you downgrade docusaurus?

It did not compile and I didn't aim to update docusaurus in this pull request; it's already large enough.

pokey:
strange was working fine for me, but yeah agreed it's orthogonal

pokey:
btw wdyt of f189f124c44f4069ab6409170ad815f388435428 😄

AndreasArvidsson:
> strange was working fine for me, but yeah agreed it's orthogonal

Yeah if it would have worked fine I wouldn't really have cared, but I ran into major problems with pnpm and just reverting to main fixed it.

AndreasArvidsson:
> btw wdyt of [f189f12](https://github.com/cursorless-dev/cursorless/commit/f189f124c44f4069ab6409170ad815f388435428) 😄

Fine with me :)

pokey:
yeah confirmed it's working; doesn't appear in sidebar https://deploy-preview-2451--cursorless.netlify.app/docs/ but can go directly to page so it is still deploying eg https://deploy-preview-2451--cursorless.netlify.app/docs/user/languages/cpp/

AndreasArvidsson:
Nice. I'm happy with that.

packages/cursorless-org-docs/src/docs/user/languages/ScopeSupport.tsx
pokey, line 1: I don't love this representation, but no point in trying to get it perfect before we actually show it to users; happy to have this as a placeholder for now. The important thing about this PR imo is that we're now able to access `common` from `.mdx` files in our docs

AndreasArvidsson: Exactly

pokey (approved):
lgtm. i fixed a bunch of broken links. see https://github.com/cursorless-dev/cursorless/issues/2609 for the future because we should really catch those 😅

AndreasArvidsson:
Totally agree!

PR #2457, opened 2024-07-01
Migrated surrounding pair scope handler

### LINKED ISSUE #2316: "Take inside" doesn't work in C++ with #pragma line
In C++, if I use "take inside" to try and take the text "-Wdeprecated-declarations", it doesn't work.

```c
#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```

AndreasArvidsson:
Fixes #2316

## Checklist

- [x] Add `@disqualifyDelimiter` for all languages
- [x] Add plural form of surrounding pair [scopes.py](https://github.com/cursorless-dev/cursorless/blob/5e9a2e955570cb43d5d07de8c861365e8361cc36/cursorless-talon/src/modifiers/scopes.py#L48)
- [x] Run talon tests
- [x] add multiline quad tests for plaintext?


AndreasArvidsson:
@pokey I'm not sure why all the scope provide tests are failing.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
```diff
@@ -0,0 +1,121 @@
...
+      yield scope;
+    }
+  }
+
+  isPreferredOver(
+    scopeA: TargetScope,
+    scopeB: TargetScope,
+    position: Position,
+  ): boolean | undefined {
+    // In the case of `( [ )| ]` we prefer the one we're actually touching
```
pokey, line 112: Is that not the case for other scopes? Surprised this was necessary

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
```diff
@@ -0,0 +1,121 @@
...
+      this.languageDefinitions.get(this.languageId),
+      editor.document,
+      individualDelimiters,
+      delimiterRegex,
+    );
+
+    const surroundingPairs = getSurroundingPairOccurrences(
+      individualDelimiters,
+      delimiterOccurrences,
+    );
```
pokey, line 71: This is also quite brute-force. So if I'm sitting next to a paren and say "take round", cursorless regexes the entire file, then iterates over all delimiters to match them up

I don't want to prematurely optimise, but elsewhere we've been quite careful to leverage the fact we have generators so that we can support big log files. I'm not sure why we don't care when it comes to pairs?

pokey (commented):
Haven't reviewed everything, but left a few comments

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
```diff
@@ -0,0 +1,121 @@
...
+      this.languageId,
+    );
+    const delimiterRegex = getDelimiterRegex(individualDelimiters);
+
+    if (this.scopeType.forceDirection != null) {
+      //  DEPRECATED @ 2024-07-01
+      void showError(
+        ide().messages,
+        "deprecatedForceDirection",
+        "forceDirection not supported. Use 'next pair' or 'previous pair' instead",
```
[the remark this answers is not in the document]
AndreasArvidsson: Yes, but I don't think we should have a specific grammar for matching pairs here. You have the same use case with touching tokens and so on.

left and right and previous and next doesn't really behave exactly the same. Next for example includes the containing scope while right doesn't.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
```diff
@@ -0,0 +1,121 @@
...
+      );
+      return;
+    }
+
+    const delimiterOccurrences = getDelimiterOccurrences(
+      this.languageDefinitions.get(this.languageId),
+      editor.document,
+      individualDelimiters,
+      delimiterRegex,
+    );
```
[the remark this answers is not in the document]
AndreasArvidsson: I don't think that is a problem. In my extension I'm regularly regexing the entire user directory. I just don't want to make it unnecessarily complicated.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
AndreasArvidsson: Do we have any other examples of overlapping scopes?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
AndreasArvidsson: I wanted to do the same thing as we do with the syntax scopes using Tree sitter. The position of each scope is defined in the document and your cursor position has no bearing on it.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
```diff
@@ -0,0 +1,121 @@
...
+  constructor(
+    private languageDefinitions: LanguageDefinitions,
+    public readonly scopeType: SurroundingPairScopeType,
+    private languageId: string,
+  ) {
+    super();
+    const iterationScopeType: ScopeType = {
+      type: "oneOf",
+      scopeTypes: [this.scopeType, { type: "document" }],
+    };
```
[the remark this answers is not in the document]
AndreasArvidsson: We already have tests relying on this.I do agree it should be inside, but that is a bit more complicated as you said.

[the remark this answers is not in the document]
AndreasArvidsson: Should we just leave it as document? We can always go back and improve it later.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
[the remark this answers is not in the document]
pokey: To be clear I agree we remove left and right. I just mean message isn't quite right and we should try to change some of the left and right tests to use next and backward

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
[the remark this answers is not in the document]
pokey: Why are we bothering with generators at all then?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
pokey: No we don't support overlapping scopes. Are you suggesting we should support them? Guessing there was a failing test?

AndreasArvidsson: `(  [  )|  ]` is an overlapping scope which we have tests for. That is why I need this work around.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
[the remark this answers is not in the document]
AndreasArvidsson: You decided you wanted generators and I don't mind, but I'm not sure we actually need them at all. 
But that is not the point here. I know of no possible way of doing this where the scope visualizer will not go out of sync with containing if we do not look at the entire document.

AndreasArvidsson:
This is really interesting 
```
if the file is over 30MB or over 300k lines, we consider it to be a very large file and proceed to turn off very many features, including tokenization, word wrapping, indent guides, anything that would consume additional memory, etc.
```
https://github.com/Microsoft/vscode/issues/32118#issuecomment-320952246

That means that Cursorless can never encounter a file with over three hundred thousand lines. I just did a test with the surrounding pair regex on a file with just below three hundred thousand lines and that clocked in at `274ms`. I would say that is a bit on the high side so we might want to do some windowing, but we definitely don't need to be as careful as we was before. Doing it at tens of thousand lines at the time probably is fine.

pokey:
is there a reason this is still marked as draft?

AndreasArvidsson:
Yes the checkboxes in the description are not done. I wasn't sure this implementation was going to get your approval so I didn't go further originally.

pokey:
Got it. So would it make sense to knock those out now that we're aligned on the high-level direction or did you want me to take a look first?

AndreasArvidsson:
Please have a look first. Some of the implementation details I'm not sure off myself. The single line strings is one thing that come to mind.

AndreasArvidsson:
@pokey That simplification looks great! :)

cursorless-talon/src/modifiers/scopes.py
```diff
@@ -24,16 +27,27 @@ def cursorless_scope_type(m) -> dict[str, str]:
...
+    return {
+        "type": "customRegex",
+        "regex": m.cursorless_custom_regex_scope_type,
+    }
 
 
 @mod.capture(
-    rule="{user.cursorless_scope_type_plural} | <user.cursorless_glyph_scope_type_plural> | {user.cursorless_custom_regex_scope_type_plural}"
+    rule="{user.cursorless_scope_type_plural}"
+    " | <user.cursorless_surrounding_pair_scope_type>"
```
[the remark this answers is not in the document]
pokey: Ah right I see you left a checkbox for that 👍

packages/cursorless-engine/src/processTargets/modifiers/ItemStage/getIterationScope.ts
```diff
@@ -147,23 +146,41 @@ function getParentSurroundingPair(
...
+    type: "containingScope",
+    scopeType: {
+      type: "surroundingPair",
+      delimiter: "collectionBoundary",
+      requireStrongContainment: true,
+    },
   });
+  try {
+    const targets = pairStage.run(target);
+    return targets.length > 0 ? targets[0] : undefined;
```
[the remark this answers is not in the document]
AndreasArvidsson: Shouldn't that be a feature of the containing scope stage? Every one who uses that shouldn't need to check to make sure that containing doesn't return more than one?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/delimiterMaps.ts
```diff
@@ -46,11 +49,13 @@ const delimiterToTextOverrides: Record<string, Partial<DelimiterMap>> = {
   },
 
   python: {
-    // FIXME: We technically can't distinguish between single and double quotes
-    // now, but we'll revisit all this; see
-    // https://github.com/cursorless-dev/cursorless/issues/1812#issuecomment-1691493746
-    singleQuotes: ["string_start", "string_end"],
-    doubleQuotes: ["string_start", "string_end"],
+    doubleQuotes: [['"', 'f"'], '"', true],
```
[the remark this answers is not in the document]
AndreasArvidsson: Sounds reasonable

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  *generateScopeCandidates(
+    editor: TextEditor,
+    position: Position,
+    direction: Direction,
+    hints: ScopeIteratorRequirements,
+  ): Iterable<TargetScope> {
+    const individualDelimiters = getIndividualDelimiters(
+      this.scopeType.delimiter,
+      this.languageId,
```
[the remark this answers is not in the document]
AndreasArvidsson: What about this scenario?
```ts
`foo ${value} bar`
```

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/getDelimiterOccurrences.ts
```diff
@@ -0,0 +1,56 @@
...
+  languageDefinition: LanguageDefinition | undefined,
+  document: TextDocument,
+  individualDelimiters: IndividualDelimiter[],
+  delimiterRegex: RegExp,
+): DelimiterOccurrence[] {
+  if (individualDelimiters.length === 0) {
+    return [];
+  }
+
+  const queryMatches = languageDefinition?.getMatches(document) ?? [];
```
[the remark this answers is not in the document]
AndreasArvidsson: That and it's a simpler interface to work with.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/getDelimiterOccurrences.ts
```diff
@@ -0,0 +1,46 @@
...
+  return matchAll(text, delimiterRegex, (match): DelimiterOccurrence => {
+    const text = match[0];
+    const range = new Range(
+      document.positionAt(match.index!),
+      document.positionAt(match.index! + text.length),
+    );
+
+    return {
+      delimiterInfo: delimiterTextToDelimiterInfoMap[text],
+      isDisqualified: disqualifyDelimiters.some((r) => r.contains(range)),
```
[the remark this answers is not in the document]
AndreasArvidsson: `matchAll` runs on all matches. Iterating over the entire list one extra time is something we could do of course, but I preferred the boolean solution.

packages/cursorless-engine/src/processTargets/modifiers/ItemStage/getIterationScope.ts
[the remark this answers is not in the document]
pokey: Sorry not following. It's perfectly acceptable for containing scope to return more than one in the general case. Eg "tags". But here, you're just silently ignoring if there's more than one

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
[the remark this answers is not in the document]
pokey: Huh interesting point. I also thought about this some more and a) doing something better here is almost certainly not worth the work and b) it might actually be nice if behaviour were consistent on commented code even though it's technically textual

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/getDelimiterOccurrences.ts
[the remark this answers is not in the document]
pokey: One advantage of using a proper scope is we could benefit from all of the query predicates and fanciness such as comment merging in your other PR that I will review one day 😅

packages/cursorless-engine/src/processTargets/modifiers/ItemStage/getIterationScope.ts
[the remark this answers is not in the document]
AndreasArvidsson: That is also the only example we have? I'm not even sure it's an error in that case maybe we should just return undefined we couldn't find a singular containing pair.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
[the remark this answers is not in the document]
AndreasArvidsson: That is probably useful as well

packages/cursorless-engine/src/processTargets/modifiers/scopeTypeStages/BoundedNonWhitespaceStage.ts
```diff
@@ -64,4 +58,20 @@ export class BoundedNonWhitespaceSequenceStage implements ModifierStage {
...
+    const pairStage = this.modifierStageFactory.create({
+      type: "containingScope",
+      scopeType: {
+        type: "surroundingPair",
+        delimiter: "any",
+        requireStrongContainment: true,
+      },
+    });
+    try {
+      return pairStage.run(target)[0];
```
pokey, line 72: As above we should probably throw here

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/getDelimiterOccurrences.ts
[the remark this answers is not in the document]
pokey: ok let's leave it

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairScopeHandler.ts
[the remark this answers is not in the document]
pokey: yeah let's leave it for now. I think it would be a lot more complex to get this one perfect so let's not waste the effort until we decide it's actually necessary

packages/cursorless-engine/src/processTargets/modifiers/ItemStage/getIterationScope.ts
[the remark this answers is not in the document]
pokey: Shall we add a utility function that takes a surrounding pair scope modifier and returns a surrounding pair target? then you could throw there, and not need to add a `!` on interior, etc. Can use that here and in short paint

pokey:
Ok I've now fully reviewed this PR. I made a bunch of changes. I think what remains are:

- [ ] Review [my changes](https://github.com/cursorless-dev/cursorless/pull/2457/files/39d2309155731c71939f3c2d25750168f734564e..db97b693a266e274de02994aae80641338d281f5) to make sure you're happy
- [ ] Finish the checkboxes in the description
- [ ] Address my remaining unresolved comments, as well

Getting really close tho! Very excited to have this one

packages/cursorless-engine/src/processTargets/modifiers/ItemStage/getIterationScope.ts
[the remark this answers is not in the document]
AndreasArvidsson: That we can do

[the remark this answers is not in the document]
AndreasArvidsson: done

packages/cursorless-engine/src/processTargets/modifiers/ItemStage/getIterationScope.ts
```diff
@@ -147,23 +146,41 @@ function getParentSurroundingPair(
...
+    type: "containingScope",
+    scopeType: {
+      type: "surroundingPair",
+      delimiter: "string",
+      requireStrongContainment: true,
+    },
   });
+  try {
+    const targets = pairStage.run(target);
+    return targets.length > 0 ? targets[0] : undefined;
```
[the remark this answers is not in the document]
AndreasArvidsson: done

packages/cursorless-engine/src/processTargets/modifiers/scopeTypeStages/BoundedNonWhitespaceStage.ts
AndreasArvidsson: done

AndreasArvidsson:
Ready for review. I have added `@disqualifyDelimiter` to all the languages that have boolean expressions to my knowledge.

data/fixtures/scopes/c/disqualifyDelimiter.scope
pokey, line 1: wow so much noise from the removal / leading / whatever. I still argue we should ignore related ranges when they are the same as token. obv out of scope here tho

i guess a simpler solution would be to ignore them for specific special scopes like `disqualifyDelimiter` and `textFragment`

data/fixtures/scopes/php/disqualifyDelimiter.scope
```diff
@@ -0,0 +1,106 @@
+<?php
```
pokey, line 1: what about this one?

data/fixtures/scopes/c/disqualifyDelimiter.scope
```diff
@@ -0,0 +1,84 @@
+1 < 2
+1 > 2
+1 <= 2
+1 >= 2
+---
```
pokey, line 5: what about pointer dereference `->`?

data/fixtures/scopes/c/disqualifyDelimiter.scope
AndreasArvidsson: Totally agree we need a clean up to the generator code.

data/fixtures/scopes/c/disqualifyDelimiter.scope
pokey: you're also missing shift operators `>>`. How did you come up with these lists?

queries/javascript.core.scm
pokey, line 1: you're missing `=>` in typescript

pokey (commented):
Unfortunately I think to do this properly we need to look through the `grammar.json` for each language. Looks like you missed a lot of operators

For example, I just grepped for `>` and `<` in https://github.com/tree-sitter/tree-sitter-c/blob/master/src/grammar.json

data/fixtures/scopes/c/disqualifyDelimiter.scope
AndreasArvidsson: added

data/fixtures/scopes/php/disqualifyDelimiter.scope
AndreasArvidsson: added. There is no node for `<` separate, but I think it's fine to just exclude that entire thing.

queries/javascript.core.scm
AndreasArvidsson: added

data/fixtures/recorded/surroundingPair/parseTree/python/clearPair5.yml
```diff
@@ -10,7 +10,7 @@ command:
           scopeType: {type: surroundingPair, delimiter: any}
   usePrePhraseSnapshot: true
 initialState:
-  documentContents: "def funk(words:list<str>):"
+  documentContents: "def funk(words:list[str]):"
```
pokey, line 13: This is unfortunate. I'm guessing this is a test you added to make sure that pairs were predictable in the event of a bad parse. Seems the approach in this PR has backfired?

data/fixtures/scopes/c/disqualifyDelimiter.scope
pokey: I think it might make sense to just add that special case for these special scopes. Should just be a few lines, no? These are quite difficult to review as is

data/fixtures/recorded/surroundingPair/parseTree/python/clearPair5.yml
pokey: We used to have logic to bail on parse-tree support if you're in an error node. Maybe we need to reintroduce that?

AndreasArvidsson: Yes that was my intent, but we don't have access to the syntax node on the `QueryCapture` interface. With our logic of merging multiple captures I'm not even sure how to get the correct node.

AndreasArvidsson: This is the last place we actually have access to the node on `match.captures`.
https://github.com/cursorless-dev/cursorless/blob/fb6b846efb96c595f16748ca5597b5e4ab441c0e/packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts#L102

data/fixtures/scopes/c/disqualifyDelimiter.scope
AndreasArvidsson: Have a look now

data/fixtures/recorded/surroundingPair/parseTree/python/clearPair5.yml
pokey: I would just add an isError field on the final match and set it to true if any of the nodes in its captures have an error. Similar to how we use `some` form `allowMultiple` below

AndreasArvidsson: Here is an implementation that restores the test by looking for error nodes. I'm not in love with how invasive this was, but I don't really know of a simpler solution.
https://github.com/cursorless-dev/cursorless/pull/2457/commits/709b19edb39e1937cc0dd0542c08a2327f9d065b

pokey: Didn't we used to walk up ancestors looking for errors too? If so we should prob do that still

AndreasArvidsson: This implementation is do in that

AndreasArvidsson: > I would just add an isError field on the final match and set it to true if any of the nodes in its captures have an error. Similar to how we use `some` for `allowMultiple` below

The reason I'm not doing that is performance. Walking up the ancestor tree all the way to the route for every captured node in every scope scenario?

AndreasArvidsson: We could of course add a method so we can do it lazily?

AndreasArvidsson: This isn't to bad?
https://github.com/cursorless-dev/cursorless/pull/2457/commits/ec3a611f9506abef4ea9f177851bf2df0673f9d5

data/fixtures/scopes/cpp/disqualifyDelimiter.scope
pokey, line 1: I would be tempted to remove the first 9 lines, as we get them for free from `c` do to https://github.com/saidelike/cursorless/blob/main/data/fixtures/scopes/cpp/index.json

AndreasArvidsson: Just as sanity check. I can do that if you prefer

AndreasArvidsson:
@pokey I have now gone through all the grammar json files an added all the operators I can find that clashes. I might have missed something some of these files are quite long and complicated, but I done my best.

data/fixtures/scopes/cpp/disqualifyDelimiter.scope
pokey: I would. No reason to unnecessarily duplicate tests. That's why we built facet test import / inheritance machinery

AndreasArvidsson: done

packages/cursorless-engine/src/languages/TreeSitterQuery/QueryCapture.ts
```diff
@@ -6,13 +6,19 @@ import { Point } from "web-tree-sitter";
...
   readonly id: number;
   readonly type: string;
+  readonly parent: SimpleSyntaxNode | null;
+  readonly children: Array<SimpleChildSyntaxNode>;
+}
+
+/**
+ * Add start and end position to the simple syntax node. Used by the `child-range!` predicate.
+ */
+interface SimpleChildSyntaxNode extends SimpleSyntaxNode {
```
pokey, line 19: just to make sure I'm understanding, you've just cleaned it up so we don't need to use the `Omit`; the types are just as strict as before

AndreasArvidsson: > just to make sure I'm understanding, you've just cleaned it up so we don't need to use the `Omit`; the types are just as strict as before

Exactly. Even a little bit more. Now we don't have start and and position on the parent for example. I was trying to make it more explicit that it's just for the children and just for this specific predicate we need to have these fields.

packages/cursorless-engine/src/languages/TreeSitterQuery/isContainedInErrorNode.ts
```diff
@@ -0,0 +1,19 @@
+import type { SyntaxNode } from "web-tree-sitter";
+
+/**
+ * Determines whether the given node or one of its ancestors is an error node
+ * @param node The node to check
+ * @returns True if the given node is contained in an error node
+ */
+export function isContainedInErrorNode(node: SyntaxNode) {
+  let currentNode: SyntaxNode | null = node.parent;
```
pokey, line 9: why do you start with the parent? Why not start with the node itself

AndreasArvidsson: The dangers of copilot. Fixed now.

pokey (commented):
wow that disqualify stuff looks like it was a lot of work. Nice job. We should prob have some negative tests, ie that we're not disqualifying delimiters that we do want, but I think we have some tests for those and this PR is growing long in the tooth so I'd be happy to leave those for now.

As discussed, let's drive this for a couple days and then ship it

AndreasArvidsson:
> wow that disqualify stuff looks like it was a lot of work. Nice job. We should prob have some negative tests, ie that we're not disqualifying delimiters that we do want, but I think we have some tests for those and this PR is growing long in the tooth so I'd be happy to leave those for now.
> 
> As discussed, let's drive this for a couple days and then ship it

It was definitely some effort. We have a few tests. If anything comes to mind please record a few. Otherwise I would like to just get this in at the stage.
I've been running it for this entire day and except for the bug that was fixed earlier it has been working fine.

pokey:
I wonder if we should make iteration scope be "line". I think being able to say "first string" / "last string" will be really useful

AndreasArvidsson:
> I wonder if we should make iteration scope be "line". I think being able to say "first string" / "last string" will be really useful

Why not?

packages/cursorless-vscode-e2e/src/suite/serializeTargetRange.ts
```diff
@@ -36,7 +36,7 @@ export function serializeTargetRange(
   const lines: string[] = [];
 
   // Number of characters in the line number + `|`
-  const startIndent = `${start.line}`.length + 1;
+  const startIndent = start.line.toString().length + 1;
```
pokey, line 39: you just have to have the last word 🙄

but yeah this is better 😄

pokey (approved):

Issue #2461, opened 2024-07-02
Don't install pre-commit hooks in nix flake


pokey:
Why do we install pre-commit hooks in our nix flake? I would be strongly inclined to remove that step, but maybe there's some reason we have to do this as setting up the nix flake? Totally missed that that was what was happening in #1908

cc/ @auscompgeek @fidgetingbits

fidgetingbits:
It's not required no. Since it's a development shell, and I assume people using it would be committing things, it made sense to install the hooks? I don't mind removing it, but also curious why you think it's bad? Does it clobber existing tools if you already had them and then test the dev shell or something?

Lots of flakes do it afaik, but also would be ok to just print a message in the shellHook saying it pre-commit install can be run if needed?

pokey:
Just feels a bit invasive. I don't have them installed locally so was surprised when I couldn't commit something and couldn't figure out why. Turned out it was because I had tested out the flake PR

fidgetingbits:
Ah interesting, that makes sense. Bad assumption on my part. I'll send a PR to remove it. 

Why don't you use pre-commit locally out of curiosity. I specifically wonder about cases where it would catch stuff that can't be auto fixed by running it in CI?

pokey:
I prefer my git commit to be dumb, fast, and predictable. It's plumbing

PR #2471, opened 2024-07-09
remove pre-commit install from nix dev shell (#2461)

fidgetingbits:
## Checklist

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -44,7 +44,7 @@
             buildInputs = [ pkgs.bashInteractive ];
             shellHook = ''
               if [ ! -f .git/hooks/pre-commit ]; then
-                pre-commit install
+                echo "You can run 'pre-commit install' to install git commit hooks if you want them."
               fi
 
               pnpm install


pokey (approved):
Thank you!

flake.nix
```diff
@@ -44,7 +44,7 @@
             buildInputs = [ pkgs.bashInteractive ];
             shellHook = ''
               if [ ! -f .git/hooks/pre-commit ]; then
-                pre-commit install
+                echo "You can run 'pre-commit install' to install git commit hooks if you want them."
               fi
 
               pnpm install
```
auscompgeek, line 50: Since the pre-commit install has been removed, should we remove the pnpm install too? It's not exactly fast when you consider how fast a shell typically starts.

pokey: I unfortunately lack some understanding about how nix works, but I would imagine that when creating the environment, we want to do a pnpm install, but not when just activating the shell. Not sure if that makes sense / is possible

Issue #2353, opened 2024-05-23
Deprecate Cursorless snippet format in favour of community snippets


pokey:
- [x] #2354
- [x] #2355
- [x] https://github.com/cursorless-dev/cursorless/pull/2733
- [x] https://github.com/cursorless-dev/cursorless/pull/2744
- [x] https://github.com/cursorless-dev/cursorless/pull/2747
- [x] https://github.com/cursorless-dev/cursorless/pull/2789
- [x] https://github.com/talonhub/community/pull/1718
- [x] https://github.com/cursorless-dev/cursorless/pull/2804

Later:
- [ ] #2858

Re-add functionality that was only supported by legacy snippets:

- [/] ~~Context specific scope matching (i.e. method inside of a class versus outside)~~
     @pokey will try to use `"snip method"`. Adding a bunch of machinery for a single snippet in a single language was deemed (probably) not worth it.

Tabled:

- [x] #2356
- [x] #2357

jaresty:
What do you think about having the migrate command optionally migrate to the directory specified as the user's snippets directory? (In case folks want to keep snippets outside of community). Asking for others like me, since I don't need to migrate 😀

pokey:
If you have a custom snippets directory, then yes that's where they'll end up

pokey:
@AndreasArvidsson reading this one again, I think this change would make the entire `Snippets` component that you moved to `VSCodeSnippets` in #2478 will just go away entirely, because Cursorless VSCode would no longer be responsible for keeping track of snippets at all, as we'd always send the body from community. Does that sound right?

So the only place we'd be using the file system would be for "snip make", but we can figure out a way to make that work. The key thing is that we will no longer need a `Snippets` component so we really don't care about moving it into file-system-common / upgrading it to new format / changing it to use file watcher. Did I get that right?

I guess there's the issue of Cursorless keyboard, but we could probably make that work external to the engine, the way the rest of cursorless keyboard-specific stuff is done in vscode side today

AndreasArvidsson:
Unless we want to support named snip beds for Cursorless keyboard, then yes.

That sounds correct.

pokey:
Yeah I think if we wanted to support named snippets with keyboard we'd do it external to engine, like the rest of keyboard support

AndreasArvidsson:
Then there is no problem :)

phillco:
After discussing this with @AndreasArvidsson, we've updated the list of planned items in the description.

[...]

PR #2478, opened 2024-07-09
Pass snippets as argument to engine

AndreasArvidsson:
Makes it so that Cursorless everywhere editors doesn't have to provide a snippet implementation

## Checklist


pokey (commented):
I haven't read in detail, but my first two reactions are as follows:

Moving snippets from engine to vscode feels like a lateral at best. We went through the effort to abstract vscode out of the snippets component, so I would prefer if we didn't reverse that. Ie this makes it harder for us to use snippets in neovim or other node contexts

Second issue is that it doesn't really make the core much leaner if we still have a bunch of references to snippets in the engine. They're only used in two actions right? I'd be tempted to think the right abstraction is to allow passing in custom actions. Then snippets are never even mentioned

But I didn't read super closely so maybe missing something

AndreasArvidsson:
I've had another look and started a more more abstracted implementation and bailed. Basically most of what is in `Snippets.ts` that I here reamed to `VscodeSnippets.ts` is vscode specific. The fact that we are polling the file system. The handling of directory error message. The fact that you can a register third party snippets to some extent. If I remove everything that isn't vscode and make an abstraction it basically becomes more code for no good reason. I stand by my original assessment that this should probably live in vscode until we make a refactor where we used the new snippet format and don't poll.

AndreasArvidsson:
Implementations like this also makes it a bit hard to abstract snippets. We are explicitly creating a file in the file system and then opening in the editor. To open the file we need a filepath so we are still stuck in the file system with snippets.
https://github.com/cursorless-dev/cursorless/blob/2e69df8b9b8852a1df6cba5a6b5b7a73d82d6c63/packages/cursorless-engine/src/actions/GenerateSnippet/openNewSnippetFile.ts#L20

pokey:
> Implementations like this also makes it a bit hard to abstract snippets. We are explicitly creating a file in the file system and then opening in the editor. To open the file we need a filepath so we are still stuck in the file system with snippets. Here my inclination is just to drill the file system instance through and add touch to that. We can later on revamp the entire snippet system and make its smarter, but for now I can a just need to be unblocked. Sounds reasonable?
> 
> https://github.com/cursorless-dev/cursorless/blob/2e69df8b9b8852a1df6cba5a6b5b7a73d82d6c63/packages/cursorless-engine/src/actions/GenerateSnippet/openNewSnippetFile.ts#L20

Wouldn't my second suggestion above solve this problem? Ie generate snippet, insert snippet and wrap with snippet are just custom actions passed into the engine. If this isn't making sense let's discuss tomorrow alongside vscode vs engine thing

AndreasArvidsson:
That sounds like a reasonable solution, but to be fair I don't really have the time for that right now. This was only moving a single file that will unblock me. And we can then later make a more flexible approach. Happy to discuss it tomorrow.

pokey:
Let's discuss tomorrow

pokey:
update from meet-up: pokey caved

pokey (commented):
The one thing I think we discussed changing here is creating a `DisabledSnippets` class so that we can confine null checks to `createEngine`. Otherwise looks good; see also https://github.com/cursorless-dev/cursorless/issues/2353#issuecomment-2222821726

packages/cursorless-engine/src/cursorlessEngine.ts
```diff
@@ -30,24 +31,22 @@ import { ScopeSupportChecker } from "./scopeProviders/ScopeSupportChecker";
...
-import { KeyboardTargetUpdater } from "./KeyboardTargetUpdater";
+import { DisabledSnippets } from "./disabledComponents/DisabledSnippets";
 
 export async function createCursorlessEngine(
   treeSitter: TreeSitter,
   ide: IDE,
   hats: Hats,
   commandServerApi: CommandServerApi | null,
   fileSystem: FileSystem,
+  snippets: Snippets | undefined = new DisabledSnippets(),
```
pokey, line 42:
```suggestion
  snippets: Snippets = new DisabledSnippets(),
```

pokey (commented):
Looks good with minor nit

pokey (approved):

PR #2480, opened 2024-07-09
Remove node references from Talon spoken form reader

AndreasArvidsson:
## Checklist


packages/cursorless-vscode/src/ide/vscode/VscodeFileSystem.ts
```diff
@@ -4,11 +4,13 @@ import {
   PathChangeListener,
   RunMode,
 } from "@cursorless/common";
+import * as fs from "fs/promises";
 import { isAbsolute, join } from "path";
 import * as vscode from "vscode";
 
 export class VscodeFileSystem implements FileSystem {
   public readonly cursorlessTalonStateJsonPath: string;
+  readonly cursorlessTalonStateJsonDirPath: string;
```
AndreasArvidsson, line 13: This isn't necessary if we go with https://github.com/cursorless-dev/cursorless/pull/2481 so that one should probably merge first

pokey:
Can you elaborate a bit on why you've gone this route? The idea was for this to be node-specific and move it into a separate package

AndreasArvidsson:
Definitely. I aim for all Cursorless features to be available in a non node environment. Since we already have a file system representation this was an easy win to share as much functionality as possible between different implementations while still supporting non node environments.

pokey:
But wouldn't talon js just get the spoken forms from talon?

AndreasArvidsson:
In the long run maybe, but that would require changes to cursorless-Talon. Today the source of truth for the spoken forms is a json file and it's much easier to just read that. 

Let's look at it from another angle. Why even have a file system abstraction if all file system interaction is gonna be via node?

I'm starting to lean towards actually removing the file system interface and have more custom interfaces for the editor to provide the specific data in a more environment agnostic way.

pokey:
> In the long run maybe, but that would require changes to cursorless-Talon. Today the source of truth for the spoken forms is a json file and it's much easier to just read that. 

Yeah I'm ok with keeping the file thing in the short term if it's a lot easier, but does feel weird to use file-based rpc when you're in the same process

> I'm starting to lean towards actually removing the file system interface and have more custom interfaces for the editor to provide the specific data in a more environment agnostic way.

Exactly. That's what I was proposing Tuesday

AndreasArvidsson:
@pokey Refactored
I was trying to remove `FileSystem.watchDir` but id notice that that actually is a vcode implementation. I'm very tempted not having this at all in node common, but having this just be accustom vscode implementation of the spoken form reader. In that case we don't even need the file system instance.

AndreasArvidsson:
This latest version I like. Great engine gets the Talon spook in form as an argument and vscode has its own implementation. Then later in a separate pull request we can I remove the file system abstraction since that is used in other places as well that needs to be updated.

pokey:
update from meet-up:

- move json reader to a new package called `file-system-common`

AndreasArvidsson:
@pokey ready

packages/cursorless-engine/src/cursorlessEngine.ts
```diff
@@ -63,10 +65,8 @@ export async function createCursorlessEngine(
   const languageDefinitions = new LanguageDefinitions(fileSystem, treeSitter);
   await languageDefinitions.init();
 
-  const talonSpokenForms = new TalonSpokenFormsJsonReader(fileSystem);
-
   const customSpokenFormGenerator = new CustomSpokenFormGeneratorImpl(
-    talonSpokenForms,
+    talonSpokenForms ?? new DisabledTalonSpokenForms(),
```
pokey, line 69: did you try forcing this to use disabled and seeing if cursorless side bar reverts to default spoken forms?

pokey (commented):
looks good but just want to make sure you had a chance to try this locally; see comment below

packages/cursorless-engine/src/cursorlessEngine.ts
AndreasArvidsson: You were actually right. Returning an empty list made it so that all spoken forms were disabled :D
I made a commit to support this new usecase

packages/cursorless-engine/src/disabledComponents/DisabledTalonSpokenForms.ts
```diff
@@ -0,0 +1,11 @@
+import type { TalonSpokenForms } from "../scopeProviders/TalonSpokenForms";
+
+export class DisabledTalonSpokenForms implements TalonSpokenForms {
+  getSpokenFormEntries() {
+    return Promise.resolve(null);
```
pokey, line 5: if you just throw here, I believe it will show you a message and fall back to defaults anyway. Any reason not to just do that?

AndreasArvidsson: Because I don't want an error message every time. This is not an error this is an intended behavior.

pokey: I don't think it's unreasonable to get an error message if they try to use custom spoken forms when they're unsupported. This only shows when you pull up scope side bar right?

AndreasArvidsson: But they're not trying to use customs spoken forms exactly. they're just trying to use the scope visualizer. For a user to get an error message that is an implementation decision by a developer feels wrong.

AndreasArvidsson: The intent behind not supplying spoken forms is to use default. This is not an error to my mind.

pokey: Yes but they should know that those spoken forms are not correct

pokey: Keep in mind that the scope side bar doesn't exist in the engine; it's defined in VSCode. So if we had some ide that wants to support scope sidebar without showing spoken forms, we could make it so that it doesn't request custom spoken forms. But to my mind, at the layer we're operating in this PR, the components would be functioning as intended if they complain when you ask them for custom spoken forms, because that's their job

pokey: if nothing else, it ends up being a smaller PR, so there's always that 😄

pokey (approved):
👍

PR #2481, opened 2024-07-09
Replace path with pathe

AndreasArvidsson:
This is a simple way of getting rid of node path references in common and engine. I thought it best to also replace it in vscode so we don't have different implementations for path in the bundle. I guess for things like running tests and ours setup scrips then just using node path is probably fine? I could go either way.

https://www.npmjs.com/package/pathe

## Checklist


pokey (approved):
I think we're on the same page that any references to `path` in the engine are a smell, but this is a fine way forward for now

AndreasArvidsson:
Totally in agreements there

PR #2488, opened 2024-07-11
Added Tree sitter provider

AndreasArvidsson:
## Checklist

- [] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-talon/src/cheatsheet)

### DIFF
--- a/packages/cursorless-engine/src/languages/LanguageDefinition.ts
+++ b/packages/cursorless-engine/src/languages/LanguageDefinition.ts
@@ -1,12 +1,12 @@
 import {
-  FileSystem,
   ScopeType,
   SimpleScopeType,
   showError,
+  type IDE,
+  type RawTreeSitterQueryProvider,
 } from "@cursorless/common";
-import { basename, dirname, join } from "pathe";
+import { dirname, join } from "pathe";
 import { TreeSitterScopeHandler } from "../processTargets/modifiers/scopeHandlers";
-import { ide } from "../singletons/ide.singleton";
 import { TreeSitter } from "../typings/TreeSitter";
 import { matchAll } from "../util/regex";
 import { TreeSitterQuery } from "./TreeSitterQuery";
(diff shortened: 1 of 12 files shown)


packages/cursorless-engine/src/languages/LanguageDefinition.ts
```diff
@@ -91,43 +90,42 @@ export class LanguageDefinition {
  * @returns The text of the query file, with all imports inlined
  */
 async function readQueryFileAndImports(
-  fileSystem: FileSystem,
-  languageQueryPath: string,
+  ide: IDE,
+  provider: RawTreeSitterQueryProvider,
+  languageQueryName: string,
```
AndreasArvidsson, line 95: The reason for switching from path to name is that if we have a non file system implementation of the language provider, path is irrelevant. I think every definition name(scm filename) should just be unique and in which folder they are located doesn't really matter.

packages/cursorless-engine/src/languages/LanguageDefinitions.ts
```diff
@@ -1,31 +1,56 @@
...
 import { LanguageDefinition } from "./LanguageDefinition";
-import { toString } from "lodash-es";
 
 /**
  * Sentinel value to indicate that a language doesn't have
  * a new-style query definition file
  */
 const LANGUAGE_UNDEFINED = Symbol("LANGUAGE_UNDEFINED");
 
+export interface LanguageDefinitions {
```
pokey, line 23: I removed the lifecycle methods from this interface (`init` and `dispose`). I don't think interfaces should expose lifecycle methods. Clients of an interface shouldn't be responsible for the lifecycle of any given instance; that is the job of whoever constructs the actual class

pokey (approved):
looks good! See my various changes and related comment below. Merge if you're happy

packages/cursorless-engine/src/languages/LanguageDefinitions.ts
AndreasArvidsson: Very true

PR #2493, opened 2024-07-12
Added `CURSORLESS_MODE` environment variable

### LINKED ISSUE #2408: Switch from `CURSORLESS_TEST` to `CURSORLESS_MODE` environment variable
It can be either `"test"`, `"develop"`, or unset. If it is unset, it means production, because we don't have an easy way to set it in production and prob best not to have extra env vars in production just in case

- [ ] We should also try to avoid using `extension.runMode`
- [ ] We could also add an assert that `extension.runMode` is equal to `CURSORLESS_MODE` for sanity checking

Note that this is how it is done in our neovim support

See also https://github.com/cursorless-dev/cursorless/pull/2301

AndreasArvidsson:
Fixes #2408

## Checklist

### DIFF
--- a/packages/cursorless-vscode/src/ide/vscode/VscodeEnabledHatStyleManager.ts
+++ b/packages/cursorless-vscode/src/ide/vscode/VscodeEnabledHatStyleManager.ts
@@ -1,4 +1,3 @@
-import { isTesting } from "@cursorless/common";
 import { pickBy } from "lodash-es";
 import * as vscode from "vscode";
 import { HatStyleInfo, HatStyleMap } from "@cursorless/common";
@@ -33,7 +32,7 @@ export default class VscodeEnabledHatStyleManager {
   hatStyleMap!: ExtendedHatStyleMap;
   private notifier: Notifier<[HatStyleMap]> = new Notifier();
 
-  constructor(extensionContext: vscode.ExtensionContext) {
+  constructor(private extensionContext: vscode.ExtensionContext) {
     this.recomputeEnabledHatStyles = this.recomputeEnabledHatStyles.bind(this);
 
     extensionContext.subscriptions.push(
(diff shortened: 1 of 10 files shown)


packages/cursorless-vscode/src/ide/vscode/VscodeEnabledHatStyleManager.ts
```diff
@@ -76,9 +75,10 @@ export default class VscodeEnabledHatStyleManager {
     colorPenalties.default = 0;
 
     // So that unit tests don't fail locally if you have some colors disabled
-    const activeHatColors = isTesting()
-      ? HAT_COLORS.filter((color) => !color.startsWith("user"))
-      : HAT_COLORS.filter((color) => colorEnablement[color]);
+    const activeHatColors =
+      this.extensionContext.extensionMode === vscode.ExtensionMode.Test
```
AndreasArvidsson, line 79: I think it's fine for the vscode ide to internally use extension mode. We are asserting that this one is equal to the environment variable.

packages/cursorless-vscode/src/ide/vscode/VscodeRunMode.ts
```diff
@@ -8,5 +8,29 @@ const EXTENSION_MODE_MAP: Record<ExtensionMode, RunMode> = {
...
+  if (envMode !== extensionMode) {
+    throw new Error(
+      `Extension mode '${extensionMode}' doesn't match environmental mode '${envMode}'`,
+    );
+  }
+
+  return extensionMode;
+}
+
+function getEnvironmentalMode(): RunMode {
```
saidelike, line 23: this should probably be moved outside of cursorless-vscode so it can be used by neovim

saidelike (commented):
sounds sane (just a minor comment)

packages/cursorless-vscode/src/ide/vscode/VscodeRunMode.ts
AndreasArvidsson: Yes this should probably move to a node package

AndreasArvidsson: Or should be us rename our existing file system package to node? I know @pokey was leaning that way originally.

AndreasArvidsson: done

packages/node-common/package.json
```diff
@@ -1,15 +1,15 @@
 {
-  "name": "@cursorless/file-system-common",
+  "name": "@cursorless/node-common",
```
pokey, line 2: I am tempted to rename this package. Our convention has been that things with `-common` don't import `cursorless-engine`, so that they can be imported in any context, eg test harness. I don't feel strongly though and I can't come up with a better name

pokey (approved):
Looks good! Left a minor comment and pushed some changes

packages/node-common/package.json
AndreasArvidsson: I don't have a better na?me either. We can always rename it in the future if we come up with a better one?

PR #2500, opened 2024-07-12
Move node dependencies out of commons and engine

AndreasArvidsson:
This is all about getting node out of commons. A lot of our file system interaction had to do with running tests and reading/writing test fixtures so I created a separate node test package. 

To avoid circular dependencies I had to move a bunch of types from the engine to commons. This make it so we now have changes in imports in a bunch of different files. Most of the changes are quite mechanical or boilerplate. The only actual new code is the two storage providers to `ScopeTestRecorder` and `TestCaseRecorder`. The latter one is quite lacking in implementation. This is just about unblocking cursorless everywhere. A better abstraction is something we can workshop together.

And before you ask: no I can't split this into smaller pr. It feels like a small miracle that I could solve all of the circular dependency problems I had already  :)

I wonder if we in a follow up want to move everything test related to the test package? Generally I prefer to have test as close to the source code as possible but in this case it's getting problematic to keeping track of all the dependencies. If we had a test package we could make sure that a lot of the dependencies only existed in that package and we could add lint rules preventing them in the other packages.

## Checklist


packages/common/src/ide/normalized/NormalizedIDE.ts
```diff
@@ -46,7 +46,7 @@ export class NormalizedIDE extends PassthroughIDEBase {
       hatStability: this.configuration.getOwnConfiguration(
         "experimental.hatStability",
       ),
-      snippetsDir: getFixturePath("cursorless-snippets"),
+      snippetsDir: this.cursorlessSnippetsDir,
```
AndreasArvidsson, line 49: This implementation I'm not quite happy with. The problem is that normalized ide is used at runtime and uses a environment fixture path. I will happily take any suggestions here.

packages/common/src/ide/types/RecordedTestStorage.ts
```diff
@@ -0,0 +1,11 @@
+import type { RecordTestCaseCommandOptions } from "../../testUtil/RecordTestCaseCommandOptions";
+
+export interface TestCaseRecorderStorage {
```
AndreasArvidsson, line 3: This is the one provider I'm not quite happy with. It's to close to the file system and had don't abstract enough. The problem is that its a lot of path logic in the test case recorder and we basically would be moving all of that into a environment specific implementation. We need to have a discussion about this.

pokey: I agree this is awkward. I thought we decided test case recorder doesn't make sense without file system so we might as well just put it in its own package and let it use the file system as it wants

We had actually discussed at some point that test case recorder might someday be used independent of cursorless, so moving it out of the engine seems reasonable

AndreasArvidsson: Yeah that is probably where I want to go in the long run. This pull request just got so large that I'm trying to keep it down, but might be worth doing here anyhow?

AndreasArvidsson: Ok tests case recorder is now moved into a separate package :)

packages/common/package.json
```diff
@@ -17,21 +17,16 @@
   "license": "MIT",
   "dependencies": {
     "lodash-es": "^4.17.21",
-    "pathe": "^1.1.2",
```
AndreasArvidsson, line 20: All path dependencies have moved out of commons and the engines so we don't need pathe anymore

packages/common/src/ide/types/ScopeTestRecorderStorage.ts
pokey, line 1: This interface is also a bit awkward, and also will be overly restrictive. Eg for #2390, we'll need to be able to iterate recorded tests, and potentially remove them.

I'd be tempted to move ScopeTestRecorder into its own package, or into the `test-case-recorder` package. It is test code anyway; shouldn't be shipped with the engine imo

packages/node-common/package.json
```diff
@@ -25,9 +25,14 @@
...
-    "pathe": "^1.1.2"
+    "lodash-es": "^4.17.21"
   },
-  "devDependencies": {}
+  "devDependencies": {
+    "@types/chai": "^4.3.14",
+    "@types/js-yaml": "^4.0.9",
+    "@types/lodash-es": "4.17.0",
+    "chai": "^5.1.0",
+    "js-yaml": "^4.1.0"
```
pokey, line 36: how come these are dev deps?

AndreasArvidsson: I just mirrored what is on main. chai is by definition a developer tool and not really in production, but how you use these two different dependence lists is a bit debatable
https://github.com/cursorless-dev/cursorless/blob/4282c4bb13a3a3f0953f9a91550e80dbb43cf83e/packages/common/package.json#L29

packages/common/src/ide/types/ScopeTestRecorderStorage.ts
AndreasArvidsson: We can move it into the test case package

packages/node-common/package.json
pokey: i'm ok with it for now; we're not particularly principled with our dev / prod deps at this point anyway, so we can revisit in the future

AndreasArvidsson: Sounds reasonable

packages/common/src/ide/types/ScopeTestRecorderStorage.ts
AndreasArvidsson: Moved

packages/cursorless-engine/src/languages/LanguageDefinition.ts
```diff
@@ -149,15 +148,13 @@ async function readQueryFileAndImports(
...
 
           if (doValidation) {
-            validateImportSyntax(ide, queryName, relativeImportPath, match[0]);
+            validateImportSyntax(ide, queryName, importName, match[0]);
           }
 
-          const importQueryPath = join(dirname(queryName), relativeImportPath);
-          rawQueryStrings[importQueryPath] =
-            rawQueryStrings[importQueryPath] ?? null;
+          rawQueryStrings[importName] = rawQueryStrings[importName] ?? null;
```
pokey, line 157: maybe i'm missing something, but haven't you broken relative imports? eg if query is in a subdir, and includes a `..`, that will no longer work. if it's going to be too annoying to support this, i'm not opposed to forcing a flat structure, but might be a bit more future-proof to support subdirs. If we are going to force flat structure, we should prob enforce it tho rather than silently trying to read the wrong file if there is a subdir

AndreasArvidsson: No relative but folder agnostic is my idea. Just glob all scm files and include by name even if they are in separate folders.

pokey: that seems a bit unintuitive. also a bit awkward if there is a subdir with several things and everything needs to specify path from root. I don't feel super strongly but if we do it we should prob explicitly catch and throw if we see `../`

alternately, we could expose a `getCanonicalName(base, relative)` function on the storage provider?

packages/cursorless-engine/src/index.ts
pokey, line 1: I wouldn't say I'm in love with the engine exposing more of its guts, but I'm fine with doing a refactor later

packages/cursorless-engine/src/languages/LanguageDefinition.ts
AndreasArvidsson: Not path from root, just name
`import javascript` should work from a parent dir, a subdir etc.
Kinda how npm does it with node modules.

packages/cursorless-engine/src/index.ts
AndreasArvidsson: Same here

packages/common/src/ide/normalized/NormalizedIDE.ts
pokey: I mean `NormalizedIDE` is really test / dev code, so probalby shouldn't be in engine / common. But I'm ok with this for now

packages/cursorless-engine/src/languages/LanguageDefinition.ts
pokey: but how do you import something in a subdir?

AndreasArvidsson: No slashes. Just `import foo` it's the providers job to find a query definition with that name. Doesn't matter if you glob files or if they are in an js map. Each definitions name just need to be unique regardless of folder structure.

pokey: Huh. That would be a bit unexpected. At that point the folders aren't really doing much

pokey: shall we just force flat for now? i'm not sure it's worth the effort to do anything fancier

AndreasArvidsson: They are just grouping like all js/ts files in the same folder, but folders have no syntactic meaning.
That was my idea. Or just a flat structure if you prepare.

pokey: if we do force flat tho would be good to ensure that the node layer is not trying to treat it as a relative path

AndreasArvidsson: Sounds good

pokey: I think just an assert here that path.name is equal to path https://github.com/cursorless-dev/cursorless/blob/main/packages/node-common/src/FileSystemLanguageDefinitionsProvider.ts#L41

AndreasArvidsson: fixed

pokey (approved):
Made a couple tweaks. Seemed better to check for relative in the engine for consistency. Merge if you're happy with my changes https://github.com/cursorless-dev/cursorless/pull/2500/files/cd71783e6aa8f73e026a7e46e6b2d78e8c867cc5..73a2f35febe5af8362c2f982070d1f28f805acd8

AndreasArvidsson:
Looks good. Thanks for reviewing this many changes so quickly :)

PR #2501, opened 2024-07-13
Remove node process from hat token map

AndreasArvidsson:
Since we're using a threshold of sixty seconds I don't think we need nanosecond precision. Instead of finding a polyfill for bigint I just switch to date.now which gives us the time in milliseconds.

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/core/HatTokenMapImpl.ts
+++ b/packages/cursorless-engine/src/core/HatTokenMapImpl.ts
@@ -5,9 +5,7 @@ import {
   ReadOnlyHatMap,
   TokenHat,
 } from "@cursorless/common";
-import { hrtime } from "process";
 import { ide } from "../singletons/ide.singleton";
-import { abs } from "../util/bigint";
 import { Debug } from "./Debug";
 import { HatAllocator } from "./HatAllocator";
 import { IndividualHatMap } from "./IndividualHatMap";
@@ -16,7 +14,7 @@ import { RangeUpdater } from "./updateSelections/RangeUpdater";
 /**
  * Maximum age for the pre-phrase snapshot before we consider it to be stale
  */
-const PRE_PHRASE_SNAPSHOT_MAX_AGE_NS = BigInt(6e10); // 60 seconds
+const PRE_PHRASE_SNAPSHOT_MAX_AGE_MS = 60000; // 60 seconds
 
 /**
  * Maps from (hatStyle, character) pairs to tokens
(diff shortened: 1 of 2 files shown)


packages/cursorless-engine/src/core/HatTokenMapImpl.ts
```diff
@@ -106,8 +104,8 @@ export class HatTokenMapImpl implements HatTokenMap {
       }
 
       if (
-        abs(hrtime.bigint() - this.prePhraseMapsSnapshotTimestamp!) >
-        PRE_PHRASE_SNAPSHOT_MAX_AGE_NS
+        Math.abs(Date.now() - this.prePhraseMapsSnapshotTimestamp!) >
```
AndreasArvidsson, line 107: Do we actually need abs here? The time now can never be before the timestamp we saved earlier?

pokey (commented):
`Date` should not be used for timestamps. It is subject to clock skew, daylight savings, etc etc. There's a reason we're using hrtime here

AndreasArvidsson:
> `Date` should not be used for timestamps. It is subject to clock skew, daylight savings, etc etc. There's a reason we're using hrtime here

I know, but for this use case here where we have a sixty second threshold is clock skew really a problem? Same with daylight saving it's running on the same machine so the only reason why it can be problematic is when it changes. If we update the hat map one time to often or to few whenever daylight saving happens is that really a problem? Otherwise we need to find a library for this because we can't use node ofcourse.

pokey:
> Also, Date.now() may have been impacted by system and user clock adjustments, clock skew, etc. as it is relative to the Unix epoch (1970-01-01T00:00:00Z) and dependent on the system clock. The performance.now() method on the other hand is relative to the timeOrigin property which is a [monotonic clock](https://w3c.github.io/hr-time/#dfn-monotonic-clock): its current time never decreases and isn't subject to adjustments.

- https://developer.mozilla.org/en-US/docs/Web/API/Performance/now

How about a9de4414dc26c52a91efb0a61e7b073036c57667?

AndreasArvidsson:
> > Also, Date.now() may have been impacted by system and user clock adjustments, clock skew, etc. as it is relative to the Unix epoch (1970-01-01T00:00:00Z) and dependent on the system clock. The performance.now() method on the other hand is relative to the timeOrigin property which is a [monotonic clock](https://w3c.github.io/hr-time/#dfn-monotonic-clock): its current time never decreases and isn't subject to adjustments.
> 
>     * https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
> 
> 
> How about [a9de441](https://github.com/cursorless-dev/cursorless/commit/a9de4414dc26c52a91efb0a61e7b073036c57667)?

Performance is also a node lib. I tried to compile without node and it didn't work.

pokey:
cool yeah let's leave it for now. it is part of the mdn spec and supported on pretty much every browser as well as node since 8.5, so I think it's pretty safe. If it turns out talon or some other js context doesn't support it, we can always inject it by exposing a `getPerformanceCounter` or `getMonotonicTime` on `ide`

pokey (approved):

PR #2502, opened 2024-07-13
Remove node from runIntegrationTests

AndreasArvidsson:
Since `runIntegrationTests` is constructed in the engine it cannot use assert. Switching assert to throwing an error I think is totally fine here. It will still fail the test and we don't have that many legacy scopes left and hopefully soon they will be gone as well.

## Checklist


pokey (approved):
We should really revisit to see if we now have a better way to get this test code out of engine. But this is fine for now

AndreasArvidsson:
> We should really revisit to see if we now have a better way to get this test code out of engine. But this is fine for now

Agreed

PR #2503, opened 2024-07-13
Cursorless everywhere in Talon

AndreasArvidsson:
## Oddities and todos
- [ ] Talon requires a javascript file to have a Talon(or another package?) import before any other code. Right now I do that in my [`esbuild.sh`](https://github.com/cursorless-dev/cursorless/blob/100f7706553fd24ce3ccbec2d1ec30c20e7f0a3f/packages/cursorless-talon-js/esbuild.sh#L20-L22) script. This does not support esbuild watch.
    - The solution here is for Talon to support mjs file endings
- [x] Restore num selections check for windows: [`cursorless_everywhere_talon_win.py`](https://github.com/cursorless-dev/cursorless/blob/824735f58e58ea3ca5bfb2448df801a4e7f28b1c/cursorless-everywhere-talon/cursorless_everywhere_talon_win.py#L49-L50) once Talon converts from js to py correctly
- [x] Is there any point in outputting source maps to Talon?
- [x] If easy, let's run in CI on all platforms

## Information
* https://bellard.org/quickjs/

## Checklist


pokey:
for bigint issue, why not just move `timeUtils` and `takeSnapshot` to new test case recorder package?

AndreasArvidsson:
> for bigint issue, why not just move `timeUtils` and `takeSnapshot` to new test case recorder package?

That is definitely a possibility I've been thinking about. If you agreed I would probably do that.

pokey:
Yeah it's only used by cursorless-vscode and test-case-recorder and logically it feels like part of test case recorder anyway

AndreasArvidsson:
@pokey What should we do about `sbd`? My inclination is to build our own implementation. sbd hasn't had an update in over three years and there are known bugs in it. I'm pretty sure we could do better :D

AndreasArvidsson:
Glad to report that we only have three dependencies that or still commonjs
moo
nearley
immutability-helper

pokey:
> @pokey What should we do about `sbd`? My inclination is to build our own implementation. sbd hasn't had an update in over three years and there are known bugs in it. I'm pretty sure we could do better :D

I'm def hesitant to maintain our own sentence boundary detector. Were you not able to get that browser condition to fire? Scroll back in our slack DMs if you're not sure what I mean

pokey:
> Glad to report that we only have three dependencies that are still commonjs
> 
> moo
> 
> nearley
> 
> immutability-helper

Nice. Is that a blocker for this PR?

AndreasArvidsson:
> > @pokey What should we do about `sbd`? My inclination is to build our own implementation. sbd hasn't had an update in over three years and there are known bugs in it. I'm pretty sure we could do better :D
> 
> I'm def hesitant to maintain our own sentence boundary detector. Were you not able to get that browser condition to fire? Scroll back in our slack DMs if you're not sure what I mean

I know what you mean and I have absolutely no idea how to get that to fire. I can't even find any good no documentation what that field is supposed to actually do. Also we are not in the browser we don't have access to window and so on.

AndreasArvidsson:
> > Glad to report that we only have three dependencies that are still commonjs
> > moo
> > nearley
> > immutability-helper
> 
> Nice. Is that a blocker for this PR?

Not at all. Just nice that we are now 98% esm.

packages/cursorless-talon-js/src/ide/TalonJsTextDocument.ts
pokey, line 1: move this and `performEdits` into a new PR

packages/cursorless-talon-js/src/ide/performEdits.ts
```diff
@@ -0,0 +1,90 @@
...
+  for (const change of changes) {
+    const { rangeOffset, rangeLength, text } = change;
+
+    result =
+      result.slice(0, rangeOffset) +
+      text +
+      result.slice(rangeOffset + rangeLength);
+  }
+
+  actions.user.cursorless_js_set_text(result);
```
pokey, line 66: move this and `ide` line below into wrapper function that is talon-specific

pokey:
- [ ] add https://github.com/AndreasArvidsson/cursorless-talon/blob/main/src/cursorless_talon_js.py to a `cursorless_everywhere_talon` at root of repo

pokey:
- [ ] Use quickjs cli to run actual production bundled payload. We would use `--include` to create the talon mocks, and then also import a function that runs a few commands manually and checks the results

.vscode/settings.json
```diff
@@ -1,7 +1,7 @@
 // Place your settings in this file to overwrite default and user settings.
 {
   "[python]": {
-    "editor.defaultFormatter": "charliermarsh.ruff"
+    "editor.defaultFormatter": "ms-python.black-formatter"
```
pokey, line 4: why? we use ruff in CI

AndreasArvidsson: Not on purpose

AndreasArvidsson: I don't have ruff so this setting completely destroys my local development

pokey: Are you hoping to avoid installing it? I would argue for using the same thing we use in CI

AndreasArvidsson: I have never even looked into what ruff or how to install it. I've been busy with other stuff :p

AndreasArvidsson:
@pokey can you run this on mac and see if the build script works. I have no idea what it's actually complaining about here.

`packages/cursorless-everywhere-talon build: sed: 1: "out/talon.js": invalid command code o`
https://github.com/cursorless-dev/cursorless/actions/runs/9973679351/job/27559540925?pr=2503#step:8:159

pokey:
Oh yeah sed on Mac is weird. Prob something like this https://stackoverflow.com/a/19457213

AndreasArvidsson:
If I follow what's said there it won't work on my machine any more. We probably need to workshop this in session so it works on both.

packages/cursorless-everywhere-talon/src/test/quickjsTest.ts
```diff
@@ -0,0 +1,93 @@
+import { getTestHelpers } from "talon";
```
pokey, line 1: why do you need these? how come you can't just directly call the talon actions from here?

cursorless-everywhere-talon/cursorless_everywhere_talon.py
pokey, line 1: there's still one more file necessary to actually switch cursorless on globally right? we should prob add that to this PR

packages/cursorless-everywhere-talon/src/test/quickjsTest.ts
AndreasArvidsson: I don't want all of this exposed to cursorless. A lot of this stuff should only be in the test runner.

cursorless-everywhere-talon/cursorless_everywhere_talon.py
AndreasArvidsson: Yes but I don't want that yet. Enabling Cursorless globally with disabled my non Cursorless commands I'm still relying on and I don't want to do that's which before this implementation is more filled out.

pokey: How are you using this then?

packages/cursorless-everywhere-talon/src/test/quickjsTest.ts
```diff
@@ -0,0 +1,93 @@
...
+function runAction(action: ActionDescriptor) {
+  const command: CommandLatest = {
+    version: 7,
+    usePrePhraseSnapshot: false,
+    action,
+  };
+  return getTestHelpers().contextActions.private_cursorless_run_rpc_command_get(
+    "cursorless.command",
+    command,
+  );
```
pokey, line 72: Here, for example, I'm not following why you can't just import actions from talon and run the action that way

AndreasArvidsson: That is not on the actions type by design. Only the test runner should ever call it.

cursorless-everywhere-talon/cursorless_everywhere_talon.py
AndreasArvidsson: Separate talon file with new commands for testing.

AndreasArvidsson:
@pokey We're now running tests in ci using the production bundle :)

packages/cursorless-everywhere-talon/src/polyfill.ts
```diff
@@ -0,0 +1,16 @@
...
+global.console = {
+  log: print,
+  error: print,
+  warn: print,
+  debug: print,
+};
+
+global.setTimeout = (callback: () => void, _delay: number) => {
+  callback();
+};
```
pokey, line 16: try throwing an error

AndreasArvidsson: No go
```
Running quickjs tests
    * testTake
Error: setTimeout is not implemented
    x testChuck
Error: setTimeout is not implemented
```

packages/cursorless-everywhere-talon/src/registerCommands.ts
```diff
@@ -0,0 +1,59 @@
...
+import { Context, actions } from "talon";
+import type { TalonJsIDE } from "./ide/TalonJsIDE";
+
+const ctx = new Context();
+
+let ide: TalonJsIDE | undefined;
+let commandApi: CommandApi | undefined;
+
+ctx.matches = `
+not app: vscode
```
AndreasArvidsson, line 12: @pokey What do we do with this matcher? For this to be valid with other editors we probably need a Cursorless everywhere tag instead?

pokey:
re source maps: let's keep them for debugging recorded tests, but not include when we create our staging dir

.vscode/settings.json
phillco: It's great, imo! Give it a try. Much faster replacement for the `flake8` linter that now also does good formatting (replacing black)

AndreasArvidsson: I'm running it since a few days back :)

AndreasArvidsson:
@pokey All tests are now passing and we are running them on as many platforms as we can.

AndreasArvidsson:
Updated according to our discussion in the meetup.

packages/test-harness/src/scripts/runTalonJsTestsCI.ts
pokey, line 1: Would this one not work locally as well? If so then maybe we should just use this one for both use cases

packages/cursorless-everywhere-talon/scripts/esbuild.sh
```diff
@@ -0,0 +1,17 @@
...
+  --platform=neutral \
+  --format=esm \
+  --main-fields=main,module \
+  --conditions=cursorless:bundler \
+  --bundle \
+  --sourcemap \
+  --external:talon \
+  "$@"
+
+# Talon javascript files needs to start with an import from Talon before any other code
```
pokey, line 15: Getting the impression this won't be fixed before we ship? Shall we file an issue / add a FIXME?

packages/cursorless-everywhere-talon/scripts/populate-dist.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+mkdir dist
+
+cp -r out/. dist
+
+cp -r ../../cursorless-everywhere-talon/. dist
```
pokey, line 7: would be easier to read if you found a way to use our `CURSORLESS_REPOSITORY_ROOT` env var. If awkward to get access to that, maybe just define it at the top so it's clearer what's happening here. For a while I thought `cursorless-everywhere-talon` was referring the the parent directory of this script

packages/cursorless-everywhere-talon/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "description": "cursorless in talon js",
+  "main": "./out/index.js",
+  "license": "MIT",
+  "type": "module",
+  "private": true,
+  "scripts": {
+    "esbuild:dev": "pnpm run esbuild:base src/mainDevelopment.ts",
+    "esbuild:prod": "pnpm run esbuild:base --minify src/mainProduction.ts",
+    "esbuild:watch": "pnpm run esbuild:base --watch src/mainDevelopment.ts",
+    "esbuild:meta": "pnpm run esbuild:base --metafile=meta.json src/mainDevelopment.ts",
```
pokey, line 13: should we call this `esbuild:profile`? At first read it's tough to know what `meta` is supposed to mean here. I know you're just borrowing esbuild terminology but I think profiling is actually what you use this for right?

packages/cursorless-everywhere-talon-core/src/polyfill.ts
```diff
@@ -0,0 +1,20 @@
...
+    log: print,
+    error: print,
+    warn: print,
+    debug: print,
+  };
+}
+
+global.setTimeout = (callback: () => void, _delay: number) => {
+  callback();
+};
```
pokey, line 20: As discussed, I'm not in love with silently ignoring the delay. Do we actually need this polyfill?

packages/cursorless-everywhere-talon-core/src/polyfill.ts
```diff
@@ -0,0 +1,20 @@
+const global = globalThis as any;
+
+if (global.process == null) {
```
pokey, line 3: prob worth a comment about why we need this?

packages/cursorless-everywhere-talon-core/src/polyfill.ts
```diff
@@ -0,0 +1,20 @@
...
+  global.process = {
+    env: {},
+  };
+}
+
+if (typeof print !== "undefined") {
```
pokey, line 9: as above, could use comment

packages/cursorless-everywhere-talon-core/src/registerCommands.ts
```diff
@@ -0,0 +1,49 @@
...
+import type { TalonJsIDE } from "./ide/TalonJsIDE";
+import type { Talon } from "./types/talon.types";
+
+export function registerCommands(
+  talon: Talon,
+  ide: TalonJsIDE,
+  commandApi: CommandApi,
+) {
+  const ctx = talon.Context();
+  ctx.matches = "not app: vscode";
```
pokey, line 12: We should do something more generic. For example this will now break cursorless neovim, which is already shipped

packages/cursorless-everywhere-talon-core/src/registerCommands.ts
```diff
@@ -0,0 +1,49 @@
...
+    command: unknown,
+  ): Promise<unknown> {
+    try {
+      if (commandId !== CURSORLESS_COMMAND_ID) {
+        throw Error(`Unknown command ID: ${commandId}`);
+      }
+
+      const editorState =
+        talon.actions.user.cursorless_everywhere_get_editor_state();
+      ide.updateTextEditors(editorState);
```
pokey, line 41: prob worth a comment about why we need to do this. Ideally we wouldn't have to use this hack, but it is the right place to start

packages/cursorless-everywhere-talon-core/src/ide/TalonJsMessages.ts
pokey, line 1: kinda cool we can directly show notifications from cursorless js using talon app.notify

packages/cursorless-everywhere-talon-core/src/ide/TalonJsTestHats.ts
```diff
@@ -0,0 +1,39 @@
...
+const HAT_COLORS = [
+  "default",
+  "blue",
+  "green",
+  "red",
+  "pink",
+  "yellow",
+] as const;
+
+export class TalonJsTestHats implements Hats {
```
pokey, line 18: docstring

packages/cursorless-everywhere-talon-core/src/ide/setSelections.ts
```diff
@@ -0,0 +1,17 @@
...
+  selections: Selection[],
+): Promise<void> {
+  const selectionOffsets = selections.map((selection) => ({
+    anchor: document.offsetAt(selection.anchor),
+    active: document.offsetAt(selection.active),
+  }));
+
+  talon.actions.user.cursorless_everywhere_set_selections(selectionOffsets);
+
+  return Promise.resolve();
```
pokey, line 16: i've seen you doing this in a couple places in this PR. Is this really better than just adding an `async` keyword to the function?

packages/cursorless-everywhere-talon-core/src/types/Global.d.ts
```diff
@@ -0,0 +1,8 @@
+declare function print(...objs: any): void;
+
+declare namespace console {
```
pokey, line 3: why do you have to do this? worth a comment

packages/cursorless-everywhere-talon-core/src/types/Global.d.ts
```diff
@@ -0,0 +1,8 @@
+declare function print(...objs: any): void;
```
pokey, line 1: what happens when you run in node during testing?

packages/cursorless-everywhere-talon-core/src/types/std.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module "std" {
```
pokey, line 1: why do you need this?

packages/cursorless-everywhere-talon-e2e/scripts/test-quickjs.sh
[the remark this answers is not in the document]
pokey: whoops looks like the above comment was waiting for review rather than immediate. But seems you did it anyway

packages/cursorless-everywhere-talon-e2e/src/types/cursorless.ts
pokey, line 1: is this stuff used anymore? Can't seem to find any references to it

packages/cursorless-everywhere-talon-e2e/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "test:quickjs": "bash scripts/test-quickjs.sh",
+    "clean": "rm -rf ./out tsconfig.tsbuildinfo ./dist ./build"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "MIT",
+  "type": "module",
+  "dependencies": {
+    "@cursorless/common": "workspace:*",
+    "@cursorless/cursorless-engine": "workspace:*",
```
pokey, line 18: we generally don't import the engine from e2e. Can this be avoided?

packages/cursorless-everywhere-talon-e2e/src/quickjsTest.ts
```diff
@@ -0,0 +1,102 @@
+import type { ActionDescriptor, CommandLatest } from "@cursorless/common";
+import { activate } from "@cursorless/cursorless-everywhere-talon-core";
+import * as std from "std";
+import talonMock from "./talonMock";
+
+let hasFailed = false;
+
+async function runTests() {
+  await activate(talonMock, "test");
```
pokey, line 9: I would prefer if we tested the real js and mocking talon imports rather than directly calling activate here, but I don't have time to think about whether there's a better way so I'll go with your instincts here

packages/cursorless-everywhere-talon-core/package.json
pokey, line 1: a bit awkward to have this core package that is imported both by real talon-everywhere and by tests, but I guess that made the machinery simpler

pokey (approved):
Left a bunch of comments; feel free to merge once addressed

.github/workflows/test.yml
```diff
@@ -43,6 +43,14 @@ jobs:
         if: runner.os == 'Linux'
       - run: pnpm --color test
         if: runner.os != 'Linux'
+      - run: xvfb-run -a pnpm -F @cursorless/test-harness test:talonJs
+        if: runner.os == 'Linux' && matrix.vscode_version == 'stable'
```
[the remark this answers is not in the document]
AndreasArvidsson: Get some kind of error. I guess the same reason why we're using that for rounding other scripts on Linux?

packages/cursorless-everywhere-talon-e2e/src/types/cursorless.ts
AndreasArvidsson: I use `canonicalizeAndValidateCommand` for the code that decides which test should be excluded, but I can work around that.

cursorless-everywhere-talon/cursorless_everywhere_talon.py
```diff
@@ -0,0 +1,43 @@
...
+
+class EditorChange(TypedDict):
+    text: str
+    rangeOffset: int
+    rangeLength: int
+
+
+class EditorChanges(TypedDict):
+    text: str
+    changes: list[EditorChange]
```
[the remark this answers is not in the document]
AndreasArvidsson: This gets passed to python where we need to do that on a javascript object. I rather not make things more complicated on that side.

cursorless-everywhere-talon/cursorless_everywhere_talon.py
```diff
@@ -0,0 +1,43 @@
...
+class Actions:
+    def cursorless_everywhere_get_editor_state() -> EditorState:  # pyright: ignore [reportReturnType]
+        """Get the focused editor element state"""
+
+    def cursorless_everywhere_set_selections(
+        selections: list[OffsetSelection],  # pyright: ignore [reportGeneralTypeIssues]
+    ):
+        """Set focused element selections"""
+
+    def cursorless_everywhere_set_text(
```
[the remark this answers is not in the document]
AndreasArvidsson: Should we call it update text instead?

cursorless-everywhere-talon/cursorless_everywhere_talon_win.py
```diff
@@ -0,0 +1,123 @@
...
+        selections: list[OffsetSelection],  # pyright: ignore [reportGeneralTypeIssues]
+    ):
+        if selections.length != 1:  # pyright: ignore [reportAttributeAccessIssue]
+            raise ValueError("Only single selection supported")
+
+        selection = selections[0]
+        anchor = selection["anchor"]
+        active = selection["active"]
+
+        print(f"Setting selection to {anchor}, {active}")
```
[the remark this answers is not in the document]
AndreasArvidsson: They are quite useful for testing right now. I will add a fix me.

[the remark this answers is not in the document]
AndreasArvidsson: I remove them. We can always readd them if we want to debug some more. For now it seems quite reliable.

cursorless-everywhere-talon/cursorless_everywhere_talon_win.py
```diff
@@ -0,0 +1,123 @@
...
+    range.move_endpoint_by_range("End", "Start", target=document_range)
+    range.move_endpoint_by_unit("End", "Character", end)
+    range.move_endpoint_by_unit("Start", "Character", start)
+    range.select()
+
+
+def get_selection(
+    document_range: TextRange, selection_range: TextRange, caret_range: TextRange
+) -> tuple[int, int]:
+    range_before_selection = document_range.clone()
```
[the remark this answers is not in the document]
AndreasArvidsson: Definitely :D

packages/cursorless-engine/src/util/selectionUtils.ts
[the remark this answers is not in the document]
AndreasArvidsson: I think so, but it should probably be done in a follow up. It will touch a lot of files and even some types that the utilities are using needs to move to common as well.

packages/test-harness/package.json
```diff
@@ -7,12 +7,15 @@
...
+    "build:base": "esbuild --sourcemap --conditions=cursorless:bundler --bundle --external:vscode --external:./reporters/parallel-buffered --external:./worker.js --external:talon --format=cjs --platform=node",
+    "build": "pnpm run build:runner:vscode && pnpm run build:runner:neovim && pnpm run build:tests && pnpm run build:unit && pnpm run build:talon && pnpm run build:talonJs && pnpm run build:talonJsCI",
     "build:runner:vscode": "pnpm run build:base ./src/runners/extensionTestsVscode.ts --outfile=dist/extensionTestsVscode.cjs",
     "build:runner:neovim": "pnpm run build:base ./src/runners/extensionTestsNeovim.ts --outfile=out/runners/extensionTestsNeovim.cjs",
     "build:unit": "pnpm run build:base ./src/scripts/runUnitTestsOnly.ts --outfile=dist/runUnitTestsOnly.cjs",
     "build:talon": "pnpm run build:base ./src/scripts/runTalonTests.ts --outfile=dist/runTalonTests.cjs",
+    "build:talonJs": "pnpm run build:base ./src/scripts/runTalonJsTests.ts --outfile=dist/runTalonJsTests.cjs",
+    "build:talonJsCI": "pnpm run build:base ./src/scripts/runTalonJsTestsCI.ts --outfile=dist/runTalonJsTestsCI.cjs",
```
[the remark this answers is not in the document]
AndreasArvidsson: Both are our full node test suite. The only difference is that the one in CI exits with error code 1 so we actually get an error in our GitHub workflow.

packages/test-harness/src/scripts/runTalonJsTestsCI.ts
AndreasArvidsson: It would, but its exiting with an error code which we're not doing in any other of our node tests. I just want to be consistent with vscode.

packages/cursorless-everywhere-talon/scripts/populate-dist.sh
AndreasArvidsson: I do agree it's a bit confusing. I'm not sure the best way of doing that in bash. Can I just leave a comment instead?

packages/cursorless-everywhere-talon/package.json
AndreasArvidsson: Sounds good

packages/cursorless-everywhere-talon-core/src/polyfill.ts
AndreasArvidsson: Yes. This gets called by our hat map debouncer.

packages/cursorless-everywhere-talon-core/src/registerCommands.ts
AndreasArvidsson: We could eider keep adding applications here or we need a new tag.

packages/cursorless-everywhere-talon-core/src/ide/TalonJsMessages.ts
AndreasArvidsson: Not bad ey? :D

packages/cursorless-everywhere-talon-core/src/ide/setSelections.ts
AndreasArvidsson: I don't know? Generally the linter will be angry if you have an async function without an await in it so I kinda Learned not to use that, but we might not be using that rule in Cursorless so here it's probably okay.

packages/cursorless-everywhere-talon-core/src/types/Global.d.ts
AndreasArvidsson: We don't call print. Print only exits so our polyfill is allowed to shunt console log to it.

packages/cursorless-everywhere-talon-core/src/types/std.d.ts
AndreasArvidsson: We don't anymore. We did in the implementation that read the Cursorless mode environment variable. I'm removing it.

packages/cursorless-everywhere-talon-e2e/src/quickjsTest.ts
AndreasArvidsson: I have no idea how I would be able to do that in node. This was the only solution I could come up with.

packages/cursorless-everywhere-talon-core/package.json
AndreasArvidsson: Definitely. You can't import from a private package of course so the test could not import from the actual application and we didn't want all the tests in the application so this was the only solution.

cursorless-everywhere-talon/cursorless_everywhere_talon.py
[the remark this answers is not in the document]
pokey: Not sure I follow, but it's still a confusing interface, so it would be good to find a way to make things clearer

cursorless-everywhere-talon/cursorless_everywhere_talon.py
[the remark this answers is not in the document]
pokey: Edit or edit text?

packages/cursorless-engine/src/util/selectionUtils.ts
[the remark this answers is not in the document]
pokey: Would prefer a pre-PR to a follow up but I won't push strongly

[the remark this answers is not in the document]
AndreasArvidsson: It would probably change 20+ files more, but if you're okay with that I can do it.

cursorless-everywhere-talon/cursorless_everywhere_talon.py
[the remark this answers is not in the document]
AndreasArvidsson: Edit text sounds fine

packages/test-harness/package.json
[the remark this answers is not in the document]
pokey: I'm not talking about the line above; I mean the test script above that also refers to the same CI entry point

packages/test-harness/src/scripts/runTalonJsTestsCI.ts
pokey: If the tests fail I think nonzero exit code is the right thing to do

packages/cursorless-everywhere-talon/scripts/populate-dist.sh
pokey: Just define a variable of that name to point to ../.. if it's hard to expose the proper one

cursorless-everywhere-talon/cursorless_everywhere_talon.py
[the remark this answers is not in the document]
AndreasArvidsson: That whole discussion we had about objects passed from javascript to python does not get converted so we get a javascript object in python. I don't want to try to do type narrowing on the javascript object in python. Don't even know if we can do that properly.

packages/cursorless-everywhere-talon-core/src/polyfill.ts
pokey: But if we pass null hats we never construct hat map right?

packages/cursorless-everywhere-talon-core/src/registerCommands.ts
pokey: I don't feel strongly, but probably better to have something where we don't need to remember to update this every time we add a new application

packages/test-harness/package.json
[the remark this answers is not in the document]
AndreasArvidsson: You mean `runTalonTests.cjs`? I believe that is for the spoken grammar.

packages/cursorless-everywhere-talon-core/src/ide/setSelections.ts
pokey: Yeah I've never seen that error

packages/cursorless-everywhere-talon/scripts/populate-dist.sh
AndreasArvidsson: Sounds good

packages/cursorless-everywhere-talon-core/src/types/Global.d.ts
pokey: But the poly fill is defined in a different package right?

packages/cursorless-everywhere-talon-e2e/src/quickjsTest.ts
pokey: For node I don't care. I mean for quickjs

packages/test-harness/src/scripts/runTalonJsTestsCI.ts
AndreasArvidsson: We don't do that for our vscode tests, but we can do that here of course.

packages/cursorless-everywhere-talon-core/package.json
pokey: Hmm. I wonder how it works in vscode, where we don't have to do this

packages/cursorless-everywhere-talon-core/src/ide/setSelections.ts
AndreasArvidsson: I think it is because we haven't enabled the full type checking.

packages/cursorless-everywhere-talon-e2e/src/quickjsTest.ts
AndreasArvidsson: Should we have different entry points and build different bundles for node and for quick js then? I'm trying to keep it as simple as possible.

packages/cursorless-everywhere-talon-core/src/registerCommands.ts
pokey: update from meet-up: have a tag, and then a README in this dir explaining how to activate it for `not vscode`

packages/cursorless-everywhere-talon-core/src/polyfill.ts
pokey: filed https://github.com/cursorless-dev/cursorless/issues/2596; let's add a FIXME as well pointing to it

packages/cursorless-everywhere-talon/scripts/populate-dist.sh
pokey: update from meet-up: delete this file

packages/test-harness/src/scripts/runTalonJsTestsCI.ts
pokey: update from meet-up: if it fails, don't return 0. even if we do the wrong thing in vscode

packages/test-harness/package.json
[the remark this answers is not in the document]
pokey: update: delete this script and should be fine because we're merging the CI and non-CI test script

AndreasArvidsson:
@pokey Updated. We have two ongoing discussions that we need to resolve then we can merge

cursorless-everywhere-talon/cursorless_everywhere_talon.py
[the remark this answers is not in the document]
pokey: Do we even support changes? Shall we just start with a string?

packages/cursorless-engine/src/util/selectionUtils.ts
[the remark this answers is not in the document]
pokey: Just to be clear, I'm not proposing you do it in this PR. But let's leave it for a follow-up; saves you some weird merge conflicts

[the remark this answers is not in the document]
AndreasArvidsson: Solved. There was only a single function in that file actually needed to be exported si moved just that to common.

cursorless-everywhere-talon/cursorless_everywhere_talon.py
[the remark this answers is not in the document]
AndreasArvidsson: The mac implementation does.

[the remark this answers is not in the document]
pokey: Got it. Could we add the mac impl in rough form? Or at least send me a link? A bit hard to understand what interface we want without seeing that

packages/cursorless-everywhere-talon-core/src/README.md
```diff
@@ -0,0 +1,12 @@
+# Cursorless everywhere Talon
+
+## Activate by enabling tag
+
+In the below example we are enabling curse es everywhere Talon if we are not in vscode. Of course update this to fit your Cursorless IDE (vscode, neovim) of choice. Or globally enabled if this is your only Cursorless implementation.
```
pokey, line 5:
```suggestion
In the below example we are enabling Cursorless everywhere if we are not in vscode. Of course update this to fit your Cursorless IDE (vscode, neovim) of choice. Or globally enabled if this is your only Cursorless implementation.
```

cursorless-everywhere-talon/cursorless_everywhere_talon.py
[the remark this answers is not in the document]
AndreasArvidsson: I don't have it. Only @phillco does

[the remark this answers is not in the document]
AndreasArvidsson: btw I have now renamed the field to edit. Edit has text and changes. I think that is clear enough.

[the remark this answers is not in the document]
pokey: added docstrings. Still not in love but at least it makes it clearer what's happening

pokey (approved):
let's ship it!

PR #2511, opened 2024-07-15
Allow whitespace lines when calculating edit indentation

AndreasArvidsson:
When calculating the edit indentation we want to skip empty lines, but whitespace lines we should actually use.

## Checklist

### DIFF
--- a/data/fixtures/recorded/actions/drinkLine2.yml
+++ b/data/fixtures/recorded/actions/drinkLine2.yml
@@ -1,0 +1,37 @@
+languageId: plaintext
+command:
+  version: 7
+  spokenForm: drink line
+  action:
+    name: editNewLineBefore
+    target:
+      type: primitive
+      modifiers:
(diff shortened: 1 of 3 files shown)


pokey (commented):
Huh. I could see a scenario where you have some empty line in the middle of a function with a single space on it screwing things up unexpectedly. I would be tempted to only have this fire for single-line case as that seems to be the problem you're observing

AndreasArvidsson:
> Huh. I could see a scenario where you have some empty line in the middle of a function with a single space on it screwing things up unexpectedly. I would be tempted to only have this fire for single-line case as that seems to be the problem you're observing

Can you elaborate on that example? I'm not against doing single line logic, but I want to understand your use case better.

AndreasArvidsson:
done

pokey (approved):
Tweaked; merge if happy

PR #2512, opened 2024-07-15
Extend ide capabilities

AndreasArvidsson:
When the ide don't have capabilities/own implementation of `indentLine`, `outdentLine` and `insertLineAfter` fallback to doing an editor edit instead. This makes it so that in a non code editor environment like the web browser we still get these three actions implemented.

Ideally we would test this, but that would involve running tests with mocked capabilities. I think that is a lot of work?

## Checklist


AndreasArvidsson:
I agree. The test are not passing, but I think this direction is nice.

AndreasArvidsson:
After another tweak now the tests are passing

pokey:
maybe it makes sense to pull in the in-memory text document and text editor from your talon js PR to use for testing here? You could just directly instantiate this action and run it

pokey:
update from meet-up: split this PR to include only the "pour" piece, as that is simple enough to not really need tests

AndreasArvidsson:
Okay this is now just pour. Here is the pr for indent
https://github.com/cursorless-dev/cursorless/pull/2518

pokey (approved):

PR #2515, opened 2024-07-16
Remove sbd dependency

AndreasArvidsson:
Note that this is still `sbd`. The reason for this PR is to get rid of the html/postcss dependences that breaks non node installations. We didn't even have that activated so for us this doesn't break anything

My intent is to replace the implementation with something better in the future, but for now this is just sbd with the html support removed. For that future implementation it's nice to already have the sbd tests in place.

## Checklist

- [/] I have updated the [docs](https://github.com/cursorless-dev/cursorless/tree/main/docs) and [cheatsheet](https://github.com/cursorless-dev/cursorless/tree/main/cursorless-taln/src/cheatsheet)


pokey (approved):
I'm not in love with vendoring in a rule-based sentence segmenter, because I don't think that's the way we want to do segmentation in the long term, but I agree it's nice to have the tests and the implementation itself is not that much code. And yeah the code actually hasn't been touched in more like 5 years if you look at their commit history; everything else was just a version bump of some sort

See my changes in 1097ce15a1ce1a07b8a5c7b8f2874566808d6d84 and merge if you're happy

PR #2523, opened 2024-07-17
Lint rule to prevent node in common and engine

AndreasArvidsson:
## Checklist

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -85,6 +85,16 @@
           }
         ]
       }
+    },
+    {
+      "files": [
+        "packages/common/**/*.ts",
+        "packages/cursorless-engine/**/*.ts"
+      ],
+      "excludedFiles": ["**/testUtil/**", "**/scripts/**", "*.test.ts"],
+      "rules": {
+        "import/no-nodejs-modules": "error"
+      }
     }
   ],
   "settings": {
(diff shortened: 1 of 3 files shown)


pokey:
I would lean towards putting one of these in each package instead of at root

Also, I'd use exact path to scripts instead of glob

And not in love with the testUtil exclusion. That feels like a pretty big loophole. But I guess the right thing to do there would be to create a new package for our testUtils, which is harder. This is probably ok for a start? We'll catch any actual problems when we try to bundle talon js or vscode web in the future, as esbuild won't be in node mode

AndreasArvidsson:
Yeah I was kinda unsure which direction you preferred. Probably better with separate config files. Updated now. 

I agree about the test utils. It's actually only a single file that is used by a single test. As we talked about before if we move tests out of the engine we need to expose more of the guts of the engine outside of the package and we don't want that either.
https://github.com/cursorless-dev/cursorless/blob/6749779d98e4a3a7b25ff3738c67a56213f51276/packages/cursorless-engine/src/test/spokenForms.talon.test.ts

I'm fine with this for now.

pokey:
Oh that file doesn't really need to live in the engine. It's not like it's close to what it's testing. But otoh I'm not sure where we'd put it

I'm ok with this for now

pokey (approved):

PR #2534, opened 2024-07-19
Replace firstNonWhitespaceCharacterIndex with rangeTrimmed

### LINKED ISSUE #2519: Replace poorly named `firstNonWhitespaceCharacterIndex` and `lastNonWhitespaceCharacterIndex`
`lastNonWhitespaceCharacterIndex` is not at all the character index. It's the character position after the last non whitespace character.
eg `" foo "` => 1, 4

This should either be renamed or actually made into a range. eg `strippedRange: Range`

AndreasArvidsson:
Fixes #2519

## Checklist


packages/common/src/ide/inMemoryTextDocument/test/InMemoryTextDocumentLineAt.test.ts
pokey, line 1: let's add tests for `rangeTrimmed` for the case where

- line is empty
- line is all whitespace

The semantics of what to do in this situation are arbitrary, so let's lock them in with a test

packages/cursorless-engine/src/processTargets/targets/DestinationImpl.ts
```diff
@@ -197,8 +197,12 @@ function getIndentationString(editor: TextEditor, range: Range) {
       continue;
     }
 
-    if (line.firstNonWhitespaceCharacterIndex < length) {
-      length = line.firstNonWhitespaceCharacterIndex;
+    const trimmedPosition = line.isEmptyOrWhitespace
+      ? line.rangeTrimmed.end
+      : line.rangeTrimmed.start;
+
+    if (trimmedPosition.character < length) {
```
pokey, line 204: why has this gotten more awkward? I feel that maybe the semantics of `trimmedRange` are maybe the opposite of what they should be? Looks like when we had the two separate attributes, they kinda formed a backward range when it was empty

packages/cursorless-vscode/src/ide/vscode/VscodeTextLineImpl.ts
```diff
@@ -21,12 +21,16 @@ export default class VscodeTextLineImpl implements TextLine {
...
   }
 
-  get firstNonWhitespaceCharacterIndex(): number {
-    return this.line.firstNonWhitespaceCharacterIndex;
-  }
-
-  get lastNonWhitespaceCharacterIndex(): number {
-    return this.line.text.trimEnd().length;
+  get rangeTrimmed(): Range {
+    return this.line.isEmptyOrWhitespace
```
pokey, line 25: Something feels awkward about all these extra is empty checks. I think maybe we need to think a bit more carefully about this one. Like maybe we should return undefined or something special when it's empty or whitespace

pokey (commented):
not sure bout this one; see comments

packages/cursorless-engine/src/processTargets/targets/DestinationImpl.ts
pokey: update from meet-up: it appears that the new api has exposed existing awkwardness here, and with some thought, we might be able to simplify / clarify this whole loop body

packages/common/src/ide/inMemoryTextDocument/InMemoryTextLine.ts
```diff
@@ -4,11 +4,10 @@ import type { TextLine } from "../../types/TextLine";
 import { getLeadingWhitespace, getTrailingWhitespace } from "../../util/regex";
 
 export class InMemoryTextLine implements TextLine {
+  readonly isEmptyOrWhitespace: boolean;
   readonly range: Range;
   readonly rangeIncludingLineBreak: Range;
-  readonly firstNonWhitespaceCharacterIndex: number;
-  readonly lastNonWhitespaceCharacterIndex: number;
-  readonly isEmptyOrWhitespace: boolean;
+  readonly rangeTrimmed: Range;
```
pokey, line 10: update from meet-up: let's make this `undefined` if `isEmptyOrWhitespace` is `true`

AndreasArvidsson:
Updated with undefined if the range is empty or whitespace.

packages/common/src/types/TextLine.ts
```diff
@@ -28,20 +28,13 @@ export interface TextLine {
   readonly rangeIncludingLineBreak: Range;
 
   /**
-   * The offset of the first character which is not a whitespace character as defined
-   * by `/\s/`. **Note** that if a line is all whitespace the length of the line is returned.
+   * The trimmed range, which excludes leading and trailing whitespace (`/\s/`).
+   * **Note** that if a line is all whitespace this is undefined so the trimmed range can never be empty.
```
phillco, line 32:
```suggestion
   * **Note:** if a line is all whitespace, the `rangeTrimmed` is undefined. Thus, a trimmed range can never be empty.
```

packages/cursorless-engine/src/processTargets/targets/DestinationImpl.ts
```diff
@@ -122,12 +122,12 @@ export class DestinationImpl implements Destination {
...
-        const nonWhitespaceCharacterIndex = this.isBefore
-          ? line.firstNonWhitespaceCharacterIndex
-          : line.lastNonWhitespaceCharacterIndex;
+        const trimmedPosition = this.isBefore
+          ? line.rangeTrimmed?.start ?? line.range.start
+          : line.rangeTrimmed?.end ?? line.range.end;
 
-        // Use the full line with included indentation and trailing whitespaces
-        if (insertionPosition.character === nonWhitespaceCharacterIndex) {
+        // Use the full line with including indentation and trailing whitespaces
```
phillco, line 129:
```suggestion
        // Use the full line, including indentation and trailing whitespaces
```

phillco (approved):

PR #2542, opened 2024-07-21
Stop using Intl.Collator

AndreasArvidsson:
`Intl.Collator` is not available in all environments. `localeCompare` gives the same result

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/actions/Sort.ts
+++ b/packages/cursorless-engine/src/actions/Sort.ts
@@ -45,13 +45,15 @@ abstract class SortBase implements SimpleAction {
 }
 
 export class Sort extends SortBase {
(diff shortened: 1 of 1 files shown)


pokey (approved):
We've got tests for this right? Ie something breaks if you remove the numeric true flag right? If so I'm happy

AndreasArvidsson:
> We've got tests for this right? Ie something breaks if you remove the numeric true flag right? If so I'm happy

Yes if I remove either of those arguments we have tests that fail.

AndreasArvidsson:
> Where's the types for `Intl.Collator` coming from? Can we remove it from our types entirely?

It's part of the typescript lib.

PR #2549, opened 2024-07-22
Properly insert CRLF in in memory text document

AndreasArvidsson:
## Checklist


pokey (approved):
really happy we're getting a solid suite of tests on this in memory text document. please keep adding them as our main test suite finds bugs. I think this is a path to starting to run our test suite on in-memory text document

Issue #729, opened 2022-06-02
Update the range update recode with better support for insertion and selection 


AndreasArvidsson:
* Support multiple edits in the same position and keeping track off their individual resulting range
* Support updating the editor's selection internally
* Think of a better api to to use different updating behavior(OpenOpen etc)

pokey:
Related to #202

pokey:
Also have a look at [`registerReplaceEditList`](https://www.cursorless.org/docs/contributing/api/classes/core_updateSelections_RangeUpdater.RangeUpdater/#registerreplaceeditlist), because that has a mechanism to associate changes that come back from VSCode with edits provided in `performDocumentEdits`.  It is not perfect; eg if two edits come in at the same place with the same text, it can mix them up.  We might need to make it more robust for this new purpose

PR #2551, opened 2024-07-23
New interface for performing edits and updating ranges

AndreasArvidsson:
This introduces a new interface to perform updates. The requirements I had was:
1. Support both selections and ranges
2. Support both edits and callback
3. Support setting range expansion behavior per range/selection
4. Support updating selection internally

What this pr doesn't try to do:
1. Improve the range updater logic

Example
```ts
const {
  selections: [updatedContentSelections],
  ranges: [updatedEditRanges]
} = await new EditsUpdater(rangeUpdater, editableEditor, edits)
  .selections(contentSelections)
  .ranges(editRanges, RangeExpansionBehavior.openOpen)
  .updateEditorSelections()
  .run();
```

## Ponderings
1. Should updating editor selections be opt out instead?
1. Here I'm using two separate classes `EditsUpdater` and `CallbackUpdater`. I'm a bit tempted to just use a single class and a polymorphic argument. `editsOrCallback: Edit[] | () => Promise<void>`. Something like that.
2. Strict necessary we don't actually need the two separate return lists. Internally the range updater works with selections so we converting ranges to selections internally. We could just return a single list of selections and since all selections are also ranges I don't think any of the actions would actually care?

If we embrace this polymorphic nature it could look something like this:
```ts
const [updatedEditSelections, updatedContentSelections] =
  await new EditsUpdater(rangeUpdater, editableEditor, edits)
    .add(contentSelections)
    .add(editRanges, RangeExpansionBehavior.openOpen)
    .updateEditorSelections()
    .run();
```

Personally I am a fan of the above since it became cleaner and simpler, but I know that @pokey generally doesn't prefer union arguments if it can be avoided.

We could of course get rid of `new` if we wanted to make it even cleaner.

```ts
const [updatedEditSelections, updatedContentSelections] =
  await performUpdates(rangeUpdater, editableEditor, edits)
    .add(contentSelections)
    .add(editRanges, RangeExpansionBehavior.openOpen)
    .updateEditorSelections()
    .run();
```

Related to #729

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/actions/BreakLine.ts
+++ b/packages/cursorless-engine/src/actions/BreakLine.ts
@@ -7,7 +7,7 @@ import {
 } from "@cursorless/common";
 import { flatten, zip } from "lodash-es";
 import type { RangeUpdater } from "../core/updateSelections/RangeUpdater";
-import { performEditsAndUpdateRanges } from "../core/updateSelections/updateSelections";
+import { EditsUpdater } from "../core/updateSelections/updateSelections";
 import { ide } from "../singletons/ide.singleton";
 import { Target } from "../typings/target.types";
 import { flashTargets, runOnTargetsForEachEditor } from "../util/targetUtils";
@@ -25,13 +25,14 @@ export class BreakLine {
       await runOnTargetsForEachEditor(targets, async (editor, targets) => {
         const contentRanges = targets.map(({ contentRange }) => contentRange);
         const edits = getEdits(editor, contentRanges);
+        const editableEditor = ide().getEditableTextEditor(editor);
 
-        const [updatedRanges] = await performEditsAndUpdateRanges(
-          this.rangeUpdater,
-          ide().getEditableTextEditor(editor),
-          edits,
-          [contentRanges],
-        );
+        const {
+          ranges: [updatedRanges],
+        } = await new EditsUpdater(this.rangeUpdater, editableEditor, edits)
+          .ranges(contentRanges)
+          .updateEditorSelections()
+          .run();
 
         return zip(targets, updatedRanges).map(([target, range]) => ({
           editor: target!.editor,
--- a/packages/cursorless-engine/src/actions/BreakLine.ts
+++ b/packages/cursorless-engine/src/actions/BreakLine.ts
@@ -7,7 +7,7 @@ import {
 } from "@cursorless/common";
 import { flatten, zip } from "lodash-es";
 import type { RangeUpdater } from "../core/updateSelections/RangeUpdater";
-import { EditsUpdater } from "../core/updateSelections/updateSelections";
+import { performEditsAndUpdateSelections } from "../core/updateSelections/updateSelections";
 import { ide } from "../singletons/ide.singleton";
 import { Target } from "../typings/target.types";
 import { flashTargets, runOnTargetsForEachEditor } from "../util/targetUtils";
@@ -27,12 +27,15 @@ export class BreakLine {
         const edits = getEdits(editor, contentRanges);
         const editableEditor = ide().getEditableTextEditor(editor);
 
-        const {
-          ranges: [updatedRanges],
-        } = await new EditsUpdater(this.rangeUpdater, editableEditor, edits)
-          .ranges(contentRanges)
-          .updateEditorSelections()
-          .run();
+        const { contentRanges: updatedRanges } =
+          await performEditsAndUpdateSelections({
+            rangeUpdater: this.rangeUpdater,
+            editor: editableEditor,
+            edits,
+            selections: {
+              contentRanges,
+            },
+          });
 
         return zip(targets, updatedRanges).map(([target, range]) => ({
           editor: target!.editor,
--- a/packages/cursorless-engine/src/actions/CallbackAction.ts
+++ b/packages/cursorless-engine/src/actions/CallbackAction.ts
@@ -114,7 +114,7 @@ export class CallbackAction {
     // Reset original selections
     if (options.setSelection && options.restoreSelection) {
       // NB: We don't focus the editor here because we'll do that at the
-      // very end. This code can run on mul:truetiple editors in the course of
+      // very end. This code can run on multiple editors in the course of
       // one command, so we want to avoid focusing the editor multiple
       // times.
       await editableEditor.setSelections(updatedOriginalSelections);
--- a/packages/cursorless-engine/src/actions/CallbackAction.ts
+++ b/packages/cursorless-engine/src/actions/CallbackAction.ts
@@ -2,7 +2,7 @@ import { EditableTextEditor, FlashStyle, TextEditor } from "@cursorless/common";
 import { flatten } from "lodash-es";
 import { selectionToStoredTarget } from "../core/commandRunner/selectionToStoredTarget";
 import { RangeUpdater } from "../core/updateSelections/RangeUpdater";
-import { CallbackUpdater } from "../core/updateSelections/updateSelections";
+import { performEditsAndUpdateSelections } from "../core/updateSelections/updateSelections";
 import { ide } from "../singletons/ide.singleton";
 import { Target } from "../typings/target.types";
 import {
@@ -97,19 +97,24 @@ export class CallbackAction {
       });
     }
 
-    const callback = () => options.callback(editableEditor, targets);
-
     const {
-      selections: [updatedOriginalSelections, updatedTargetSelections],
-    } = await new CallbackUpdater(this.rangeUpdater, editableEditor, callback)
-      .selections(originalSelections)
-      .selections(targetSelections)
-      .run();
+      originalSelections: updatedOriginalSelections,
+      targetSelections: updatedTargetSelections,
+    } = await performEditsAndUpdateSelections({
+      rangeUpdater: this.rangeUpdater,
+      editor: editableEditor,
+      callback: () => options.callback(editableEditor, targets),
+      preserveEditorSelections: true,
+      selections: {
+        originalSelections,
+        targetSelections,
+      },
+    });
 
     // Reset original selections
     if (options.setSelection && options.restoreSelection) {
       // NB: We don't focus the editor here because we'll do that at the
-      // very end. This code can run on multiple editors in the course of
+      // very end. This code can run on mul:truetiple editors in the course of
       // one command, so we want to avoid focusing the editor multiple
       // times.
       await editableEditor.setSelections(updatedOriginalSelections);
--- a/packages/cursorless-engine/src/actions/CallbackAction.ts
+++ b/packages/cursorless-engine/src/actions/CallbackAction.ts
@@ -104,7 +104,7 @@ export class CallbackAction {
       rangeUpdater: this.rangeUpdater,
       editor: editableEditor,
       callback: () => options.callback(editableEditor, targets),
-      preserveEditorSelections: true,
+      preserveCursorSelections: true,
       selections: {
         originalSelections,
         targetSelections,
--- a/packages/cursorless-engine/src/actions/CallbackAction.ts
+++ b/packages/cursorless-engine/src/actions/CallbackAction.ts
@@ -2,7 +2,7 @@ import { EditableTextEditor, FlashStyle, TextEditor } from "@cursorless/common";
 import { flatten } from "lodash-es";
 import { selectionToStoredTarget } from "../core/commandRunner/selectionToStoredTarget";
 import { RangeUpdater } from "../core/updateSelections/RangeUpdater";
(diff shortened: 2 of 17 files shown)


pokey:
Proposed new interface:

```ts
const {
            sourceEditRanges: updatedSourceEditRanges,
            destinationEditRanges: updatedDestinationEditRanges,
          } = await edit.performEdits({
            editor: editableEditor,
            edits: filteredEdits.map(({ edit }) => edit),
            // callback: async () => {...do whatever},
            selections: {
              sourceEditRanges,
              destinationEditRanges: {
                selections: destinationEditRanges,
                behavior: RangeExpansionBehavior.openOpen,
              },
            },
            // preserveEditorSelections: true,
          });
```

AndreasArvidsson:
@pokey Updated according to the new interface now

packages/cursorless-engine/src/actions/BreakLine.ts
```diff
@@ -25,13 +25,17 @@ export class BreakLine {
...
-        );
+        const { contentRanges: updatedRanges } =
+          await performEditsAndUpdateSelections({
+            rangeUpdater: this.rangeUpdater,
+            editor: editableEditor,
+            edits,
+            selections: {
+              contentRanges,
+            },
+          });
```
pokey, line 38: tempting to have shorthand
```suggestion
        const updatedRanges =
          await performEditsAndUpdateSelections({
            rangeUpdater: this.rangeUpdater,
            editor: editableEditor,
            edits,
            selections: contentRanges,
          });
```

but that might be getting a bit too overloaded 😅

packages/cursorless-engine/src/actions/CallbackAction.ts
```diff
@@ -97,18 +97,24 @@ export class CallbackAction {
...
+        originalSelections,
+        targetSelections,
+      },
+    });
 
     // Reset original selections
     if (options.setSelection && options.restoreSelection) {
       // NB: We don't focus the editor here because we'll do that at the
-      // very end. This code can run on multiple editors in the course of
+      // very end. This code can run on mul:truetiple editors in the course of
```
pokey, line 117:
```suggestion
      // very end. This code can run on multiple editors in the course of
```

packages/cursorless-engine/src/actions/EditNew/runEditTargets.ts
```diff
@@ -59,44 +51,45 @@ export async function runEditTargets(
...
+    thatRanges: updatedThatRanges,
+    cursorRanges: updatedCursorRanges,
+    editRanges: updatedEditRanges,
+  } = await performEditsAndUpdateSelections({
     rangeUpdater,
     editor,
     edits,
-    [thatSelections, cursorSelections, editSelections],
-  );
+    preserveEditorSelections: true,
```
pokey, line 65: I wonder if the word `cursor` is easier to understand, eg `preserveCursors` or `preserveCursorSelections`. I don't feel strongly, though; we're not super consistent. Just that `editorSelection` doesn't stronlgy evoke cursors to me

packages/cursorless-engine/src/core/updateSelections/updateSelections.ts
```diff
@@ -14,33 +14,114 @@ import {
...
+type UpdaterProps<K extends string> = EditsProps<K> | CallbackProps<K>;
+
+export async function performEditsAndUpdateSelections<K extends string>({
+  rangeUpdater,
+  editor,
+  selections,
+  preserveEditorSelections,
+  ...rest
+}: UpdaterProps<K>): Promise<Record<K, Selection[]>> {
+  const keys = Object.keys(selections) as K[];
```
pokey, line 48: I think you can use our `unsafeKeys` utility here

packages/cursorless-engine/src/actions/Remove.ts
pokey, line 1: this one will def benefit from an api that returns the edited ranges but agreed this is a good step forward

packages/cursorless-engine/src/actions/PasteFromClipboard.ts
```diff
@@ -29,32 +26,38 @@ export class PasteFromClipboard {
...
-      this.rangeUpdater,
-      async () => {
-        await this.actions.editNew.run(destinations);
-      },
-      editor.document,
-      [editor.selections],
-    );
+    const callbackEdit = async () => {
+      await this.actions.editNew.run(destinations);
+    };
```
pokey, line 31: i def seem to prefer inlining a bit more than you do, but I don't think it matters much 😊. Fwiw I think we could simplify this one by removing the `async` and `await`, in which case it is pretty strong candidate for inlining

packages/cursorless-engine/src/actions/Wrap.ts
```diff
@@ -50,67 +45,45 @@ export default class Wrap {
...
+          selections: {
+            boundariesStartSelections: {
+              selections: boundariesStartSelections,
+              behavior: RangeExpansionBehavior.openClosed,
+            },
+            boundariesEndSelections: {
+              selections: boundariesEndSelections,
+              behavior: RangeExpansionBehavior.closedOpen,
+            },
+            sourceSelections: contentSelections,
```
pokey, line 75: I know it saves lines, but I wonder if we want to rely so heavily on default behavior here. In this case I think it would actually be easier to read if it were parallel to the other entries

packages/cursorless-engine/src/actions/Wrap.ts
```diff
@@ -50,67 +45,45 @@ export default class Wrap {
...
+              behavior: RangeExpansionBehavior.openOpen,
+            },
+          },
+        });
 
-        await editableEditor.setSelections(cursorSelections);
+        const delimiterSelections = [
+          ...delimiterStartSelections,
+          ...delimiterEndSelections,
+        ];
```
pokey, line 86: ah interesting. our interface doesn't allow for a single list to have multiple range behaviours. I think your solution here is fine. Not worth making our interface more complex

packages/cursorless-engine/src/actions/Wrap.ts
```diff
@@ -50,67 +45,45 @@ export default class Wrap {
...
-
-        const cursorSelectionInfos = editor.selections.map((selection) =>
-          getSelectionInfo(
-            document,
-            selection,
-            RangeExpansionBehavior.closedClosed,
-          ),
-        );
+        const boundariesStartSelections = boundaries.map(({ start }) => start);
+        const boundariesEndSelections = boundaries.map(({ end }) => end);
```
pokey, line 49: I think I would just inline these. The variable names actually make things more confusing imo

packages/cursorless-engine/src/core/updateSelections/updateSelections.ts
```diff
@@ -14,33 +14,114 @@ import {
...
+    getSelectionInfoInternal(
+      document,
+      selection,
+      selection instanceof Selection ? !selection.isReversed : true,
+      rangeBehavior,
+    ),
   );
 }
 
+function getsSelectionsWithBehavior(
```
pokey, line 113:
```suggestion
function getSelectionsWithBehavior(
```

packages/cursorless-engine/src/core/updateSelections/updateSelections.ts
```diff
@@ -14,33 +14,114 @@ import {
...
+  edits: Edit[];
+}
+
+interface CallbackProps<K extends string> extends BaseProps<K> {
+  callback: () => Promise<void>;
+}
+
+type UpdaterProps<K extends string> = EditsProps<K> | CallbackProps<K>;
+
+export async function performEditsAndUpdateSelections<K extends string>({
```
pokey, line 41: docstring

packages/cursorless-engine/src/core/updateSelections/updateSelections.ts
```diff
@@ -14,33 +14,114 @@ import {
...
- * @param selection The selection
- * @param rangeBehavior How selection should behave with respect to insertions on either end
- * @returns An object that can be used for selection tracking
- */
-export function getSelectionInfo(
+interface BaseProps<K extends string> {
+  rangeUpdater: RangeUpdater;
+  editor: EditableTextEditor;
+  preserveEditorSelections?: boolean;
+  selections: Record<K, SelectionsOrRanges | SelectionsWithBehavior>;
```
pokey, line 28: docstrings for these and the ones in the below interface

packages/cursorless-engine/src/actions/Replace.ts
```diff
@@ -63,33 +63,41 @@ export default class Replace {
...
-          selections: edits.map(({ edit }) => edit.range.toSelection(false)),
-          rangeBehavior: RangeExpansionBehavior.openOpen,
-        };
+      async (editor, editWrappers) => {
+        const edits = editWrappers.map(({ edit }) => edit);
+        const contentSelections = editWrappers.map(
+          ({ target }) => target.contentSelection,
+        );
+        const editRanges = edits.map(({ range }) => range);
+        const editableEditor = ide().getEditableTextEditor(editor);
```
pokey, line 72: just a stylistic nit but i would personally inline this one as well; there is something to be said for making the function a bit smaller as you can fit it on a screen. Always a trade-off tho and very much a stylistic choice so feel free to ignore

my guess is your thinking is that these `performEditsAndUpdateSelections` calls tend to be pretty big so trying to keep the call itself as simple as possible, which makes sense

packages/cursorless-engine/src/actions/Replace.ts
```diff
@@ -63,33 +63,41 @@ export default class Replace {
...
       (edit) => edit.editor,
-      async (editor, edits) => {
-        const contentSelections = {
-          selections: edits.map(({ target }) => target.contentSelection),
-        };
-        const editSelections = {
-          selections: edits.map(({ edit }) => edit.range.toSelection(false)),
-          rangeBehavior: RangeExpansionBehavior.openOpen,
-        };
+      async (editor, editWrappers) => {
```
pokey, line 66: whoah this `editWrapper` thing is awkward. I don't love the name `editWrapper` but I agree not sure what to call it. A bigger refactor is prob in order here in the future

pokey (approved):
bunch of very minor nits. Feel free to merge once you've addressed the ones that seem important enough to address 😅

Happy with how the interface turned out tho; nice to rip out all those extra functions

packages/cursorless-engine/src/actions/BreakLine.ts
AndreasArvidsson: Yeah I think that is to much

packages/cursorless-engine/src/actions/CallbackAction.ts
AndreasArvidsson: Talon woopsie :p

packages/cursorless-engine/src/core/updateSelections/updateSelections.ts
AndreasArvidsson: Good call

packages/cursorless-engine/src/actions/PasteFromClipboard.ts
AndreasArvidsson: Unfortunately we can't do that since they have different return values. editNew returns an action results, but our callback expects void.

packages/cursorless-engine/src/actions/Wrap.ts
AndreasArvidsson: I can do that for for this specific action, but I don't think we should enforce it in the updated interface.

packages/cursorless-engine/src/actions/Wrap.ts
AndreasArvidsson: This is the only example of this so I think it's fine. We don't want increased complexity.

packages/cursorless-engine/src/actions/Replace.ts
AndreasArvidsson: Totally agree

packages/cursorless-engine/src/actions/Replace.ts
AndreasArvidsson: Yeah I was going for readability but I'm happy to inline them

packages/cursorless-engine/src/actions/Wrap.ts
pokey: yeah I agree the shorthand is useful sometimes

packages/cursorless-engine/src/actions/EditNew/runEditTargets.ts
AndreasArvidsson: renamed

PR #2552, opened 2024-07-23
Proper return type on HatMap.getToken()

AndreasArvidsson:
In some places we rely on the fact that the hat map can return undefined for a token and in other places we didn't handle this because the return type was wrong.

## Checklist


pokey (approved):

auscompgeek:
Eep. Should we turn on https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess ?

AndreasArvidsson:
Sounds very reasonable

PR #2558, opened 2024-07-24
Batch building/compiling tests

AndreasArvidsson:
We have to many tests in to many packages now. On windows I get `The commandline is too long`. Solving this by batching 50 tests at the time.

## Checklist


AndreasArvidsson:
@pokey Appears that `mapfile` doesn't work on the bash version used by default in macOS. Should we try this?
https://github.com/jitterbit/get-changed-files/issues/15#issuecomment-724909010

pokey (approved):
wow this is some serious bash-foo. guessing chatgpt helped you here? 😅

pokey:
looks like you'll need a different trick on mac tho; see CI 😕

pokey:
> @pokey Appears that `mapfile` doesn't work on the bash version used by default in macOS. Should we try this? [jitterbit/get-changed-files#15 (comment)](https://github.com/jitterbit/get-changed-files/issues/15#issuecomment-724909010)

hmm not sure. seems slightly strange to install new bash and overwrite default one. here's what chatgpt suggests:

```sh
#!/usr/bin/env bash
set -euo pipefail

# Initialize an empty array to hold the test files
TESTS=()

# Read the output of the find command into the TESTS array
while IFS= read -r file; do
    TESTS+=("$file")
done < <(find .. -name '*.test.ts')

BATCH_SIZE=50
SIZE=${#TESTS[@]}

for ((i = 0; i < SIZE; i += BATCH_SIZE)); do
    BATCH=("${TESTS[@]:i:BATCH_SIZE}")
    pnpm run build:base --outdir=dist --out-extension:.js=.cjs "${BATCH[@]}"
done
```

But I don't feel strongly; probably fine to use that bash upgrade trick?

pokey:
alternately, you could rewrite it in typescript so it can be read by humans? 😅

AndreasArvidsson:
> wow this is some serious bash-foo. guessing chatgpt helped you here? 😅

Maybe  :D

AndreasArvidsson:
> alternately, you could rewrite it in typescript so it can be read by humans? 😅

Using esbuild js api?

pokey:
actually i think `xargs` can do this for us:

```sh
#!/usr/bin/env bash
set -euo pipefail

find .. -name '*.test.ts' | xargs -n 50 pnpm run build:base --outdir=dist --out-extension:.js=.cjs
```

AndreasArvidsson:
Oh xargs was one of the things I tested first, but I couldn't get it to work. I must have done something weird because this looks fine :)

PR #2590, opened 2024-07-27
Support escape sequence in js string text fragments

### LINKED ISSUE #2393: Fix typescript text fragments when there is an escape sequence
In the following code:

```ts
"hello\nworld"
```

`textFragment` currently looks like:

image

We should have one text fragment for the entire inside of the string

Here's parse tree:

```scm
(string
    """
    (string_fragment)
    (escape_sequence)
    (string_fragment)
    """
)
```

should be easy with range

Though we need to be a bit careful because we don't want to eat interpolations, only escape sequences. Although I wonder how we can support

```ts
`aaa (bbb ${ccc} ddd) eee`
```

I would expect "take round" to work from `bbb`. Probably also from `ccc`, tho I could see a case against that as we're in a syntactic region so might not want text-based pairs to match

AndreasArvidsson:
Fixes #2393

## Checklist

### DIFF
--- a/data/fixtures/scopes/javascript.core/textFragment.comment.block.scope
+++ b/data/fixtures/scopes/javascript.core/textFragment.comment.block.scope
@@ -1,0 +1,13 @@
+/* hello
+world */
+---
+
+[Content] =
+[Removal] =
+[Domain] = 0:0-1:8
+  >--------
+0| /* hello
+1| world */
+   --------<
+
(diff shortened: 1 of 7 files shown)


pokey (approved):
Seems good enough. Technically includes interpolated regions, which should probably not be text fragments, but I think the proper way to do that may be to have those be nested, which is a bigger problem

PR #2593, opened 2024-07-27
Added literal mark

AndreasArvidsson:
I'm not sure this is a direction we want to go or if this is a good implementation, but for editors where we don't have hats it would be nice if you could just speak the literal text instead.
`"take hello world"`
`"chuck state get value"`

In this implementation I'm only looking in the viewport and I'm solving ambiguity by selecting the match closest to a cursor selection.

## Checklist

- [x] Add list for literal prefix/keyword. Empty by default. test that this behaves properly
- [x] Tag to remove prefix and have literal as a replacement for hats. Check dfa compilation
- [x] Tag to default to preferred scope instead of containing scope


pokey:
probably better to do this talon-side using dynamic lists and then send range. I believe I saw a demo of using dynamic lists for referring to text onscreen. We probably don't want to use a free text capture, and at the point we fix that, we'll have the ranges, so I think we should use them

AndreasArvidsson:
We can definitely implement this as a dynamic list. I still feel that there is an advantage in sending the actual words to Cursorless instead of the range. First of all I rather put the logic in typescript compared to python, but also we would make sure that different implementations of Cursorless everywhere had the same logic when it comes to disambiguation.

pokey:
What if this were a scope? In fact you could even just reuse custom regex scope by generating a literal regex the fly. Then you could say "take next literal hello" or "take previous literal hello"

AndreasArvidsson:
> What if this were a scope? In fact you could even just reuse custom regex scope by generating a literal regex on the fly. Then you could say "take next literal hello" or "take previous literal hello"

That was my first thought, but I kinda envisioned this as a replacement for hats in talonjs for example. You don't want something that is too verbose in that case. I also think that a disambiguation with distance to selections you could go either back or forward using the same spoken form is quite nice. We could of course both have a mark and a scope if we prefer that.

pokey:
I wonder if we should have a modifier that takes a scope and then applies your heuristic. Eg "good", so you could say "take good literal hello" and it would be equivalent to what you have here. We could then reuse this machinery so you could say "take good link" to take nearest url etc

pokey:
I guess one problem with having it be a scope is it balloons the grammar. The advantage of a mark is it has to appear at the end. I guess we could treat it as a mark grammatically in talon and then construct a "good instance" modifier to send to vscode

AndreasArvidsson:
If they can't actually use the full scope grammar I don't really see a point in doing that. For now I think we should just start with a mark.

pokey:
Right but if we can get the same effect by just adding a new modifier that we could also use with other scopes, that seems like a win, no?

pokey:
We might also experiment with a setting where that "good <scope>" modifier is used as a fallback for containing. That way we get the serenity behaviour where you just say "round" or "funk" and it uses a heuristic to figure out which one you mean

pokey:
update from meet-up:

- Add `"good"` modifier
- Support `"literal hello world"` as follows:
  - grammatically it functions as a mark, but it generates a modifier
  - Don't make it customizable for now
  - It generates the following modifier:

        {
            "type": "good",
            "scopeType": {
                "type": "customRegex",
                "regex": "hello world",
            }
        }
  - one reason `customRegex` might not be exact fit is we want case-insensitive flags which I don't think we can do using `customRegex`. So maybe we just want a new scope type called `literal` or just add `flags` field to `customRegex`

phillco:
One thing I didn't quite catch from the discussion -- where does "good" come from as the name for the modifier and what does it mean?

AndreasArvidsson:
`good` is a potential spoken form for the preferred scope modifier, but we might just implement this as the default instead of containing instead.

cursorless-talon/src/targets/primitive_target.py
```diff
@@ -11,7 +11,13 @@
...
 )
 def cursorless_primitive_target(m) -> PrimitiveTarget:
-    return PrimitiveTarget(
-        getattr(m, "cursorless_mark", None),
-        getattr(m, "cursorless_modifier_list", None),
-    )
+    mark = getattr(m, "cursorless_mark", None)
+    modifiers = getattr(m, "cursorless_modifier_list", None)
+
+    if mark is not None and mark["type"] == "literal":
```
AndreasArvidsson, line 17: This is the closest place where we have both the mark and the modifiers. I try to do as much as possible constructing the modifier in the literal mark. Basically only moving it from a mark to a modifier is done at this stage.

AndreasArvidsson:
Now this is implemented as a preferred scope modifier. Turned out really well.

The only thing that is left to figure out is how to disable this so we can try it out privately. As far as I know there's no way with a tag to disable a capture. You will just get an error if the rule is empty. We could of course put the word literal in the list and have that list be empty. That I think will work.

cursorless-talon/src/marks/literal_mark.py
```diff
@@ -0,0 +1,28 @@
...
+                "flags": "gui",
+            },
+        },
+    }
+
+
+def construct_fuzzy_regex(text: str) -> str:
+    parts = text.split(" ")
+    # Between each word there can be nothing(camelCase) or a non character symbol.
+    # Escape characters. eg: \t\r\n, are also acceptable.
```
phillco, line 27: I had a bit of trouble parsing `nothing(camelCase)`. Does it mean
> there can be nothing (e.g., when searching for a camelCase identifier)

Is this version of the comment equally accurate?

```suggestion
    # Each word can optionally be separated by a sequence of non-letter characters (as well as \t, \r, or \n).
```

cursorless-talon/src/marks/literal_mark.py
```diff
@@ -0,0 +1,28 @@
+from talon import Module
+
+from .mark_types import LiteralMark
+
+mod = Module()
+
+
+# FIXME: Replace literal *heh* with a Talon list before we make this public
```
phillco, line 8: \*heh\* is a joke because it's a command literal named literal, right? It makes sense but took a couple of parses

phillco (commented):
> We could of course put the word literal in the list and have that list be empty. That I think will work.

I _think_ that sounds good to me. The intention is to ship this to main, and test it in normal use, so we don't have to be on a branch?

AndreasArvidsson:
> > We could of course put the word literal in the list and have that list be empty. That I think will work.
> 
> I _think_ that sounds good to me. The intention is to ship this to main, and test it in normal use, so we don't have to be on a branch?

Correct

cursorless-talon/src/marks/literal_mark.py
AndreasArvidsson: More or less. Each word can be separated by nothing. `helloWorld` are an example of two words separated by nothing.

cursorless-talon/src/marks/literal_mark.py
AndreasArvidsson: Exactly. I wasn't sure if anyone else would get it :D

AndreasArvidsson:
@phillco This one is now ready

To activate literal with a prefix just supply a list
```talon
list: user.private_cursorless_literal_mark
-
literal
```

To activate literal without a prefix enable tag
`tag(): user.private_cursorless_literal_mark_no_prefix`

To use preferred scope instead of containing scope enable setting
`user.private_cursorless_use_preferred_scope = true`

All of this is working. The only drawback was that I had to switch from `<user.text>` work  `<phrase>`. The dfa increased about a second with user text. With phrase it appears to be negligible.

phillco:
Great, I'll take a look in the morning.

> The only drawback was that I had to switch from <user.text> to <phrase>. The dfa increased about a second with user text. With phrase it appears to be negligible.

Interesting, any idea why user.text is so much more expensive?

AndreasArvidsson:
> Great, I'll take a look in the morning.
> 
> > The only drawback was that I had to switch from <user.text> to . The dfa increased about a second with user text. With phrase it appears to be negligible.
> 
> Interesting, any idea why user.text is so much more expensive?

It's a complex capture. Specifically mine has things like abbreviations, date and time, etc in it.

cursorless-talon/src/targets/primitive_target.py
phillco, line 17:
```suggestion
    # for grammar performance reasons, the literal modifier is exposed to Talon as a mark,
    # but is converted to a modifier in the engine.
    if mark is not None and mark["type"] == "literal":
```

packages/cursorless-engine/src/processTargets/modifiers/PreferredScopeStage.ts
```diff
@@ -0,0 +1,103 @@
...
+import type { ModifierStage } from "../PipelineStages.types";
+import { getContainingScopeTarget } from "./getContainingScopeTarget";
+import type { TargetScope } from "./scopeHandlers/scope.types";
+import type { ScopeHandler } from "./scopeHandlers/scopeHandler.types";
+import type { ScopeHandlerFactory } from "./scopeHandlers/ScopeHandlerFactory";
+
+/**
+ * This modifier tries to find the preferred scope for the target.
+ * It does this by first looking for a containing scope, and if that fails,
+ * it looks for the closest scope.
```
phillco, line 13:
```suggestion
 * This modifier tries to find the "preferred" scope for the target.
 *
 * It does this by first looking for a containing scope, and if one doesn't exist, the closest scope.
```

packages/cursorless-engine/src/processTargets/modifiers/PreferredScopeStage.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    if (scopeHandler == null) {
+      throw Error(`Couldn't create scope handler for: ${scopeType.type}`);
+    }
+
+    const containingTargets = getContainingScopeTarget(target, scopeHandler);
+
+    if (containingTargets != null) {
+      return containingTargets;
+    }
```
phillco, line 38:
```suggestion
    const containingTargets = getContainingScopeTarget(target, scopeHandler);
    if (containingTargets) {
      return containingTargets;
    }
```

phillco: Since the if statement it is tightly bound to the variable declaration I would remove the blank line for readability

Or if there is an equivalent to the Python syntax:

```
    if (containingTargets := getContainingScopeTarget(target, scopeHandler)) {
      return containingTargets;
    }
```

packages/cursorless-engine/src/processTargets/modifiers/PreferredScopeStage.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    const closestTargets = getClosestScopeTargets(target, scopeHandler);
+
+    if (closestTargets != null) {
+      return closestTargets;
+    }
```
phillco, line 44: Same here

packages/cursorless-engine/src/processTargets/modifiers/PreferredScopeStage.ts
```diff
@@ -0,0 +1,103 @@
...
+      break;
+    }
+  }
+
+  return { scope: closestScope, distance: closestDistance };
+}
+
+function distanceBetweenPositions(a: Position, b: Position): number {
+  return (
+    Math.abs(a.line - b.line) * 10000 + Math.abs(a.character - b.character)
```
phillco, line 99:
```suggestion
    // 10000 is arbitrary to always pick same-line occurrences first
    Math.abs(a.line - b.line) * 10000 + Math.abs(a.character - b.character)
```

cursorless-talon/src/marks/literal_mark.py
```diff
@@ -0,0 +1,49 @@
...
+    desc="Tag for enabling literal mark without prefix",
+)
+
+ctx_no_prefix = Context()
+ctx_no_prefix.matches = r"""
+tag: user.private_cursorless_literal_mark_no_prefix
+"""
+
+
+@mod.capture(rule="{user.private_cursorless_literal_mark} <phrase>")
```
phillco, line 21:
```suggestion
# NB: <phrase> is used over <user.text> for DFA performance reasons
# (we intend to replace is with a dynamic list of document contents eventually)
@mod.capture(rule="{user.private_cursorless_literal_mark} <phrase>")
```

cursorless-talon/src/marks/literal_mark.py
```diff
@@ -0,0 +1,49 @@
...
+                "flags": "gui",
+            },
+        },
+    }
+
+
+def construct_fuzzy_regex(text: str) -> str:
+    parts = text.split(" ")
+    # Between each word there can be nothing(camelCase) or a non character symbol.
+    # Escape characters. eg: \t\r\n, are also acceptable.
```
phillco, line 48:
```suggestion
    # Between each word there can be any number of non-alpha symbols (including escape characters: \t\r\n). No separator at all is also valid -- for example, when searching for a camelCase identifier.
```

cursorless-talon/src/modifiers/simple_scope_modifier.py
```diff
@@ -13,6 +13,13 @@
     desc="Cursorless ancestor scope modifiers",
 )
 
+# This is a private setting and should not be used by non Cursorless developers
+mod.setting(
+    "private_cursorless_use_preferred_scope",
+    bool,
+    desc="Use preferred scope instead of containing scope",
```
phillco, line 20:
```suggestion
    desc="Use preferred scope instead of containing scope for all scopes by default (EXPERIMENTAL)",
```

phillco (approved):
lgtm, with the previous suggestions

cursorless-talon/src/marks/literal_mark.py
```diff
@@ -0,0 +1,50 @@
...
+)
+
+ctx_no_prefix = Context()
+ctx_no_prefix.matches = r"""
+tag: user.private_cursorless_literal_mark_no_prefix
+"""
+
+
+# NB: <phrase> is used over <user.text> for DFA performance reasons
+# (we intend to replace is with a dynamic list of document contents eventually)
```
phillco, line 22:
```suggestion
# (we intend to replace this with a dynamic list of document contents eventually)
```

phillco (commented):
One typo in my own suggestion :P

phillco (approved):

PR #2595, opened 2024-07-28
Added `boundedNonWhitespaceSequence` scope handler

### LINKED ISSUE #903: Add `"short block"` scope type
Analogously to the `"short paint"` scope type we'd like to add a `"short block"` scope type that stops when it hits the end of a surrounding pair

Should be able to share some of the implementation of `"short paint"`

AndreasArvidsson:
Fixes #903

## Checklist


packages/cursorless-engine/src/processTargets/modifiers/EveryScopeStage.ts
```diff
@@ -143,7 +143,7 @@ function getScopesOverlappingRange(
   return Array.from(
     scopeHandler.generateScopes(editor, start, "forward", {
       distalPosition: end,
-      skipAncestorScopes: true,
+      includeDescendantScopes: true,
```
AndreasArvidsson, line 146: This is necessary if we want to use ordinals with short paint

AndreasArvidsson: I guess the other option is to either yield the full range or the sub ranges not both. We could make this decision based on the position. Probably something we need to discuss.

AndreasArvidsson: This made a lot of other test fails I'm restoring it for now.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+import { compareTargetScopes } from "./compareTargetScopes";
+import type { TargetScope } from "./scope.types";
+import type { ScopeHandler } from "./scopeHandler.types";
+import type { ScopeHandlerFactory } from "./ScopeHandlerFactory";
+
+export class BoundedNonWhitespaceSequenceScopeHandler extends BaseScopeHandler {
+  public readonly scopeType = { type: "boundedNonWhitespaceSequence" } as const;
+  public readonly iterationScopeType: ScopeType = {
+    type: "line",
+  } as const;
```
pokey, line 20: I believe this should be a `oneOf` between `line` and strictly contained inside surrounding pair

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+      },
+      this.languageId,
+    )!;
+
+    const surroundingPairScopeHandler = this.scopeHandlerFactory.create(
+      {
+        type: "surroundingPair",
+        delimiter: "any",
+        requireStrongContainment: true,
+      },
```
pokey, line 48: I think we should have a version of this scope where content range is interior and requires strong containment. Then we can use that here and as part of the `oneOf` for iteration scope above

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+      this.languageId,
+    )!;
+
+    const paintScopes = paintScopeHandler.generateScopes(
+      editor,
+      position,
+      direction,
+      {
+        allowAdjacentScopes: true,
+      },
```
pokey, line 58: why not just use `hints`?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+function getSurroundingPairInteriors(
+  surroundingPairScopeHandler: ScopeHandler,
+  target: Target,
+): Range[] {
+  const surroundingPairScopes = surroundingPairScopeHandler.generateScopes(
+    target.editor,
+    target.contentRange.start,
+    "forward",
```
pokey, line 92: I think you might be able to just use direction and hints from main function call?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+  );
+  return Array.from(surroundingPairScopes).map(
+    (scope) => assertSingleTarget(scope).getInterior()![0].contentRange,
+  );
```
pokey, line 96: and yeah if we have a proper scope type that yields interior we don't need all this stuff

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+    editor: TextEditor,
+    position: Position,
+    direction: Direction,
+  ): Iterable<TargetScope> {
+    const paintScopeHandler = this.scopeHandlerFactory.create(
+      {
+        type: "nonWhitespaceSequence",
+      },
+      this.languageId,
+    )!;
```
pokey, line 41: I believe if we change this to `block` we get https://github.com/cursorless-dev/cursorless/issues/903 for free; could prob just use a shared base class

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+      editor,
+      position,
+      direction,
+      {
+        allowAdjacentScopes: true,
+      },
+    );
+
+    const scopes = Array.from(paintScopes)
+      .flatMap((scope) => {
```
pokey, line 62: do you need to do this? If you follow all the hints and directions, I think you could keep everything in iterator space

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
AndreasArvidsson: I don't want the surrounding pair containing the position. I want the surrounding pair containing the paint target.
Using the distal position is a good idea.

pokey: you do want the one containing hte position. eg `aaa(bb|b)ccc`

AndreasArvidsson: No I don't think so. 
eg: `|(a) (b)`
I want to yield `b` then  I need to find the surrounding pairs in each paint.

AndreasArvidsson:
Updated with `surroundingPairInterior` scope and implemented using `oneOf` scope.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
AndreasArvidsson: With the implementation using `oneOf` it's not that much code to add short block in a similar matter.
https://github.com/cursorless-dev/cursorless/blob/9bf56f2053cf062391a27b5279fc6d5a524570ba/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts

AndreasArvidsson:
I thought that was what you suggested, but I read it again and you were only talking about iteration scope. I will go back to my previous implementation.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
AndreasArvidsson: We now support short block

data/fixtures/scopes/textual/boundedParagraph.scope
```diff
@@ -0,0 +1,17 @@
+{
```
AndreasArvidsson, line 1: These scope tests didn't really turn out to well since the position for our recorder is 0,0. I wonder if we should do something like `$0` in the future to indicate position?
For now I think it's fine to just have the files there and we can look into an improvement to our scope recorder in the future.

pokey: This is a bug in your implementation. You're not yielding nested scopes

AndreasArvidsson: I punted on that. It made multiple test fail in unexpected ways. I don't know how it even should behave. 
Our test recorder should still be able to catch the scope that is there was my point.

pokey (commented):
This is pretty broken. Your original impl was actually much closer. Maybe let's move back there and go from there?

AndreasArvidsson:
> This is pretty broken. Your original impl was actually much closer. Maybe let's move back there and go from there?

This one passes all the tests. I couldn't get the other one to do that, but sure.

pokey:
Yes but the tests you added are wrong

AndreasArvidsson:
> Yes but the tests you added are wrong

I'm not sure it's possible to get it correct. The amount of time I spent on this and it's just between breaking existing behavior or getting nested scopes afaik.

pokey:
I'll have a look

AndreasArvidsson:
Okay now I have something that is a bit better but it quickly becomes complicated. I don't know if there is a way to simplify it, but now at least the tests are passing and we do have nested scopes.

data/fixtures/scopes/textual/boundedParagraph2.scope
```diff
@@ -0,0 +1,211 @@
...
+
+[#1 Insertion delimiter] = "\n\n"
+
+
+[#2 Content] =
+[#2 Domain] = 2:4-3:7
+      >---
+2|     ccc
+3|     ddd
+   -------<
```
pokey, line 53: as above, I think domain here should extend to at least start of line, maybe all the way to end of opening `{`

data/fixtures/scopes/textual/boundedParagraph2.scope
```diff
@@ -0,0 +1,211 @@
...
+
+[#2 Removal] = 2:0-5:0
+  >-------
+2|     ccc
+3|     ddd
+4|
+5|     eee
+   <
```
pokey, line 61: I think this removal looks good though

data/fixtures/scopes/textual/boundedParagraph2.scope
```diff
@@ -0,0 +1,211 @@
...
+  >
+7|
+8|     ggg
+   <
+
+[#3 Domain] = 5:0-6:7
+  >-------
+5|     eee
+6|     fff
+   -------<
```
pokey, line 111: this domain looks good

data/fixtures/scopes/textual/boundedParagraph2.scope
```diff
@@ -0,0 +1,211 @@
...
+
+[#3 Removal] = 5:0-8:0
+  >-------
+5|     eee
+6|     fff
+7|
+8|     ggg
+   <
```
pokey, line 87: this looks good

data/fixtures/scopes/textual/boundedParagraph2.scope
```diff
@@ -0,0 +1,211 @@
...
+
+[#3 Insertion delimiter] = "\n\n"
+
+
+[#4 Content] =
+[#4 Domain] = 8:0-9:7
+  >-------
+8|     ggg
+9|     hhh
+   -------<
```
pokey, line 121: this content range is wrong, but the domain seems ~good. Tempted to have end go to start of `}` but I don't feel strongly

data/fixtures/scopes/textual/boundedParagraph2.scope
```diff
@@ -0,0 +1,211 @@
...
+
+[#4 Removal] = 7:0-10:0
+  >
+7|
+8|     ggg
+9|     hhh
+10| } iii
+    <
```
pokey, line 129: this looks good

data/fixtures/scopes/textual/boundedParagraph2.scope
```diff
@@ -0,0 +1,211 @@
...
+
+[#5 Removal] = 8:0-13:0
+  >-------
+8|     ggg
+9|     hhh
+10| } iii
+11| jjj
+12|
+13| kkk
+    <
```
pokey, line 159: kind of an edge case, but I think this looks good

data/fixtures/scopes/textual/boundedParagraph2.scope
```diff
@@ -0,0 +1,211 @@
...
+
+[#1 Removal] = 0:0-5:0
+  >---
+0| aaa
+1| bbb {
+2|     ccc
+3|     ddd
+4|
+5|     eee
+   <
```
pokey, line 34: edge case, but I think I'm happy with this one

data/fixtures/scopes/textual/boundedParagraph3.scope
```diff
@@ -0,0 +1,122 @@
...
+  >-----<
+2|   aaa
+
+[#2 Insertion delimiter] = "\n\n"
+
+
+[#3 Content] =
+[#3 Domain] = 2:2-2:5
+    >---<
+2|   aaa
```
pokey, line 71: this seems to be a duplicate of scope 2 above?

data/fixtures/scopes/textual/boundedParagraph4.scope
```diff
@@ -0,0 +1,43 @@
...
+[#2 Content] =
+[#2 Domain] = 0:4-0:7
+      >---<
+0| aaa(bbb) {
+
+[#2 Removal] = 0:0-1:0
+  >----------
+0| aaa(bbb) {
+1|   ccc
+   <
```
pokey, line 27: Hmm. I'm not sure we want to surface this one at all. If we do, the content and removal range have a very strange relationship 😅

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
```diff
@@ -0,0 +1,178 @@
...
+        return [
+          new ParagraphTarget({
+            editor,
+            isReversed,
+            contentRange,
+          }),
+        ];
+      },
+    };
+  }
```
pokey, line 177: I wonder how things work in this case. Because paragraph targets kind of expect to actually be a full block. See the test cases I added

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
```diff
@@ -0,0 +1,178 @@
...
+        editor,
+        position,
+        direction,
+        {
+          ...hints,
+          // For every (skipAncestorScopes=true) we don't want to go outside of the surrounding pair
+          containment:
+            hints.containment == null && hints.skipAncestorScopes
+              ? "required"
+              : hints.containment,
```
pokey, line 90: what happens if you leave this out? A bit tough to reason about what's happening here 😅

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
```diff
@@ -0,0 +1,178 @@
...
+      {
+        ...hints,
+        // Don't skip containing paint since it might have non contained nested scopes.
+        containment:
+          hints.containment !== "disallowed" ? hints.containment : undefined,
+      },
+    );
+
+    const pairScopes = Array.from(
+      this.surroundingPairInteriorScopeHandler.generateScopes(
```
pokey, line 80: I think this is very reasonable. My first instinct would have been to do it per paint, but this is more efficient with our current impl of pairs and is nice and simple

pokey (commented):
nice work! pretty much all of my complaints are with short block, so I would be fine with us pushing that to a follow-up PR.

data/fixtures/recorded/scopes/boundedParagraph/chuckShortBlock.yml
```diff
@@ -0,0 +1,30 @@
...
+        bar
+    }
+  selections:
+    - anchor: {line: 2, character: 4}
+      active: {line: 2, character: 4}
+  marks: {}
+finalState:
+  documentContents: |-
+    {
+    }
```
pokey, line 27: as mentioned elsewhere, I'd be tempted to close these `{}` but I don't feel strongly

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairInteriorScopeHandler.ts
```diff
@@ -0,0 +1,72 @@
...
+      yield {
+        editor,
+        domain: interiorRange,
+        getTargets(isReversed) {
+          return [
+            new TokenTarget({
+              editor,
+              isReversed,
+              contentRange: interiorRange,
+            }),
```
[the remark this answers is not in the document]
AndreasArvidsson: That was my original intents, but interior targets expects the full interior range which we don't have here.
We need to apply the `isReversed` argument. Otherwise we could have done that.

data/fixtures/scopes/textual/boundedParagraph.scope
```diff
@@ -0,0 +1,34 @@
...
+
+[#1 Insertion delimiter] = "\n\n"
+
+
+[#2 Content] =
+[#2 Domain] = 1:4-2:7
+      >---
+1|     foo
+2|     bar
+   -------<
```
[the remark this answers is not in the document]
AndreasArvidsson: To distort of the line is no problem, but from an empty line you should not be able to take a paragraph? That sounds weird.

data/fixtures/scopes/textual/boundedParagraph.iteration.scope
```diff
@@ -0,0 +1,22 @@
...
+3| }
+   -<
+
+
+[#2 Range] =
+[#2 Domain] = 1:4-2:7
+      >---
+1|     foo
+2|     bar
+   -------<
```
pokey, line 22: would be tempted for domain to extend all the way up to surrounding pairs? At the very least should extend to start of line

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
AndreasArvidsson: `every short block` takes outside of the containing pair

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
AndreasArvidsson: Exactly. My original implementation was to get it per paint, but this simplified enough that we don't have to. And for performance it's nice that every short paint doesn't have to do surrounding pair ten times :D

data/fixtures/scopes/textual/boundedParagraph.scope
```diff
@@ -0,0 +1,34 @@
...
+
+[#2 Removal] = 1:0-3:0
+  >-------
+1|     foo
+2|     bar
+3| }
+   <
```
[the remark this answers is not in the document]
AndreasArvidsson: It still a paragraph and paragraph removes entire lines. I think it would be strange if we all of the sudden removed part of lines in a paragraph.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairInteriorScopeHandler.ts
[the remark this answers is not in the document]
pokey: I went ahead and did it and it seemed to work well. It broke some of the new tests but they're all due to the fact we're using paragraph removal which I don't think is correct anyway

data/fixtures/scopes/textual/boundedParagraph.scope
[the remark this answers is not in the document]
pokey: I don't feel strongly; I'd lean towards whatever is easiest to impl and then we can tweak if it feels awkward in practice

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
```diff
@@ -0,0 +1,179 @@
...
+    return {
+      editor,
+      domain: new Range(domainStart, contentRange.end),
+      getTargets(isReversed) {
+        return [
+          new ParagraphTarget({
+            editor,
+            isReversed,
+            contentRange,
+          }),
```
pokey, line 174: I think this is breaking things. Removal range for `ParagraphTarget` expects it to not be touching other lines

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/BoundedNonWhitespaceSequenceScopeHandler.ts
```diff
@@ -161,9 +161,10 @@ export class BoundedParagraphScopeHandler extends BoundedBaseScopeHandler {
     editor: TextEditor,
     contentRange: Range,
   ): TargetScope {
+    const domainStart = new Position(contentRange.start.line, 0);
     return {
       editor,
-      domain: contentRange,
+      domain: new Range(domainStart, contentRange.end),
```
pokey: i'm not sure this is necessary anymore now we're using real interior targets

AndreasArvidsson:
The problem with short block is that since we started using surrounding pair interior we don't have access to the boundaries any more. I think what we need to do is use a full surrounding pair targets we can actually get the boundaries and then we can update the domain as use a suggests.
If it okay with you I think we should remove short block from this pull request and do that in a separate one?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairInteriorScopeHandler.ts
```diff
@@ -0,0 +1,65 @@
...
+        position,
+        direction,
+        hints,
+      ),
+      (scope) => ({
+        editor,
+        domain: scope.domain,
+        getTargets(isReversed) {
+          return scope
+            .getTargets(isReversed)
```
AndreasArvidsson, line 59: We are fine with just ignoring this argument?

pokey:
> The problem with short block is that since we started using surrounding pair interior we don't have access to the boundaries any more. I think what we need to do is use a full surrounding pair targets we can actually get the boundaries and then we can update the domain as use a suggests. If it okay with you I think we should remove short block from this pull request and do that in a separate one?

hmm I think the domain is actually fine after my changes, but I agree we can leave this for a follow-up because "short paint" is working really well so let's ship that

AndreasArvidsson:
Ok bounded paragraph is now removed so once we have this merge I can make a follow up with that code so we can focus on it specifically.

pokey:
One reason to fix it here is all the comments I left on this PR. Shall I take a quick swing? I think we're close

AndreasArvidsson:
Please do!

data/fixtures/scopes/textual/boundedParagraph4.scope
```diff
@@ -0,0 +1,44 @@
...
+   -<
+
+[#1 Insertion delimiter] = "\n\n"
+
+
+[#2 Content] =
+[#2 Removal] =
+[#2 Domain] = 0:4-0:7
+      >---<
+0| aaa(bbb) {
```
pokey, line 22: I would be tempted to remove this one? Seems odd to get trapped inside parens on a single line when you say "short block", but maybe we should see how it feels in practice?

data/fixtures/scopes/textual/boundedParagraph5.scope
```diff
@@ -0,0 +1,132 @@
...
+[#2 Content] =
+[#2 Domain] = 0:1-0:4
+   >---<
+0| {aaa
+
+[#2 Removal] = 0:1-1:0
+   >---
+0| {aaa
+1|
+   <
```
pokey, line 41: Arguably this should extend to the beginning of `bbb`, but I can't imagine this edge case is too common

data/fixtures/scopes/textual/boundedParagraph5.scope
```diff
@@ -0,0 +1,132 @@
...
+[#4 Content] =
+[#4 Domain] = 4:0-4:3
+  >---<
+4| ccc}
+
+[#4 Removal] = 3:0-4:3
+  >
+3|
+4| ccc}
+   ---<
```
pokey, line 97: As above, there's a case for this to extend to end of `bbb`, but prob not worth tackling

packages/cursorless-engine/src/processTargets/targets/BoundedParagraphTarget.ts
```diff
@@ -0,0 +1,135 @@
...
+      : undefined;
+  }
+
+  getTrailingDelimiterTarget() {
+    return this.endLineGap > 1
+      ? this.paragraphTarget.getTrailingDelimiterTarget()
+      : undefined;
+  }
+
+  getRemovalRange(): Range {
```
pokey, line 52: Copied from `ParagraphTarget` with addition of `startLineGap` check in the middle

packages/cursorless-engine/src/processTargets/targets/BoundedParagraphTarget.ts
```diff
@@ -0,0 +1,135 @@
...
+      editor: this.editor,
+      isReversed: this.isReversed,
+    }).getRemovalRange();
+  }
+
+  private get fullLineContentRange() {
+    return expandToFullLine(this.editor, this.contentRange);
+  }
+
+  getRemovalHighlightRange() {
```
pokey, line 87: As above, copied from `ParagraphTarget` with boundary check

pokey (approved):
Ok I got it working but it was a bit more involved than I expected 😅. Happy to make it into a separate PR, though I left a bunch of comments on my impl so I don't want to lose them

pokey:
fwiw here's my changes https://github.com/cursorless-dev/cursorless/pull/2595/files/f0de01a3fc2250b360dba3fa006a69f78ae9f29e..f0755690d063dc2141372a390dc08b8adde9e787

PR #2599, opened 2024-07-29
Initial work towards next-gen collectionItem

### LINKED ISSUE #1059: Migrate `collectionItem` to use a scope handler
Use a one off scope handler to combine the treesitter based implementation with the text based one

pokey:
Fixes #1059

## Checklist


packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/CollectionItemScopeHandler/CollectionItemScopeHandler.ts
```diff
@@ -0,0 +1,59 @@
...
+  public readonly iterationScopeType: ScopeType = {
+    type: "oneOf",
+    scopeTypes: [
+      { type: "line" },
+      {
+        type: "surroundingPairInterior",
+        delimiter: "any",
+      },
+    ],
+  };
```
pokey, line 30: Actually this shouldn't be `oneOf`, it should be what I think we call `fallback`, where it uses inside if it can find it and only uses line if it finds no containing inside

PR #2602, opened 2024-07-29
Make hat forcing more aggressive

pokey:
## Checklist

### DIFF
--- a/packages/cursorless-engine/src/util/allocateHats/getRankedTokens.ts
+++ b/packages/cursorless-engine/src/util/allocateHats/getRankedTokens.ts
@@ -1,6 +1,10 @@
-import { TextEditor } from "@cursorless/common";
+import {
+  CompositeKeyMap,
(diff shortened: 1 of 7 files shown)


packages/cursorless-engine/src/util/allocateHats/getRankedTokens.ts
```diff
@@ -44,7 +49,33 @@ export function getRankedTokens(
...
     );
 
-    return tokens.map((token, index) => ({ token, rank: -index }));
+    return tokens;
+  });
+
+  return moveForcedHatsToFront(forcedHatMap, tokens).map((token, index) => ({
+    token,
+    rank: -index,
+  }));
```
pokey, line 58: notice that we generate the ranks at the end now. It seems we had a bug where we were treating hats as having the same rank if they were in other editors

AndreasArvidsson (approved):

PR #2611, opened 2024-07-31
feat: Add lua-language-server to CI

### LINKED ISSUE #2576: neovim: add lua language server lint checks to CI
Atm warnings from vscode triggered by sumneko.lua plugin (using lua language server) don't fire in CI. This should ideally added via .pre-commit so people that run that locally can also see the errors in advance.

fidgetingbits:
- Implement lua-language-server linter in pre-commit
- Install via Nix due to lack of Ubuntu package
- Apply patch for --check CLI bug using Nix overlay
- Centralize linting config for VSCode, pre-commit, and CI

Fixes #2576

## Checklist


pokey:
I wonder if we want to do this in pre-commit or as part of test. This feels like running tsc, which we do as part of our test CI step. I'm not in love with needing nix set up in order for pre-commit to work locally, so doing it as part of test makes it easier to just be CI-only / manually triggered locally

fidgetingbits:
Ok. I moved it into test.yml. I went the route of having another combined action again so we don't have to litter multiple entries with 'if linux'. Didn't manually test these new changes yet, so will see how the CI goes.

flake.nix
```diff
@@ -45,7 +46,21 @@
                     };
                   };
                   neovim = prev.neovim.override { withNodeJs = true; };
-
+                  # There is a recent bug that prevents cli --check invocation:
+                  # See: https://github.com/LuaLS/lua-language-server/pull/2775
```
pokey, line 50: Shall we file an issue to remove this once that PR ships?

scripts/lint-lua-ls.sh
```diff
@@ -0,0 +1,42 @@
...
+# lua-language-server should be installed automatically by the flake.nix dev shell
+# or .github/workflows/pre-commit.yml
+if ! type lua-language-server &>/dev/null; then
+  echo "ERROR: lua-language-server is not installed. Please run 'nix develop' or install it manually."
+  exit 1
+fi
+
+if [ ! -e "${CURSORLESS_REPO_ROOT-nonexistent}" ]; then
+  CURSORLESS_REPO_ROOT=$(git rev-parse --show-toplevel)
+fi
```
pokey, line 13: @AndreasArvidsson this is what we should use for that script you removed when we reintroduce it. Not sure why we didn't think of this

scripts/lint-lua-ls.sh
```diff
@@ -0,0 +1,42 @@
...
+    fi
+    return 1
+
+  fi
+  return 0
+}
+
+# lua-language-server doesn't support single file parsing, so check entire folder
+exit_code=0
+if ! check_file .; then
```
pokey, line 38: Should we pass repo root here instead of . to make it more robust?

pokey (approved):
Looks good with minor comments. Feel free to merge once addressed. I dig that sub-action thing. We could do more of that

flake.nix
fidgetingbits: https://github.com/cursorless-dev/cursorless/issues/2613

pokey, line 50:
```suggestion
                  # See #2613
```

fidgetingbits:
Any idea why the 'ubuntu-latest, stable' CI test would've skipped here? That's the only one that will run the new lua test and was running before, so seems a bit weird... but details don't show anything. I don't seem to be able to force re-run it

pokey:
Hmm strange. Maybe try pushing an empty commit?

Issue #2618, opened 2024-08-01
Enforce `type` in front of type-only imports


pokey:
We want to proceed in the following steps, each in its own PR, in the following exact order:

- [ ] enable `"@typescript-eslint/consistent-type-imports": "error"`
- [ ] enable [`--verbatimModuleSyntax` TS flag](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax)
- [ ] enable `"import/consistent-type-specifier-style": ["error", "prefer-top-level"]`
- [ ] enable `"import/no-duplicates": "error"` (note we don't need `prefer-inline` because we don't use inline)

The reasoning is that we want `import type` to be like its own kind of statement that we know always gets dropped in the output Javascript, and in addition, if we were to have a rule where `import {type foo}` => `import type {foo}` only when there are _no_ non-type imports in the statement, it would feel a bit inconsistent that it doesn't happen when there _are_ non-type imports. The proposed rule is simpler: types always get imported in their own statement, and never have side effects

Each step should be its own PR, and they should be in the above order. Note that 1) must come before 2) because without 2), 1) should have no side effects, and then the effects of 2) should become more clear once 1) is in

Note that we don't need [`no-import-type-side-effects` lint rule](https://typescript-eslint.io/rules/no-import-type-side-effects/) because 2 and 3 won't allow us to have inline `type` anywhere

_Originally posted by @pokey in https://github.com/cursorless-dev/cursorless/pull/1832#discussion_r1339948173_

pokey:
_Originally posted by @auscompgeek in https://github.com/cursorless-dev/cursorless/pull/1832/files#r1399814887_:

I still never got a chance to test that matrix, but up until TypeScript 5.3 whether auto-import used `import type` depends on the `tsconfig`. TypeScript 5.3 adds an IDE setting to prefer auto-importing with `import type` https://devblogs.microsoft.com/typescript/announcing-typescript-5-3/#settings-to-prefer-type-auto-imports

The presence of this lint rule will obviously not affect auto-import behaviour since eslint doesn't change the tsconfig. The auto-import behaviour probably isn't too big of an issue anyway since we all have eslint autofix on save, I believe.

PR #2619, opened 2024-08-01
consistent-type-imports rule

AndreasArvidsson:
related #2618
revives https://github.com/cursorless-dev/cursorless/pull/1832

vscode quick fix import and auto import appears to be doing the right thing in regards to type only or runtime imports. Imports organize or imports fix unfortunately do not convert a runtime import to a type import if it's only used as a type. We probably need to rely on the linter to do that part for us.

## Checklist


AndreasArvidsson:
> When you save, shouldn't the eslint fixer correct it tho?

Yes if you have that enabled of course. I was just testing what vscode by itself would do.

phillco (approved):
lgtm, reviewed the base commit and a bit of the automatic changes

tbh kind of surprised this is a linter feature not a compiler one, given the benefits for transpilation

auscompgeek:
> tbh kind of surprised this is a linter feature not a compiler one, given the benefits for transpilation

TypeScript's `verbatimModuleSyntax` config enforces types are only imported with `import type`. The lint rule provides an autofixer though.

PR #2626, opened 2024-08-03
Remove references to left and right in surrounding pair documentation

### LINKED ISSUE #2625: Docs reference deprecated left/right grammar for surrounding pairs
Per discussion with @AndreasArvidsson, these have been esoteric for a while since you can use previous or next, and were explicitly deprecated in https://github.com/cursorless-dev/cursorless/pull/2457. Neither pokey or Andreas used them and they were not referenced in any videos. 

However the documentation still refers to them:

https://www.cursorless.org/docs/#ambiguous-delimiters----etc

> Position your cursor directly next to the delimiter (or just use the delimiter as a mark), and then prefix the name of the delimiter pair with "left" or "right" to force cursorless to expand the selection to the left or right, respectively.
> 
> For example:
> 
> "take left quad" (with your cursor adjacent to a quote)
> "take left pair green double" (with your cursor anywhere)
> "take inside right quad"
> If your cursor / mark is between two delimiters (not adjacent to one), then saying either "left" or "right" will cause cursorless to just expand to the nearest delimiters on either side, without trying to determine whether they are opening or closing delimiters.

AndreasArvidsson:
These are now deprecated and now error out after https://github.com/cursorless-dev/cursorless/pull/2457. We missed this reference in the documentation.

Fixes #2625

## Checklist

### DIFF
--- a/cursorless-talon/src/modifiers/surrounding_pair.py
+++ b/cursorless-talon/src/modifiers/surrounding_pair.py
@@ -9,7 +9,7 @@ ctx = Context()
 
 mod.list(
     "cursorless_delimiter_force_direction",
-    desc="Can be used to force an ambiguous delimiter to extend in one direction",
+    desc="DEPRECATED: Can be used to force an ambiguous delimiter to extend in one direction",
 )
 # FIXME: Remove type ignore once Talon supports list types
 # See https://github.com/talonvoice/talon/issues/654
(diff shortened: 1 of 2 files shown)


phillco (approved):
Removing for now is fine, but should we file an issue to see if we can replace these suggestions with next and previous?

AndreasArvidsson:
I'm not sure we need that; we already have documentation for previous and next and they are not surrounding pair specific. Left and right were just documented here because they only worked with surrounding pairs.

PR #2628, opened 2024-08-04
Flatten scope type list

### LINKED ISSUE #2614: Investigate DFA compilation times
They are up over 700ms for both me and @AndreasArvidsson. We should do another git bisect to see what caused it

AndreasArvidsson:
Brings our dfa compilation time from around 700ms down to about 320ms

Fixes #2614

I had forgot, but we actually do parse `<user.any_alphanumeric_key>` se we can show the users spoken form in the tutorial. I've now utilize this further and actually add the spoken forms for the glyph scope type in the flattened list.

Note that this implementation is somewhat hacky on purpose. Basically I didn't want to touch our csv parser without first talking to pokey since a lot of these list and spoken forms are used in places like the cheat sheet and the tutorial. What I'm instead doing is keeping all the existing lists and then creating a new list that is a flattened version of them. That way the sheet sheet and other places can still use the individual list and we are only using this flattened larger list for the actual scope type capture. This is probably something we want to revisit later, but for now we're getting a huge boast in dfa compilation time with no changes to the speakable grammar and that I think is a clear win.

## Checklist

- [x] Run Talon grammar tests


cursorless-talon-dev/src/default_vocabulary.py
```diff
@@ -96,14 +102,17 @@
...
     **{digits[i]: str(i) for i in range(10)},
     **punctuation_words,
     **symbol_key_words,
 }
 
 
-@ctx.capture("user.any_alphanumeric_key", rule="|".join(any_alphanumeric_keys.keys()))
+@ctx.capture(
+    "user.any_alphanumeric_key",
+    rule="{user.cursorless_default_any_alphanumeric_key}",
```
AndreasArvidsson, line 115: Our capture parser can only deal with lists and not literals
https://github.com/cursorless-dev/cursorless/blob/e9160e95cca55ab8b86ab290acf5f35254cd062f/cursorless-talon/src/get_grapheme_spoken_form_entries.py#L37-L68

cursorless-talon/src/modifiers/modifiers.py
```diff
@@ -27,7 +27,6 @@ def cursorless_simple_modifier(m) -> dict[str, str]:
     "<user.cursorless_simple_scope_modifier>",  # funk, state, class, every funk
     "<user.cursorless_ordinal_scope>",  # first past second word
     "<user.cursorless_relative_scope>",  # next funk, 3 funks
-    "<user.cursorless_surrounding_pair_force_direction>",  # DEPRECATED "left quad" / "right quad"
```
AndreasArvidsson, line 30: This also added to the compilation time and I didn't want to create a second flattened list just for something that's deprecated and no one is using afaik.

cursorless-talon/src/modifiers/scopes.py
```diff
@@ -4,6 +4,25 @@
 
 mod.list("cursorless_scope_type", desc="Supported scope types")
 mod.list("cursorless_scope_type_plural", desc="Supported plural scope types")
+
+mod.list(
```
AndreasArvidsson, line 8: Since we're removing a bunch of captures it was easier to just move the scope lists here and remove those other files

cursorless-talon/src/modifiers/scopes.py
```diff
@@ -13,60 +32,49 @@
...
-    try:
-        return m.cursorless_glyph_scope_type
-    except AttributeError:
-        pass
 
-    return {
-        "type": "customRegex",
-        "regex": m.cursorless_custom_regex_scope_type,
-    }
+@mod.capture(rule="{user.cursorless_scope_type_flattened}")
```
AndreasArvidsson, line 45: This is the change where the performance increase actually comes from. Instead of referencing multiple captures and list we now just have a single list.

cursorless-talon/src/modifiers/scopes.py
```diff
@@ -13,60 +32,49 @@
...
-    try:
-        return {"type": m.cursorless_scope_type_plural}
-    except AttributeError:
-        pass
-
-    try:
-        return m.cursorless_surrounding_pair_scope_type_plural
-    except AttributeError:
-        pass
+    return creates_scope_type(m.cursorless_scope_type_flattened_plural)
```
AndreasArvidsson, line 54: Note that the singular and plural implementation actually unified to a common utility function which is a nice side effect

cursorless-talon/src/spoken_scope_forms.py
```diff
@@ -0,0 +1,76 @@
+from talon import Context
+
+from .csv_overrides import csv_get_ctx, csv_get_normalized_ctx
+
+
+def init_scope_spoken_forms(graphemes_talon_list: dict[str, str]):
+    create_flattened_talon_list(csv_get_ctx(), graphemes_talon_list, True)
+    create_flattened_talon_list(csv_get_normalized_ctx(), graphemes_talon_list, False)
```
AndreasArvidsson, line 8: The normalized context is only used for the Cursorless Talon grammar tests so we can run test without user customization.

cursorless-talon/src/spoken_scope_forms.py
```diff
@@ -0,0 +1,76 @@
...
+        "cursorless_selectable_only_paired_delimiter",
+    )
+    append_talon_lists(
+        ctx,
+        scope_types_singular,
+        scope_types_plural,
+        "surroundingPair",
+        "cursorless_wrapper_selectable_paired_delimiter",
+    )
+    if include_custom_regex:
```
AndreasArvidsson, line 41: The default grammar have no custom regexes and that would throw an error when we couldn't find the list in the register

cursorless-talon/src/get_grapheme_spoken_form_entries.py
```diff
@@ -32,7 +43,7 @@ def generate_lists_from_capture(capture_name) -> Iterator[str]:
     if capture_name.startswith("self."):
         capture_name = "user." + capture_name[5:]
     try:
-        rule = registry.captures[capture_name][0].rule.rule
+        rule = registry.captures[capture_name][-1].rule.rule
```
AndreasArvidsson, line 46: -1/the last item is the active list and of course the one we want to use. This was actually a bug on main that would show incorrect spoken form in the tutorial.

cursorless-talon/src/get_grapheme_spoken_form_entries.py
```diff
@@ -57,19 +68,26 @@ def generate_lists_from_capture(capture_name) -> Iterator[str]:
...
-def get_id_to_spoken_form_map(list_name: str) -> Mapping[str, list[str]]:
+def get_id_to_talon_list(list_name: str) -> dict[str, str]:
     """
-    Given the name of a Talon list, return a mapping from the values in that
-    list to the list of spoken forms that map to the given value.
+    Given the name of a Talon list, return that list
     """
     try:
-        raw_list = typing.cast(dict[str, str], registry.lists[list_name][0]).copy()
+        return typing.cast(dict[str, str], registry.lists[list_name][-1]).copy()
```
AndreasArvidsson, line 76: Same here. Use the active list instead of the first one

cursorless-talon/src/spoken_scope_forms.py
phillco, line 8:
```suggestion
    create_flattened_talon_list(csv_get_ctx(), graphemes_talon_list, include_custom_regex=True)
    create_flattened_talon_list(csv_get_normalized_ctx(), graphemes_talon_list, include_custom_regex=False)
```

phillco, line 8: If we can easily check if we are in the test mode (reading from scope), I would probably put this behind an if statement as a small step towards detangling that code from here

cursorless-talon/src/spoken_scope_forms.py
```diff
@@ -0,0 +1,76 @@
...
+        ctx,
+        scope_types_singular,
+        scope_types_plural,
+        "surroundingPair",
+        "cursorless_surrounding_pair_scope_type",
+    )
+    append_talon_lists(
+        ctx,
+        scope_types_singular,
+        scope_types_plural,
```
phillco, line 30: You can reduce the duplication here by making a dict and iterating through it. It will also make the a little bit different further down more obvious

```
    x = {
        "simple": "cursorless_scope_type",
...
     }
```

cursorless-talon/src/get_grapheme_spoken_form_entries.py
```diff
@@ -57,19 +68,26 @@ def generate_lists_from_capture(capture_name) -> Iterator[str]:
...
     except Error:
         app.notify(f"Error getting list {list_name}")
         return {}
 
+
+def talon_list_to_spoken_form_map(
+    talon_list: dict[str, str],
+) -> Mapping[str, list[str]]:
+    """
+    Given the Talon list, return a mapping from the values in that
```
phillco, line 86:
```suggestion
    Given a Talon list, return a mapping from the values in that
```

phillco (commented):
Looks great! A few suggestions

cursorless-talon/src/spoken_scope_forms.py
```diff
@@ -0,0 +1,54 @@
...
+    if is_cursorless_test_mode():
+        create_flattened_talon_list(
+            csv_get_normalized_ctx(), graphemes_talon_list, include_custom_regex=False
+        )
+
+
+def create_flattened_talon_list(
+    ctx: Context, graphemes_talon_list: dict[str, str], include_custom_regex: bool
+):
+    list_to_merge = {
```
phillco, line 19:
```suggestion
    lists_to_merge = {
```

phillco (approved):

PR #2631, opened 2024-08-04
Added setting to restore system default microphone after Talon grammar tests

AndreasArvidsson:
There is no way of telling if Talon currently is using the system default microphone. Every time I run the Talon grammar tests this setting changes and I have to manually reset it which is really annoying.

## Checklist

### DIFF
--- a/cursorless-talon-dev/src/spoken_form_test.py
+++ b/cursorless-talon-dev/src/spoken_form_test.py
@@ -13,7 +13,7 @@ mod.mode(
 mod.setting(
     "cursorless_spoken_form_test_restore_microphone",
     str,
-    desc="The microphone to revert to after the spoken form tests are done. If not provided the previously active microphone will be restored.",
+    desc="The microphone to switch to after the spoken form tests are done. If unset, the microphone that was active before tests started is restored. (If you want to switch back to 'System Default', you should set that as the value here)",
 )
 
 ctx = Context()
--- a/cursorless-talon-dev/src/spoken_form_test.py
+++ b/cursorless-talon-dev/src/spoken_form_test.py
@@ -1,13 +1,20 @@
 import json
 from typing import Any, Optional
 
-from talon import Context, Module, actions, scope
+from talon import Context, Module, actions, scope, settings
 
 mod = Module()
 
 mod.mode(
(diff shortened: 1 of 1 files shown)


cursorless-talon-dev/src/spoken_form_test.py
```diff
@@ -10,6 +10,13 @@
     "Used to run tests on the Cursorless spoken forms/grammar",
 )
 
+mod.setting(
```
phillco, line 13:
```suggestion
# This setting is necessary because we can't detect if you're using "System Default", as `actions.sound.active_microphone()` will return what microphone that is pointing to.
mod.setting(
```

cursorless-talon-dev/src/spoken_form_test.py
```diff
@@ -72,7 +79,12 @@ def private_cursorless_spoken_form_test_mode(enable: bool):  # pyright: ignore [
 
         if enable:
             saved_modes = scope.get("mode")
-            saved_microphone = actions.sound.active_microphone()
+            if settings.get(
+                "user.cursorless_spoken_form_test_restore_default_microphone"
+            ):
+                saved_microphone = "System Default"
+            else:
+                saved_microphone = actions.sound.active_microphone()
```
phillco, line 87: Per vc discussion, I think it's simpler to just have a string setting that the user can set to "system default" (or any other microphone), rather than a boolean that hardcodes a reference to "system default"

cursorless-talon-dev/src/spoken_form_test.py
```diff
@@ -10,6 +10,12 @@
     "Used to run tests on the Cursorless spoken forms/grammar",
 )
 
+mod.setting(
+    "cursorless_spoken_form_test_restore_microphone",
+    str,
+    desc="The microphone to revert to after the spoken form tests are done. If not provided the previously active microphone will be restored.",
```
phillco, line 16:
```suggestion
    desc="The microphone to switch to after the spoken form tests are done. If unset, the microphone that was active before tests started is restored. (If you want to switch back to 'System Default', you should set that as the value here)",
```

phillco (approved):
Cleaner!

PR #2637, opened 2024-08-05
Fixed bug where we tried to read missing custom regex list

AndreasArvidsson:
## Checklist

### DIFF
--- a/cursorless-talon/src/get_grapheme_spoken_form_entries.py
+++ b/cursorless-talon/src/get_grapheme_spoken_form_entries.py
@@ -14,16 +14,6 @@ grapheme_capture_name = "user.any_alphanumeric_key"
 def get_grapheme_spoken_form_entries(
     grapheme_talon_list: dict[str, str],
 ) -> list[SpokenFormOutputEntry]:
-    if grapheme_capture_name not in registry.captures:
-        # We require this capture, and expect it to be defined. We want to show a user friendly error if it isn't present (usually indicating a problem with their community.git setup) and we think the user is going to use Cursorless.
-        # However, sometimes users use different dictation engines (Vosk, Webspeech) with entirely different/smaller grammars that don't have the capture, and this code will run then, and falsely error. We don't want to show an error in that case because they don't plan to actually use Cursorless.
-        if "en" in scope.get("language", {}):
-            app.notify(f"Capture <{grapheme_capture_name}> isn't defined")
-            print(
-                f"Capture <{grapheme_capture_name}> isn't defined, which is required by Cursorless. Please check your community setup"
-            )
-        return []
-
     return [
         {
             "type": "grapheme",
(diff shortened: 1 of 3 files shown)


cursorless-talon/src/get_grapheme_spoken_form_entries.py
```diff
@@ -37,6 +27,16 @@ def get_grapheme_spoken_form_entries(
 
 
 def get_graphemes_talon_list() -> dict[str, str]:
+    if grapheme_capture_name not in registry.captures:
```
AndreasArvidsson, line 30: After the merge of literal mark this if statement ended up in the wrong place. It should of course be here where we actually read it from the registry.

cursorless-talon/src/spoken_forms.py
```diff
@@ -125,6 +125,7 @@ def handle_new_values(csv_name: str, values: list[SpokenFormEntry]):
         if initialized:
             # On first run, we just do one update at the end, so we suppress
             # writing until we get there
+            init_scope_spoken_forms(graphemes_talon_list)
```
AndreasArvidsson, line 128: Without this we don't regenerate the flattened list whenever a csv file changes

cursorless-talon/src/spoken_scope_forms.py
```diff
@@ -4,25 +4,20 @@
...
+def create_flattened_talon_list(ctx: Context, graphemes_talon_list: dict[str, str]):
     lists_to_merge = {
         "cursorless_scope_type": "simple",
         "cursorless_surrounding_pair_scope_type": "surroundingPair",
         "cursorless_selectable_only_paired_delimiter": "surroundingPair",
         "cursorless_wrapper_selectable_paired_delimiter": "surroundingPair",
     }
-    if include_custom_regex:
+    # If the user have no custom regex scope type, then that list is missing from the context
+    if "user.cursorless_custom_regex_scope_type" in ctx.lists.keys():  # noqa: SIM118
```
AndreasArvidsson, line 20: This is really weird. ruff wants me to remove `.keys()` but if I do this condition is always false. I wonder if this has anything to do with how Talon implements the lists/dict?

pokey: yeah SIM118 is fairly aggressive imo. Fine with a noqa but if it keeps biting us we might consider just shutting it off

pokey (approved):
it's a bit surprising that the list just doesn't get created at all if it's empty, but best to get this fix in and can re-evaluate that after

AndreasArvidsson:
> it's a bit surprising that the list just doesn't get created at all if it's empty, but best to get this fix in and can re-evaluate that after

Yeah that was a bit surprising. I think we should have a real look over the entire csv handling and usage of list when you are back.

cursorless-talon/src/spoken_scope_forms.py
AndreasArvidsson: I think it should be equivalent generally in python and this is something Talon is doing

Issue #2538, opened 2024-07-19
OSX tests sometimes temporarily fail on CI


saidelike:
macos tests fails randomly sometimes on CI. 

![image](https://github.com/user-attachments/assets/eb4cf3ee-8c64-47db-b22a-c9671c2bab69)

Pokey tried bumping the sleep but hasn’t been able to get it to pass reliably. The workaround in the meantime is to re-run the job.

An example is the failed job in [attempt 1](https://github.com/saidelike/cursorless/actions/runs/10006602088/job/27659621847?pr=6) which when we re-run the job works for [attempt 2](https://github.com/saidelike/cursorless/actions/runs/10006602088/job/27659954809?pr=6).

pokey:
We should probably use a fake `talonSpokenForms` component here https://github.com/cursorless-dev/cursorless/blob/a09177bcc492529d943ab92dfc43d3abe3a84bda/packages/cursorless-vscode/src/extension.ts#L98

AndreasArvidsson:
Agreed

PR #2638, opened 2024-08-05
Don't use file system for custom spoken form tests

AndreasArvidsson:
https://github.com/cursorless-dev/cursorless/issues/2538

This should help with the ci tests failing on mac

## Checklist


pokey:
No you'd need to create a new class that you use for mocking

AndreasArvidsson:
> No you'd need to create a new class that you use for mocking. Then you return that class in test helpers and tell it what custom spoken forms should be rather than using file system

I understood that, but that tests now are writing to the file system so we need to do a bit more than just create a new class. We also need to update the tests and then comes the question are we just now testing our mocks?

AndreasArvidsson:
I will have a look at it when I have some spare time

packages/common/src/ide/fake/FakeTalonSpokenForms.ts
```diff
@@ -0,0 +1,25 @@
...
+} from "../../types/TalonSpokenForms";
+import { Notifier } from "../../util/Notifier";
+
+export class FakeTalonSpokenForms implements TalonSpokenForms {
+  private notifier = new Notifier();
+  private entries: SpokenFormEntry[] | null = null;
+
+  async getSpokenFormEntries(): Promise<SpokenFormEntry[]> {
+    if (this.entries == null) {
+      throw new NeedsInitialTalonUpdateError("FakeTalonSpokenForms");
```
AndreasArvidsson, line 14: This is what the file system based implementation does if the json file doesn't exist. The custom spoken form implementation has a special case for this error where it defaults to spoken forms without showing a user error.

packages/cursorless-vscode-e2e/src/suite/scopeProvider/runCustomRegexScopeInfoTest.ts
```diff
@@ -26,30 +28,17 @@ export async function runCustomRegexScopeInfoTest() {
...
 
-    await writeFile(
-      cursorlessTalonStateJsonPath,
-      JSON.stringify(spokenFormJsonContents),
-    );
+    talonSpokenForms.mockSpokenFormEntries(spokenFormEntries);
     await assertCalledWithScopeInfo(fake, unsupported);
 
     await openNewEditor(contents);
+    await sleep(10);
```
AndreasArvidsson, line 35: Added a short sleep to just make sure that the document open event had triggered which is one of the events the scope provider uses to update available scopes.

packages/cursorless-vscode-e2e/src/suite/scopeProvider/runCustomRegexScopeInfoTest.ts
```diff
@@ -59,17 +48,13 @@ hello world
 
 const regex = "[a-zA-Z]+";
 
-const spokenFormJsonContents = {
-  version: 0,
```
AndreasArvidsson, line 63: The version is only used by the file system implementation

packages/cursorless-vscode/src/extension.ts
```diff
@@ -76,25 +77,29 @@ export async function activate(
   const parseTreeApi = await getParseTreeApi();
 
   const { vscodeIDE, hats, fileSystem } = await createVscodeIde(context);
+  const isTesting = vscodeIDE.runMode === "test";
```
AndreasArvidsson, line 80: We're doing this exact check in multiple places I think storing it to a variable increases readability

AndreasArvidsson:
Updated with fake spoken forms

packages/cursorless-vscode-e2e/src/suite/scopeProvider/runCustomRegexScopeInfoTest.ts
phillco: I would document that in a comment above the sleep, since the PR comment will be lost to time

packages/cursorless-vscode-e2e/src/suite/scopeProvider/runCustomRegexScopeInfoTest.ts
```diff
@@ -26,30 +28,17 @@ export async function runCustomRegexScopeInfoTest() {
...
 
-    await writeFile(
-      cursorlessTalonStateJsonPath,
-      JSON.stringify(spokenFormJsonContents),
-    );
+    talonSpokenForms.mockSpokenFormEntries(spokenFormEntries);
     await assertCalledWithScopeInfo(fake, unsupported);
 
     await openNewEditor(contents);
+    await sleep(100);
```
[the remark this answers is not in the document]
phillco: Same as above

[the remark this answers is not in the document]
AndreasArvidsson: Done

phillco (commented):
Since the sleep seems to be the thing that fixes the issue, let's split that into a smaller we can land right away, and the change in memory files as a separate we can discuss separately/later

AndreasArvidsson:
Separate pull request with just the sleep
https://github.com/cursorless-dev/cursorless/pull/2645

AndreasArvidsson:
Closing this

PR #2639, opened 2024-08-05
Use containing scope stage in preferred scope stage

AndreasArvidsson:
ContainingScopeStage have some logic around how to treat legacy scopes that we were missing from PreferredScopeStage. 

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/PreferredScopeStage.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/PreferredScopeStage.ts
@@ -35,7 +35,7 @@ export class PreferredScopeStage implements ModifierStage {
     try {
       return containingScopeStage.run(target);
     } catch (ex) {
-      // NoContainingScopeError is thrown if no containing scope was found.
+      // NoContainingScopeError is thrown if no containing scope was found, which is fine.
       if (!(ex instanceof NoContainingScopeError)) {
         throw ex;
       }
--- a/packages/cursorless-engine/src/processTargets/modifiers/PreferredScopeStage.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/PreferredScopeStage.ts
@@ -2,7 +2,7 @@ import { type Position, type PreferredScopeModifier } from "@cursorless/common";
 import type { Target } from "../../typings/target.types";
 import type { ModifierStageFactory } from "../ModifierStageFactory";
 import type { ModifierStage } from "../PipelineStages.types";
-import { getContainingScopeTarget } from "./getContainingScopeTarget";
+import { ContainingScopeStage } from "./ContainingScopeStage";
 import type { TargetScope } from "./scopeHandlers/scope.types";
 import type { ScopeHandler } from "./scopeHandlers/scopeHandler.types";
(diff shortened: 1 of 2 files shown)


packages/cursorless-engine/src/processTargets/modifiers/PreferredScopeStage.ts
```diff
@@ -22,6 +22,18 @@ export class PreferredScopeStage implements ModifierStage {
...
+    const containingScopeStage = new ContainingScopeStage(
+      this.modifierStageFactory,
+      this.scopeHandlerFactory,
+      { type: "containingScope", scopeType },
+    );
+
+    try {
+      return containingScopeStage.run(target);
+    } catch (_ex) {
+      // Do nothing
```
phillco, line 34: Can we clarify why it's okay to swallow this here?

Otherwise we should just remove the comment, because it's redundant to the empty catch block

phillco (commented):
lgtm, but I'd like to talk through it in the meetup too just so I know all the background. I'll approve after that

packages/cursorless-engine/src/processTargets/modifiers/PreferredScopeStage.ts
AndreasArvidsson: `Empty block statement.eslint[no-empty](https://eslint.org/docs/latest/rules/no-empty)`
But sure I can elaborate

packages/cursorless-engine/src/processTargets/modifiers/PreferredScopeStage.ts
```diff
@@ -22,6 +22,18 @@ export class PreferredScopeStage implements ModifierStage {
...
+    const containingScopeStage = new ContainingScopeStage(
+      this.modifierStageFactory,
+      this.scopeHandlerFactory,
+      { type: "containingScope", scopeType },
+    );
+
+    try {
+      return containingScopeStage.run(target);
+    } catch (_ex) {
+      // Do nothing. We'll try the closest scope next.
```
phillco, line 34:
```suggestion
      // This is expected if one isn't found. We'll move onto closest scope next.
```

phillco: If there's a more specific exception class that we can catch, that would be preferable, since it would make the exception more readable / the comment redundant (although we still need it because of the linter :P), but also prevent us from wrongly catching something else

phillco (approved):

packages/cursorless-engine/src/processTargets/modifiers/PreferredScopeStage.ts
```diff
@@ -22,6 +26,21 @@ export class PreferredScopeStage implements ModifierStage {
...
+    const containingScopeStage = new ContainingScopeStage(
+      this.modifierStageFactory,
+      this.scopeHandlerFactory,
+      { type: "containingScope", scopeType },
+    );
+
+    try {
+      return containingScopeStage.run(target);
+    } catch (ex) {
+      // NoContainingScopeError is thrown if no containing scope was found.
```
phillco, line 38:
```suggestion
      // NoContainingScopeError is thrown if no containing scope was found, which is fine.
```

phillco (approved):

Issue #2635, opened 2024-08-04
Pokey "return from leave" discussion list


phillco:
Pokey needs to do:

- [x] https://github.com/cursorless-dev/vscode-parse-tree/issues/88

Discussion:

- [ ] https://github.com/cursorless-dev/cursorless/pull/2628, "Basically I didn't want to touch our csv parser without first talking to pokey since a lot of these list and spoken forms are used in places like the cheat sheet and the tutorial."
- [ ] https://github.com/cursorless-dev/cursorless/pull/2638 Is this still worth doing even if it doesn't affect test reliability?
- [ ] https://github.com/cursorless-dev/cursorless/pull/2651 - do we want to improve how this would handle multiple separate targets? Is that even desirable?

PR #2651, opened 2024-08-09
Added the ability to remove whitespaces with `join token`

AndreasArvidsson:
## Checklist

### DIFF
--- a/changelog/2024-08-joinTokens.md
+++ b/changelog/2024-08-joinTokens.md
@@ -3,4 +3,4 @@ tags: [enhancement]
 pullRequest: 2651
 ---
 
-Added the ability to remove whitespaces with join token. `"join two tokens air"` will remove the whitespace between the tokens instead of joining together two lines.
+`join token` now behaves as expected. `"join two tokens air"` will now remove the whitespace between the tokens; previously, it simply joined together their lines, which was usually not expected.
--- a/changelog/2024-08-joinTokens.md
+++ b/changelog/2024-08-joinTokens.md
@@ -1,0 +1,6 @@
+---
+tags: [enhancement]
+pullRequest: 2651
+---
+
+Added the ability to remove whitespaces with join token. `"join two tokens air"` will remove the whitespace between the tokens instead of joining together two lines.
--- a/packages/cursorless-engine/src/actions/JoinLines.ts
+++ b/packages/cursorless-engine/src/actions/JoinLines.ts
@@ -1,39 +1,53 @@
 import type { Edit, TextEditor } from "@cursorless/common";
-import { FlashStyle, Range } from "@cursorless/common";
+import { FlashStyle, Range, zipStrict } from "@cursorless/common";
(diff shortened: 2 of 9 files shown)


packages/cursorless-engine/src/typings/target.types.ts
```diff
@@ -54,6 +54,9 @@ export interface Target {
   /** If true this target should be treated as a word */
   readonly isWord: boolean;
 
+  /** If true this target should be treated as a line when using the join action */
```
phillco, line 57:
```suggestion
  /** If true this target should be treated as a line when using the join action; otherwise a token */
```

phillco: Convert to a type checked string or enum type?

packages/cursorless-engine/src/actions/JoinLines.ts
```diff
@@ -42,28 +56,59 @@ export default class JoinLines {
...
-        isReplace: true,
-      });
-    }
+    edits.push(...targetsEdits);
   }
 
   return edits;
 }
+
+function getTokenTargetEditEdits(target: Target): Edit[] {
```
phillco, line 73: editedits?

packages/cursorless-engine/src/actions/JoinLines.ts
```diff
@@ -42,28 +56,59 @@ export default class JoinLines {
...
+    "forward",
+  );
+
+  return Array.from(pairwise(matches)).map(
+    ([range1, range2]): Edit => ({
+      range: new Range(range1.end, range2.start),
+      text: "",
+      isReplace: true,
+    }),
+  );
```
phillco, line 89: Check the text between the tokens and verify that it's actually just whitespace. Otherwise if you try to join tokens from two discrete ranges the document text in the middle will get clobbered. 

Update: this is not an issue because it only operates on the target, and those would be separate targets. This means we have the reverse problem if you specify multiple targets, but this already affects existing commands, like sort. We should probably file an issue for that but it shouldn't block this change

phillco (changes requested):
- [x] Change from boolean to enum or string key
- [x] Let's try to find that existing issue about `sort` with multiple targets so we can link it - added to https://github.com/cursorless-dev/cursorless/issues/2635

phillco (approved):

changelog/2024-08-joinTokens.md
```diff
@@ -0,0 +1,6 @@
+---
+tags: [enhancement]
+pullRequest: 2651
+---
+
+Added the ability to remove whitespaces with join token. `"join two tokens air"` will remove the whitespace between the tokens instead of joining together two lines.
```
phillco, line 6:
```suggestion
`join token` now behaves as expected. `"join two tokens air"` will now remove the whitespace between the tokens; previously, it simply joined together the lines containing them, which was usually not expected.
```

phillco (approved):
lgtm

PR #2652, opened 2024-08-10
Change default head/tail scope to be line bounded by matching pairs

### LINKED ISSUE #2434: Change default nested modifier for tail to stop at "inside"
Instead of defaulting to "line", it would stop before it hits a closing delimiter

Start with it behind a feature flag

AndreasArvidsson:
If no scope is specified for head/tail instead of defaulting to just line we now use the smallest of line or surrounding pair interior. This means that `"take head"` would be bounded by the surrounding pair just like `"short paint"`. If there is no surrounding pair the line is used just like today. You can also override this behavior by specifying line. `"take head line"` would always use the line regardless if you are in a surrounding pair or not.

Fixes #2434

The issue mentions a tag for this. I guess since this change is fully extension side it would be a vscode setting, but in tempted to just have this be enabled by default since it's a really useful feature.

## Checklist

### DIFF
--- a/changelog/2024-08-boundedHeadTail.md
+++ b/changelog/2024-08-boundedHeadTail.md
@@ -3,4 +3,4 @@ tags: [enhancement]
 pullRequest: 2652
 ---
 
-If no scope is specified for head/tail instead of defaulting to just line we now use the smallest of line or surrounding pair interior. This means that "take head" would be bounded by the surrounding pair just like "short paint". If there is no surrounding pair the line is used just like today. You can also override this behavior by specifying line. "take head line" would always use the line regardless if you are in a surrounding pair or not.
+If you're inside of a surrounding pair, the "head" and "tail" modifiers now only expand to the interior of that pair, instead of the whole line. This is more useful, akin to how "short paint" works. You can explicitly say "head line" or "tail line" to get the old behavior.
--- a/changelog/2024-08-boundedHeadTail.md
+++ b/changelog/2024-08-boundedHeadTail.md
@@ -1,0 +1,6 @@
+---
+tags: [enhancement]
+pullRequest: 2652
+---
+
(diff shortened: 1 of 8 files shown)


phillco:
Do we have a feeling for how many users would be surprised by change to the `take head` / `take tail` default behavior?

phillco:
I misunderstood, in my head I was thinking of `pre line` or just the commonly defined `head` or `tail` commands to go to the beginning or the end the line, which won't be affected by this.

I agree it's much more likely that people will want to select up to the bounding of a surrounding pair if they're in one, rather than the beginning of the line. Andreas has mentioned the use case of inside of a Python docstring, which is a good example. The fact that they can still use `take head line` to get the original behavior seems like a reasonable compromise if anyone is annoyed by that.

As a bonus you can also now say `pre head` to go to the beginning of either the surrounding pair interior or the line, which now behaves more like a smart selector, instead of having to think about whether you want `pre line` or `pre core` (interior, etc)

phillco (approved):

phillco:
Also note: it won't always select the surrounding pair; it will select the containing scope that's smaller. So if the surrounding pair covers multiple lines the line will always be smaller.

phillco:
Let's add this to release notes because it's a useful change

changelog/2024-08-boundedHeadTail.md
```diff
@@ -0,0 +1,6 @@
+---
+tags: [enhancement]
+pullRequest: 2652
+---
+
+If no scope is specified for head/tail instead of defaulting to just line we now use the smallest of line or surrounding pair interior. This means that "take head" would be bounded by the surrounding pair just like "short paint". If there is no surrounding pair the line is used just like today. You can also override this behavior by specifying line. "take head line" would always use the line regardless if you are in a surrounding pair or not.
```
phillco, line 6:
```suggestion

If you're inside of a surrounding pair, the "head" and "tail" modifiers now only expand to the interior of that pair, instead of the whole line. This is more useful, akin to how "short paint" works. You can explicitly say "head line" or "tail line" to get the old behavior.
```

phillco (approved):

PR #2653, opened 2024-08-12
Add `.DS_Store` to `.gitignore` for `cursorless-talon`

bentitmus:
When this is synced to [`cursorless-talon`](https://github.com/cursorless-dev/cursorless-talon) the `.gitignore` in the monorepo doesn't apply so copy the `.DS_Store` ignore for those on macOS to prevent spurious `untracked content` messages in `git`.

I don't think the checklist below applies in this case.

## Checklist

### DIFF
--- a/cursorless-talon/.gitignore
+++ b/cursorless-talon/.gitignore
@@ -1,3 +1,4 @@
 *.flac
 data/
 .vscode/settings.json
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ node_modules
 .vscode-test/
 *.vsix
 /package-lock.json
-*.DS_Store
+.DS_Store
 meta.json
 .luacheckcache
 


AndreasArvidsson (approved):

cursorless-talon/.gitignore
```diff
@@ -1,3 +1,4 @@
 *.flac
 data/
 .vscode/settings.json
+*.DS_Store
```
phillco, line 4:
```suggestion
.DS_Store
```

I think this will work as well, no? The most common gitignores I've seen just use `.DS_Store` since we only need to match the literal name of the file.

phillco (commented):
Small nit otherwise lgtm

cursorless-talon/.gitignore
phillco: Ah, I guess this comes from here: https://github.com/cursorless-dev/cursorless/blob/main/.gitignore#L8

We should probably fix it there as well.

bentitmus: Agreed that it should be fixed. I copied what was there in the existing `.gitignore` but prefer without the unnecessary wildcard. Personally I tend to generate `.gitignore` files using something like [gitignore.io](https://www.toptal.com/developers/gitignore/) which cover files you might not have thought of, although sometimes too many. In any case this suggests `.DS_Store` for `macOS`.

phillco: Do you want to just make that change as part of this PR? If so we can stamp and merge this one pretty quickly. If you're not comfortable I can do that as a separate change, but I don't think it's contentious.

bentitmus: Sorry. Done. Force pushed the update.

phillco (approved):

phillco:
Thanks!

PR #2655, opened 2024-08-16
Show better exception message when vscode settings json is not found

AndreasArvidsson:
Today you get a `ValueError: max() arg is an empty sequence` message that is quite confusing.

## Checklist

### DIFF
--- a/cursorless-talon/src/apps/vscode_settings.py
+++ b/cursorless-talon/src/apps/vscode_settings.py
@@ -69,7 +69,7 @@ def pick_path(paths: list[Path]) -> Path:
     existing_paths = [path for path in paths if path.exists()]
     if not existing_paths:
         raise FileNotFoundError(
-            f"Couldn't find vscode settings json in the following paths: {paths}"
+            f"Couldn't find VSCode's settings JSON. Tried these paths: {paths}"
         )
     return max(existing_paths, key=lambda path: path.stat().st_mtime)
 
--- a/cursorless-talon/src/apps/vscode_settings.py
+++ b/cursorless-talon/src/apps/vscode_settings.py
@@ -67,6 +67,10 @@ class Actions:
 
 def pick_path(paths: list[Path]) -> Path:
     existing_paths = [path for path in paths if path.exists()]
+    if not existing_paths:
+        raise FileNotFoundError(
+            f"Couldn't find vscode settings json in the following paths: {paths}"
+        )
     return max(existing_paths, key=lambda path: path.stat().st_mtime)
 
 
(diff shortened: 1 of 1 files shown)


cursorless-talon/src/apps/vscode_settings.py
```diff
@@ -67,6 +67,10 @@ def vscode_get_setting_with_fallback(
 
 def pick_path(paths: list[Path]) -> Path:
     existing_paths = [path for path in paths if path.exists()]
+    if not existing_paths:
+        raise FileNotFoundError(
+            f"Couldn't find vscode settings json in the following paths: {paths}"
```
phillco, line 72:
```suggestion
            f"Couldn't find VSCode's settings JSON (tried these paths: {', '.join(paths)})"
```

phillco (approved):

cursorless-talon/src/apps/vscode_settings.py
AndreasArvidsson: You don't need to do a join; python interpolates list of strings.

phillco: It's nicer to omit the array literals and wrapping strings if they're not necessary, imo.

```
>>> print(', '.join(["a", "b", "c"]))
a, b, c
>>> print(str(["a", "b", "c"]))
['a', 'b', 'c']
```

AndreasArvidsson: I could go either ways if you prefer that let's go with it. But this isn't a list of strings its list of paths which gives me the following error.
`Argument of type "list[Path]" cannot be assigned to parameter "iterable" of type "Iterable[str]" in function "join"`

phillco: Cast them individually to strings?

phillco:
```
>>> ', '.join(Path('.').iterdir())
Traceback (most recent call last):
  File "<console>", line 1, in <module>
TypeError: sequence item 0: expected str instance, PosixPath found
>>> ', '.join([str(x) for x in Path('.').iterdir()])
'talon.logecho, .DS_Store, recordings, draconity.toml, bin, plugins, old-logs, talon.log.bak, talon.log.2, user, user.old, talon.log.3, recordings-bad, talon.log.1.bak, calib.bin, .venv, .sys, talon.log, talon.log.1'
```

phillco: I think at this point it's more verbose than it's worth, but the error you cited is easily overcome, no?

AndreasArvidsson: I have now updated with a list comprehension

phillco (approved):

PR #2659, opened 2024-09-08
Don't use multiline pairs for head/tail modifier.

AndreasArvidsson:
`change tail` did the wrong thing in this case
```py
"""|foo
 bar
"""
```

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairInteriorScopeHandler.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairInteriorScopeHandler.ts
@@ -45,17 +45,19 @@ export class SurroundingPairInteriorScopeHandler extends BaseScopeHandler {
     );
 
     for (const scope of scopes) {
-      if (!this.scopeType.requireSingleLine || scope.domain.isSingleLine) {
-        yield {
-          editor,
-          domain: scope.domain,
-          getTargets(isReversed) {
-            return scope
-              .getTargets(isReversed)
-              .flatMap((target) => target.getInterior()!);
-          },
-        };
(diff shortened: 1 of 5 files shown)


AndreasArvidsson:
@saidelike Looks like our neovim lua tests are failing
https://github.com/cursorless-dev/cursorless/blob/49f70840f0b11d4fca1744f88af1cd3d91cbd3d5/.github/actions/test-neovim-lua/action.yml

Here is a suggested solution. Do you want to look at this?
https://github.com/leafo/gh-actions-lua/issues/49

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairInteriorScopeHandler.ts
```diff
@@ -32,28 +31,31 @@ export class SurroundingPairInteriorScopeHandler extends BaseScopeHandler {
...
-      }),
+    const scopes = this.surroundingPairScopeHandler.generateScopes(
+      editor,
+      position,
+      direction,
+      hints,
     );
+
+    for (const scope of scopes) {
+      if (!this.scopeType.requireSingleLine || scope.domain.isSingleLine) {
```
phillco, line 48: This reads a touch easier to me.
```suggestion
      if (scope.domain.isSingleLine || !this.scopeType.requireSingleLine) {
```

phillco (approved):
Thanks, sorry I missed this.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/SurroundingPairInteriorScopeHandler.ts
AndreasArvidsson: I think the intent is clearer the way I wrote it, but it's not to important either way.

PR #2673, opened 2024-11-05
Add declare to class fields in child class that overrides a base clas…

bjaspan:
…s field without explicitly reinitializing the field

According to an internal Google audit of TS code: When public class fields are enabled and target output is set to ES2022, TypeScript reports the following error -

```
class A {
    x: SomeType = new SomeType();
}

class B extends A {

    override x!: NarrowType;      // Property 'x' will overwrite the base property in 'A'. If this is intentional, add an initializer. Otherwise, add a 'declare' modifier or remove the redundant declaration.

}
```


pokey (approved):
Seems fine to me. Any idea how we could catch this one in CI?

PR #2675, opened 2024-11-09
Add item scope to JS

AndreasArvidsson:
```js
const value = [1, 2];
```
This example is problematic. Currently we test the language specific implementation before the text based one. This means that if your cursor is inside the curly brackets `item` is `value = [1, 2]`. I don't think this can be solved until we have a better `oneOf` implementation of collection item.
Same problem with Java https://github.com/cursorless-dev/cursorless/pull/2674

To fix that above I made a change so that the text based item can be used even if there is a syntax tree one available if it's smaller. This needs to merge before the java pr. 

## Checklist


pokey (approved):
Seems good for now. I wonder if we have other examples we can use to test this?

AndreasArvidsson:
> Seems good for now. I wonder if we have other examples we can use to test this?

Good question. Got anything in mind?

pokey:
No, I couldn't come up with anything 🤷‍♂️

PR #2681, opened 2024-11-16
Proper handing of raw strings in rust

### LINKED ISSUE #2678: rust raw string breaks scopes
If a [rust raw string](https://doc.rust-lang.org/reference/tokens.html#raw-string-literals) (e.g. `r#"foobar"#`) is present anywhere in a rust file, cursorless scopes stop working with the error: `Cannot read properties of undefined (reading 'endPosition')`

AndreasArvidsson:
Fixes #2678

## Checklist

### DIFF
--- a/data/fixtures/scopes/rust/string.singleLine.scope
+++ b/data/fixtures/scopes/rust/string.singleLine.scope
@@ -1,0 +1,20 @@
+r#"hello"#
+"world"
+---
+
+[#1 Content] =
+[#1 Removal] =
+[#1 Domain] = 0:0-0:10
+  >----------<
+0| r#"hello"#
+
+[#1 Insertion delimiter] = " "
+
+
+[#2 Content] =
+[#2 Removal] =
+[#2 Domain] = 1:0-1:7
+  >-------<
+1| "world"
+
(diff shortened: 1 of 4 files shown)


auscompgeek (changes requested):
Could we also have a test for more hashes, like `r##"hello"##`?

AndreasArvidsson:
> Could we also have a test for more hashes, like `r##"hello"##`?

Oh that is valid? I did not add support for that.

auscompgeek:
> Oh that is valid?

Per [the language reference](https://doc.rust-lang.org/reference/tokens.html#raw-string-literals) (also linked in the issue), yes :)

AndreasArvidsson:
> > Oh that is valid?
> 
> Per [the language reference](https://doc.rust-lang.org/reference/tokens.html#raw-string-literals) (also linked in the issue), yes :)

Tree sitter gives us no help here so all those different combinations with hashes and backslashes is something we need to solve with regex in that case.

AndreasArvidsson:
Now we support multiple hashes :)

auscompgeek (approved):

pokey (approved):

Issue #1052, opened 2022-10-21
Migrate remaining text based scope types to new scope handler setup


pokey:
We'd like to migrate all of our scope types to use the new scope handler setup introduced in #1031.  See

- existing scope handlers such as `TokenScopeHandler`,
- the `ScopeHandler` interface definition, and
- the `NestedScopeHandler` class, which might be of use

## Scope types to migrate

- [x] #1053
- [x] #1054
- [x] #1055
- [x] #1056
- [x] #1057
- [x] #1058
- [x] #1059
- [x] #1060
- [x] #1061

AndreasArvidsson:
@phillco @pokey 

I'm looking into migrating the notebook cell scope and just realize that we are actually using `notebookCell` in markdown.
https://github.com/cursorless-dev/cursorless/blob/8cd7c211930e1ad252eaf399cb38bab7900bcc68/queries/markdown.scm#L54-L63

The problem of course is that you can have markdown cells in a python notebook. What would `"take next cell"` do in a markdown cell?

AndreasArvidsson:
But what if you are not in a code block in markdown? Should `next cell` refer to a markdown code block further down or a next notebook cell?

Implementation wise I think it simplest to just ask the language scope provider and if we get no results go for a notebook cell. Does that sound good?

pokey:
I think the consistent thing to do if we're treating it as nested scope would be to go to next markdown code block if we find one in current cell, otherwise next cell

Tho I do think that might be surprising, which kinda argues they should really be different scopes

AndreasArvidsson:
I think we will get that behavior if we do as I specified above.

But a different scope for notebook cells is of course another option. Maybe we should start with a combined one and see if anyone runs into problem?

phillco:
All of these are done.

Issue #1770, opened 2023-08-12
improve comment/node association


josharian:
We could start by stealing some of the [Go heuristics](https://pkg.go.dev/go/ast#NewCommentMap), which definitely have some shortcomings but work well enough for common cases in practice:

> A CommentGroup represents a sequence of comments with no other tokens and no empty lines between.

> A comment group g is associated with a node n if:
>
> * g starts on the same line as n ends
> * g starts on the line immediately following n, and there is at least one empty line after g and before the next node
> * g starts before n and is not associated to the node before n via the previous rules
>
> [It] tries to associate a comment group to the "largest" node possible: For instance, if the comment is a line comment trailing an assignment, the comment is associated with the entire assignment rather than just the last operand in the assignment.

pokey:
Use cases:

- `"item"` scope could include the comment associated with the item
- Other scopes could include their associated comments

PR #2683, opened 2024-11-19
Collection item scope provider

### LINKED ISSUE #1059: Migrate `collectionItem` to use a scope handler

AndreasArvidsson:
Fixes #1059

Related https://github.com/cursorless-dev/cursorless/issues/1052

## Checklist


AndreasArvidsson:
@pokey The nested loops are gone and performance is now improved.

AndreasArvidsson:
Needs to be updated after https://github.com/cursorless-dev/cursorless/pull/2700 has merged

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/CollectionItemScopeHandler/collectionItemIterationScopeHandler.ts
```diff
@@ -0,0 +1,24 @@
...
+import type { ComplexScopeType } from "../scopeHandler.types";
+import { separatorRegex } from "./getSeparatorOccurrences";
+
+export const collectionItemIterationScopeHandler: ComplexScopeType = {
+  type: "fallback",
+  scopeTypes: [
+    {
+      type: "surroundingPairInterior",
+      delimiter: "collectionBoundary",
+    },
```
phillco, line 12: NTS: This will probably be the majority case because most lists are indeed inside of a surrounding pair somewhere

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/CollectionItemScopeHandler/collectionItemIterationScopeHandler.ts
```diff
@@ -0,0 +1,24 @@
...
+      type: "conditional",
+      scopeType: {
+        type: "line",
+      },
+      predicate: (scope: TargetScope) => {
+        const text = scope.editor.document.getText(scope.domain);
+        return testRegex(separatorRegex, text);
+      },
+    },
+  ],
```
phillco, line 23: NTS: This is triggered for cases like having a comma separated list inside of a comment or prose without a bounding pair.

This does mean that you cannot iterate over lists that span multiple lines if they aren't in a surrounding pair of some sort (say in plaintext)

phillco (commented):
.

data/fixtures/recorded/languages/rust/changeItemOne.yml
```diff
@@ -23,7 +23,7 @@ initialState:
       end: {line: 0, character: 21}
 finalState:
   documentContents: |
-    let x = [None, ];
+    let x = [None, Some()];
```
phillco, line 26: NTS: this is an expected change because of way we evaluate the textual and language sources with the change. Previously the language source would always be used if it was available; now we use both (`OneOf...`) and compare the text ranges as a heuristic.

In this situation the language server says `Some(1)` is the item, the textual definition says `1` is the item, and we prefer that because of the nuances in this algorithm, which can frequently prefer the shorter range: https://github.com/cursorless-dev/cursorless/blob/85fd827c25c9e6b7c8bb67e11c4bd355f21769eb/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/compareTargetScopes.ts#L72

(A related example, with the roles reversed, in Python; `import x, y, z`, the language server will not include the import keyword; the original textual range would think `import x` is the first item.). The comparison is applied to each item.

This existing comparison logic is already used by all the other scope providers, such as surrounding pairs, so it's pretty well tested. There will be other cases where this changes though since legacy scopes were using a much simpler algorithm that iterates forward and returns the first containing item. This is not a design decision in this change, everything that uses the next generation scope providers uses this comparison function.

NB: Andreas says the Rust language definition possibly shouldn't even have a list item provider, because they're only necessary when the textual definition isn't sufficient (i.e., unusual delimiters); it's not clear why it is necessary but the language definition was provided by a contributor so we're not sure.

data/fixtures/recorded/languages/clojure/clearEveryItem.yml
```diff
@@ -27,11 +27,10 @@ finalState:
 
     {
         ,
-        ;; hello
```
phillco, line 30: NTS: This is an example of the new code performing worse, but we understand why. This is because textual scope defines the item as:
```
;; hello
:baz "whatever",
```

In this case the language does identify that the comment is not part of the item, but we are preferring the textual scope, and as above, we no longer blind take the language server's view.

This will affect every language will you have an inline comment next to an item, we just don't have many tests in our more common languages that show that. There's an issue for improving the handling of comments around items. (Might be https://github.com/cursorless-dev/cursorless/issues/1770?). It's a complex change.

phillco: NB: the newer code uses more queries and predicates and less imperative TypeScript, so it's also something that would have been easier to monkey patch in the older code, although arguably, for a worse long term outcome

data/fixtures/recorded/languages/clojure/clearItem.yml
```diff
@@ -26,8 +26,11 @@ initialState:
 finalState:
   documentContents: |-
     {
-        
+        :bongo {
+            :foo "bar",
+            ,
+        }
     }
```
phillco, line 33: NTS: the old behavior was interesting: the cursor was on the comment so presumably the language server returned the entire :bongo item as the "item".

With the new behavior the textual scope includes the comment as part of the next item (see other thread), which is generally slightly worse, but in this case leads to a better outcome.

data/fixtures/recorded/languages/clojure/clearItemFine.yml
```diff
@@ -21,7 +21,7 @@ initialState:
       start: {line: 0, character: 2}
       end: {line: 0, character: 5}
 finalState:
-  documentContents: "{ :baz \"whatever\"}"
+  documentContents: "{}"
```
phillco, line 24: NTS: this outcome is worse but there's not much we can do about it at the current time. Problem is that closure does not use explicit list limiters and the textual base system can obviously not figure out if a space is part of an item or a separator here. The textual system thinks the entire `:foo "bar" :baz "whatever"` is one "item".

Andreas' idea: allow a language to specify whether it uses explicit delimiters for lists, and if it does not, go back to the old implementation of always taking the language servers opinion.

phillco: https://github.com/cursorless-dev/cursorless/issues/2723

data/fixtures/recorded/languages/clojure/chuckItemZip.yml
```diff
@@ -29,9 +29,7 @@ finalState:
   documentContents: |-
     {
         :foo "bar",
-        ;; hello
-    ,
```
phillco, line 33: Same as the other ones

data/fixtures/scopes/python/collectionItem.unenclosed.iteration3.scope
```diff
@@ -2,10 +2,25 @@ def foo():
...
 
-[Range] = 1:11-1:19
+[#1 Range] =
+[#1 Domain] = 0:8-0:8
+          ><
+0| def foo():
+
+
+[#2 Range] = 1:4-1:19
+      >---------------<
```
phillco, line 12: This syntax is indicating where Cursorless might _search_ for items (Andreas says: that's the definition of iteration scope), although not necessarily were they could be returned. The >/< are not part of the range.

data/fixtures/scopes/python/collectionItem.unenclosed.iteration3.scope
```diff
@@ -2,10 +2,25 @@ def foo():
     global bar, baz
 ---
 
-[Range] = 1:11-1:19
+[#1 Range] =
+[#1 Domain] = 0:8-0:8
+          ><
+0| def foo():
```
phillco, line 8: NB: this range is technically empty. We don't consider it worthwhile to omit empty ranges from this test file.

Andreas says: conceivably it could be useful to be able to boot strew up an empty list by saying "pour item" and targeting this range, so empty ranges are not necessarily useless.

data/fixtures/scopes/python/collectionItem.unenclosed.iteration3.scope
```diff
@@ -2,10 +2,25 @@ def foo():
...
+[#2 Range] = 1:4-1:19
+      >---------------<
+1|     global bar, baz
+
+[#2 Domain] = 1:0-1:19
+  >-------------------<
+1|     global bar, baz
+
+
+[#3 Range] = 1:11-1:19
```
phillco, line 20: NB: this is the old range from the language server, it's not marked as such, but you know because it's identical to the old code

data/fixtures/scopes/python/collectionItem.unenclosed.iteration3.scope
```diff
@@ -2,10 +2,25 @@ def foo():
...
+0| def foo():
+
+
+[#2 Range] = 1:4-1:19
+      >---------------<
+1|     global bar, baz
+
+[#2 Domain] = 1:0-1:19
+  >-------------------<
+1|     global bar, baz
```
phillco, line 17: This domain is extra but we know from the previous code it will be eliminated when items are actually requested (`global` will be whittled away).

The intent of this type of test file is just to show the iteration search scopes, so this is proper.

data/fixtures/scopes/python/collectionItem.unenclosed2.scope
```diff
@@ -2,12 +2,12 @@ import foo, bar
 ---
 
 [#1 Content] =
-[#1 Domain] = 0:7-0:10
-         >---<
+[#1 Domain] = 0:0-0:10
```
phillco, line 5: NB: #1 on the right side is new; the old 1 has become 2, the old 2 has become 3

The new one here is the extra scope generated by the textual source.

As above this won't actually change you behavior because when items are actually specifically requested we are going to end up the same list, `import ` won't be included

data/fixtures/scopes/textual/collectionItem.textual.iteration6.scope
```diff
@@ -0,0 +1,49 @@
+[
+    1,
+    2,
+    3,
+]
+---
+
+[#1 Range] = 1:4-3:6
+      >--
```
phillco, line 9: `>--` Doesn't include the trailing whitespace before the "1" item, since we don't consider that to be part of the list scope, but the whitespace before "3" is; that's why these look unbalanced.

If there was trailing whitespace after the "3," item that would similarly not be part of the iteration range, and not included in closing marker.

Andreas says this is the same logic that handles the "inside" code, we haven't seen it in this review.

phillco: Andreas says: if you have only whitespace between delimiters and you say take inside, you get all of the whitespace; if there's any non whitespace characters it gets trimmed.

data/fixtures/scopes/textual/collectionItem.textual.iteration6.scope
```diff
@@ -0,0 +1,49 @@
...
+   >
+0| [
+1|     1,
+2|     2,
+3|     3,
+4| ]
+   <
+
+
+[#2 Range] = 1:4-1:6
```
phillco, line 25: NB: all of these ranges are from the textual source

data/fixtures/scopes/textual/collectionItem.textual8.scope
```diff
@@ -0,0 +1,33 @@
...
+---
+
+[#1 Content] =
+[#1 Domain] = 0:1-0:2
+   >-<
+0| [1, "2, 3"]
+
+[#1 Removal] = 0:1-0:4
+   >---<
+0| [1, "2, 3"]
```
phillco, line 11: There is specific logic to discard stuff inside of strings (text fragments)

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/OneOfScopeHandler.ts
```diff
@@ -64,6 +88,14 @@ export class OneOfScopeHandler extends BaseScopeHandler {
     direction: Direction,
     hints: ScopeIteratorRequirements,
   ): Iterable<TargetScope> {
+    // If we have used the iteration scope handler we only want to yield from it's handler
```
phillco, line 91:
```suggestion
    // If we have used the iteration scope handler, we only want to yield from its handler.
```

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/CollectionItemScopeHandler/createTargetScope.ts
```diff
@@ -0,0 +1,54 @@
...
+    previousRange != null
+      ? new Range(previousRange.end, contentRange.start)
+      : undefined;
+  const trailingDelimiterRange =
+    nextRange != null
+      ? new Range(contentRange.end, nextRange.start)
+      : undefined;
+
+  // We have both leading and trailing delimiter ranges
+  // The leading one is longer/more specific so prefer to use that for removal.
```
phillco, line 24:
```suggestion
  // If the leading one is longer/more specific, prefer to use that for removal;
  // otherwise use undefined to fallback to the default behavior (often trailing)
```

phillco (approved):

PR #2684, opened 2024-11-19
Assert creation of scope handlers

AndreasArvidsson:
## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ScopeHandlerFactoryImpl.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ScopeHandlerFactoryImpl.ts
@@ -46,6 +46,7 @@ import type { CustomScopeType, ScopeHandler } from "./scopeHandler.types";
 export class ScopeHandlerFactoryImpl implements ScopeHandlerFactory {
   constructor(private languageDefinitions: LanguageDefinitions) {
     this.tryCreate = this.tryCreate.bind(this);
+    this.create = this.create.bind(this);
   }
 
   tryCreate(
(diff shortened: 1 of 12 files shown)


packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ScopeHandlerFactoryImpl.ts
```diff
@@ -45,10 +45,11 @@ import type { CustomScopeType, ScopeHandler } from "./scopeHandler.types";
  */
 export class ScopeHandlerFactoryImpl implements ScopeHandlerFactory {
   constructor(private languageDefinitions: LanguageDefinitions) {
+    this.tryCreate = this.tryCreate.bind(this);
     this.create = this.create.bind(this);
   }
 
-  create(
+  tryCreate(
```
pokey, line 52: I believe we usually use `maybe` for things like this, eg `maybeCreate`

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ScopeHandlerFactoryImpl.ts
```diff
@@ -114,4 +115,15 @@ export class ScopeHandlerFactoryImpl implements ScopeHandlerFactory {
           ?.getScopeHandler(scopeType);
     }
   }
+
+  create(
```
pokey, line 119: I would be tempted to call this `createStrict` or `createOrThrow`. Once we've migrated all scopes, this will never throw, right? So maybe we just leave it 🤷‍♂️

pokey (approved):
looks good subject to minor comments

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ScopeHandlerFactoryImpl.ts
AndreasArvidsson: Fixed

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/ScopeHandlerFactoryImpl.ts
AndreasArvidsson: Yes this shouldn't really throw. It's just there because I don't like doing `!` assertion in typescript if we can get around it.

Issue #2685, opened 2024-11-19
Cannot read properties of undefined (reading 'run')


tankorsmash:
Hi there, Cursorless has been wonderful.

I'm using Cursorless to edit Python and Vue2 SFC Typescript files, on `d8a602e`. `take line this` throws an error now, "Cannot read properties of undefined (reading 'run')", which _seems_ like it's because `action.run` (`r.run` in the compiled JS) is undefined.

I bisected from 10 commits ago to now, and it showed `b3b1abf3ce2c94bc09c9d87bb288b61572ed805d` as the bad commit, but that might only because that commit, `take line this` stopped being recognized.

```
2024-11-19 13:07:05.372 [error] TypeError: Cannot read properties of undefined (reading 'run')
	at c:\Users\username\.vscode\extensions\pokey.cursorless-0.28.1110\extension.cjs:128:9129
	at Array.flatMap (<anonymous>)
	at c:\Users\username\.vscode\extensions\pokey.cursorless-0.28.1110\extension.cjs:128:9116
	at Array.forEach (<anonymous>)
	at Vw (c:\Users\username\.vscode\extensions\pokey.cursorless-0.28.1110\extension.cjs:128:9100)
	at qw.processPrimitiveTarget (c:\Users\username\.vscode\extensions\pokey.cursorless-0.28.1110\extension.cjs:128:9014)
	at qw.processTarget (c:\Users\username\.vscode\extensions\pokey.cursorless-0.28.1110\extension.cjs:128:8031)
	at qw.run (c:\Users\username\.vscode\extensions\pokey.cursorless-0.28.1110\extension.cjs:128:7819)
	at iy.run (c:\Users\username\.vscode\extensions\pokey.cursorless-0.28.1110\extension.cjs:128:7681)
	at Kb.getTargets (c:\Users\username\.vscode\extensions\pokey.cursorless-0.28.1110\extension.cjs:140:5397)
	at Kb.runAction (c:\Users\username\.vscode\extensions\pokey.cursorless-0.28.1110\extension.cjs:140:5297)
	at Kb.run (c:\Users\username\.vscode\extensions\pokey.cursorless-0.28.1110\extension.cjs:140:3004)
	at async Object.run (c:\Users\username\.vscode\extensions\pokey.cursorless-0.28.1110\extension.cjs:140:30329)
	at async yE (c:\Users\username\.vscode\extensions\pokey.cursorless-0.28.1110\extension.cjs:140:19925)
[...]

PR #2686, opened 2024-11-20
Validate action description in factories

AndreasArvidsson:
Since we don't have input validation we get these kind of errors to frequently:
https://github.com/cursorless-dev/cursorless/issues/2685

## Checklist


pokey (commented):
I think validation should be done during the validation stage in https://github.com/cursorless-dev/cursorless/blob/8cd7c211930e1ad252eaf399cb38bab7900bcc68/packages/common/src/util/ensureCommandShape.ts#L12, rather than sprinkled all over our codebase. Is there a reason we can't do the validation there? I'm guessing it's simpler to it in these places where we've already drilled into the payload, but I still think it pollutes the code base and we'd have to remove it once we do proper schema validation

PR #2689, opened 2024-11-21
Use app.notify action

AndreasArvidsson:
The fewer module imports from Talon the better

## Checklist

### DIFF
--- a/cursorless-talon/src/actions/homophones.py
+++ b/cursorless-talon/src/actions/homophones.py
@@ -1,6 +1,6 @@
 from typing import Optional
 
-from talon import actions, app
+from talon import actions
 
 from ..targets.target_types import (
     CursorlessExplicitTarget,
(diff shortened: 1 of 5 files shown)


pokey (commented):
I'm not sure I agree with this one, but if you feel strongly I'm not opposed. It makes our code longer and in at least two files it doesn't remove any imports

AndreasArvidsson:
We had this discussion in community and all agreed that we should use a Talon action if it exists. Importing modules to get access to features that Talon already exposes as an action is bad code. For example the user might have redefined notified to send its someplace else and we should respect that. But it's not really something I care to much about when it comes to notify in particular. I think in the two files where we actually get rid of an import I think it's an improvement at least.

pokey:
update from meet-up: we didn't like the following

- no type hints
- can't use `body` kwarg so would have to switch back for cases where you need it

PR #2693, opened 2024-11-23
Notebook cell scope provider

### LINKED ISSUE #1053: Migrate `notebookCell` to use a scope handler

AndreasArvidsson:
Fixes #1053

Related #1052

I'm sure how we're gonna support the python comment style cells

```py
# %%
print("hello")
```

We have tests for them: https://github.com/cursorless-dev/cursorless/blob/56ad4c5cef3f8c8267a098a375cf9139cbbfb0ff/data/fixtures/recorded/selectionTypes/pourCell.yml
But they are not a cell scope you can actually target. Tree sitter just sees these as a comment and a function call. We could
1. Remove them. I'm probably leaning towards this.
2. Try to use regular expression to actually support them properly. That means we're back to language specific implementations in typescript.
3. Do what we did before and just yield the current position as a notebook target. Feels very hacky and not something I would like to do.

## Checklist


pokey:
This one probably warrants some discussion.

AndreasArvidsson:
I went with solution one and just removed those four tests

phillco (approved):

PR #2700, opened 2024-12-22
Fix surrounding pair performance

AndreasArvidsson:
With a ~10000 line json. `"take pair"` went from about 12 seconds to 200ms on my computer

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/languages/LanguageDefinition.ts
+++ b/packages/cursorless-engine/src/languages/LanguageDefinition.ts
@@ -3,6 +3,7 @@ import type {
   ScopeType,
   SimpleScopeType,
   SimpleScopeTypeType,
+  StringRecord,
   TreeSitter,
 } from "@cursorless/common";
 import {
@@ -97,6 +98,28 @@ export class LanguageDefinition {
       .matches(document)
       .map((match) => match.captures.find(({ name }) => name === captureName))
       .filter((capture) => capture != null);
+  }
+
+  getMultipleCaptures(
+    document: TextDocument,
+    captureNames: SimpleScopeTypeType[],
+  ): StringRecord<QueryCapture[]> {
+    const matches = this.query.matches(document);
+    const result: StringRecord<QueryCapture[]> = {};
+    const captureNamesSet = new Set<string>(captureNames);
+
+    for (const match of matches) {
+      for (const capture of match.captures) {
+        if (captureNamesSet.has(capture.name)) {
+          if (result[capture.name] == null) {
+            result[capture.name] = [];
+          }
+          result[capture.name]!.push(capture);
+        }
+      }
+    }
+
+    return result;
   }
 }
 
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -76,6 +76,7 @@ export * from "./types/Selection";
 export * from "./types/snippet.types";
 export * from "./types/SpokenForm";
 export * from "./types/SpokenFormType";
+export * from "./types/StringRecord";
 export * from "./types/TalonSpokenForms";
 export * from "./types/TestCaseFixture";
 export * from "./types/TestHelpers";
--- a/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/RangeIterator.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/RangeIterator.ts
@@ -6,7 +6,14 @@ import type { Range } from "@cursorless/common";
 export class RangeIterator<T extends { range: Range }> {
   private index = 0;
 
-  constructor(public items: T[]) {}
+  constructor(
+    public items: T[],
+    sortItems = false,
+  ) {
+    if (sortItems) {
+      this.items.sort((a, b) => a.range.start.compareTo(b.range.start));
+    }
+  }
 
   contains(searchItem: Range): boolean {
     return this.advance(searchItem);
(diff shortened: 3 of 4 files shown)


packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/RangeLookupList.ts
```diff
@@ -0,0 +1,49 @@
+import type { Range } from "@cursorless/common";
+
+/**
+ * An iterator that allows for efficient lookup of ranges that contain a search item.
+ * The items must be sorted in document order.
+ */
+export class RangeLookupList<T extends { range: Range }> {
```
pokey, line 7: Is this correct? If you accept my name change, don't forget to rename doc

```suggestion
/**
 * Given a list of ranges (the haystack), allows the client to search for a sequence of ranges (the needles).
 * Has the following requirements:
 * - the haystack must be sorted in document order
 * - **the needles must be in document order as well**. This enables us to avoid backtracking as you search for a sequence of items.
 * - the haystack entries must not overlap. Adjacent is fine
 */
export class OneWayRangeFinder<T extends { range: Range }> {
```

packages/cursorless-engine/src/languages/LanguageDefinition.ts
```diff
@@ -93,10 +99,34 @@ export class LanguageDefinition {
...
-      .filter((capture) => capture != null);
+    if (!this.cache.isValid(document)) {
+      this.cache.update(document, this.getCapturesMap(document));
+    }
+
+    return this.cache.get(captureName);
+  }
+
+  clearCache(): void {
+    this.cache.clear();
```
pokey, line 110: i'd personally just make a new one but 🤷‍♂️

packages/cursorless-engine/src/languages/LanguageDefinitions.ts
```diff
@@ -32,6 +32,11 @@ export interface LanguageDefinitions {
    */
   get(languageId: string): LanguageDefinition | undefined;
 
+  /**
+   * Clear the cache of all language definitions. This is run at the start of a command.
```
pokey, line 36: worth saying why this is necessary. I'm also concerned that if you need to do this, then the cache is breaking things that don't happen as a result of a command, eg does hot reloading still work with the scope visualizer?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/RangeLookupTree.ts
```diff
@@ -0,0 +1,73 @@
...
+  constructor(items: T[]) {
+    this.children = createNodes(items);
+  }
+
+  getSmallLestContaining(separator: Range): T | undefined {
+    return this.children
+      .getContaining(separator)
+      ?.getSmallLestContaining(separator);
+  }
+}
```
pokey, line 23:
```suggestion
  getSmallestContaining(separator: Range): T | undefined {
    return this.children
      .getContaining(separator)
      ?.getSmallestContaining(separator);
  }
}
```

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/RangeLookupTree.ts
```diff
@@ -0,0 +1,73 @@
+import type { Range } from "@cursorless/common";
+import { RangeLookupList } from "./RangeLookupList";
+
+/**
+ * A tree of ranges that allows for efficient lookup of ranges that contain a search item.
+ * The items must be sorted in document order.
+ */
+export class RangeLookupTree<T extends { range: Range }> {
```
pokey, line 8: As above, if correct, don't forget to rename doc.

```suggestion
/**
 * Given a list of ranges (the haystack), allows the client to search for smallest range containing a range (the needle).
 * Has the following requirements:
 * - the haystack must be sorted in document order
 * - **the needles must be in document order as well**. This enables us to avoid backtracking as you search for a sequence of items.
 * - the haystack entries **may** be nested, but one haystack entry cannot partially contain another
 */
export class OneWayNestedRangeFinder<T extends { range: Range }> {
```

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/RangeLookupTree.ts
pokey, line 1: A few unit tests would be nice here. It's complex enough, and would function as a form of documentation as well

pokey (approved):
Good stuff; see comments

packages/cursorless-engine/src/languages/LanguageDefinition.ts
```diff
@@ -93,10 +99,34 @@ export class LanguageDefinition {
...
+  private getCapturesMap(document: TextDocument): StringRecord<QueryCapture[]> {
+    const matches = this.query.matches(document);
+    const result: StringRecord<QueryCapture[]> = {};
+
+    for (const match of matches) {
+      for (const capture of match.captures) {
+        if (result[capture.name] == null) {
+          result[capture.name] = [];
+        }
+        result[capture.name]!.push(capture);
```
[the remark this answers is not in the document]
AndreasArvidsson: I kinda prefer it this way to be honest

packages/cursorless-engine/src/languages/LanguageDefinitions.ts
pokey: I guess we blow away the language def every time it hot reloads so that's why it works?

pokey:
Feel free to merge when you're happy

PR #2704, opened 2024-12-31
Added support for browser rp

AndreasArvidsson:
When using Cursorless everywhere in Talon js the accessibility is not good enough for browsers. I'm working on my own browser extension to mitigate this. 

What this pull request is doing is adding a new Cursorless everywhere context. When the new tag is set Cursorless everywhere will use the command client instead of access ability.

## Checklist

### DIFF
--- a/cursorless-everywhere-talon/cursorless_everywhere_talon_browser.py
+++ b/cursorless-everywhere-talon/cursorless_everywhere_talon_browser.py
@@ -1,0 +1,53 @@
+from talon import Context, Module, actions
+
+from .cursorless_everywhere_types import EditorEdit, EditorState, SelectionOffsets
+
+mod = Module()
+
+mod.tag(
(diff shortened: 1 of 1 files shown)


cursorless-everywhere-talon/cursorless_everywhere_talon_win.py
```diff
@@ -24,6 +24,9 @@ class Actions:
     def cursorless_everywhere_get_editor_state() -> EditorState:
         el = ui.focused_element()
 
+        print(el.patterns)
+        print(el)
+
```
pokey, line 29: did you mean to leave these print statements?

pokey (approved):
I haven't looked closely, but it is well contained so I won't stand in the way. Maybe @phillco has opinions here

Also of course looking at how rango does things for browser rpc as that is quite battle tested at this point

PR #2705, opened 2024-12-31
Proper handling of surrounding pair delimiter prefixes

AndreasArvidsson:
On main `"take inside"` on `"command server"` will only select `command serve` because `r"` is considered a delimiter. This pull request fixes this.

## Checklist


data/fixtures/recorded/surroundingPair/parseTree/python/changePair3.yml
```diff
@@ -11,10 +11,10 @@ command:
           scopeType: {type: surroundingPair, delimiter: any}
   usePrePhraseSnapshot: true
 initialState:
-  documentContents: "\" r\""
```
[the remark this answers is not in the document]
AndreasArvidsson: No
I will give it another go to try to support this, but this kind of edge case is less important they're not being able to take the inside of `"command server"`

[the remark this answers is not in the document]
AndreasArvidsson: Now working

data/fixtures/recorded/surroundingPair/parseTree/typescript/changePair.yml
```diff
@@ -12,9 +12,9 @@ command:
   usePrePhraseSnapshot: true
 initialState:
   documentContents: |-
-    (
-    // )
-    )
+    {
```
AndreasArvidsson, line 15: The old example is not valid typescript and gives an error node. This pull request changes so we actually check if the text fragment is erroring or not.

queries/python.scm
```diff
@@ -652,3 +652,8 @@ operator: [
 (function_definition
   "->" @disqualifyDelimiter
 )
+
+(
+  (string_start) @pairDelimiter
+  (#match? @pairDelimiter "^\\w")
```
AndreasArvidsson, line 658: No need to qualify the default delimiters

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/delimiterMaps.ts
```diff
@@ -102,26 +59,8 @@ const delimiterToTextOverrides: Record<string, Partial<DelimiterMap>> = {
...
-    tripleSingleQuotes: [
-      pythonPrefixes.map((prefix) => `${prefix}'''`),
-      "'''",
-      { isUnknownSide: true },
-    ],
-    tripleDoubleQuotes: [
-      pythonPrefixes.map((prefix) => `${prefix}"""`),
-      '"""',
-      { isUnknownSide: true },
-    ],
```
phillco, line 124: Note to self: the approach still uses the regex first and then uses the parse tree to validate the matches. Thus we are still using this but we are simplifying and removing all of the Python string prefix hardcoding because it shouldn't be necessary most of the time with the new querying mechanism (plus it's worse in the case where a string ends with r). But if the parse tree is invalid the regex are still going to be the primary source of truth.

queries/python.scm
phillco:
```suggestion
  (#match? @pairDelimiter "^[a-zA-z]+")
```

We don't need numbers since the Python special string prefixes can only be letters

queries/python.scm
```diff
@@ -652,3 +652,8 @@ operator: [
 (function_definition
   "->" @disqualifyDelimiter
 )
+
+(
+  (string_start) @pairDelimiter
+  (#match? @pairDelimiter "^[a-zA-z]+")
```
phillco, line 658:
```suggestion
  (#match? @pairDelimiter "^[a-zA-Z]+")
```

phillco (approved):

Issue #2157, opened 2023-12-21
Support implicit list targets


pokey:
Similar to the way we support "take past bat", we should support "take and bat". This won't do auto token expansion because it's an implicit target, which is desirable

phillco:
Merged #2707 instead; we found "take and" to feel clunkier.

phillco:
Conceivable that you might want "chuck and bat" instead of "chuck this and bat", but we consider that pretty obtuse and marginal, closing.

PR #2707, opened 2025-01-04
Add action to select new targets

AndreasArvidsson:
Adds a new selection while preserving existing ones
`"append <target>"` replaces `"take this and <target>"`

You can also use `append pre` and `append post` to add a cursor before or after the given target (rather than selecting that target), while keeping your current selections unchanged. (Previously you could do something like `pre this and <target>` but this would also reduce your current selection to a cursor before the current target.)

## Checklist


cursorless-talon/src/spoken_forms.json
```diff
@@ -2,9 +2,12 @@
   "NOTE FOR USERS": "Please don't edit this json file; see https://www.cursorless.org/docs/user/customization",
   "actions.csv": {
     "simple_action": {
+      "add post": "addSelectionAfter",
+      "add pre": "addSelectionBefore",
+      "add": "addSelection",
       "bottom": "scrollToBottom",
-      "break": "breakLine",
       "break point": "toggleLineBreakpoint",
+      "break": "breakLine",
```
AndreasArvidsson, line 10: These have moved because I sorted the list

cursorless-talon/src/spoken_forms.json
```diff
@@ -2,9 +2,12 @@
   "NOTE FOR USERS": "Please don't edit this json file; see https://www.cursorless.org/docs/user/customization",
   "actions.csv": {
     "simple_action": {
+      "add post": "addSelectionAfter",
+      "add pre": "addSelectionBefore",
+      "add": "addSelection",
```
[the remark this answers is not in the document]
phillco: My notes: Sounds like this was discussed but the consensus was that it's better this way because the user could change it to a single keyword. The concern with the cross product was that the "add" prefix would have to be changed three times (@pokey if I missed something let me know)

phillco:
Sounds like #2157 was discussed but this was considered to be a more elegant syntax (I agree)

packages/cursorless-org-docs/src/docs/user/README.md
```diff
@@ -531,9 +531,12 @@ Despite the name cursorless, some of the most basic commands in cursorless are f
 
 Note that when combined with list targets, `take`/`pre`/`post` commands will result in multiple cursors.
 
+- `"take <TARGET>"`: Selects the given target.
 - `"pre <TARGET>"`: Places the cursor before the given target.
 - `"post <TARGET>"`: Places the cursor after the given target.
-- `"take <TARGET>"`: Selects the given target.
+- `"add <TARGET>"`: Selects the given target while preserving existing selections.
+- `"add pre <TARGET>"`: Places the cursor before the given target while preserving existing selections.
+- `"add post <TARGET>"`: Places the cursor after the given target while preserving existing selections.
```
phillco, line 539:
```suggestion
- `"append <TARGET>"`: Selects the given target while preserving existing selections.
- `"append pre <TARGET>"`: Adds a new cursor before the given target while preserving your existing selections.
- `"append post <TARGET>"`: Adds a new cursor after the given target, while preserving your existing selections.
```

packages/cursorless-org-docs/src/docs/user/README.md
```diff
@@ -531,9 +531,12 @@ Despite the name cursorless, some of the most basic commands in cursorless are f
 
 Note that when combined with list targets, `take`/`pre`/`post` commands will result in multiple cursors.
 
+- `"take <TARGET>"`: Selects the given target.
 - `"pre <TARGET>"`: Places the cursor before the given target.
 - `"post <TARGET>"`: Places the cursor after the given target.
-- `"take <TARGET>"`: Selects the given target.
+- `"append <TARGET>"`: Selects the given target, while preserving your existing selections.
+- `"append pre <TARGET>"`: Adds a new cursor before the given target while preserving your existing selections.
```
phillco, line 538:
```suggestion
- `"append pre <TARGET>"`: Adds a new cursor before the given target, while preserving your existing selections.
```

phillco (approved):

phillco:
Going to merge, @pokey let us know if you have followup comments

PR #2709, opened 2025-01-08
Pass individual changes to browser extension

AndreasArvidsson:
## Checklist

### DIFF
--- a/cursorless-everywhere-talon/cursorless_everywhere_talon_browser.py
+++ b/cursorless-everywhere-talon/cursorless_everywhere_talon_browser.py
@@ -1,6 +1,11 @@
 from talon import Context, Module, actions
 
-from .cursorless_everywhere_types import EditorEdit, EditorState, SelectionOffsets
+from .cursorless_everywhere_types import (
+    EditorChange,
+    EditorEdit,
+    EditorState,
+    SelectionOffsets,
+)
 
 mod = Module()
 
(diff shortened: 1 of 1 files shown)


cursorless-everywhere-talon/cursorless_everywhere_talon_browser.py
```diff
@@ -71,3 +77,17 @@ def get_serializable_selections(selections: list[SelectionOffsets]):
...
+    for i in range(changes.length):  # pyright: ignore [reportAttributeAccessIssue]
+        change = changes[i]
+        result.append(
+            {
+                "text": change["text"],
+                "rangeOffset": change["rangeOffset"],
+                "rangeLength": change["rangeLength"],
+            }
+        )
+    return result
```
phillco, line 93: We can't use a list comprehension here because `changes` it is not a true Python list, it's a QuickJS array which doesn't support that. We also don't have the ability to convert these JavaScript objects to Python with a one shot utility because certain reflection data is missing per Andreas.

But what we could do is make simple wrapping utilities that will convert one level of the tree:

* JavaScript array to Python list (just does a iteration like this without looking at the objects in the array)
* JavaScript dictionary to Python dictionary (you would have to pass exactly the keys you expect)

With this approach, we can contain the weirdness of the JavaScript objects to where we get them, requiring a couple of utility calls, and use proper Python code thereafter.

phillco (approved):

phillco (approved):

PR #2711, opened 2025-01-09
Added Tree sitter query cache

AndreasArvidsson:
Uses the cache added in #2700 to also optimize all tree sitter queries, not just surrounding pairs.

## Checklist


packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts
```diff
@@ -70,6 +71,18 @@ export class TreeSitterQuery {
...
     end?: Position,
+  ): QueryMatch[] {
+    if (!treeSitterQueryCache.isValid(document, start, end)) {
+      const matches = this.getAllmatches(document, start, end);
+      treeSitterQueryCache.update(document, start, end, matches);
+    }
+    return treeSitterQueryCache.get();
+  }
+
+  private getAllmatches(
```
phillco, line 82:
```suggestion
  private getAllMatches(
```

packages/cursorless-engine/src/languages/TreeSitterQuery/treeSitterQueryCache.ts
```diff
@@ -0,0 +1,72 @@
...
+  private matches: QueryMatch[] = [];
+
+  isValid(
+    document: TextDocument,
+    startPosition: Position | undefined,
+    endPosition: Position | undefined,
+  ) {
+    return (
+      this.documentUri === document.uri.toString() &&
+      this.documentVersion === document.version &&
```
phillco, line 19: Note to self: this prevents the cache from being racy relying on the document change callback to invalidate it

packages/cursorless-engine/src/languages/TreeSitterQuery/treeSitterQueryCache.ts
```diff
@@ -0,0 +1,72 @@
...
+  },
+
+  get() {
+    return cache.get();
+  },
+
+  clear() {
+    cache = new Cache();
+  },
+};
```
phillco: I would just put `clear()` on the cache object itself so that we're not duplicating the access patterns here.

packages/cursorless-engine/src/languages/TreeSitterQuery/treeSitterQueryCache.ts
```diff
@@ -0,0 +1,72 @@
...
+  ) {
+    this.documentUri = document.uri.toString();
+    this.documentVersion = document.version;
+    this.documentLanguageId = document.languageId;
+    this.startPosition = startPosition;
+    this.endPosition = endPosition;
+    this.matches = matches;
+  }
+
+  get(): QueryMatch[] {
```
phillco, line 49: Note to self: I was surprised we are caching a lower level tree sitter object rather than something that we have post processed but it looks like we use this later on in other parts of the codebase, so we need to keep it around. It's probably not worth the work to change this.

https://github.com/cursorless-dev/cursorless/blob/1bffae54938cc5702000d72812bc04ea1ac0dab8/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler/TreeSitterIterationScopeHandler.ts#L40

https://github.com/cursorless-dev/cursorless/blob/main/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler/captureUtils.ts#L16-L36

phillco (approved):

Issue #2656, opened 2024-08-29
Improve performance when using large files


AndreasArvidsson:
When using large files. eg a json file with a 100 000 lines our scopes that looks at the entire file are quite slow. That two main cases for this is:
1. Tree sitter scope handler
2. Surrounding pair scope handler

The surrounding pair scope handler could be improved by parsing the file in chunks. Initially I been focused on debugging the Tree sitter scope handler and its matches function. 
https://github.com/cursorless-dev/cursorless/blob/49f70840f0b11d4fca1744f88af1cd3d91cbd3d5/packages/cursorless-engine/src/languages/TreeSitterQuery/TreeSitterQuery.ts#L69-L131

I added a time log between each step in the return statement. The below is the result for `"take key"` in a 26MB json file.

```ts
TreeSitterQuery.matches: query.matches: 4.579s
TreeSitterQuery.matches: map1: 3.303s
TreeSitterQuery.matches: filter: 160.087ms
TreeSitterQuery.matches: map2: 4.269s
TreeSitterQuery.matches: 12.313s
Tree sitter generateScopeCandidates: 12.917s
collectionKey getContainingScopeTarget: 12.985s
```

It's clear that the Tree sitter query match takes some time. This could be improved by not actually checking all patterns/captures and instead focusing on the one that's actually match to the requested scope type. It's also clear that our post processing in typescript is quite slow. I'm referring to the two map stages.

PR #2713, opened 2025-01-10
Increase performance for Tree sitter matches

AndreasArvidsson:
First, were checking the captures against the SCM files in production, which was an oversight - should only be happening in debugging.

Secondarily, the existing implementation multiple map and filters. This part of the code is very time sensitive and can run quite long for large files since we are iterating every capture in the entire file.

This implementation aims to reduce iteration steps

Fixes https://github.com/cursorless-dev/cursorless/issues/2656

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/languages/LanguageDefinition.ts
+++ b/packages/cursorless-engine/src/languages/LanguageDefinition.ts
@@ -74,7 +74,7 @@ export class LanguageDefinition {
    * legacy pathways
    */
   getScopeHandler(scopeType: ScopeType) {
-    if (!this.query.captureNames.includes(scopeType.type)) {
+    if (!this.query.hasCapture(scopeType.type)) {
       return undefined;
     }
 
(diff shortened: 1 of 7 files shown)


phillco:
We're going to test the performance of the test change and the captured to for loop rewrite separately so we can get a better sense for where the improvement is. The loop rewrite is still a nice code cleanliness change on its own, but we'd like to know how much more efficient it is to better calibrate JavaScript performance mental.

AndreasArvidsson:
Now I have some data

529k lines json, 11 iterations, median value
* main, checking captures: 4201ms
* main, don't check captures: 3944ms
* pr, checking captures: 3878ms
* pr, don't check captures: 3754ms

phillco:
Total performance improvement: 11.45%

- Not checking captures: 8.3%
- Loop rewriting: 3.2%

phillco (approved):

PR #2714, opened 2025-01-10
Use narrow search range instead of full document for Tree sitter queries

### LINKED ISSUE #1769: Narrow search range for tree-sitter queries
In #1800, we introduced a workaround for https://github.com/tree-sitter/tree-sitter/issues/2484, which means that we no longer restrict the search range for our tree-sitter queries.  That workaround has no impact on correctness, but could have performance implications for large files, as we get all matches and then filter them after the fact, rather than restricting to the range we care about.  Once https://github.com/tree-sitter/tree-sitter/issues/2484 is fixed, we'd like to reactivate our code that restricts the search range, by rolling back #1800

AndreasArvidsson:
Today we generate all captures in the entire document whenever we do a Tree sitter query. This is of course quite expensive for large documents and it would be better if we could narrow the search range. That is what's implemented in this pull request.

I also found some incorrect scm patterns that broke with this implementation that I have fixed.

Fixes #1769

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler/BaseTreeSitterScopeHandler.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler/BaseTreeSitterScopeHandler.ts
@@ -3,12 +3,13 @@ import { showError } from "@cursorless/common";
 import { uniqWith } from "lodash-es";
 import type { TreeSitterQuery } from "../../../../languages/TreeSitterQuery";
 import type { QueryMatch } from "../../../../languages/TreeSitterQuery/QueryCapture";
+import { ide } from "../../../../singletons/ide.singleton";
 import { BaseScopeHandler } from "../BaseScopeHandler";
 import { compareTargetScopes } from "../compareTargetScopes";
 import type { TargetScope } from "../scope.types";
 import type { ScopeIteratorRequirements } from "../scopeHandler.types";
+import { getQuerySearchRange } from "./getQuerySearchRange";
 import { mergeAdjacentBy } from "./mergeAdjacentBy";
-import { ide } from "../../../../singletons/ide.singleton";
 
 /** Base scope handler to use for both tree-sitter scopes and their iteration scopes */
 export abstract class BaseTreeSitterScopeHandler extends BaseScopeHandler {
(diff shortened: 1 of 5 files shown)


queries/c.scm
```diff
@@ -182,7 +182,7 @@
 ) @_.domain
 
 ;;!! aaa = 0;
-(
+(_
```
AndreasArvidsson, line 185: We need a common ancestor for all captures. Nodes not intersecting with the search range will not be matched.

AndreasArvidsson: For more information
https://github.com/cursorless-dev/cursorless/issues/1769#issuecomment-2582673327

phillco:
Note to self: this makes it more efficient by only searching forward or backwards based on the type of query specified, since many expensive queries are often directional ("take next funk" without an iteration scope for example).

phillco:
Possible future direction: We could make querying more incremental so it doesn't lock up with a gigantic query, but that's tricky because we don't necessarily know the tradeoff point. Multiple queries could be slower overall.

Some of this is limited by the fact that the actual grammar allows you to select an infinite amount of targets if you ask for it. So there's some limit to what we can do. We discussed the idea of a configuration default limit, but it's pretty hazy. I seem to recall least one person running into a problem where VS Code became extremely laggy after they accidentally asked for hundreds of thousands of selections. OTOH it's conceivable you could want to use this to do a gigantic refactor across the whole file?

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler/getQuerySearchRange.ts
```diff
@@ -0,0 +1,89 @@
...
+    end: document.positionAt(end),
+  };
+}
+
+/** Helper function for {@link getQuerySearchRange} */
+function getQuerySearchRangeCore(
+  offset: number,
+  distalOffset: number,
+  direction: Direction,
+  containment: ContainmentPolicy | null,
```
phillco, line 45: Any reason you don't use this syntax?

 ```suggestion
  containment: ContainmentPolicy?,
```

Comment (phillco) [reply packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler/getQuerySearchRange.ts:45]: Answer: ? defaults to undefined not null, we don't want to deal with both everywhere

File: packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler/getQuerySearchRange.ts [hunk 3]
```diff
@@ -0,0 +1,89 @@
...
+  // If containment is disallowed, we can shift the position forward by a
+  // character to avoid matching scopes that touch position. Otherwise, we
+  // shift the position backward by a character to ensure we get scopes that
+  // touch position, if we allow adjacent scopes.
+  const proximalShift = containment === "disallowed" ? 1 : -adjacentShift;
+
+  // FIXME: Don't go all the way to end of document when there is no
+  // distalPosition? Seems wasteful to query all the way to end of document for
+  // something like "next funk" Might be better to start smaller and
+  // exponentially grow
```
phillco, line 79:
```suggestion
  // FIXME: Don't go all the way to end of document when there is no
  // distalOffset? Seems wasteful to query all the way to end of document for
  // something like "next funk". Might be better to start smaller and
  // exponentially grow
```

phillco (approved):

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/TreeSitterScopeHandler/BaseTreeSitterScopeHandler.ts
```diff
@@ -20,18 +21,20 @@ export abstract class BaseTreeSitterScopeHandler extends BaseScopeHandler {
     editor: TextEditor,
     position: Position,
     direction: Direction,
-    _hints: ScopeIteratorRequirements,
```
phillco, line 23: NTS: _ meant it was unused, satisfies linter

PR #2717, opened 2025-01-11
Added performance test for `select previous scope`

AndreasArvidsson:
With our latest performance updates there is a clear difference in performance between containing and relative scopes. To keep track of this I have added relative scopes to the performance tests. I also added a few tests for every scope.

## Checklist

### DIFF
--- a/packages/cursorless-vscode-e2e/src/suite/performance.vscode.test.ts
+++ b/packages/cursorless-vscode-e2e/src/suite/performance.vscode.test.ts
@@ -101,7 +101,7 @@ async function removeToken(thresholdMs: number) {
 async function selectScopeType(
   scopeType: ScopeType,
   thresholdMs: number,
-  modifierType: ModifierType = "containing",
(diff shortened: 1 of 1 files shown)


packages/cursorless-vscode-e2e/src/suite/performance.vscode.test.ts
```diff
@@ -11,9 +12,10 @@ import { endToEndTestSetup } from "../endToEndTestSetup";
 
 const testData = generateTestData(100);
 
-const textBasedThresholdMs = 100;
-const parseTreeThresholdMs = 500;
-const surroundingPairThresholdMs = 500;
+const smallThresholdMs = 100;
+const largeThresholdMs = 500;
```
phillco, line 16: NTS: these are scaled to the GitHub runners that these will run on in CI, so much thresholds then you'll see locally

phillco (approved):

PR #2720, opened 2025-01-11
Show webview about missing installation dependencies

### LINKED ISSUE #1953: Detect common installation mistakes and show error message
- [x] User forgets to clone community => Look for action we depend on and show error message during init if not detected
- [x] User clones monorepo instead of cursorless-talon => Have python file at root of mono-repo that shows error message
- [ ] user hasn't installed talon?
- [ ] user hasn't installed cursorless-talon (note that we can look for `.cursorless` directory, tho old installations won't have it)
- [ ] user missing command-server

Note that if they just want cursorless keyboard, they don't need any of these things, so figure out how to handle

See https://github.com/cursorless-dev/cursorless/issues/1951 for an example where a user tried to use cursorless without Talon community repo

AndreasArvidsson:
If the user is missing some installation dependencies talon, cursorless-talon, command server we now show a web view informing them

Fixes #1953

Screenshot 2025-01-11T15-11-03 - Visual Studio Code

## Checklist

### DIFF
--- a/packages/cursorless-vscode/resources/installationDependencies.js
+++ b/packages/cursorless-vscode/resources/installationDependencies.js
@@ -1,24 +1,29 @@
+const vscode = acquireVsCodeApi();
+
 const root = document.getElementById("root");
 
 const header = document.createElement("h2");
 header.textContent = "Cursorless extension is now running!";
 
 const description = document.createElement("p");
-description.innerHTML =
-  "<a href='https://www.cursorless.org/docs/user/installation'>Click here to learn how to install Cursorless</a>";
+const aDocs = link(
+  "https://www.cursorless.org/docs/user/installation",
+  "Click here to learn how to install Cursorless",
+);
+description.innerHTML = `<p>Lets check if all dependencies are installed.</p>${aDocs}`;
 
-const div = document.createElement("div");
+const messages = document.createElement("div");
 
-root.append(header, description, div);
+root.append(header, description, messages);
 
 const keyboardUserMessage =
   "<p><i>If you're using Cursorless by keyboard you can ignore this message.</i></p>";
 
 window.addEventListener("message", (event) => {
-  const { data } = event;
+  const { dontShow, dependencies } = event.data;
   const children = [];
 
-  if (!data.talon) {
+  if (!dependencies.talon) {
     const a = link("https://talonvoice.com", "talonvoice.com");
     children.push(
       getChild(
@@ -26,7 +31,7 @@ window.addEventListener("message", (event) => {
         `Cursorless requires Talon to function by voice.</br>You can download Talon from ${a}.${keyboardUserMessage}`,
       ),
     );
-  } else if (!data.cursorlessTalon) {
+  } else if (!dependencies.cursorlessTalon) {
     const a = link(
       "https://github.com/cursorless-dev/cursorless-talon",
       "github.com/cursorless-dev/cursorless-talon",
@@ -39,7 +44,7 @@ window.addEventListener("message", (event) => {
     );
   }
 
-  if (!data.commandServer) {
+  if (!dependencies.commandServer) {
     const a = link(
       "https://marketplace.visualstudio.com/items?itemName=pokey.command-server",
       "vscode marketplace",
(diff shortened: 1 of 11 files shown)


packages/cursorless-vscode/resources/installationDependencies.js
```diff
@@ -0,0 +1,99 @@
...
+
+const header = document.createElement("h2");
+header.textContent = "Cursorless extension is now running!";
+
+const description = document.createElement("p");
+const aDocs = link(
+  "https://www.cursorless.org/docs/user/installation",
+  "Click here to learn how to install Cursorless",
+);
+description.innerHTML = `<p>Lets check if all dependencies are installed.</p>${aDocs}`;
```
phillco, line 13:
```suggestion
description.innerHTML = `<p>Let's check if all dependencies are installed.</p>${aDocs}`;
```

packages/cursorless-vscode/resources/installationDependencies.js
```diff
@@ -0,0 +1,99 @@
...
+
+const messages = document.createElement("div");
+
+root.append(header, description, messages);
+
+const keyboardUserMessage =
+  "<p><i>If you're using Cursorless by keyboard you can ignore this message.</i></p>";
```
phillco, line 20:
```suggestion
  "<p><i>If you're using Cursorless by keyboard, you can ignore this message.</i></p>";
```

packages/cursorless-vscode/src/InstallationDependencies.ts
```diff
@@ -0,0 +1,141 @@
...
+<head>
+    <meta http-equiv="Content-Security-Policy" content="script-src ${webview.cspSource};" />
+</head>
+
+<body>
+    <div id="root"></div>
+    <script src="${jsUri}"></script>
+</body>
+</html>`;
+}
```
phillco, line 141: * Extract this to a file

phillco (changes requested):
Rather than building up the HTML manually with JavaScript, since our layout is quite simple, let's just move this to a HTML file that the script can load and hydrate. I think the result will be more maintainable.

We don't have access to a templating library, but since we only have a couple of variables to inject we can just do this manually. There can still be some dynamic JavaScript as necessary but we don't need to build the entire template that way.

This also makes it easier to use a CSS block in <head> rather than inlining all the styles for each block. We could also consider putting that in its own dedicated file and have the script load and inject it, although the style is pretty simple that it may not be worth it.

It would be nice if you could just open the html file locally (file://) to iterate on the design, and this code would just make it work within the VS Code webview context.

packages/cursorless-vscode/src/InstallationDependencies.ts
```diff
@@ -0,0 +1,141 @@
...
+  return Object.values(dependencies).some((value) => !value);
+}
+
+function talonHomeExists() {
+  return fs.existsSync(getTalonHomePath());
+}
+
+function cursorlessTalonExists() {
+  const talonUserPath = path.join(getTalonHomePath(), "user");
+  const files = globSync("*/src/cursorless.talon", { cwd: talonUserPath });
```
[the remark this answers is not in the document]
AndreasArvidsson: Done

packages/cursorless-vscode/src/InstallationDependencies.ts
AndreasArvidsson: Done

packages/cursorless-vscode/resources/installationDependencies.html
```diff
@@ -0,0 +1,80 @@
...
+      </p>
+    </div>
+
+    <div id="msg-cursorless-talon" class="hide">
+      <h4>Cursorless Talon scripts missing</h4>
+      <p>
+        Cursorless requires Talon user scripts to function by voice.
+        <br />
+        The scripts are available at
+        <a href="https://github.com/cursorless-dev/cursorless-talon">
```
phillco, line 42:
```suggestion
        <a href="https://www.cursorless.org/docs/user/installation/#installing-the-talon-side">
```

packages/cursorless-vscode/resources/installationDependencies.html
```diff
@@ -0,0 +1,80 @@
...
+        </i>
+      </p>
+    </div>
+
+    <div id="msg-cursorless-talon" class="hide">
+      <h4>Cursorless Talon scripts missing</h4>
+      <p>
+        Cursorless requires Talon user scripts to function by voice.
+        <br />
+        The scripts are available at
```
phillco, line 41:
```suggestion
        The installation steps for the scripts are available at
```

phillco (approved):

PR #2728, opened 2025-01-14
Properly handle exception from missing vscode config json file

AndreasArvidsson:
Today on main this exception is not caught and the Cursorless list containing all the scopes is not filled which means that Cursorless is not working if we can't find the vscode json config file.

## Checklist

### DIFF
--- a/cursorless-talon/src/apps/vscode_settings.py
+++ b/cursorless-talon/src/apps/vscode_settings.py
@@ -61,7 +61,6 @@ class Actions:
             return actions.user.vscode_get_setting(key, default_value), False
         except Exception:
             print(fallback_message)
-            traceback.print_exc()
             return fallback_value, True
 
 
--- a/cursorless-talon/src/marks/decorated_mark.py
+++ b/cursorless-talon/src/marks/decorated_mark.py
@@ -153,7 +153,11 @@ slow_reload_job = None
 def init_hats(hat_colors: dict[str, str], hat_shapes: dict[str, str]):
     setup_hat_styles_csv(hat_colors, hat_shapes)
 
-    vscode_settings_path: Path = actions.user.vscode_settings_path().resolve()
+    try:
+        vscode_settings_path: Path = actions.user.vscode_settings_path().resolve()
+    except Exception as ex:
+        print(ex)
+        return lambda: None
 
     def on_watch(path, flags):
         global fast_reload_job, slow_reload_job
--- a/cursorless-talon/src/marks/decorated_mark.py
+++ b/cursorless-talon/src/marks/decorated_mark.py
@@ -153,11 +153,12 @@ slow_reload_job = None
 def init_hats(hat_colors: dict[str, str], hat_shapes: dict[str, str]):
     setup_hat_styles_csv(hat_colors, hat_shapes)
 
+    vscode_settings_path: Path | None = None
+
     try:
-        vscode_settings_path: Path = actions.user.vscode_settings_path().resolve()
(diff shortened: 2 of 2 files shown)


cursorless-talon/src/apps/vscode_settings.py
```diff
@@ -61,7 +61,6 @@ def vscode_get_setting_with_fallback(
             return actions.user.vscode_get_setting(key, default_value), False
         except Exception:
             print(fallback_message)
-            traceback.print_exc()
```
AndreasArvidsson, line 64: This just created a lot of noise in the Talon log. We know were the problem is without this.

cursorless-talon/src/marks/decorated_mark.py
```diff
@@ -153,7 +153,11 @@ def setup_hat_styles_csv(hat_colors: dict[str, str], hat_shapes: dict[str, str])
 def init_hats(hat_colors: dict[str, str], hat_shapes: dict[str, str]):
     setup_hat_styles_csv(hat_colors, hat_shapes)
 
-    vscode_settings_path: Path = actions.user.vscode_settings_path().resolve()
+    try:
+        vscode_settings_path: Path = actions.user.vscode_settings_path().resolve()
+    except Exception as ex:
+        print(ex)
+        return lambda: None
```
AndreasArvidsson, line 160: This exception was bubbling up the call stack and breaking Cursorless

cursorless-talon/src/marks/decorated_mark.py
```diff
@@ -153,7 +153,12 @@ def setup_hat_styles_csv(hat_colors: dict[str, str], hat_shapes: dict[str, str])
 def init_hats(hat_colors: dict[str, str], hat_shapes: dict[str, str]):
     setup_hat_styles_csv(hat_colors, hat_shapes)
 
-    vscode_settings_path: Path = actions.user.vscode_settings_path().resolve()
+    vscode_settings_path: Path | None = None
+
+    try:
+        vscode_settings_path = actions.user.vscode_settings_path().resolve()
+    except Exception as ex:
+        print(ex)
```
AndreasArvidsson, line 161: This is the exception that was uncaught and broke Cursorless logic further up the call stack.

phillco, line 161:
```suggestion
    try:
        vscode_settings_path = actions.user.vscode_settings_path().resolve()
    except Exception as e:
        print(f"Could not resolve VS Code settings path at {vscode_settings_path} (not fatal): {e}")
```

phillco: Try to avoid just doing raw printing of exception objects, because then you see the exception message in the log without much context / able to debug them.

phillco (approved):

cursorless-talon/src/marks/decorated_mark.py
AndreasArvidsson: The exception message is already well formed
`2025-01-16 18:00:29.801    IO Couldn't find VSCode's settings JSON. Tried these paths: C:\Users\andre\AppData\Roaming\Code2\User\settings.json, C:\Users\andre\AppData\Roaming\VSCodium\User\settings.json`

AndreasArvidsson: It's defined here
https://github.com/cursorless-dev/cursorless/blob/0bad81a015c08e2976728df09fe31d8a9105ec84/cursorless-talon/src/apps/vscode_settings.py#L66-L73

PR #2739, opened 2025-01-21
Fix for markdown cells

AndreasArvidsson:
The latest update for the notebook cell handler broke cells in markdown.
Also implements notebook scope handler with a one of scope handler the same way we do for collection item. We can have both language specific implementations as well as the ide notebook api so this is a more proper implementation.

## Checklist

### DIFF
--- a/data/fixtures/recorded/languages/markdown/changeCell.yml
+++ b/data/fixtures/recorded/languages/markdown/changeCell.yml
@@ -1,0 +1,27 @@
+languageId: markdown
+command:
+  version: 7
+  spokenForm: change cell
+  action:
+    name: clearAndSetSelection
+    target:
+      type: primitive
+      modifiers:
+        - type: containingScope
+          scopeType: {type: notebookCell}
+  usePrePhraseSnapshot: false
+initialState:
+  documentContents: |
+    ```
+    code
+    ```
+  selections:
+    - anchor: {line: 1, character: 0}
+      active: {line: 1, character: 0}
+  marks: {}
+finalState:
+  documentContents: |+
+
+  selections:
+    - anchor: {line: 0, character: 0}
+      active: {line: 0, character: 0}
--- a/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/LineScopeHandler.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/LineScopeHandler.ts
@@ -15,7 +15,7 @@ export class LineScopeHandler extends BaseScopeHandler {
     type: "paragraph",
   } as const;
   protected readonly isHierarchical = false;
-  public readonly includeAdjacentInEvery: boolean = true;
+  public readonly includeAdjacentInEvery = true;
 
   constructor(_scopeType: ScopeType, _languageId: string) {
     super();
--- a/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/NotebookCellApiScopeHandler.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/NotebookCellApiScopeHandler.ts
@@ -12,7 +12,7 @@ import type { TargetScope } from "./scope.types";
 import type { ScopeIteratorRequirements } from "./scopeHandler.types";
 
 /**
- * This is the scope handler for the actual notebook api in the ide.
+ * This is the scope handler for the actual notebook API in the IDE.
  */
 export class NotebookCellApiScopeHandler extends BaseScopeHandler {
   public readonly scopeType = { type: "notebookCell" } as const;
--- a/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/NotebookCellApiScopeHandler.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/NotebookCellApiScopeHandler.ts
@@ -1,0 +1,116 @@
+import {
+  Range,
+  type Direction,
+  type NotebookCell,
+  type Position,
+  type TextEditor,
+} from "@cursorless/common";
+import { ide } from "../../../singletons/ide.singleton";
+import { NotebookCellTarget } from "../../targets";
+import { BaseScopeHandler } from "./BaseScopeHandler";
+import type { TargetScope } from "./scope.types";
+import type { ScopeIteratorRequirements } from "./scopeHandler.types";
+
+/**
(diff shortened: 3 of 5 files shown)


packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/LineScopeHandler.ts
```diff
@@ -15,7 +15,7 @@ export class LineScopeHandler extends BaseScopeHandler {
     type: "paragraph",
   } as const;
   protected readonly isHierarchical = false;
-  public readonly includeAdjacentInEvery: boolean = true;
```
AndreasArvidsson, line 18: Just noticed this unnecessary type

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/NotebookCellScopeHandler.ts
```diff
@@ -9,35 +9,46 @@ import {
...
+import { BaseScopeHandler } from "./BaseScopeHandler";
 import type { TargetScope } from "./scope.types";
 import type {
+  ComplexScopeType,
   ScopeHandler,
   ScopeIteratorRequirements,
 } from "./scopeHandler.types";
 
-export class NotebookCellScopeHandler implements ScopeHandler {
+export class NotebookCellScopeHandler extends BaseScopeHandler {
```
AndreasArvidsson, line 20: The problem was that we were not using the base scope handler and without it we don't always have a distal position which broke our query search range logic.

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/NotebookCellApiScopeHandler.ts
```diff
@@ -0,0 +1,116 @@
...
+  type TextEditor,
+} from "@cursorless/common";
+import { ide } from "../../../singletons/ide.singleton";
+import { NotebookCellTarget } from "../../targets";
+import { BaseScopeHandler } from "./BaseScopeHandler";
+import type { TargetScope } from "./scope.types";
+import type { ScopeIteratorRequirements } from "./scopeHandler.types";
+
+/**
+ * This is the scope handler for the actual notebook api in the ide.
```
phillco, line 15:
```suggestion
 * This is the scope handler for the actual notebook API in the IDE.
```

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/NotebookCellScopeHandler.ts
```diff
@@ -1,132 +1,70 @@
...
-    private languageDefinitions: LanguageDefinitions,
-    _scopeType: ScopeType,
-    private languageId: string,
-  ) {}
-
-  *generateScopes(
-    editor: TextEditor,
-    position: Position,
-    direction: Direction,
-    hints: ScopeIteratorRequirements,
```
phillco, line 33: NTS: The actual type of the interface is `requirements?: Partial<ScopeIteratorRequirements>`. This should really be a tsc compile error but it wasn't. We therefore started to read properties on `hints` when some of those properties could be undefined which threw an exception

phillco: We needed something like what the base class does:

```
  *generateScopes(
    editor: TextEditor,
    position: Position,
    direction: Direction,
    requirements: Partial<ScopeIteratorRequirements> = {},
  ): Iterable<TargetScope> {
    const hints: ScopeIteratorRequirements = {
      ...DEFAULT_REQUIREMENTS,
      ...requirements,
      distalPosition:
        requirements.distalPosition ??
        (direction === "forward"
          ? editor.document.range.end
          : editor.document.range.start),
    };
```

Rather than copying the logic it was easier to just rewrite this class to use the base class, which Andreas says should have been done originally, just didn't think of it

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/NotebookCellScopeHandler.ts
```diff
@@ -1,132 +1,70 @@
...
+          scopeTypes: [
+            languageScopeHandler.scopeType,
+            apiScopeHandler.scopeType,
+          ],
+        },
+        [languageScopeHandler, apiScopeHandler],
+        languageId,
+      );
+    })();
+  }
```
phillco, line 60: `OneOfScopeHandler` means that sources from the two could be merged, so if you have markdown cells inside of note cells something like `next fifth cell` could refer to a markdown cell and the second notebook cell, previously this could by referred to one or the other.

(NB: not `fifth item`; this uses the iteration scope, and the iteration scope is more tightly bound to the data structure that you are in. It doesn't combined sources like that. `next fifth...` instead queries the scope generator five times, and this use of the scope generator does combine sources.)

Four different types of modifiers:

* containing
* every
* ordinal
* relative

Every and ordinal use the iteration scope; containing and relative consume from the scope generator but not the iteration scope.

Why don't iteration scopes combine sources? We discussed this and the behavior makes sense when you think about the use cases. When asking for ordinal or every scope ("every item" for example), you are not likely to want a creative interpretation of item that expands across the data structure that you are in.

`containing` does query the generator (so it could combine sources in theory) but only once so it behaves like every and ordinal.

In the context of notebook cells that might contain markdown cells, the pattern's that "every" never crosses a document boundary. (Imagine if you had conventional editors split and asked to take every markdown cell -- it's only going to act on the current editor.)

In short, the behavior generally works well for our existing use cases but it is not explicitly specified. `OneOfScopeHandler` will combine multiple sources when yielding items but when you query the iteration scope only one source will be used, because that's how iteration scopes work.

data/fixtures/recorded/languages/markdown/changeCell.yml
```diff
@@ -0,0 +1,27 @@
...
+command:
+  version: 7
+  spokenForm: change cell
+  action:
+    name: clearAndSetSelection
+    target:
+      type: primitive
+      modifiers:
+        - type: containingScope
+          scopeType: {type: notebookCell}
```
phillco, line 11: NB: `notebookCell` just means the "cell" scope type. It doesn't actually mean a notebook cell (and in fact here it means a markdown table cell), but we added notebooks first and we don't want to rename it.

phillco (approved):

PR #2740, opened 2025-01-21
Remove unnecessary catch which caused confusing error when iteration scopes were missing

AndreasArvidsson:
Today's implementation when you say something like `take every item` will show an error message `Couldn't find containing collectionItem.` even if it's the iteration scope that is missing. 

This change makes it so that we get the error message `Couldn't find containing iteration scope for collectionItem.` instead. This greatly helps us with diagnosing what actually went wrong.

For program languages were a scope is defined in Tree sitter the current error message makes little sense. You could use `take scope` and it works fine, but when you say `take every scope` you get an error message that there is no scope. Much better to actually show that this language is missing an iteration scope.

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/EveryScopeStage.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/EveryScopeStage.ts
@@ -73,20 +73,13 @@ export class EveryScopeStage implements ModifierStage {
     if (scopes == null) {
       // If target had no explicit range, or was contained by a single target
       // instance, expand to iteration scope before overlapping
-      try {
-        scopes = this.getDefaultIterationRange(
-          scopeHandler,
-          this.scopeHandlerFactory,
-          target,
-        ).flatMap((iterationRange) =>
-          getScopesOverlappingRange(scopeHandler, editor, iterationRange),
-        );
-      } catch (error) {
-        if (!(error instanceof NoContainingScopeError)) {
-          throw error;
-        }
-        scopes = [];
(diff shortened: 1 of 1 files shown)


packages/cursorless-engine/src/processTargets/modifiers/EveryScopeStage.ts
```diff
@@ -73,20 +73,13 @@ export class EveryScopeStage implements ModifierStage {
...
-        }
-        scopes = [];
-      }
+      scopes = this.getDefaultIterationRange(
+        scopeHandler,
+        this.scopeHandlerFactory,
+        target,
+      ).flatMap((iterationRange) =>
+        getScopesOverlappingRange(scopeHandler, editor, iterationRange),
+      );
```
phillco, line 82: This basically reverts the logic added in https://github.com/cursorless-dev/cursorless/pull/2081/files#diff-a47f14c27291690e54ca5b2e034c62193e9da0615bdb7b3be4153ede7668914f.\

It was necessary when there was this fallback code after it:

```
    if (scopes.length === 0) {
      if (scopeType.type === "collectionItem") {
        // For `collectionItem`, fall back to generic implementation
        return this.modifierStageFactory
          .getLegacyScopeStage(this.modifier)
          .run(target);
      }

      throw new NoContainingScopeError(scopeType.type);
    }
```

We have since removed the legacy code but didn't know that this code also needed to be removed

phillco (approved):

PR #2741, opened 2025-01-21
Fix bug with collection item scope

AndreasArvidsson:
We didn't properly pop this stack of iteration states. This led to items being grouped incorrectly.

## Checklist

### DIFF
--- a/data/fixtures/scopes/textual/collectionItem.textual15.scope
+++ b/data/fixtures/scopes/textual/collectionItem.textual15.scope
@@ -1,0 +1,180 @@
+{
+    language: {},
+    choices: {
+        disabled: true,
+        items: {
+            colWidth: [2, 10]
+        }
+    },
+    versions: {}
+}
+---
+
+[#1 Content] =
+[#1 Domain] = 1:4-1:16
+      >------------<
+1|     language: {},
+
+[#1 Removal] = 1:4-2:4
+      >-------------
+1|     language: {},
+2|     choices: {
+   ----<
+
+[#1 Trailing delimiter] = 1:16-2:4
+                  >-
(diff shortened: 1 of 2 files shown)


data/fixtures/scopes/textual/collectionItem.textual15.scope
```diff
@@ -0,0 +1,180 @@
+{
```
AndreasArvidsson, line 1: Added a really complex test just capture this behavior

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/CollectionItemScopeHandler/CollectionItemTextualScopeHandler.ts
```diff
@@ -68,51 +68,46 @@ export class CollectionItemTextualScopeHandler extends BaseScopeHandler {
         continue;
       }
 
-      const currentIterationState =
-        iterationStatesStack[iterationStatesStack.length - 1];
+      let currentIterationState: IterationState | undefined;
+
+      // Find current iteration state and pop all states not containing the separator
+      while (iterationStatesStack.length > 0) {
```
AndreasArvidsson, line 74: This is the logic that was missing before

data/fixtures/scopes/textual/collectionItem.textual15.scope
```diff
@@ -0,0 +1,180 @@
...
+    language: {},
+    choices: {
```
phillco, line 3: User behavior was with the cursor after the `{`, "take item" selected both language and choices instead of just choices.

PR #2744, opened 2025-01-21
Generate snippets for community format

AndreasArvidsson:
If the user has set the `user.cursorless_use_community_snippets` tag we create a snippet in the community snippet format instead of the Cursorless one. 

The legacy Cursorless snippets will be removed in a final pull request.

## Checklist


cursorless-talon/src/actions/generate_snippet.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@mod.action_class
+class Actions:
+    def private_cursorless_generate_snippet_action(target: CursorlessExplicitTarget):  # pyright: ignore [reportGeneralTypeIssues]
+        """Generate a snippet from the given target"""
+        actions.user.private_cursorless_command_no_wait(
+            {
+                "name": "generateSnippet",
+                "target": target,
```
AndreasArvidsson, line 23: By default don't send dirPath argument. This is for the legacy Cursorless snippet format. The vscode setting for snippet directory will be used.

cursorless-talon/src/actions/generate_snippet.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@ctx.action_class("user")
+class UserActions:
+    def private_cursorless_generate_snippet_action(target: CursorlessExplicitTarget):  # pyright: ignore [reportGeneralTypeIssues]
+        actions.user.private_cursorless_command_no_wait(
+            {
+                "name": "generateSnippet",
+                "target": target,
+                "dirPath": get_dir_path(),
```
AndreasArvidsson, line 35: If the community snippet tag has been set send dirPath argument. A new community snippet will be added to this folder.

data/fixtures/recorded/actions/snippets/snipMakeFunk2.yml
```diff
@@ -0,0 +1,55 @@
+languageId: typescript
```
AndreasArvidsson, line 1: Double up on existing snippet make tests in the community format

data/fixtures/recorded/actions/snippets/snipMakeFunk2.yml
```diff
@@ -0,0 +1,55 @@
...
+command:
+  version: 7
+  spokenForm: snippet make funk
+  action:
+    name: generateSnippet
+    dirPath: ""
```
AndreasArvidsson, line 7: Including the dirPath argument indicates that this is a community snippet.

cursorless-talon/src/actions/generate_snippet.py
```diff
@@ -0,0 +1,71 @@
...
+        if not setting_dir:
+            return None
+
+        dir = Path(str(setting_dir))
+
+        if not dir.is_absolute():
+            user_dir = Path(actions.path.talon_user())
+            dir = user_dir / dir
+
+        return str(dir.resolve())
```
phillco, line 59: I'd make all of these return `Path` and only coalesce to a string when you need to stick it into the JSON dictionary

packages/cursorless-engine/src/actions/GenerateSnippet/GenerateSnippetCommunity.ts
```diff
@@ -0,0 +1,237 @@
...
+
+function getsSnippetSelections(editor: TextEditor, range: Range): Selection[] {
+  const selections = editor.selections.filter((selection) =>
+    range.contains(selection),
+  );
+  selections.sort((a, b) => a.start.compareTo(b.start));
+  return selections;
+}
+
+const PLACEHOLDER = "PLACEHOLDER_VFA77zcbLD6wXNmfMAay";
```
phillco, line 224:
```suggestion
// Used to temporarily escape the $1, $2 snippet holes (the "meta snippet" holes that become live snippets when the user edits) so we can use traditional backslash escaping for the holes in the underlying snippet itself (the "user snippet" holes that will be saved as part of their new template).
const PLACEHOLDER = "PLACEHOLDER_VFA77zcbLD6wXNmfMAay";
```

packages/cursorless-engine/src/actions/GenerateSnippet/GenerateSnippet.ts
```diff
@@ -1,263 +1,25 @@
...
-      // Otherwise, we create and open a new document for the snippet in the
-      // user snippets dir
-      await this.snippets.openNewSnippetFile(snippetName);
+    if (dirPath == null) {
+      const action = new GenerateSnippetLegacy(this.snippets);
+      return action.run(targets, snippetName);
     }
 
-    // Insert the meta-snippet
-    await editableEditor.insertSnippet(snippetText);
```
phillco, line 234: NB: this was technically a bug and relying on odd Visual Studio Code behavior because we aren't using the new editor that was created inside of `this.snippets.openNewSnippetFile(snippetName);`. It just so happens that `insertSnippet` always uses the current editor even if you call it on a different one. The new code fixes this by making the function return a promise with the actual new editor which we needed for appending.

phillco: NB: Andreas fixed this because we now want to append the text to the file if it already exists, and reading the text from an editor that's not the current one does work, and that's when he found that the editor object was actually technically wrong

packages/cursorless-engine/src/actions/actions.types.ts
```diff
@@ -124,7 +124,11 @@ export interface ActionRecord extends Record<SimpleActionName, SimpleAction> {
   };
 
   generateSnippet: {
-    run(targets: Target[], snippetName?: string): Promise<ActionReturnValue>;
+    run(
+      targets: Target[],
+      dirPath?: string,
```
phillco, line 129: Not thrilled about `dirPath` but it makes sense after discussion -- we sometimes have the directory name `dirName` so it's needed to disambiguate.

If we had a `Path` type of class it would be nicer so we didn't have to put the type information in the name itself, but we don't. We could make a wrapper ourselves or just do two type aliases. We might make a followup for this.

`storageDirectory` would be my choice for the name but not if it has to be `storageDirectoryPath`, I think that's worse than the status quo

phillco (commented):
.

packages/cursorless-engine/src/actions/actions.types.ts
phillco: Settled on `directory` for now

packages/cursorless-engine/src/disabledComponents/DisabledSnippets.ts
```diff
@@ -17,7 +17,10 @@ export class DisabledSnippets implements Snippets {
     throw new Error("Snippets are not implemented.");
   }
 
-  openNewSnippetFile(_snippetName: string): Promise<void> {
+  openNewSnippetFile(
+    _snippetName: string,
+    _dirPath?: string,
```
phillco, line 22: .

PR #2747, opened 2025-01-23
Added voice command to migrate Cursorless snippet to community format

### LINKED ISSUE #2355: Create "migrate snippets" command that will automatically migrate all of a user's snippets
That can generate internal rep from above and use same output machinery. Will need to have a close look at schema of cursorless snippets vs community to understand what might be missing. @AndreasArvidsson may add support for scope-specific snippets (eg method v funk), but at the very least will not error out when the given field appears

As with #2354, if they have set custom snippet directory in community, use that dir instead of default

AndreasArvidsson:
2 fields in the Cursorless snippets are not available in the community format

1. scopeTypes. To my knowledge the only use case is omitting the `function` keyword when in a javascript class
2. excludeDescendantScopeTypes: Used in conjunction with `scopeTypes` to include the function keyword when in a function inside a class.

scopeTypes and excludeDescendantScopeTypes I'm not to sure we want to add to community. First of all they have a single use case and needing two separate fields just to describe this behavior fields a bit verbose.
As well community doesn't have support for sending multiple snippets. 

I'm planning this as a follow up. 

Reference
https://github.com/cursorless-dev/cursorless/blob/57acd6c0fcc32e5df122ce09bb26bde5f09495ca/packages/common/src/types/snippet.types.ts#L4-L24

Fixes #2355

## Checklist


packages/cursorless-vscode/src/migrateSnippets.ts
```diff
@@ -0,0 +1,106 @@
...
+    await migrateFile(targetDirectory, file);
+  }
+
+  await vscode.window.showInformationMessage(
+    `${files.length} snippet files migrated successfully!`,
+  );
+}
+
+async function migrateFile(targetDirectory: string, filePath: string) {
+  const fileName = path.basename(filePath, ".cursorless-snippets");
```
phillco, line 32: CURSORLESS_SNIPPETS_SUFFIX?

packages/cursorless-vscode/src/migrateSnippets.ts
```diff
@@ -0,0 +1,109 @@
...
+  const content = await fs.readFile(filePath, "utf8");
+  if (content.length === 0) {
+    return {};
+  }
+  return JSON.parse(content);
+}
+
+async function writeCommunityFile(snippetFile: SnippetFile, filePath: string) {
+  const snippetText = serializeSnippetFile(snippetFile);
+  const file = await fs.open(filePath, "w");
```
phillco, line 106: Checking for file existence separately from opening the file is generally racy, best to just do this:

```suggestion
  const file = await fs.open(filePath, "wx");
```

> 'wx': Like 'w' but fails if the path exists.

Then catch and put your conflict filename behavior in the catch block.

packages/cursorless-vscode/src/migrateSnippets.ts
```diff
@@ -0,0 +1,102 @@
...
+  try {
+    const destinationPath = path.join(targetDirectory, `${fileName}.snippet`);
+    await writeCommunityFile(communitySnippetFile, destinationPath);
+  } catch (_error) {
+    const destinationPath = path.join(
+      targetDirectory,
+      `${fileName}_CONFLICT.snippet`,
+    );
+    await writeCommunityFile(communitySnippetFile, destinationPath);
+  }
```
phillco, line 69:
```suggestion
  } catch (error: any) {
  if (error.code === 'EEXIST') {
    const destinationPath = path.join(
      targetDirectory,
      `${fileName}_CONFLICT.snippet`,
    );
    await writeCommunityFile(communitySnippetFile, destinationPath);
  } else {
    throw error;
  }
}
```

packages/cursorless-vscode/src/migrateSnippets.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+async function writeCommunityFile(snippetFile: SnippetFile, filePath: string) {
+  const snippetText = serializeSnippetFile(snippetFile);
+  const file = await fs.open(filePath, "wx");
+  try {
+    await file.write(snippetText);
+  } finally {
+    await file.close();
+  }
```
phillco, line 108: `using` is actually on the horizon, filed here:https://github.com/cursorless-dev/cursorless/issues/2786

phillco (approved):

pokey:
What does this tool do with the `scopeTypes` and `excludeDescendantScopeTypes` fields if you're using them?

AndreasArvidsson:
> What does this tool do with the `scopeTypes` and `excludeDescendantScopeTypes` fields if you're using them?

Discards them. These fields are not really used that much (There is only one use case as far as I know). And since we haven't decided on the format we are going to use in the future I want to avoid a community to community migration script if we decide to use something else.

pokey:
I would bail if you encounter them. Silently dropping data is not good ux

AndreasArvidsson:
You mean bail on the entire migration or just on that snippet?

pokey:
Maybe just that snippet? I don't feel strongly, as long as we raise a flag rather than silently losing data

AndreasArvidsson:
Totally agree. Here is the follow up.
https://github.com/cursorless-dev/cursorless/pull/2789

PR #2749, opened 2025-01-24
Disable Cursorless for vscode file dialogs

AndreasArvidsson:
File dialogs interfere with the command server keypress. Commands like `"paste to line"` just ends up with a timeout error in the talon log and nothing happens. 

I have been running with this in my own fork of cursorless Talon for a while to test it out and I'm happy with the result. 

## Checklist

### DIFF
--- a/cursorless-talon/src/apps/cursorless_vscode.py
+++ b/cursorless-talon/src/apps/cursorless_vscode.py
@@ -4,7 +4,7 @@ ctx = Context()
 
 ctx.matches = r"""
 app: vscode
-# Disable Cursorless for vscode file dialogs
+# Disable Cursorless for VS Code file dialogs (at least on Windows)
 not win.title: /^(Open Folder|Open File|Save As|Open Workspace from File|Add Folder to Workspace|Save Workspace)$/i
 """
 
--- a/cursorless-talon/src/apps/cursorless_vscode.py
+++ b/cursorless-talon/src/apps/cursorless_vscode.py
@@ -4,7 +4,8 @@ ctx = Context()
 
 ctx.matches = r"""
 app: vscode
-# Disable Cursorless for VS Code file dialogs (at least on Windows)
+# Disable Cursorless when VS Code is displaying a native OS dialog during which the command server
+# hotkey will not work.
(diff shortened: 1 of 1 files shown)


cursorless-talon/src/apps/cursorless_vscode.py
```diff
@@ -4,6 +4,8 @@
 
 ctx.matches = r"""
 app: vscode
+# Disable Cursorless for vscode file dialogs
+not title: /^(Open Folder|Open File|Save As|Open Workspace from file|Add Folder to Workspace|Save Workspace)$/i
```
phillco, line 8:
```suggestion
not win.title: /^(Open Folder|Open File|Save As|Open Workspace from File|Add Folder to Workspace|Save Workspace)$/i
```

phillco: I think either work fine and we should resolve the inconsistency in community as a followup, slight reference for this personally

cursorless-talon/src/apps/cursorless_vscode.py
```diff
@@ -4,6 +4,8 @@
 
 ctx.matches = r"""
 app: vscode
+# Disable Cursorless for vscode file dialogs
```
phillco, line 7:
```suggestion
# Disable Cursorless for VS Code file dialogs (at least on Windows)
```

cursorless-talon/src/apps/cursorless_vscode.py
```diff
@@ -4,6 +4,8 @@
 
 ctx.matches = r"""
 app: vscode
+# Disable Cursorless for VS Code file dialogs (at least on Windows)
```
phillco, line 7:
```suggestion
# Disable Cursorless when VS Code is displaying a native OS dialog during which the command server
# hotkey will not work.
```

phillco:
NB: quickly tested on macOS; there, they don't use the operating system pickers for these, they show a custom fuzzy picker UI. I was actually surprised that they don't do this on windows. When this view is shown the window title doesn't change, but we also don't need to change the command ever behavior, so it works out.

Screen Shot 2025-01-25 at 11 23 31 AM

phillco (approved):

PR #2754, opened 2025-01-25
Prevent overlapping pairs

### LINKED ISSUE #1318: <> pairs in files without a parse tree
Given this document:

```
(a < b)
(c > d)
```

"chuck pair cap" deletes between < and >. I would have expected it to delete between ( and ). I've been getting this false positive a lot because a `<` character anywhere in the file can trigger it. Maybe it would make sense to deprioritize <> pairs for the generic "pair" target if the opening/closing brackets are not on the same line.

AndreasArvidsson:
Matching pairs can be nested, but they can't be overlapping
eg `( [ ) ]`
In this example the parentheses are a actual pair since they came first, but the square brackets are not a pair. This corresponds to the syntax highlighting in vscode.

Fixes #1318

## Checklist


packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/getSurroundingPairOccurrences.ts
```diff
@@ -29,48 +22,29 @@ export function getSurroundingPairOccurrences(
...
+        openingDelimitersStack,
+        (o) =>
+          o.delimiterInfo.delimiterName === delimiterName &&
+          isSameTextFragment(o.textFragmentRange, textFragmentRange) &&
+          isValidLine(isSingleLine, o.range, range),
+      );
 
-      if (openingDelimiter == null) {
+      if (openingDelimiterIndex === -1) {
+        // When side is unknown and we can't find an opening delimiter, that means this is the opening delimiter.
```
phillco, line 37:
```suggestion
        // When side is unknown and we can't find an opening delimiter, that means this *is* the opening delimiter.
```

phillco (approved):

PR #2755, opened 2025-01-25
Improve out of range error

### LINKED ISSUE #1540: "Scope index out of range" error is confusing
image
I was trying to replace `CAPITALIZE_FIRST_WORD` with `capitalize_all_words` and said:

`bring second through fourth word blue trap to red cap`

I got:
```py
 > user.cursorless_multiple_target_command(cursorless_move_bring_action, cursorless_move_bring_targets)
Exception: Scope index out of range
```
And had to ask on Slack for help.  From what I understand Cursorless was trying to do something like:

`bring second through fourth word blue trap to second through fourth word red cap`

And as there was no fourth word in `CAPITALIZE_FIRST_WORD`, I got "Scope index out of range".

Things I would have loved to have understood about the error, ideally using the words that I used when specifying it.

- Which scope? (e.g., "word")
- Which index? (e.g., "4" or "fourth")
- Where was the scope being used (e.g, "in the target of a bring/move")

The last I can tell a bit from the stack trace but this would require me to have the Talon log visible.

And even better:
- What was **implied** or **assumed** in my command that triggered this error? (e.g., that the scope qualifiers used for the source of the move should also apply to the target)
- What could I do to tell Cursorless that I did not want it to assume this? (e.g., use "token" before "red cap".

AndreasArvidsson:
`Scope index out of range` is a bit confusing. At least add the scope type and witch index to the error message. 

Fixes #1540

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/listUtils.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/listUtils.ts
@@ -2,8 +2,8 @@ import type { ScopeType } from "@cursorless/common";
 
 export class OutOfRangeError extends Error {
   constructor(scopeType: ScopeType, index?: number) {
-    const indexStr = index != null ? ` ${index} ` : "";
-    super(`Index ${indexStr} out of range for '${scopeType.type}' scope`);
+    const numberStr = index != null ? ` #${index + 1} ` : "";
+    super(`Scope '${scopeType.type}'${numberStr} is out of range`);
     this.name = "OutOfRangeError";
   }
 }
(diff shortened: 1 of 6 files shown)


packages/cursorless-engine/src/processTargets/modifiers/listUtils.ts
```diff
@@ -1,27 +1,35 @@
+import type { ScopeType } from "@cursorless/common";
+
 export class OutOfRangeError extends Error {
-  constructor() {
-    super("Scope index out of range");
+  constructor(scopeType: ScopeType, index?: number) {
+    const indexStr = index != null ? ` ${index} ` : "";
+    super(`Index ${indexStr} out of range for '${scopeType.type}' scope`);
```
phillco, line 6:
```suggestion
    const numberStr = index != null ? ` #${index + 1} ` : "";
    super(`Scope '${scopeType.type}'${numberStr} is out of range`);
```

phillco (approved):

phillco (approved):

PR #2757, opened 2025-01-25
Added iteration scope for python arguments

### LINKED ISSUE #2729: python: No iteration scope for `"name"` and `"value"` in function call args
In the following python code:

```py
aaa(bbb=0, ccc=1, ddd=2)
```

Saying `"take every name bat"` should target `bbb`, `ccc`, and `ddd`, but doesn't, because we don't have an iteration scope inside those parens

AndreasArvidsson:
Fixes #2729

## Checklist


queries/python.scm
```diff
@@ -626,9 +626,9 @@
 ) @argumentOrParameter.iteration.domain
 
 (argument_list
-  "(" @argumentOrParameter.iteration.start.endOf
-  ")" @argumentOrParameter.iteration.end.startOf
-) @argumentOrParameter.iteration.domain
+  "(" @argumentOrParameter.iteration.start.endOf @name.iteration.start.endOf @value.iteration.start.endOf
```
phillco, line 629: NB: Andreas says there's a limit of three captures on each node and more are silently discarded by tree-sitter, so beware.

phillco: Ideally we should lint this - or adding to SCM tests

packages/common/src/scopeSupportFacets/python.ts
```diff
@@ -7,10 +7,16 @@ export const pythonScopeSupport: LanguageScopeSupportFacetMap = {
   "name.foreach": supported,
   "name.resource": supported,
   "name.resource.iteration": supported,
+  "name.argument.actual": supported,
```
phillco, line 10: NB: we could extend this to other languages that support keyword arguments like ruby, but we need to have familiarity with them. A quick check of https://github.com/cursorless-dev/cursorless/blob/b62d901a8be2529b537224268ffdd26d9fdd389d/queries/ruby.scm#L4 shows that it doesn't even support argument yet

data/fixtures/scopes/python/name.argument.actual.iteration.scope
```diff
@@ -0,0 +1,16 @@
+aaa(bbb=0, ccc=1)
```
phillco, line 1: Names come from compiler history -- an "actual" parameter is where it is passed, a formal parameter is where it is defined in the function definition (if there is one)

phillco (approved):

PR #2759, opened 2025-01-25
 Change insertion delimiter for javascript call scope 

### LINKED ISSUE #2650: When you 'clone call' in javascript it results in a syntax error
#### Description:
When using the "clone call" feature in TypeScript, the new call is separated from the old call with spaces. This behavior results in a syntax error, as TypeScript does not permit multiple function calls separated by spaces without proper delimiters. Consequently, users must manually intervene to correct this by adding a line break or a semicolon after cloning the call.

#### Example:
Consider the following TypeScript code:
```
function example() {
  myFunction();
  myFunction(); // Cloned call
}
```

Upon using "clone call," the result is:
```
function example() {
  myFunction();
  myFunction() myFunction(); // Syntax error
}
```

This occurs because the cloned call is appended with spaces, causing a syntax error.

#### Expected Behavior:
The expected behavior when using "clone call" is either to:
1. Insert a line break after the cloned call:
```
function example() {
  myFunction();
  myFunction();
  myFunction(); // Cloned call with line break
}
```

3. Append a semicolon after the cloned call:
```
function example() {
  myFunction();
  myFunction();; // Cloned call with semicolon
}
```

#### Impact:
This issue frequently arises when editing files with multiple function calls, such as in vitest files where each spec is its own call. The need for manual correction after each "clone call" operation disrupts workflow and reduces productivity.

#### Proposed Solution:
To address this issue, modify the "clone call" functionality to automatically insert a line break or a semicolon after the cloned call. While adding semicolons may be more complex to implement, it could provide a more robust solution.

#### Additional Notes:
This issue requires careful consideration as multiple function calls on the same line are possible, and the chosen solution should not break valid code patterns.

AndreasArvidsson:
Fixes #2650

## Checklist

### DIFF
--- a/data/fixtures/scopes/javascript.core/functionCall.constructor.scope
+++ b/data/fixtures/scopes/javascript.core/functionCall.constructor.scope
@@ -7,4 +7,4 @@ new Foo.bar(0);
   >--------------<
 0| new Foo.bar(0);
 
-[Insertion delimiter] = " "
+[Insertion delimiter] = ";\n"
--- a/data/fixtures/scopes/javascript.core/functionCall.scope
+++ b/data/fixtures/scopes/javascript.core/functionCall.scope
@@ -7,4 +7,4 @@ foo.bar(0);
   >----------<
 0| foo.bar(0);
 
-[Insertion delimiter] = " "
+[Insertion delimiter] = ";\n"
--- a/queries/javascript.core.scm
+++ b/queries/javascript.core.scm
@@ -503,28 +503,27 @@
   (#child-range! @private.switchStatementSubject 0 -1 true true)
 ) @_.domain
 
-;;!! foo()
-;;!  ^^^^^
-;;!! new Foo()
-;;!  ^^^^^^^^^
-[
-  (call_expression)
-  (new_expression)
-] @functionCall
-
-;;!! foo()
+;;!! foo();
+;;!  ^^^^^^
 ;;!  ^^^
 ;;!  -----
-(call_expression
-  function: (_) @functionCallee
-) @_.domain
+(
+  (call_expression
+    function: (_) @functionCallee
+  ) @functionCall @_.domain
+  (#insertion-delimiter! @functionCall ";\n")
+)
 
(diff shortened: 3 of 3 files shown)


phillco (approved):
tldr: easy fix, was always wrong, but Andreas (and presumably pokey) would just use "clone state" instead of "clone call" so they didn't run into this

phillco:
@pokey I was good on merging this, but then after discussing the underlying issue with Andreas, sort of realized it's a bit more ambiguous then I realized. The problem is that a call could be inside of the list, or a statement on its own, and there's not a single insertion delimiter that covers both (list would be `, `, etc).

Then we considered that the user should really just be targeting the specific scope that they want in the first place (item, statement, etc), rather than clone call assuming that they want a statement and inserting `;\n`. In general if the user is wanting clone call to insert a statement they are probably just using the wrong scope, and they should learn about statement.

On the other hand, the majority of the time, probably duplicating a statement is what they are trying to do and maybe this is a probabilistic quality life improvement. What's the precedence for this sort of design question?

pokey:
yeah in my opinion there is no natural delimiter for call, as it is not naturally a part of any delimited sequence. As you point out, a call could be an argument or could be the rhs of a statement just as easily as it could be a complete statement. But there is always the counterargument that it is better to do something useful than nothing at all, even if it is not well-founded from a theoretical perspective. My inclination would be not to have an insertion delimiter, but I don't feel strongly

phillco:
sgtm

jaresty:
fwiw, if the call was in an argument list or the right hand side of an assignment, 'clone call' wouldn't make much sense to do-in the former I think one would clone the argument (rather than the call), and in the latter I think you would clone the statement most likely (or perhaps the value)-I can hardly think of a situation where you would want to clone the call and add a space between them. I think there is a difference between clone and other operations here where cloning a call is a special case and the behavior may make sense for it to be different because it is a clone operation rather than something else. The most sensible place to insert a call would be another statement, though I recognize that it is an ambiguous case.

PR #2760, opened 2025-01-25
Make cheat sheet handle missing spoken forms

### LINKED ISSUE #2647: Make cheatsheet more robust
We had user that disabled a spoken form in the csv file which broke our cheatsheet
```
 > user.private_cursorless_cheat_sheet_show_html()
KeyError: 'applyFormatter'
```

Since we actually instruct people how to do that in our docs this needs to be made more robust.
https://www.cursorless.org/docs/user/customization/#example-of-custom-formatter-command

There's also a question if we actually want to read from the Talon registry or if we should read from the same data structure that constructs the spoken form json file?

AndreasArvidsson:
Today the cheat sheet generation crashes if the user has disabled certain spoken forms

Fixes #2647

## Checklist


cursorless-talon/src/cheatsheet/sections/actions.py
```diff
@@ -28,111 +28,144 @@ def get_actions():
...
+                        "description": "Copy <target> to <destination>",
+                    },
+                    {
+                        "spokenForm": f"{complex_actions['replaceWithTarget']} <target>",
+                        "description": "Insert copy of <target> at cursor",
+                    },
+                ],
+            }
+        )
+
```
phillco, line 60: There's a bit of duplication here and this switches from defining a data structure to a fair bit of imperative code, but I think we could reduce it.

Why not something like a dictionary of tuples (one description and one lambda that generates the spoken form?):

```python
{
    "replaceWithTarget": ("Copy <target> to <destination>", lambda value: f"{value} <target> <destination>"),
}
```

Then we could just use a loop to generate all of these and remove some the duplication. I've got into the habit of writing it using the dictionary and the loop to transform in the first place wherever possible because I'm lazy

Not sure if you could refer to stuff outside of the scope like `swap_connective` but I think it's helpful either way, if not that means the more complicated ones will stay as imperative code, which I think itself helps with readability.

AndreasArvidsson: Better like this?

cursorless-talon/src/cheatsheet/sections/modifiers.py
```diff
@@ -45,157 +50,238 @@ def get_modifiers():
...
+            }
+        )
+
+    if "forward" in complex_modifiers:
+        variations.append(
+            {
+                "spokenForm": f"<scope> {complex_modifiers['forward']}",
+                "description": "single instance of <scope> including target, going forwards",
+            }
+        )
```
phillco, line 198: Can we make these shorter per the method done above?

phillco (commented):
lgtm let's do the same magic on the others

cursorless-talon/src/cheatsheet/sections/modifiers.py
AndreasArvidsson: done

phillco (approved):

phillco (approved):

PR #2764, opened 2025-01-25
Flash targets action

### LINKED ISSUE #2271: As a user i would like a way to flash a target so i can practice cursorless language easily
Problem: there currently isn’t a way to test my understanding of the language that leaves the editor in the same state from which I started. Highlight is close, but it still requires another command to disable the highlight that was applied. I would like a command, similar to highlight but that just flashes the targets and then turns the highlight off. This would enable me to try out various targets in quick succession to validate my understanding of the language.

AndreasArvidsson:
Fixes #2271

## Checklist

### DIFF
--- a/data/fixtures/recorded/actions/flashToken.yml
+++ b/data/fixtures/recorded/actions/flashToken.yml
@@ -1,0 +1,30 @@
+languageId: plaintext
+command:
+  version: 7
+  spokenForm: flash token
+  action:
+    name: flashTargets
+    target:
+      type: primitive
+      modifiers:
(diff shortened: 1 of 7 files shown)


data/fixtures/recorded/actions/flashToken.yml
```diff
@@ -0,0 +1,30 @@
+languageId: plaintext
```
phillco, line 1: This is a funny test but it does actually test more than you would think

phillco (approved):

PR #2766, opened 2025-01-26
Use generalized ranges in breakpoint action

AndreasArvidsson:
We had a specific type for breakpoint descriptions which is more or less the same format as our generalized ranges that we use in other places. 

## Checklist

### DIFF
--- a/packages/common/src/types/TextEditor.ts
+++ b/packages/common/src/types/TextEditor.ts
@@ -1,5 +1,6 @@
 import type {
   Edit,
+  GeneralizedRange,
   Range,
   RevealLineAt,
   Selection,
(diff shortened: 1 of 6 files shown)


packages/common/src/types/TextEditor.ts
```diff
@@ -237,21 +238,3 @@ export interface EditableTextEditor extends TextEditor {
...
-   */
-  endLine: number;
-}
-
-interface InlineBreakpointDescriptor {
-  type: "inline";
-  range: Range;
-}
-
-export type BreakpointDescriptor =
```
AndreasArvidsson, line 255: These have been replaced by `GeneralizedRange`

packages/cursorless-engine/src/actions/ToggleBreakpoint.ts
```diff
@@ -22,25 +25,13 @@ export default class ToggleBreakpoint implements SimpleAction {
...
-                endLine: range.end.line,
-              }
-            : {
-                type: "inline",
-                range,
-              };
-        },
+      const generalizedRanges = targets.map((target) =>
+        toGeneralizedRange(target),
       );
```
phillco, line 30:
```suggestion
      const generalizedRanges = targets.map(toGeneralizedRange(target));
```

phillco (approved):

PR #2768, opened 2025-01-26
Fix bug with `"every token inside"`

### LINKED ISSUE #1806: "last token inside" refers to tokens outside when there is only one token
```
"aaa"
```

`"take last token inside air"` selects the trailing `"`, not `aaa`. Similarly for first token. `take every token inside` is also entertaining.

Pretty sure this is a bug.

AndreasArvidsson:
Fixes #1806

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/EveryScopeStage.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/EveryScopeStage.ts
@@ -61,7 +61,8 @@ export class EveryScopeStage implements ModifierStage {
 
       if (
         scopes.length === 1 &&
-        scopes[0].domain.contains(target.contentRange)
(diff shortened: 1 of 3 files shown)


packages/cursorless-engine/src/processTargets/modifiers/EveryScopeStage.ts
```diff
@@ -61,7 +61,8 @@ export class EveryScopeStage implements ModifierStage {
 
       if (
         scopes.length === 1 &&
-        scopes[0].domain.contains(target.contentRange)
+        scopes[0].domain.contains(target.contentRange) &&
+        !target.hasExplicitScopeType
```
phillco, line 65: hats + this are implicit, others are explicit

phillco (approved):

PR #2776, opened 2025-01-26
Added scope type for `"inside"` 

### LINKED ISSUE #1046: Add a proper scope type for `"inside"`
It will be a cascading scope type (#1045) combining the following:

One of the constituent scope types will be just be surrounding pair, but where we ask pair for `inside` before returning it in `getTarget()`; domain is just the surrounding pair

The other constituent scope type is a tree-sitter syntactic scope handler with a tree-sitter pattern that languages can define however they want.  For example, Python would define it to be [`body`](https://github.com/tree-sitter/tree-sitter-python/blob/de221eccf9a221f5b85474a553474a69b4b5784d/grammar.js#L473)

Note that the `"inside"` modifier won't just be simply containing instance of this new scope.  It will first take its input target and see if it has an `interior` attribute.  If so, it will use it, if not it will be containing instance of this scope

AndreasArvidsson:
Inside/interior today is a modifier and will stay so in the future, but we are also introducing interior as a scope type. This will allow us to do `"take inside"` when on a if statement, python function and so on.

It now takes the closest of the nearest tree-sitter interior _or_ surrounding pair. This is most useful for Python; in other languages with {} statement-blocks, the braces would already be easy to target, but now "take inside" behaves intuitively there as well.

The scm syntax `@scope.interior` have been replaced by `@interior` as its own scope type.

Fixes #1046

## Checklist

## Release notes

You can now use "inside" to refer to the inside body of a language scope (function, if statement, etc). This is especially useful for Python, because previously you had to think about what scope you were targeting -- "take inside" now just works.

It also works nicely for targeting the inside of C-style statements that don't have `{}` braces, like single line conditionals or loops.

In Python, this also means that you can now target the inside of the function ("take inside") even while the cursor is in a related location (such as in its name). Note that this behavior currently only exists for Python; we plan to port it to other languages as well since it is useful.


packages/common/src/scopeSupportFacets/scopeSupportFacetInfos.ts
```diff
@@ -705,6 +705,35 @@ export const scopeSupportFacetInfos: Record<
...
+  "interior.lambda": {
+    description: "The body of a lambda/anonymous function",
+    scopeType: { type: "interior" },
+  },
+  "interior.branch": {
+    description: "The body of an if/elif/else branch",
+    scopeType: { type: "interior" },
+  },
+  "interior.element": {
+    description: "The interior/children of an xml element",
```
phillco, line 725:
```suggestion
    description: "The interior/children of an XML element",
```

packages/common/src/scopeSupportFacets/scopeSupportFacetInfos.ts
```diff
@@ -705,6 +705,35 @@ export const scopeSupportFacetInfos: Record<
...
+    scopeType: { type: "interior" },
+  },
+  "interior.command": {
+    description: "The body of a Talon command",
+    scopeType: { type: "interior" },
+  },
+  "interior.cell": {
+    description: "The body of code cell in markdown",
```
phillco, line 733:
```suggestion
    description: "The body of a code cell in markdown",
```

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/InteriorScopeHandler.ts
```diff
@@ -0,0 +1,149 @@
...
+      },
+      this.languageId,
+    );
+
+    // If the scope type is explicit (ie, the user has specified a scope
+    // type), then we want to prioritize language scopes. The user might
+    // have said something like "inside element" and then we don't want to
+    // yield the interior of the `<div>` pair first.
+
+    if (this.scopeType.explicitScopeType) {
```
phillco, line 87:
```suggestion
    if (this.scopeType.explicitScopeType) {
```

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/InteriorScopeHandler.ts
```diff
@@ -0,0 +1,147 @@
...
+      [languageScopeHandler, pairScopeHandler],
+      this.languageId,
+    );
+  }
+
+  private shouldYield(targetDomain: Range, scope: TargetScope): boolean {
+    // For an explicit scope type we only yield scopes that are contained within
+    // the target domain. E.g the user said "inside token", then we don't want
+    // to yield scopes that are larger than the token. The definition of an
+    // interior is that it's inside the scope.
```
phillco, line 118:
```suggestion
    // For an explicit scope type we only yield scopes that are contained within
    // the target domain. For example, if the user said "inside token", we don't
    // want to yield scopes that are larger than the token.
```

packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/SurroundingPairScopeHandler/InteriorScopeHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+        type: "surroundingPair",
+        delimiter: "any",
+      },
+      this.languageId,
+    );
+
+    // If the scope type is explicit (ie, the user has specified a scope
+    // type), then we want to prioritize language scopes. The user might
+    // have said something like "inside element" and then we don't want to
+    // yield the interior of the `<div>` pair first.
```
phillco, line 85:
```suggestion
    // If the scope type is explicit (ie, the user has specified a scope
    // type), then we want to prioritize language scopes. For example,
    // if the user says "inside element" inside a `<div>` tag, the angle
    // brackets of the tag are also a surrounding pair which should have
    // lower priority.
```

data/fixtures/recorded/modifiers/interior/changeInside4.yml
```diff
@@ -0,0 +1,22 @@
...
+  spokenForm: change inside
+  action:
+    name: clearAndSetSelection
+    target:
+      type: primitive
+      modifiers:
+        - {type: interiorOnly}
+  usePrePhraseSnapshot: false
+initialState:
+  documentContents: <div>content</div> (hello)
```
phillco, line 13: It would be neat if this would give you this:

```<div>|</div> (|)```

phillco (approved):

PR #2778, opened 2025-01-26
Update ranges for markdown item

### LINKED ISSUE #2384: In markdown "chuck item <target>" and "take item <target>" select different things
In markdown "chuck item <target>" and "take item <target>" select different things at times. I noticed this when I had a file with a markdown list. "chuck item" would drop the first level and second level of the list item. "Take item" would only select the first level of the list. My file looked something like the following:

- This is the first level of a list.
  - This is the second level of a list. 

This may be related to #585

AndreasArvidsson:
Fixes #2384

## Checklist

### DIFF
--- a/data/fixtures/recorded/languages/markdown/changeLeadingItem.yml
+++ b/data/fixtures/recorded/languages/markdown/changeLeadingItem.yml
@@ -24,8 +24,9 @@ initialState:
 finalState:
   documentContents: |-
     - values
-        - 0- 1
+        - 0
+        - 1
         - 2
   selections:
-    - anchor: {line: 1, character: 7}
-      active: {line: 1, character: 7}
+    - anchor: {line: 2, character: 4}
+      active: {line: 2, character: 4}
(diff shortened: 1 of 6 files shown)


data/fixtures/recorded/languages/markdown/changeLeadingItem.yml
```diff
@@ -24,8 +24,9 @@ initialState:
 finalState:
   documentContents: |-
     - values
-        - 0- 1
```
phillco, line 27: lol that before

data/fixtures/recorded/languages/markdown/roundWrapItemAir.yml
```diff
@@ -26,8 +26,8 @@ initialState:
       end: {line: 0, character: 5}
 finalState:
   documentContents: |
-    - (aaa)
-      - bbb
+    - (aaa
```
phillco, line 29: You now need to use `line item` to get the old behavior

phillco: Probably better to include the children by default (the new way) because you can easily narrow the scope but it's harder to widen it arbitrarily

phillco (approved):

PR #2792, opened 2025-01-30
Clean up one of scope handler

AndreasArvidsson:
Passing in the scope type is only necessary if you're creating the scope handlers from that scope type. In the case where you're creating a one of handler from a list of scope handlers. Passing in the scope type as well just makes the code more verbose for no reason.

## Checklist

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/OneOfScopeHandler.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/OneOfScopeHandler.ts
@@ -81,7 +81,7 @@ export class OneOfScopeHandler extends BaseScopeHandler {
     direction: Direction,
     hints: ScopeIteratorRequirements,
   ): Iterable<TargetScope> {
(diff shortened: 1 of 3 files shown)


packages/cursorless-engine/src/processTargets/modifiers/scopeHandlers/OneOfScopeHandler.ts
```diff
@@ -74,21 +75,13 @@ export class OneOfScopeHandler extends BaseScopeHandler {
...
-  }
-
   *generateScopeCandidates(
     editor: TextEditor,
     position: Position,
     direction: Direction,
     hints: ScopeIteratorRequirements,
   ): Iterable<TargetScope> {
-    // If we have used the iteration scope handler, we only want to yield from its handler.
+    // If we have used a iteration scope handler, we only want to yield additional scopes from its handler.
```
phillco, line 84:
```suggestion
    // If we have used an iteration scope handler, we only want to yield additional scopes from its handler.
```

phillco (approved):

PR #2806, opened 2025-02-01
Fixed bug with edit new

AndreasArvidsson:
If you said `"pour state air and line bat"` where `air` was a multiline statement you got an undefined exception at runtime.

I have corrected the problem with our logic and also removed a not null assertion in favor of actually checking for null.

## Release notes
`"pour state air and line bat"` (if `air` was a multiline statement) now works instead of raising an exception.

### DIFF
--- a/packages/cursorless-engine/src/actions/EditNew/EditNew.ts
+++ b/packages/cursorless-engine/src/actions/EditNew/EditNew.ts
@@ -35,6 +35,7 @@ export class EditNew {
      */
     let state: State = {
       destinations,
+      actionTypes: destinations.map((d) => d.getEditNewActionType()),
       thatRanges: destinations.map(
         ({ target }) => target.thatTarget.contentRange,
       ),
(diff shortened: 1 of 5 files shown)


packages/cursorless-engine/src/actions/EditNew/EditNew.ts
```diff
@@ -63,9 +64,14 @@ export class EditNew {
       !useInsertLineAfter,
     );
 
-    const newSelections = state.destinations.map((destination, index) =>
-      state.cursorRanges[index]!.toSelection(destination.target.isReversed),
```
AndreasArvidsson, line 67: `state.cursorRanges[index]` was undefined

packages/cursorless-engine/src/actions/EditNew/EditNew.types.ts
```diff
@@ -29,6 +32,13 @@ export interface State {
    */
   destinations: Destination[];
 
+  /**
+   * The action types for each destination. Important to read from this and not
+   * the destinations themself since they get updated during the process am we want
+   * the default value for this.
+   */
+  actionTypes: EditNewActionType[];
+
```
phillco, line 41:
```suggestion
  /**
   * The action types for each destination. It's important to read from this and not
   * the destinations themselves, since they are changed as the action runs and
   * we make multiple passes.
   */
  actionTypes: EditNewActionType[];

```

phillco (approved):

Issue #2391, opened 2024-06-06
Revisit `compile` / `build` npm scripts


pokey:
We're not very consistent with our terminology around `compile` and `build` in the various packages of our monorepo. We should

- [ ] Look at `compile` / `build` in each of our packages, as well as the root package, to determine what they're doing
- [ ] Come up with consistent terminology. I'd be tempted to remove `compile` entirely, and instead have `build` for anything we're building and `lint:tsc` to replace the way in which `compile` does type checking today. Because we don't actually emit anything from typescript, it's a bit of a misnomer to call it `compile`
- [ ] Revisit whether the contributor really needs to run `pnpm compile` during initial install (see https://github.com/cursorless-dev/cursorless/blob/1183dd3cd4991db871536599252ec782293a1512/docs/contributing/CONTRIBUTING.md#L23
- [ ] Investigate impact on CI

PR #2808, opened 2025-02-01
Make work flow files more readable

### LINKED ISSUE #2796: Increase readability of git workflows
Today our yaml files are getting a bit hard to read. Actually adding names to each step and adding some empty lines would greatly increase readability.

AndreasArvidsson:
* Added name to each work flow step
* Added padding between steps

Fixes #2796

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -87,7 +87,7 @@ jobs:
           CURSORLESS_DEPLOY: true
 
       - name: Checkout cursorless.nvim plugin repo
-       uses: actions/checkout@v4
+        uses: actions/checkout@v4
         with:
           token: ${{ secrets.CURSORLESS_BOT_TOKEN }}
           repository: hands-free-vim/cursorless.nvim
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -13,26 +13,40 @@ jobs:
     environment: production
     env:
       CURSORLESS_REPO_ROOT: ${{ github.workspace }}
+
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy.yaml
```diff
@@ -13,26 +13,40 @@ jobs:
...
+
+      - name: Install dependencies
+        run: pnpm --color install
+
+      - name: Compile
+        run: pnpm --color compile
+
+      - name: Build
+        run: pnpm --color --filter '!cursorless-org' --filter '!cursorless-org-*' build
         env:
```
phillco, line 40: https://github.com/cursorless-dev/cursorless/issues/2391

It makes sense to run compile separately if we think it's faster / higher signal/ROI to read. There's some question about whether `build` would do everything necessary if compile was skipped, which is a bit concerning, but we did not test this.

.github/workflows/pre-commit.yml
```diff
@@ -16,19 +17,37 @@ jobs:
...
+        uses: leafo/gh-actions-lua@v9
+
+      - name: Install luarocks
+        uses: leafo/gh-actions-luarocks@v4
+
+      - name: Run pre-commit hooks
+        uses: pre-commit/action@v3.0.1
+
+      - name: Auto-fix pre-commit issues
+        uses: pre-commit-ci/lite-action@v1.0.2
```
phillco, line 52: Open questions: why do we run both the lite sweet and the normal one? Why `if: always()`?

For the former guessing that the normal suite doesn't automatically correct issues but we have some tests that can't run in it.

phillco (approved):

PR #2809, opened 2025-02-01
Log fail test to file went running tests locally

### LINKED ISSUE #2805: Ability to copy paths to failing tests
When your run the test locally would be nice if there was a button you could click to copy the paths to all failing tests that you could paste into `testSubsetGrep.properties`

AndreasArvidsson:
The issue suggested showing a pop-up, but this wasn't optimal. It was either blocking the tests from terminating or it closed immediately by itself. I also don't want to clobber the clipboard all the time.

Let's log to a file because it seems like the best of the options.

Fixes #2805

### DIFF
--- a/packages/test-harness/src/runAllTests.ts
+++ b/packages/test-harness/src/runAllTests.ts
@@ -1,8 +1,13 @@
+import { getCursorlessRepoRoot } from "@cursorless/node-common";
+import { glob } from "glob";
 import Mocha from "mocha";
 import * as path from "node:path";
-import { getCursorlessRepoRoot } from "@cursorless/node-common";
-import { runTestSubset, testSubsetGrepString } from "./testSubset";
-import { glob } from "glob";
+import {
+  logFailedTests,
+  runTestSubset,
+  shouldLogFailedTests,
+  testSubsetGrepString,
+} from "./testSubset";
 
 /**
  * Type of test to run, eg unit, vscode, talon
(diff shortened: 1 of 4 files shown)


packages/test-harness/src/runAllTests.ts
```diff
@@ -68,14 +73,25 @@ async function runTestsInDir(
...
         } else {
-          c();
+          resolve();
         }
       });
+
+      if (shouldLogFailedTests()) {
+        runner.on("fail", (test) => {
+          failedTests.push(test.fullTitle());
+        });
```
phillco, line 93:
```suggestion
        runner.on("fail", (test) => failedTests.push(test.fullTitle()));
```

packages/test-harness/src/testSubset.ts
```diff
@@ -37,3 +51,11 @@ export function testSubsetFilePath() {
 export function runTestSubset() {
   return process.env.CURSORLESS_RUN_TEST_SUBSET === "true";
 }
+
+/**
+ * Determine whether we should log the failed tests.
+ * @returns `true` if we should log the failed tests
```
phillco, line 57:
```suggestion
 * Determine whether we should log the failed tests to a file. This makes it easier to put them in `testSubsetGrep.properties` for faster iterating.
 * @returns `true` if we should log failed tests to `packages/test-harness/failedTests.properties`
```

phillco (approved):

PR #2810, opened 2025-02-01
Actions: attempt to validate PR description tests

phillco:
GitHub includes the description comment when it formulates the commit message, which we don't want as it's pretty messy. We will ask users to remove the comment but we need to validate it before allowing merging.

Stolen from https://github.com/orgs/community/discussions/84771#discussioncomment-8039595.

### DIFF
--- a/.github/workflows/pr-description.yaml
+++ b/.github/workflows/pr-description.yaml
@@ -5,24 +5,24 @@ on:
     types: [opened, edited, reopened]
 
 jobs:
-  validate-pr-description:  
+  validate-pr-description:
     runs-on: ubuntu-latest
     steps:
-    - name: Set up workspace
-      uses: actions/checkout@v2
+      - name: Set up workspace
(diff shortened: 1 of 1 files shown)


AndreasArvidsson (approved):

phillco:
A few follow up changes were required for this to work:

```
  pull_request:
    types: [synchronize]
```

... Otherwise, if you pushed a change but did not edit the description or attributes of the pull request, the status would be missing on the most recent commit. Because it's a required status, it needs to be present on the latest commit before you can queue to merge. The synchronize event runs after new commits are pushed to the pull request branch.

```
  merge_group:
    branches: [main]
```

.. Otherwise the job wouldn't be retriggered after the change is added to the merge queue

phillco:
Tested with https://github.com/cursorless-dev/cursorless/pull/2809 that it won't be queued to automatically merge until the check passes

PR #2825, opened 2025-02-05
Use pnpm/action-setup in workflows

AndreasArvidsson:


.github/workflows/pr-description.yaml
```diff
@@ -10,8 +10,8 @@ jobs:
   validate-pr-description:
     runs-on: ubuntu-latest
     steps:
-      - name: Set up workspace
-        uses: actions/checkout@v2
+      - name: Checkout repository
```
AndreasArvidsson, line 13: This is for consistency with the other workflows

.github/workflows/deploy.yaml
```diff
@@ -17,15 +17,9 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-        with:
```
AndreasArvidsson, line 20: We don't need fetch depth 0. Just a waste of resources.

auscompgeek: Don't we need it to count the number of commits here?

AndreasArvidsson: You might be right. This line would require fetch depth 0?
https://github.com/cursorless-dev/cursorless/blob/0795f9d52848bff38617e65dbd6bc47ad59bd540/packages/cursorless-vscode/src/scripts/populateDist/transformPackageJson.ts#L27

AndreasArvidsson: Restored

phillco (approved):

PR #2827, opened 2025-02-05
Add missing javascript scope facets

AndreasArvidsson:


packages/common/src/scopeSupportFacets/typescript.ts
```diff
@@ -7,8 +7,6 @@ const { supported } = ScopeSupportFacetLevel;
 export const typescriptScopeSupport: LanguageScopeSupportFacetMap = {
   ...javascriptCoreScopeSupport,
 
-  "name.field": supported,
```
AndreasArvidsson, line 10: These are already defined by javascript core

pokey (approved):

PR #2837, opened 2025-02-07
Support legacy snippets

AndreasArvidsson:
Still support Cursorless (legacy) snippets extension side


pokey (approved):
Painful to see this code come back, but unfortunately I think it's necessary. Presumably the VscodeSnippets.ts file was just around to perform migration, and will disappear entirely in the future?

AndreasArvidsson:
> Painful to see this code come back, but unfortunately I think it's necessary. Presumably the VscodeSnippets.ts file was just around to perform migration, and will disappear entirely in the future?

Used for the migration, but also `openNewSnippetFile` is used for the generate snippet action. That is a single function I'm tempted to my grate that to the ide surface once the legacy snippet files are removed again.

PR #2842, opened 2025-02-10
Remove snippets from spoken form json

### LINKED ISSUE #2841: Remove snippet csv support
Is there a reason we're keeping the snippet csv support around? From what I can tell everything else has been removed. See eg https://github.com/cursorless-dev/cursorless/blob/be953a55e2fe17a1701ced1dab56ff4cae91ef6f/cursorless-talon/src/spoken_forms.json#L236-L257

Without it being removed, the files keep coming back after I deleted them

AndreasArvidsson:
Read from deprecated snippet csv files if the exists, but do not create them if they don't.

Fixes #2841

### DIFF
--- a/cursorless-talon/src/csv_overrides.py
+++ b/cursorless-talon/src/csv_overrides.py
@@ -59,7 +59,7 @@ def csv_get_normalized_ctx():
 
 def init_csv_and_watch_changes(
     filename: str,
-    default_values: ListToSpokenForms,
+    default_values: ListToSpokenForms | None,
     handle_new_values: Optional[Callable[[list[SpokenFormEntry]], None]] = None,
     *,
     extra_ignored_values: Optional[list[str]] = None,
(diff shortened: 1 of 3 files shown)


cursorless-talon/src/csv_overrides.py
```diff
@@ -151,7 +158,7 @@ def on_watch(path, flags):
                 handle_new_values=handle_new_values,
             )
 
-    fs.watch(str(file_path.parent), on_watch)
+    fs.watch(file_path.parent, on_watch)
```
AndreasArvidsson, line 161: Type change in Talon. You can pass either a path instance or a string and both will work, but only a path object does not give type errors.

cursorless-talon/src/spoken_forms.json
```diff
@@ -233,32 +233,6 @@
       "-from": "experimental.setInstanceReference"
     }
   },
-  "experimental/wrapper_snippets.csv": {
```
pokey, line 236: I wonder if we want to be more explicit. Just allowing us to leave out things feels like it's making it easier to make mistakes. Maybe we use a string `"DEPRECATED"`?

pokey (approved):
looks good with minor comment

cursorless-talon/src/spoken_forms.json
AndreasArvidsson: We still need to specify the file name otherwise we don't know which csv file to read if it exists. I could of course add an extra argument to init_csv_and_watch_changes instead of passing none as the default values?

AndreasArvidsson: I went ahead with a boolean solution

PR #2853, opened 2025-02-12
Migrate css and scss to Tree sitter queries

AndreasArvidsson:


queries/css.scm
```diff
@@ -1,34 +1,140 @@
...
+      "("
+      ","
+      (
+        (plain_value) @_dummy
+        (#eq? @_dummy "at")
+      )
+    ]
+    .
+    (_) @argumentOrParameter
+    (
```
pokey, line 75: how come you can't do something like the following instead of adding the new `grow-to-named-siblings!` predicate?

```suggestion
    (_) @argumentOrParameter.start
    (_)? @argumentOrParameter.end
    (#not-eq? argumentOrParameter.end "at")
    .
    (
```

AndreasArvidsson: I could never get something like that to work. Your example here doesn't work either. I would happily accept a solution like this that works, but I couldn't find it on my own.

PR #2854, opened 2025-02-12
Update `failedTests.properties` on success

AndreasArvidsson:
I got tricked by `failedTests.properties` not updating when tests were successful.

### DIFF
--- a/packages/test-harness/src/runAllTests.ts
+++ b/packages/test-harness/src/runAllTests.ts
@@ -77,10 +77,11 @@ async function runTestsInDir(
       const failedTests: string[] = [];
 
       const runner = mocha.run((failures) => {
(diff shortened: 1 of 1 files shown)


pokey:
Sorry, can you elaborate a bit?

AndreasArvidsson:
Certainly. Now whenever you run a test that fails we update `failedTests.properties` with the test names. This makes it easier to one by one open all the fail test and manually rerun and evaluate them deeper.
The problem was that Iran test that failed. I then fixed the bug,. I ran the test once again and this time they succeeded, but the fail test filed wasn't updated so when I looked at that I still fought that the test had failed. I definitely believe that `failedTests.properties` should be updated on success as well.

pokey (approved):
Ah cool didn't know bout that functionality

PR #2859, opened 2025-02-19
Show languages in the docs sidebar

AndreasArvidsson:
So far our implementation of scope facets have been lacking and therefore we made the decision to hide languages in the docs sidebar. Now our support for scope facets in the most common languages is quite good so I would like to show languages.


pokey:
I don't think these pages are particularly helpful in their current form. I don't think we can expect users to know how to read these raw facet ids. We should probably organise them by scope, and then make the names easier to read / show descriptions. As it stands I think they'll be more confusing than useful

AndreasArvidsson:
I think my initial intent was to add them to the developer section of the docs. It's still really useful for a developer to be able to see what is missing from a particular language. It's hard in the source code to see which keys are missing since we have so many.

pokey:
> Grouping them by scope and showing description shouldn't be any problems though. I can definitely do that.

Yeah and maybe splitting on `.` and removing scope name to make the names look a bit more human-readable

Also I'd prob hide the unknown ones in a details expander. And unsupported I wonder how useful to show to users. If we do we should probably make not too prominent and link to contributor docs if they wanna roll up their sleeves

AndreasArvidsson:
Sounds good. I start working on that.

AndreasArvidsson:
@pokey I have now updated the scope facets with expandable cards and grouped scopes.

packages/common/src/util/capitalize.ts
```diff
@@ -0,0 +1,3 @@
+export function capitalize(str: string) {
```
AndreasArvidsson, line 1: Ludicrous that this is not part of the javascript standard.

packages/cursorless-org-docs/src/css/custom.css
```diff
@@ -28,3 +28,19 @@
 .hidden {
   display: none;
 }
+
+.card {
```
AndreasArvidsson, line 32: As far as I could tell we don't really have a css library for the docs page? I didn't want to add one for this simple use case.

packages/cursorless-org-docs/src/docs/user/languages/c.mdx
```diff
@@ -1,4 +1,4 @@
-import Language from "./Language";
+import { Language } from "./components/Language";
```
AndreasArvidsson, line 1: We're getting more components so I move them into a component's folder. I also prefer named imports.

packages/cursorless-org-docs/src/docs/user/languages/components/ScopeSupportForLevel.tsx
```diff
@@ -0,0 +1,119 @@
...
+}
+
+function prettifyScopeType(scopeType: string): string {
+  return capitalize(camelCaseToAllDown(scopeType));
+}
+
+function prettifyFacet(facet: ScopeSupportFacet): string {
+  const parts = facet.split(".").map(camelCaseToAllDown);
+  if (parts.length === 1) {
+    return capitalize(parts[0]);
```
AndreasArvidsson, line 102: Some facets just have the same name as the scope. In that case I just keep it.

image
image

packages/cursorless-org-docs/src/docs/user/languages/c.mdx
pokey: agreed on all counts

AndreasArvidsson:
When I try details I don't get the same formatting as in the examples

image

Probably because we're using react instead of putting the details tag directly in markdown. When I try to put the details tag directly in markdown it works.

PR #2860, opened 2025-02-22
Don't discard nodes with non adjacent error siblings

AndreasArvidsson:
Today we discard parse tree nodes if any of the ancestral nodes has a error. That means that any of the ancestral nodes has a descendant that is an error node. That could be something a hundred lines below that isn't really relevant to the node we are evaluating. This updates changes so we only look at error nodes that are adjacent to our ascending path.

### DIFF
--- a/data/fixtures/recorded/surroundingPair/changePair.yml
+++ b/data/fixtures/recorded/surroundingPair/changePair.yml
@@ -1,0 +1,25 @@
+languageId: typescriptreact
+command:
+  version: 7
+  spokenForm: change pair
+  action:
+    name: clearAndSetSelection
+    target:
+      type: primitive
+      modifiers:
+        - type: containingScope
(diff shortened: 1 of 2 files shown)


data/fixtures/recorded/surroundingPair/changePair.yml
```diff
@@ -0,0 +1,25 @@
...
+    name: clearAndSetSelection
+    target:
+      type: primitive
+      modifiers:
+        - type: containingScope
+          scopeType: {type: surroundingPair, delimiter: any}
+  usePrePhraseSnapshot: false
+initialState:
+  documentContents: |
+    <div bbb={() => null}>hello &</div>
```
AndreasArvidsson, line 15: This test fails on main. `&` is not a valid syntax in jsx. Because of this `>` in  `=>` was incorrectly used as a pair delimiter.

PR #2861, opened 2025-02-22
Make missing snippet message include languages

AndreasArvidsson:
image


packages/cursorless-engine/src/core/getPreferredSnippet.ts
```diff
@@ -37,12 +37,22 @@ export function getPreferredSnippet(
...
   }
 
   return preferredSnippet;
 }
 
+function getUniqueLanguagesString(snippets: CustomInsertSnippetArg[]): string {
+  const languages = new Set(
+    snippets.flatMap((snippet) => snippet.languages ?? []),
+  );
+  return Array.from(languages).sort().join(", ");
```
phillco, line 53: If this is just building the list for this one use case to construct the error message, I would just inline it. It's only two lines

phillco (approved):

PR #2870, opened 2025-03-04
Release 1.0.0

AndreasArvidsson:


packages/cursorless-org-docs/src/docs/user/release-notes/0.28.0.md
```diff
@@ -1,7 +1,6 @@
 ---
 version: "0.28.0"
 release_date: "2023-09-01"
-sidebar_position: -2
```
AndreasArvidsson, line 4: Having to change all of these every time we add a release notes is somewhat of a hassle

packages/cursorless-org-docs/src/docs/user/release-notes/1.0.md
```diff
@@ -0,0 +1,19 @@
+---
+version: "1.0"
+release_date: "2025-03-04"
+---
+
+# Release Notes for 1.0
+
+Cursorless if finally ready for its 1.0 release!
+
+Over the last year the Cursorless team has focused primarily on the user experience and migrating legacy language scopes. We are now at a point where we feel Cursorless is ready for its first official release.
```
AndreasArvidsson, line 10: I went with the terminology `official release`. I didn't want to use `stable` because that implies that Cursorless was unstable before.

packages/cursorless-vscode/package.json
```diff
@@ -20,7 +20,7 @@
       "description": "Tests"
     }
   ],
-  "version": "0.29.0",
+  "version": "1.0.0",
```
AndreasArvidsson, line 23: I think this should be enough for the new version to be used.

pokey (commented):
Will take a closer look but first thing is should be semantic versioning so 1.0.0

packages/cursorless-vscode/package.json
pokey: Not sure have a look at the last release should be the same

AndreasArvidsson: Yeah that was all

AndreasArvidsson:
> Will take a closer look but first thing is should be semantic versioning so 1.0.0

Please do. I've updated all references to `1.0.0`

pokey (approved):
Ok I made a bunch of changes because turned out most of the AI-generated stuff was plausible sounding, but wrong 😅. Wasn't planning to invest in this, but figured I might as well be involved in the 1.0.0 release; kind of an important milestone for the project

PR #2872, opened 2025-03-11
Bump all dependencies

pokey:


pokey:
Ok I've run out of steam. We should prob downgrade eslint and go from there

pokey:
Ah ok looks like we need to merge all our eslint configs into the root one for eslint 9; see https://github.com/eslint/eslint/discussions/16960#discussioncomment-5212286

pokey:
ok I give up. Closing but @AndreasArvidsson feel free to reopen if you want to take a crack

AndreasArvidsson:
btw do you get the same error locally on macOS? Or is this a GitHub problem?

PR #2876, opened 2025-03-23
Add initial CLAUDE.md

pokey:
We can ask claude to keep updating it as we make changes


pokey:
my one concern is it duplicates stuff from contributing guide. I wonder if there's a way to get claude to just always read that 🤔

AndreasArvidsson (approved):

PR #2890, opened 2025-04-16
Add missing JS scope facets

AndreasArvidsson:


pokey (commented):
In this code:

```
class Aaa {
  bbb() {
    ccc;
  }
  ddd() {}
}
```

Will `"every funk name cap"` still return `bbb` and `ddd`?

AndreasArvidsson:
> In this code:
> 
> ```ts
> class Aaa {
>   bbb() {
>     ccc;
>   }
>   ddd() {}
> }
> ```
> 
> Will `"every funk name cap"` still return `bbb` and `ddd`?

No. Is `namedFunction.iteration.block` and `functionName.iteration.block` really facets we want? Shouldn't it just be `namedFunction.iteration.class`?

pokey:
Yeah maybe we remove those? Where are they used? I mean I can see the use case in certain scenarios but it's a bummer to lose every funk within a method

AndreasArvidsson:
@pokey Ready for another view. I haven't renamed any of the facets. Besides that I'm feel this one is done now.

PR #2891, opened 2025-04-16
Remove PR template

AndreasArvidsson:
We haven't gotten a parser for the pull requests made yet and so far we are still using the change log directory. Removing this for now and we can readd it once we actually have a parser for our pull requests.


pokey:
Do we not want to just delete that file? Seems like you've just emptied it

AndreasArvidsson:
Just a provision in case we want to add something in the future, but I can remove the file instead if you prefer?

pokey:
I see no reason to keep an empty file around. We can always just add it back if we need it

AndreasArvidsson:
Ok template file is now removed

auscompgeek:
> I see no reason to keep an empty file around.

I'd imagine checking in an empty template would still override GitHub's default behaviour of using the commit message. Which if anything would probably surprise contributors.

pokey (approved):

AndreasArvidsson:
> > I see no reason to keep an empty file around.
> 
> I'd imagine checking in an empty template would still override GitHub's default behaviour of using the commit message. Which if anything would probably surprise contributors.

Probably true. We can always readd it if it's annoying though.

PR #2893, opened 2025-04-17
Don't discard everyScope selections for multiple targets

### LINKED ISSUE #2889: Unexpected token appears to be included when I say 'flash every token this'
When I have several tokens selected in the tale of three lines, and I say 'flash every token this' it includes the first token in the last line which is not in the selection. I expected that it would only flash the tokens that were within the selection. I'm including a video so you can see what I'm talking about. Is this the expected behavior?

https://github.com/user-attachments/assets/eecf8077-5fd5-4e7a-91a8-d720c9fbe1c7

AndreasArvidsson:
Unfortunately became quite a lot of changes to pass the modifier options, but I think this is the only way. It also opens up for more context aware modifiers in the future.

Fixes #2889

### DIFF
--- a/data/fixtures/recorded/everyScope/takeEveryToken.yml
+++ b/data/fixtures/recorded/everyScope/takeEveryToken.yml
@@ -1,0 +1,33 @@
+languageId: plaintext
+command:
+  version: 7
+  spokenForm: take every token
+  action:
+    name: setSelection
+    target:
+      type: primitive
+      modifiers:
+        - type: everyScope
+          scopeType: {type: token}
+  usePrePhraseSnapshot: false
+initialState:
+  documentContents: |-
+    aaa bbb ccc
+    ddd eee
+  selections:
+    - anchor: {line: 0, character: 4}
+      active: {line: 0, character: 11}
+    - anchor: {line: 1, character: 4}
+      active: {line: 1, character: 7}
+  marks: {}
+finalState:
+  documentContents: |-
+    aaa bbb ccc
+    ddd eee
+  selections:
+    - anchor: {line: 0, character: 4}
+      active: {line: 0, character: 7}
+    - anchor: {line: 0, character: 8}
+      active: {line: 0, character: 11}
+    - anchor: {line: 1, character: 4}
+      active: {line: 1, character: 7}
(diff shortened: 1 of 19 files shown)


pokey (commented):
Wow this touched a lot of code for such a small edge case. Would it make any sense to pass into the stage constructor instead? Would that help?

I'm also not totally convinced by this fix. Eg what if you have

```
aaa bbb
ccc ddd
```

And you say "take bat and drum" and then say "take first token". Would you really expect that to behave differently from "take bat" followed by "take first token"?

I think a better heuristic would be whether all the inputs have the same number of scopes. Ie our current hack/heuristic should either apply to all input targets or none. But that does start to feel maybe hard to reason about?

AndreasArvidsson:
> Wow this touched a lot of code for such a small edge case. Would it make any sense to pass into the stage constructor instead? Would that help?
> 

I don't think that would touch less files. There's a lot of instances where we create a modifier stage and then call it directly after.

> I'm also not totally convinced by this fix. Eg what if you have
> 
> ```
> aaa bbb
> ccc ddd
> ```
> 
> And you say "take bat and drum" and then say "take first token". Would you really expect that to behave differently from "take bat" followed by "take first token"?
> 

True

> I think a better heuristic would be whether all the inputs have the same number of scopes. Ie our current hack/heuristic should either apply to all input targets or none. But that does start to feel maybe hard to reason about?

You mean we would actually run the modify stage on each target and then compare the results. If needed rerun the state with different settings? This would be a level of interaction between the different paths of our pipeline we have never done before.

AndreasArvidsson:
@pokey I thought some more on the idea of the different instances of the modifier being more aware of each other. My first was to add a optional function `runMultiple` on the modifier interface. Then the every scope modifier could handle that internally. That would be no problem at all if it was only called from the pipeline, but we have lovely pieces of code like `getEveryScopeTargets` where the ordinal modifier relies on the every modifier. At that point we only have a single target. Passing an option object through the way I'm already doing in this pull request is fine, but bubbling an exception or similar all the way from the every modifier all the way through getEveryScopeTargets to the ordinal modifier increases complexity quite a lot.
 
https://github.com/cursorless-dev/cursorless/blob/e5e9929c05e2607cc7bbddf1d22013b5aa89c772/packages/cursorless-engine/src/processTargets/modifiers/targetSequenceUtils.ts#L38-L48

AndreasArvidsson:
I think this solution is good enough for now. We can always revisit with a more complex solution that utilizes the result from multiple stages to make its decision later.

PR #2899, opened 2025-04-21
Make notebook tests more reliable

AndreasArvidsson:
Original problem is that test run crashes on windows with false negative.
https://github.com/cursorless-dev/cursorless/actions/runs/14509376075/job/40704611423?pr=2893

It appears that it's `vscode.window.showTextDocument` in our `openNewEditor` utility that causes this. It's not always on the same test case. By default it crashes in the scope tests, but if I disable the scope test the same error occurs in our recorded tests that of course uses the same `openNewEditor` function. I can't recreate this locally. Looks like a memory leak or similar.

In this pull request I managed to get the notebook tests more reliable though so I think that is words merging and I will continue working on the crashing problem.
Continues here https://github.com/cursorless-dev/cursorless/pull/2901

### DIFF
--- a/packages/test-harness/src/launchVscodeAndRunTests.ts
+++ b/packages/test-harness/src/launchVscodeAndRunTests.ts
@@ -85,10 +85,6 @@ export async function launchVscodeAndRunTests(extensionTestsPath: string) {
         useLegacyVscode || os.platform() === "win32"
           ? undefined
           : [`--crash-reporter-directory=${crashDir}`, `--logsPath=${logsDir}`],
-      extensionTestsEnv: {
-        ["NODE_OPTIONS"]:
-          "--max-old-space-size=4096 --force-node-api-uncaught-exceptions-policy=true",
-      },
     });
 
     console.log(`Returned from "runTests" with value: ${code}`);
--- a/packages/test-harness/src/launchVscodeAndRunTests.ts
+++ b/packages/test-harness/src/launchVscodeAndRunTests.ts
@@ -81,10 +81,7 @@ export async function launchVscodeAndRunTests(extensionTestsPath: string) {
       // don't bother.  Can be re-enabled if we ever need it; on windows it only
       // hangs some of the time, so might be enough to get a crash dump when you
       // need it.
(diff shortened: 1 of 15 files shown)


packages/test-harness/src/launchVscodeAndRunTests.ts
```diff
@@ -81,10 +81,7 @@ export async function launchVscodeAndRunTests(extensionTestsPath: string) {
       // don't bother.  Can be re-enabled if we ever need it; on windows it only
       // hangs some of the time, so might be enough to get a crash dump when you
       // need it.
-      launchArgs:
-        useLegacyVscode || os.platform() === "win32"
-          ? undefined
-          : [`--crash-reporter-directory=${crashDir}`, `--logsPath=${logsDir}`],
```
pokey, line 87: Why not leave these? Could be useful for debugging

packages/test-harness/src/launchVscodeAndRunTests.ts
```diff
@@ -81,10 +81,7 @@ export async function launchVscodeAndRunTests(extensionTestsPath: string) {
...
+      launchArgs: ["--force-node-api-uncaught-exceptions-policy=true"],
```
pokey, line 84: This seems like a good thing but curious what prompted it

packages/test-harness/src/launchVscodeAndRunTests.ts
AndreasArvidsson: Oh I'm just testing right now. Shouldn't make it a draft.

packages/test-harness/src/launchVscodeAndRunTests.ts
AndreasArvidsson: https://github.com/cursorless-dev/cursorless/actions/runs/14509376075/job/40704611423?pr=2893

AndreasArvidsson: Looks like this argument isn't available in electron. Electron apparently enforces note options quite strictly.

PR #2918, opened 2025-05-21
Hide output from visible text editors

AndreasArvidsson:
Today we waste hats on the output which shouldn't really be a text editor


pokey:
Do you have a screenshot of where these hats are?

AndreasArvidsson:
Output panel
image

pokey:
I actually consider that a feature. They only show up if the panel is visible right?

AndreasArvidsson:
If you think we should keep this functionality it's fine with me?

pokey:
I'd be inclined to keep today's functionality. If we can put hats on visible things, I think we should. The active editor gets preferred hats anyway I assume so seems like there's not much lost

PR #2935, opened 2025-05-29
Disable latex parser on vscode versions >= 1.98.0

AndreasArvidsson:


packages/cursorless-vscode/src/extension.ts
```diff
@@ -266,6 +267,10 @@ function createTreeSitter(parseTreeApi: ParseTreeApi): TreeSitter {
 
     loadLanguage: parseTreeApi.loadLanguage,
     createQuery: parseTreeApi.createQuery,
+
+    disableLatexMsg: semver.lt(vscode.version, "1.98.0")
+      ? undefined
+      : "Latex is disabled on vscode versions >= 1.98.0. issues/2879",
```
pokey, line 273:
```suggestion
      : "Latex is disabled on vscode versions >= 1.98.0. See https://github.com/cursorless-dev/cursorless/issues/2879",
```

pokey (commented):
this didn't work for me. I'm assuming we need to have a similar check in the vscode parse tree extension?

AndreasArvidsson:
Oh I forgot that when you're run vscode in debug this problem doesn't occur. I think we actually only need this in the parse tree extension.

PR #2936, opened 2025-05-29
Include callee in domain for argument iteration scope

### LINKED ISSUE #2930: Inconsistent domain for argument iteration scope
For a formal arguments we use the entire function as the iteration scope to main:
https://github.com/cursorless-dev/cursorless/blob/82318903bfb430f5fd309f69c819e425b7f90d55/data/fixtures/scopes/javascript.core/argument.formal.iteration.scope#L8-L10

But for actual arguments we only use the argument list (including parentheses):
https://github.com/cursorless-dev/cursorless/blob/82318903bfb430f5fd309f69c819e425b7f90d55/data/fixtures/scopes/javascript.core/argument.actual.iteration.scope#L8-L10

I can't remember the discussion around this decision, but I think the reason why the actual argument domain doesn't include the function name is the case when you have a function call as an argument:
```ts
foo(|bar(1), baz(2))
```

`"every arg"` refers to `bar(1)` and `baz(2)` today, but if `bar` was part of the iteration domain the result would just have been just `1` instead.

That is of course useful. On the other hand it creates an inconsistency and you can always use `item` instead of `arg`. Maybe the fact that item and argument does the same thing in this case is an indication that argument is not correct according to the syntax tree? 

Also in this scenario I think people would be surprised what `"every arg"` refers to.
```ts
function foo(value: number) {

    b|ar(1, 2, 3);

}
```

AndreasArvidsson:
Fixes #2930


queries/csharp.scm
```diff
@@ -274,8 +274,10 @@
 
 (_
   (parameter_list
+    .
```
AndreasArvidsson, line 277: I'm not sure if I prefer `.` in these cases. It doesn't really change the result and it is more verbose. On the other hand it is more explicit.

PR #2941, opened 2025-05-31
Another try at changing open notebook time out

AndreasArvidsson:


packages/vscode-common/src/testUtil/openNewEditor.ts
```diff
@@ -109,7 +109,7 @@ function waitForEditorToOpen(): Promise<void> {
     const interval = setInterval(() => {
       if (vscode.window.activeTextEditor != null) {
         clearInterval(interval);
-        resolve();
+        setTimeout(resolve, 300);
```
AndreasArvidsson, line 112:
```suggestion
        setTimeout(resolve, 500);
```

PR #2949, opened 2025-06-01
Migrate ruby scope implementations

AndreasArvidsson:


data/fixtures/recorded/languages/ruby/chuckArg3.yml
```diff
@@ -19,7 +19,7 @@ initialState:
   marks: {}
 finalState:
   documentContents: |
-    [1, 2, 3].each_with_the_next { |n, | }
+    [1, 2, 3].each_with_the_next { |n| }
```
auscompgeek, line 22: Nice!

PR #2950, opened 2025-06-01
Migrate rust scope implementations

AndreasArvidsson:


data/fixtures/recorded/languages/rust/changeNameDrum2.yml
```diff
@@ -1,35 +0,0 @@
...
-    target:
-      type: primitive
-      modifiers:
-        - type: containingScope
-          scopeType: {type: name}
-      mark: {type: decoratedSymbol, symbolColor: default, character: d}
-  usePrePhraseSnapshot: true
-initialState:
-  documentContents: |
-    let user = User {
```
AndreasArvidsson: This is `key` not `name`
@pokey Any objections to this?

AndreasArvidsson:
@pokey Here is another one that is questionable. Type arguments as `arg`. I can see the logic in this, but we don't support this in any other languages.
https://github.com/cursorless-dev/cursorless/blob/834c7aa77699c06e4e9e1ce28a0c7c36114f5b17/data/fixtures/recorded/languages/rust/changeArgBlueAir.yml
https://github.com/cursorless-dev/cursorless/blob/834c7aa77699c06e4e9e1ce28a0c7c36114f5b17/data/fixtures/recorded/languages/rust/changeArgEight.yml#L17

data/fixtures/recorded/languages/rust/changeArgBlueAir.yml
```diff
@@ -21,9 +21,4 @@ initialState:
     blue.a:
       start: {line: 0, character: 18}
       end: {line: 0, character: 19}
-finalState:
```
AndreasArvidsson: This file will be removed after we have consensus that this is not an argument

packages/cursorless-engine/src/languages/LegacyLanguageId.ts
```diff
@@ -2,6 +2,6 @@
  * The language IDs that we have full tree-sitter support for using our legacy
  * modifiers.
  */
-export const legacyLanguageIds = ["rust"] as const;
+export const legacyLanguageIds = [""] as const;
```
AndreasArvidsson:
```suggestion
export const legacyLanguageIds = ["dummy"] as const;
```

packages/cursorless-engine/src/languages/getNodeMatcher.ts
```diff
@@ -39,7 +38,7 @@ export const languageMatchers: Record<
   LegacyLanguageId,
   Partial<Record<SimpleScopeTypeType, NodeMatcher>>
 > = {
-  rust,
+  "": {},
```
AndreasArvidsson:
```suggestion
  "dummy": {},
```

data/fixtures/recorded/languages/rust/clearValue3.yml
AndreasArvidsson, line 1: In no other language is the domain of the return value the entire function. Especially problematic since you can have multiple returns statements.

packages/cursorless-engine/src/languages/LegacyLanguageId.ts
```diff
@@ -2,6 +2,6 @@
  * The language IDs that we have full tree-sitter support for using our legacy
  * modifiers.
  */
-export const legacyLanguageIds = ["rust"] as const;
+export const legacyLanguageIds = ["dummy"] as const;
```
AndreasArvidsson, line 5: Compilation error if this list is empty. Cleaning up all the legacy machinery will be done in a follow up.

data/fixtures/recorded/languages/rust/changeNameDrum2.yml
```diff
@@ -1,14 +1,14 @@
...
-  spokenForm: change name drum
+  spokenForm: change key drum
   action:
     name: clearAndSetSelection
     target:
       type: primitive
       modifiers:
         - type: containingScope
-          scopeType: {type: name}
+          scopeType: {type: collectionKey}
```
AndreasArvidsson, line 11: This looks more like a key / value map than call arguments

data/fixtures/recorded/languages/rust/clearValue3.yml
AndreasArvidsson, line 1: In no other language is the domain for the return value the entire function. This is also extra problematic since you can have multiple return statements.

data/fixtures/recorded/languages/rust/changeArgTrapAndBlueTrap.yml
```diff
@@ -13,6 +13,9 @@ command:
               scopeType: {type: argumentOrParameter}
```
auscompgeek, line 13: Should this also be a type? (How do we handle tuple types in TypeScript?)

AndreasArvidsson: Which one?
You mean typed arrays like `[number, string]`? Those are just collection items and not usable as `type`.

PR #2975, opened 2025-06-13
Implement pnpm overrides

### LINKED ISSUE #2974: Start using pnpm overrides
After this we can remove syncpack

https://app.studyraid.com/en/read/14351/488164/using-overrides-to-manage-problematic-dependencies
https://www.squash.io/how-to-use-pnpm-overrides-for-dependency-management/

AndreasArvidsson:
Fixes #2974


AndreasArvidsson:
I'm closing this. Unfortunately I can't get anything similar to may evens dependency management to work with pnpm:

overrides: Specifies exact version, but a fixed transitive dependencies which breaks build
allowedVersions: Doesn't enforce the exact version. More like a suggestion.

PR #3008, opened 2025-07-04
Cleaned up scope facet infos

AndreasArvidsson:


.github/workflows/test.yml
```diff
@@ -34,6 +34,9 @@ jobs:
       NODE_OPTIONS: "--max-old-space-size=4096"
 
     steps:
+      - name: Git config
```
AndreasArvidsson, line 37: Necessary or meta-updater test will fail on `scopeSupportFastedInfos.md`

PR #3062, opened 2025-08-03
Enable cursorless tag for jetbrains app

asoee:
Enables the user.cursorless tag, when jetbrains app is active.

My reasoning is, that if you have cursorless-talon installed, it is because you want to use it where it is supported.

### DIFF
--- a/cursorless-talon/src/apps/cursorless_jetbrains.py
+++ b/cursorless-talon/src/apps/cursorless_jetbrains.py
@@ -1,0 +1,9 @@
+from talon import Context, actions
+
+ctx = Context()
+
+ctx.matches = r"""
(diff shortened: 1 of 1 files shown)


AndreasArvidsson:
I don't think we should activate Cursorless for jetbrains in this repository since we don't have the actual jetbrains extension implementation here. 

And since It's literally one line to just add `tag(): user.cursorless` two existing `jetbrains.talon` I think it's better to just add that to your documentation.

Joshua-Piper:
@asoee @AndreasArvidsson Hey - just want to see what happened with this and if jetbrains support is still going to be added in the future.

Was your suggestion Andreas to add the plugin implementation into this repo?

PR #3064, opened 2025-08-11
Fix bug wit R arguments and incidental multiple matches

AndreasArvidsson:
Fixes
image

### DIFF
--- a/queries/r.scm
+++ b/queries/r.scm
@@ -43,6 +43,7 @@
   (arguments
     (
       (argument) @_.leading.endOf
+      .
       (comma)
     )?
     .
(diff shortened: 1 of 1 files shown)


pokey:
...test? 😄

AndreasArvidsson:
> ...test? 😄

I was a bit hesitant on this one. We already have argument tests for this language. The problem only occurred when you had it three or more arguments. . I could of course add that test, but right now we have test parity between the languages due to our scope facets. If I add an argument test with three arguments here should I do it for all languages? Or maybe I should just add it here because the Tree sitter looks a bit different from other languages? Should that be a special facet for this and may be other languages? To be honest I'm not sure so I went with the old "When in doubt: do nothing" :D

AndreasArvidsson:
Sounds reasonable :)

https://github.com/cursorless-dev/cursorless/pull/3066

PR #3068, opened 2025-08-12
Change interior scope handler priority to favor surrounding pair interiors

### LINKED ISSUE #3067: typescript: `"inside"` broken for strings inside ternary branches
In the following code:

```ts
const aaa = true ? "bbb" : "ccc";
```

The scopes `"inside bat"` and `"inside cap"` return the whole strings `"bbb"` and `"ccc"` respectively, rather than `bbb` and `ccc` as expected

AndreasArvidsson:
Fixes #3067


pokey (commented):
This seems wrong to
me. I would argue lang scope should take precedence over generic text scope. Why are we even matching lang scope here?

That being said, I'm not going to get in the way of merging to get the unit test in and filing an issue to follow up and understand what's going on

PR #3073, opened 2025-08-24
Fix CONTRIBUTING.MDX formatting snafu introduced in PR 2989

rinOfTheStars:
From what I understand, the auto-formatting/prettifier plugin used by Cursorless really didn't like how CONTRIBUTING.MDX needed to be formatted to have the separation between vscode and vscodium specific commands show properly, and as a result broke a bunch of the code blocks, transforming them from
```bash
echo hello yes this is dog
```
to
`bash echo hello yes this is dog`
though ofc the commands were actual commands and not an echo example. I forget if any other formatting was broken, it has been a while since I actually finished the work on this and I just forgot to make a PR for it, sorry (been busy with getting ready for the new semester, health nonsense, and non-CS hobbies). This PR fixes the aforementioned issue and excludes CONTRIBUTING.MDX from being processed by the auto-formatting plugin, allowing it to contain all of the weird extended markdown jank the plugin seems to really not like. If there's a more elegant way to do this (changing the rules the plugin follows?) please lmk and commit it if you have the time, this is very much not my usual wheelhouse.


AndreasArvidsson (approved):

packages/cursorless-org-docs/src/docs/contributing/CONTRIBUTING.mdx
```diff
@@ -138,11 +150,14 @@ Once you have your package then you can install it into the sandbox using the fo
...
   <TabItem value="vscode" label="VSCode">
-    `bash code --profile=cursorlessDevelopment --install-extension bundle.vsix`
+    ```bash
+    code --profile=cursorlessDevelopment --install-extension bundle.vsix
+    ```
   </TabItem>
   <TabItem value="vscodium" label="VSCodium">
-    `bash codium --profile=cursorlessDevelopment --install-extension
-    bundle.vsix`
+    ```bash
```
auscompgeek, line 158: I suspect if we put blank lines around the code blocks prettier will have a better time

auscompgeek (commented):
Have we reported the bug upstream yet?

packages/cursorless-org-docs/src/docs/contributing/CONTRIBUTING.mdx
AndreasArvidsson: Thank you! That worked.

PR #3085, opened 2025-09-06
Add containment icons to scope visualizer

AndreasArvidsson:
The new category shows all scopes that intersect with your current selection. Scopes smaller than the current selection are excluded and then sorted by smallest first. The result is that the scopes at the top are the ones that match your current selection the best. If you're ever interested in different ways of targeting a particular text you just need to select that and see which scopes match.

I'm not sure about the name/label. Any suggestions for something better than `Selected`?

image


AndreasArvidsson:
Okay the latest version actually uses icons. Became a lot simpler. Note that I only use two separate icons. 

🎯scope is equal to selection
📦 scope contains selection

I think the other two options are not really useful and will just make things unnecessarily complicated

image

PR #3092, opened 2025-10-11
Add GDScript to supported languages

rinOfTheStars:
Doesn't actually implement any scopes (though it does define a few as being `notApplicable`), but defines gdscript (`.gd`) as a supported language, easing further development by not requiring me to run a local build of Cursorless when working on *actual* scope implementations in the future. This has been done for quite a while but I've not had the time to submit the PR until now due to a generally stressful fall semester and catching COVID during the first week of classes, so I thought I'd get the PR in now before I forget again >.<


packages/common/src/extensionDependencies.ts
```diff
@@ -7,6 +7,7 @@ export const extensionDependencies = [
   "mrob95.vscode-talonscript", // talon
   "jrieken.vscode-tree-sitter-query", // scm
   "mathiasfrohlich.kotlin", // kotlin
+  "geequlim.godot-tools", //gdscript
```
AndreasArvidsson:
```suggestion
  "geequlim.godot-tools", // gdscript
```

packages/common/src/scopeSupportFacets/gdscript.ts
```diff
@@ -0,0 +1,19 @@
+import type { LanguageScopeSupportFacetMap } from "./scopeSupportFacets.types";
+import { ScopeSupportFacetLevel } from "./scopeSupportFacets.types";
+
+//eslint-disable-next-line
```
AndreasArvidsson:
```suggestion
// eslint-disable-next-line
```

queries/gdscript.scm
AndreasArvidsson, line 1: Not a big fan of committing an empty file. Please add at least one scope to it. You can do something simple like `comment` for example.

AndreasArvidsson (dismissed):

queries/gdscript.scm
rinOfTheStars: Going to need to make another PR to add another codium cli naming related fix first so I can more easily do that, but once that's done that should be easy enough

rinOfTheStars: See #3093 for the fixes mentioned above

AndreasArvidsson: I don't follow why you would need to update the scripts first? Personally I don't use those scripts at all.
You just run the Cursorless extension from inside of vscode/codium when developing.

rinOfTheStars: It's required in this very specific instance because gdscript isn't supported by prod yet, and I can't get developing from the `cursorlessDevelopment` dev profile to work for some reason. Probably need to delete and remake the thing? 

I can probably bootstrap a local version using the scripts update I made by building from this PR, at the very least, so getting that PR in isn't urgent for now

AndreasArvidsson: I would definitely recommend getting the development profile working. That will hot reload and apply any changes whenever you save a scm file.
Removing the profile/sandbox and installing it again sounds like something to try.

AndreasArvidsson:
We only need the language definition. The rest we get from Tree sitter.

Maybe we should just create a cursorless-dev extension that has all of the language definition needed for testing. In that case we could remove several other extensions from our sandbox.

PR #3093, opened 2025-10-13
Manual cli tool name mode for local/from PR plugin install scripts

rinOfTheStars:
Adds the ability to manually define the IDE cli tool name to `./packages/cursorless-vscode/scripts/install-from-pr.sh`, `[...]/install-local.sh`, and `[...]/uninstall-local.sh`, and updates `./packages/cursorless-org-docs/src/docs/contributing/CONTRIBUTING.mdx` to reflect these changes. 

Soft-blocks #3092 since I use VSCodium and can't (un)install a local plugin version without this fix. 

This requires changing `set -euo pipefail` to `set -eo pipefail` at the start of all three scripts, which might have unforeseen consequences I can't think of right now; I'm also certain there's a better way to demarcate an argument as being optional than how I did it in `install-from-pr.sh` but IDK what that would be, so feel free to change it to something less garbage!

### DIFF
--- a/packages/cursorless-org-docs/src/docs/contributing/CONTRIBUTING.mdx
+++ b/packages/cursorless-org-docs/src/docs/contributing/CONTRIBUTING.mdx
@@ -190,9 +190,22 @@ Once you have your package then you can install it into the sandbox using the fo
 
 You can install a local build of the Cursorless extension by running the following command:
 
-```bash
-pnpm -F cursorless-vscode install-local
-```
+<Tabs groupId="vsx">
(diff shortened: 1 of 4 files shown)


AndreasArvidsson:
Do you still need this? I personally don't use these scripts, but if you do I have no problem with updating them to support more editors.

Regarding `-u` I would like to keep this. ChatGPT had this to say.

image

auscompgeek:
I agree we should keep `set -u`. It seems like this PR has overcomplicated the approach, where a single variable should do the job.

PR #3095, opened 2025-10-31
Added full line scope 

AndreasArvidsson:
Line scope that includes leading and trailing whitespace
`"change full line"`


pokey (dismissed):

AndreasArvidsson:
I'm started thinking that maybe I should convert this to scope instead? Might be useful to be able to do `next expand`? @pokey I think you had the idea of doing a scope `full line` and I think that might be a better solution.

AndreasArvidsson:
@pokey Now converted to a scope

PR #3099, opened 2025-11-10
Change interior domain

AndreasArvidsson:
* Interior domain is now same as content range (before shrinking to text content). `"take inside"` with the cursor on a function name will not select function body anymore. 
* Interior of typed target now utilities every scope. `"take inside funk"` will select function body. `"take inside state"` will select all branch bodies in an if statement.

### DIFF
--- a/packages/cursorless-engine/src/processTargets/modifiers/InteriorStage.ts
+++ b/packages/cursorless-engine/src/processTargets/modifiers/InteriorStage.ts
@@ -18,12 +18,12 @@ export class InteriorOnlyStage implements ModifierStage {
   run(target: Target, options: ModifierStateOptions): Target[] {
     const interior = target.getInterior();
 
-    // eg `inside pair`
+    // eg "inside pair"
     if (interior != null) {
       return interior;
     }
 
-    // eg `inside funk`
+    // eg "inside funk"
     // When you say "inside funk", in an ideal world, the function target would
     // have a defined interior property that we could just use directly.
     // However, it is painful to define interiors for every single scope in
@@ -44,7 +44,7 @@ export class InteriorOnlyStage implements ModifierStage {
       return everyModifier.run(target, options);
     }
 
-    // eg `inside air`
+    // eg "inside air"
     try {
       return createContainingInteriorStage(this.modifierHandlerFactory).run(
         target,


data/fixtures/recorded/surroundingPair/parseTree/python/clearMatching10.yml
```diff
@@ -20,7 +20,7 @@ initialState:
 finalState:
   documentContents: |-
 
-    "'' 'fsd'"
+    ""
   selections:
-    - anchor: {line: 1, character: 2}
-      active: {line: 1, character: 2}
+    - anchor: {line: 1, character: 1}
+      active: {line: 1, character: 1}
```
pokey: i think i would just push the initial cursor 1 to the right like the other tests

packages/common/src/types/command/PartialTargetDescriptor.types.ts
```diff
@@ -236,16 +237,6 @@ export interface CustomRegexScopeType {
...
-export interface InteriorScopeType {
-  type: "interior";
-
-  // The user has specified a scope type. eg "inside element".
-  explicitScopeType?: boolean;
-
-  // Gets passed to surrounding pair scope handler
-  requireStrongContainment?: boolean;
-}
-
```
pokey, line 248: presumably this is never sent from talon? otherwise we need a migration

AndreasArvidsson: It's not. This entire scope type is just for internal use. From talon it's only a modifier.

data/fixtures/recorded/surroundingPair/parseTree/python/clearMatching10.yml
AndreasArvidsson: Done

packages/cursorless-engine/src/processTargets/modifiers/InteriorStage.ts
```diff
@@ -19,49 +15,40 @@ export class InteriorOnlyStage implements ModifierStage {
...
+    // eg `inside funk`
+    if (target.hasExplicitScopeType) {
+      const everyModifier = this.modifierHandlerFactory.create({
+        type: "everyScope",
+        scopeType: {
+          type: "interior",
+        },
+      });
+
+      return everyModifier.run(target, options);
```
pokey: this seems fine, but I think I would consider it a hack. Should we add a comment giving a bit more context on what's happening here and why?

packages/cursorless-engine/src/spokenForms/defaultSpokenFormMapCore.ts
```diff
@@ -107,6 +107,7 @@ export const defaultSpokenFormMapCore: DefaultSpokenFormMapDefinition = {
     textFragment: isPrivate("text fragment"),
     disqualifyDelimiter: isPrivate("disqualify delimiter"),
     pairDelimiter: isPrivate("pair delimiter"),
+    interior: isPrivate("inside"),
```
pokey: do we want to use a different default name for this? I don't feel strongly, but might be useful to differentiate from the modifier "inside"

packages/cursorless-org-docs/src/docs/contributing/scopes/surroundingPairInterior.mdx
pokey, line 1: do we really want this to show up in docs?

changelog/2025-11-changeInteriorDomain.md
```diff
@@ -0,0 +1,7 @@
+---
+tags: [enhancement]
+pullRequest: 3099
+---
+
+- Interior domain is now same as content range (before shrinking to text content). `"take inside"` with the cursor on a function name will not select function body anymore.
+- Interior of typed target now utilities every scope. `"take inside funk"` will select function body. `"take inside state"` will select all branch bodies in an if statement.
```
pokey: not sure we want to advertise this, as I think it's kind of an experimental fix. I believe users shouldn't notice any change

packages/cursorless-org-docs/src/docs/contributing/scopes/surroundingPairInterior.mdx
AndreasArvidsson: It's flagged as an internal scope. Is basically only for developers to look at.

packages/cursorless-engine/src/spokenForms/defaultSpokenFormMapCore.ts
AndreasArvidsson: We could call it interior if you prefer? I believe the only place where this is used is for the test recorder and I don't think we're going to record any tests for the scope specifically.

packages/cursorless-engine/src/processTargets/modifiers/InteriorStage.ts
AndreasArvidsson: Comment added

changelog/2025-11-changeInteriorDomain.md
AndreasArvidsson: When I for once remembered to actually make a chance log file :D
File deleted.

packages/cursorless-engine/src/spokenForms/defaultSpokenFormMapCore.ts
pokey: I think also used for scope visualizer during dev. I don't feel strongly, but I think if you enabled the scope so you could use it for scope visualizer in dev, you'd end up getting "inside" the scope clashing with "inside" the modifier? So yeah maybe "interior" would be fine?

AndreasArvidsson: Right now it's excluded from the sidebar. It doesn't have an iteration scope which throw errors otherwise. And I think that even if we would add it it would be based on the user csv file. I think the spoken form we put in this file is only actually used for the recorded tests. 
I have changed it now.

pokey: Cool yeah good pont I think you are right that to get it to work with visualizer, you'd be manually adding it to csv so you could use whatever term you want.

packages/cursorless-org-docs/src/docs/contributing/scopes/surroundingPairInterior.mdx
pokey: Right yeah I guess that's pretty clear at the top of https://deploy-preview-3099--cursorless.netlify.app/docs/contributing/scopes/interior/ 👍

packages/cursorless-engine/src/processTargets/modifiers/InteriorStage.ts
```diff
@@ -19,49 +18,57 @@ export class InteriorOnlyStage implements ModifierStage {
...
+    // The reason for this being an every scope when we have an explicit scope
+    // type is because we are looking for interiors inside of the scope. We
+    // don't want a normal containing behavior that expands.
+    if (target.hasExplicitScopeType) {
+      const everyModifier = this.modifierHandlerFactory.create({
+        type: "everyScope",
+        scopeType: {
+          type: "interior",
+        },
+      });
```
pokey: I'm a bit surprised that this actually works. For example, in:

```ts
function aaa(bbb: ccc) {
   ddd;
}
```

If I say `"clear inside funk air"`, how come you don't get:

```ts

function aaa(|) {
   |
}
```

pokey: ohh I guess it's because `(bbb: ccc)` is `surroundingPairInterior` whereas this is explicitly requestiont `interior`?

pokey: So that means that even for `{}`-style languages, we still need to define `interior` in tree-sitter if we want `"inside funk"` to work?

pokey: Ok here's a pathological case:

```ts
function aaa(bbb = () => { return "ccc"; }) {
   ddd;
}
```

In this case, you'd expect `"clear inside air"` to give:

```ts
function aaa(bbb = () => { return "ccc"; }) {
   |
}
```

but instead you'd get:

```ts
function aaa(bbb = () => { | }) {
   |
}
```

I think this is an ok compromise, but just figured I'd flag to demonstrate why this approach is really just a hack that approximates what we actually want

packages/cursorless-engine/src/processTargets/modifiers/InteriorStage.ts
```diff
@@ -19,49 +18,57 @@ export class InteriorOnlyStage implements ModifierStage {
...
-    return containingModifier.run(target, options);
-  }
-}
-
-export class ExcludeInteriorStage implements ModifierStage {
-  private containingSurroundingPairIfNoBoundaryStage: ModifierStage;
+    // eg `inside funk`
+    // The reason for this being an every scope when we have an explicit scope
+    // type is because we are looking for interiors inside of the scope. We
+    // don't want a normal containing behavior that expands.
```
pokey: Maybe something like:

> When you say `"inside funk"`, in an ideal world, the function target would have a defined `interior` property that we could just use directly. However, it is painful to define interiors for every single scope in every single language, particularly scopes like `"if"`, which could have multiple interiors and tends to be defined in a tricky nested way in the parse tree.
>
> Instead, we just define `interior` as a scope, and then call `every interior` on the input target here. This will work as expected in most cases, as long as the nearest interior is what we expect, which it usually is.

packages/cursorless-engine/src/processTargets/modifiers/InteriorStage.ts
AndreasArvidsson: Exactly. `inside scope` completely rely on Tree sitter.

Maybe I can work around this. It's not perfect, but I think this is a decent solution for now.

packages/cursorless-engine/src/processTargets/modifiers/InteriorStage.ts
AndreasArvidsson: Thanks. Updated now.

packages/cursorless-engine/src/processTargets/modifiers/InteriorStage.ts
AndreasArvidsson: I had look at this and there's nothing in Tree sitter that can help us with this particular case with my suggested implementation. With that said I'm happy with this behavior  for now. I think it's much better than today's implementation.

The only way to properly fix this is to use interior properties. I guess we could do both? Where it matters like function bodies we could use a property and for other things like if statements we use a separate scope? That would probably give us the best of both? Right now I can't think of any other case except functions where you could create the above problem.  The question is if it worth the increased complexity for this edge case?

PR #3104, opened 2025-11-17
Add skipDuplicate to Open VSX publish step

AndreasArvidsson:
Makes it possible to rerun the deploy workflow when the vscode market place failed previously


auscompgeek:
Would it make more sense to put the build and deploys in separate jobs?

PR #3106, opened 2025-11-17
Separate VSCode publish jobs

auscompgeek:
Implements the suggestion from #3104 to separate the publish jobs.

- Split VSCode publish into two separate jobs (Open VSX and VS Marketplace)
- Ensures the same VSIX file is published to both registries
- Build and Open VSX publish happen together, VS Marketplace publishes sequentially to the same VSIX
- Allows independent reruns of each marketplace publish job
- Removed skipDuplicate as it's no longer needed with separate jobs

Amp-Thread-ID: https://ampcode.com/threads/T-c5112243-7633-492f-911d-5c3e5998c847

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -7,7 +7,7 @@ on:
     branches: [main]
 
 jobs:
-  publish-extension:
+  publish-to-open-vsx:
     runs-on: ubuntu-latest
     if: ${{ github.event.workflow_run.conclusion == 'success' }}
     environment: production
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -47,14 +47,37 @@ jobs:
...
           pat: ${{ secrets.OPEN_VSX_TOKEN }}
           packagePath: packages/cursorless-vscode/dist
-          skipDuplicate: true
+
+      - name: Upload VSIX
+        uses: actions/upload-artifact@v4
+        with:
+          name: vscode-extension
+          path: ${{ steps.publishToOpenVSX.outputs.vsixPath }}
+          retention-days: 1
```
AndreasArvidsson, line 56: Do we need `overwrite: true`? Advise I think it will fail if we do to deployments in the same day.

.github/workflows/deploy.yaml
```diff
@@ -116,7 +139,7 @@ jobs:
   push-cursorless-talon:
     name: Push cursorless-talon subrepo
     runs-on: ubuntu-latest
-    needs: publish-extension
+    needs: [publish-to-vs-marketplace, publish-neovim-extension]
```
AndreasArvidsson, line 142: My gut feeling is that if we successfully update the vscode marketplace we should also update the Talon repository regardless of neovim status. Historically the neovim publishing have been a bit flaky.

AndreasArvidsson:
I wouldn't say I'm against go in this direction. At the same time is it really that problematic to rebuild the binary in case the vscode market plays publishing doesn't work? We're already rebuilding the binary between testing and publishing. I think the best solution would be to build the binary once and then reuse it for testing and publishing, and if we're not going to do that the current set up is fine. How do you feel?

pokey:
If we're going this direction I think I'd opt for 3 steps: build, publish openvsix, and publish vscode. The latter two can run in parallel and both depend on build. Ideally one marketplace being down shouldn't block the other, and we should be able to rerun either.

.github/workflows/deploy.yaml
pokey: Agreed

AndreasArvidsson (changes requested):

AndreasArvidsson:
The publishing has actually been quite stable as of late. I think I'm going to close this and if the need arises I will create a new one based on pokeys suggestion with three three separate steps.

PR #3113, opened 2025-11-27
Update es target version in tsconfig.json files

AndreasArvidsson:
* `es2022 ` for code that targets editors or development
* `es2022` for code that targets browsers

Also did some general cleanup

### DIFF
--- a/packages/cheatsheet/tsconfig.json
+++ b/packages/cheatsheet/tsconfig.json
@@ -1,10 +1,11 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    "target": "es2022",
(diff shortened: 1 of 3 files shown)


auscompgeek:
We still target VSCode 1.95, which shipped with Node 20.18. It looks like Node 20 is missing some ES2024 features?

https://github.com/cursorless-dev/cursorless/blob/2fd5b85845eb38a8656565e3cc614e8db1147137/packages/test-harness/src/launchVscodeAndRunTests.ts#L38

AndreasArvidsson:
You're probably right. All the tests passed for 1.95, but it's probably best to er on the side of caution.
 I have now changed everything to `es2022`.

PR #3116, opened 2025-11-29
build: Update nix flake

BlueDrink9:
Brings the node js version up to date with the required one And removes outdated patches

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -53,6 +53,7 @@
                 pkgs.corepack
                 pkgs.vsce
                 pkgs.nodejs_20
+                pkgs.pre-commit
 
                 python
                 pkgs.lua-language-server # language server used by pre-commit hooks


flake.nix
```diff
@@ -90,14 +52,8 @@
               [
                 pkgs.corepack
                 pkgs.vsce
-                pkgs.nodejs
+                pkgs.nodejs_20
 
-                # https://github.com/NixOS/nixpkgs/pull/251418
-                (pkgs.pre-commit.overrideAttrs (previousAttrs: {
```
auscompgeek, line 96: Isn't this change removing the `pre-commit` package entirely?

BlueDrink9: oops, good catch ty

AndreasArvidsson (approved):

PR #3122, opened 2025-12-13
Utilize interior in relative scope modifier

### LINKED ISSUE #2627: Special handling of previous/next scope when in a code block/body
Today's behavior given is a bit awkward around nested scopes. eg
```py
if |value == 0:
   return None
print("after")
```
In this scenarios saying `"take next state"` I would expect to select the print statement, but you actually get the return statement. This is because we yield the statements in document order and we don't skip descendant scopes.

The reason why we can't just skip the containing scope altogether is this example
```py
if|value == 0:
   |
   return None
print("after")
```
Here the if statement is containing so the statement after containing would be print, but in this scenario I definitely would expect the return statement for `"take next state"`.

One solution to this is to have specific behavior when inside of a code block/body.

AndreasArvidsson:
By utilizing interior scope in relative scope modifier we can skip the interior of scopes. See new test.

Fixes #2627

### DIFF
--- a/packages/cursorless-neovim-e2e/src/shouldRunTest.ts
+++ b/packages/cursorless-neovim-e2e/src/shouldRunTest.ts
@@ -14,6 +14,8 @@ const failingFixtures = [
   "recorded/implicitExpansion/cloneThat2",
   "recorded/implicitExpansion/cloneThis",
   "recorded/implicitExpansion/cloneThis2",
+  // Incorrect final state
+  "recorded/relativeScopes/changePreviousPair",
 ];
 
 function isFailingFixture(name: string, fixture: TestCaseFixtureLegacy) {


packages/cursorless-engine/src/processTargets/modifiers/RelativeScopeStage.ts
```diff
@@ -130,12 +140,79 @@ function generateScopesExclusive(
...
+    {
+      skipAncestorScopes: true,
+      distalPosition: containingDistalPosition,
+    },
+  );
+
+  // Interiors containing the initial position are excluded
+  const interiorRanges = Array.from(interiorScopes)
+    .filter((s) => !s.domain.contains(initialPosition))
+    .map((s) => s.domain);
```
pokey: Nitpicking here but might as well do map / filter on the stream then convert to list at the end

packages/cursorless-engine/src/processTargets/modifiers/RelativeScopeStage.ts
```diff
@@ -130,12 +140,79 @@ function generateScopesExclusive(
...
+    editor,
+    containingInitialPosition,
+    direction,
+    {
+      skipAncestorScopes: true,
+      distalPosition: containingDistalPosition,
+    },
+  );
+
+  // Interiors containing the initial position are excluded
```
pokey: Does this actually happen in practice? Prob no harm leaving it but I wouldn't expect it to actually happen

packages/cursorless-engine/src/processTargets/modifiers/RelativeScopeStage.ts
```diff
@@ -130,12 +140,79 @@ function generateScopesExclusive(
...
+    scopes = ifilter(
+      scopes,
+      (s) => !interiorRanges.some((r) => r.contains(s.domain)),
+    );
+  }
+
+  return islice(scopes, offset - 1, offset + desiredScopeCount - 1);
+}
+
+function getInteriorRanges(
```
pokey: Could use a jsdoc and a better name. Maybe like "getScopeInteriorRanges"? It's getting the top-level interior ranges for this scope right?

pokey (approved):
Hard to say whether this will be nice or annoying in practice. I think worth shipping to find out

packages/cursorless-engine/src/processTargets/modifiers/RelativeScopeStage.ts
AndreasArvidsson: This happens when your cursor is in the body off an if statement for example and you say `next state`. I added test specifically for this use case.

packages/cursorless-engine/src/processTargets/modifiers/RelativeScopeStage.ts
AndreasArvidsson: You mean using itertools? I can have a look at that.

packages/cursorless-engine/src/processTargets/modifiers/RelativeScopeStage.ts
AndreasArvidsson: Agreed
Exactly nested interiors are excluded.

packages/cursorless-engine/src/processTargets/modifiers/RelativeScopeStage.ts
pokey: Ah right yeah sorry that's kinda the whole point isn't it? 😅

AndreasArvidsson: More or less :D

PR #3136, opened 2026-01-15
Preserve number padding in increment/decrement actions

trillium:
## Overview

Preserve leading zeros when incrementing or decrementing a number.

Before:

```py
# increment this
01 -> 2
```

After:
```py
# increment this
01 -> 02
```

- Add `hasLeadingZeros()` helper to detect numbers with leading zeros
- Add `formatNumber()` function to preserve original number formatting (leading zeros and decimal places)
- Update `updateInteger()` and `updateFloat()` to return formatted strings instead of raw numbers
- Preserve formatting for numbers like `007` → `008`, `3.0` → `4.0`, and `004.60` → `005.60`
- Add new fixtures files for increment/decrement

### DIFF
--- a/packages/cursorless-engine/src/actions/incrementDecrement.ts
+++ b/packages/cursorless-engine/src/actions/incrementDecrement.ts
@@ -121,7 +121,7 @@ function formatNumber(
   const absValue = Math.abs(value);
 
   if (hasLeadingZeros(text)) {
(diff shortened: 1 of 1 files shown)


trillium:
🙌

Thanks for the cleanup @AndreasArvidsson

PR #3300, opened 2026-08-23
feat: Add swift support

rinOfTheStars:
Very much a work in progress right now. The three things that need to be done before this is ready are finishing filling out the list of `unsupported` facets, getting the test fixtures at  `resources/fixtures/scopes/swift/branch.if.else.scope`, `resources/fixtures/scopes/swift/branch.if.scope`, and `resources/fixtures/scopes/swift/condition.if.scope` to generate properly, and figuring out the language plugin situation for VSCode testing specifically. Apparently, Swift shares the problem gdscript has where the language support extension only runs in a particular environment, so it's going to need the bare-bones language support extension trick as well. I've created a draft for such a thing already and it seems to work fine; should that be put in its own repository and uploaded to the VSCode and OpenVSC marketplaces, or should it live as part of the monorepo?


rinOfTheStars:
There's also currently a TODO [here](https://github.com/rinOfTheStars/cursorless/blob/259b3df59acad0bbd02a838eacffbc7eac3fd357/packages/lib-common/src/scopeSupportFacets/swift.ts#L18); as I'm not sure as to which scope facets should be used there

AndreasArvidsson:
This is a for each loop. A for loop refers to the c-style for loops with an incrementing integer.

```swift
for val in sequence {}
```

rinOfTheStars:
> This is a for each loop. A for loop refers to the c-style for loops with an incrementing integer.
> 
> 
> 
> ```swift
> 
> for val in sequence {}
> 
> ```

That's what I thought, thanks—I just wanted to make sure. I'll remove the TODO and mark the c-style for as `notApplicable` when I have time later.

rinOfTheStars:
As far as I can tell, I'm not doing anything too incorrectly when defining the if-else scopes in `swift.scm`; despite this, the tests for some of the scope facets _still_ fail to generate. I even went and added the for-each loop scope facets in much the same way just to check, and those generated fine! Gonna link to this in the Slack in case someone else can either figure out what I'm doing wrong, or get it to generate on their machine despite it failing to generate here.

AndreasArvidsson:
You can use the parse tree command from cursorless-talon-dev to have a look at the parse tree for the tests that are failing and try to figure out why you don't get a match.

https://github.com/cursorless-dev/cursorless/blob/c090b01794609e96c1650b68152a3f41f67661bb/cursorless-talon-dev/src/cursorless_dev.talon#L38-L39

Also I would like you to have a look at how the scope tests are formatted for other languages as well as our format guide. Right now you don't use the standard names or formatting.
https://github.com/cursorless-dev/cursorless/blob/c090b01794609e96c1650b68152a3f41f67661bb/packages/app-web-docs/src/docs/contributing/scope-test-format.md

Also all the recorded tests are not necessary. I don't really know what's up with the folder names, but we only need scope tests when you add a language. The recorded test are either for actions modifiers or just legacy language before we had the scope tests.

rinOfTheStars:
> You can use the parse tree command from cursorless-talon-dev to have a look at the parse tree for the tests that are failing and try to figure out why you don't get a match.
> 
> https://github.com/cursorless-dev/cursorless/blob/c090b01794609e96c1650b68152a3f41f67661bb/cursorless-talon-dev/src/cursorless_dev.talon#L38-L39

It wasn't that the tests were failing in of themselves, but failing to generate in the first place. I was able to figure out what was going on there and got it working as of the most recent commit.

> Also I would like you to have a look at how the scope tests are formatted for other languages as well as our format guide. Right now you don't use the standard names or formatting. https://github.com/cursorless-dev/cursorless/blob/c090b01794609e96c1650b68152a3f41f67661bb/packages/app-web-docs/src/docs/contributing/scope-test-format.md

I didn't realize there was a style guide for those, sorry! It's quite late here so I'll fix those tomorrow.

> Also all the recorded tests are not necessary. I don't really know what's up with the folder names, but we only need scope tests when you add a language. The recorded test are either for actions modifiers or just legacy language before we had the scope tests.

I have no clue what's up with those either, they're very much not intentionally added and I pretty quickly realized they weren't supposed to be there and stopped including them in commits. I'll go through and remove those manually tomorrow as well. Very good chance they're generating due to something weird in my personal setup, but if other people (on macOS or more generally) start seeing similar behavior when adding scopes to languages/adding new languages to Cursorless, it might be worth investigating separately.

rinOfTheStars:
Just realized five minutes after committing that the ancestor checker predicate is overkill for the purpose I created it for, and that the single-parent check would've worked fine. I'll switch that the next time I have time and capacity to work on this, then, which might not be for a week or so.

Lmk if I should also remove the new predicate outright or if it should stay in, since I'm not quite sure if it's possible for someone to make a language feature where it would be necessary?

AndreasArvidsson:
We can always add the predicate if it's needed. For now please remove any unused code.

packages/lib-common/src/scopeSupportFacets/swift.ts
rinOfTheStars: Calling attention to this todo, since the description in scopeSupportFacetInfos.ts isn't clear

packages/lib-common/src/scopeSupportFacets/swift.ts
```diff
@@ -237,7 +260,7 @@ export const swiftScopeSupport: LanguageScopeSupportFacetMap = {
   "name.iteration.interface": supported,
   "type.iteration.interface": supported,
 
-  // per-block iteration -- todo: do classlikes, functions, protocols, branches, etc. count as blocks?
+  // per-block iteration -- todo: do classlikes, functions, protocols, etc. count as blocks? Or are only if states/branches/loops/etc. included here?
```
rinOfTheStars: Not sure if this distinction actually matters on the implementation side, but I thought I would ask anyway

rinOfTheStars: (ignore the git diff; the newest version of it is the more correct one)

packages/lib-common/src/scopeSupportFacets/swift.ts
AndreasArvidsson: Its for static blocks. Please call out if any descriptions are unclear and I will fix it.

packages/lib-common/src/scopeSupportFacets/swift.ts
AndreasArvidsson: Since it's the iteration scope for statements: all code blocks that can have statements in them is applicable.

resources/queries/swift.scm
```diff
@@ -1,10 +1,10 @@
 ;; https://github.com/alex-pinkus/tree-sitter-swift/blob/with-generated-files/src/grammar.json
 
-;; document-wide
-(
-  (source_file) @class.iteration @statement.iteration @name.iteration @value.iteration @type.iteration
-  (#document-range! @class.iteration @statement.iteration @name.iteration @value.iteration @type.iteration)
-)
+;; document-wide -- seems like there might be a problem with tree-sitter-swift that causes this to not work correctly?
```
rinOfTheStars, line 3: So I honestly have no clue what is going on here? I tested using this capture all by itself and even then it didn't work, failing with a `Couldn't find capture [CAPTURE]`; where `[CAPTURE]` is any of the captures listed, seemingly chosen randomly/non-deterministically? Not sure if this is a problem in the .scm here or in tree-sitter-swift itself.

rinOfTheStars:
~~Git appears to have completely broken _something_ during one of the merges from upstream, and now the docs won't build for a reason which does not occur on main.~~

~~Lovely.~~

...I needed to `pnpm install`

resources/queries/swift.scm
AndreasArvidsson: scm only supports 3 captures per node. All the extra ones will just be ignored. Could that be the problem you are seeing?

rinOfTheStars:
Good enough for now, I'll fix coverage later (and the comment in swift.scm saying the bug fixed in f6ecfb2 is still present as well)

resources/queries/swift.scm
rinOfTheStars: I think it still happened with just one? Let me double-check

rinOfTheStars: Huh, yeah, that was the issue. Wonder why trying that didn't work before? Probably forgot to save the file or something

AndreasArvidsson: Great.
I still can't understand why they don't just throw an exception instead of silently dropping the captures.

rinOfTheStars:
Will fix the `statement.if` scope later, this is good for now

rinOfTheStars:
This is roughly where I want to leave things for the initial swift support PR; I'll be following it up as quickly as my health allows with another (or multiple anothers) to continue filling out scope support, ofc.

Do we still need a rudimentary extension for swift (and gdscript) support for people running local tests in VSC? I have one made; I just need to clean it up a bit and publish it on both extension marketplaces.

AndreasArvidsson:
Thank you. I will have a look at it and if it not to much work I might finish it myself if you don't mind?

You don't need that anymore. Before all tests where ran in vscode, but now all the scope tests are just run directly against the Tree sitter definition as unit tests. This greatly increased performance and reliability, but also made it so we don't need to have each language definition in vscode for the test to pass.

rinOfTheStars:
> Thank you. I will have a look at it and if it not to much work I might finish it myself if you don't mind?

There's still quite a bit to do still, but feel free to tackle any low-hanging fruit if you have the time lol


> You don't need that anymore. Before all tests where ran in vscode, but now all the scope tests are just run directly against the Tree sitter definition as unit tests. This greatly increased performance and reliability, but also made it so we don't need to have each language definition in vscode for the test to pass.

Oh, that's good to hear! With that confirmed I can set this as a 'ready'/non-draft PR later today.

rinOfTheStars:
To be clear, the main reason for wanting to merge this now vs later is to make working on swift support a little bit more ergonomic than it currently is; if you'd all rather I keep it as a PR until it's close to feature complete, I _can_ make that work.

AndreasArvidsson:
@rinOfTheStars Swift is now feature complete and has full test coverage. Thanks you for your contribution! :)

AndreasArvidsson (approved):

